PR #16, opened 2021-03-19
feat(volume share): add share volume REST API

paulyoong:
Add the ability to share a volume to the REST API.

Modify the share API for all resources (nexuses and replicas) such that
to unshare the protocol just needs to be set to off.

Resolves: CAS-798


gila (approved):

paulyoong:
Closing for now - need to make some changes.

PR #18, opened 2021-03-23
fix: map actix builtin errors to our control plane errors

tiagolobocastro:
- refactor(rest): simplify the openapi macros

  Simplify the openapi macros which at the moment have the full and relative
   path for a handler. This is possible thanks to the paperclip change which
   trims the base path of all handlers if it matches the api base path.
  
  This means we can now use scopes to avoid repeating the base path on each
   handler's macro.

- fix(errors): map actix builtin errors

  The Path and Json body validation are handle by the actix; in case of
   error we should make it to our own error types.
  Now that we have scopes we can easily add an error to the v0 scope.
  Example of Invalid JSON bodies error:
  400: Bad Request
  {
    "error": "Deserialize",
    "details": "Json deserialize error: missing field `size` at line 4 column 1"
  }

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2231,7 +2231,7 @@ dependencies = [
 [[package]]
 name = "paperclip"
 version = "0.5.0"
-source = "git+https://github.com/MayastorControlPlane/paperclip?branch=develop#2220457419e62d42d9b93e43d9b34b510925dd26"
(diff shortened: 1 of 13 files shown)


paulyoong (approved):

tiagolobocastro:
bors merge

PR #19, opened 2021-03-29
chore(filter): remove filter from only_one macro

paulyoong:
The only_one macro was not using the provided filter as a means of
finding an item in a list; it was only being recorded in the error
messages.

### DIFF
--- a/control-plane/mbus-api/src/message_bus/v0.rs
+++ b/control-plane/mbus-api/src/message_bus/v0.rs
@@ -12,14 +12,14 @@ pub type BusError = ReplyError;
 pub type BusResult<T> = Result<T, BusError>;
 
 macro_rules! only_one {
-    ($list:ident, $resource:expr, $filter:expr) => {
+    ($list:ident, $resource:expr) => {
         if let Some(obj) = $list.first() {
             if $list.len() > 1 {
                 Err(ReplyError {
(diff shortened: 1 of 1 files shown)


control-plane/mbus-api/src/message_bus/v0.rs
```diff
@@ -12,14 +12,14 @@ pub type BusError = ReplyError;
...
-    ($list:ident, $resource:expr, $filter:expr) => {
+    ($list:ident, $resource:expr) => {
         if let Some(obj) = $list.first() {
             if $list.len() > 1 {
                 Err(ReplyError {
                     kind: ReplyErrorKind::FailedPrecondition,
                     resource: $resource,
                     source: "".to_string(),
-                    extra: $filter.to_string(),
+                    extra: "".to_string(),
```
tiagolobocastro, line 22: wouldn't it still be useful to have the actual filter used in the error? Eg, we know if was about a resource, but we don't know which particular one?

paulyoong: The REST calls that return these errors were already called with an ID so they are already aware of the resource that they were requesting.

tiagolobocastro (approved):

paulyoong:
bors merge

PR #20, opened 2021-03-30
feat(watcher): add new watcher agent

tiagolobocastro:
The watcher is currently part of the core agent and it processes
 get/create/delete calls
On creation it makes use of the store to create a watch on the specified
 resource notifying the requester of changes everytime the resource
 changes.
At the moment the only supported type of watch are http callbacks. This
 is meant to allow the operator to watch for changes without polling.
Our own agents will also be using this method (eg, the control loop)
 might also need to watch for changes so that it can reconcile though
 that will likely not be notified via http (maybe via nats or even
 same binary callback).


tiagolobocastro:
bors try

tiagolobocastro:
bors try

tiagolobocastro:
bors ping

tiagolobocastro:
bors ping

tiagolobocastro:
bors try

tiagolobocastro:
bors try

tiagolobocastro:
bors try

control-plane/rest/src/versions/v0.rs
```diff
@@ -176,6 +177,37 @@ pub struct GetBlockDeviceQueryParams {
     pub all: Option<bool>,
 }
 
+/// The difference types of watches
```
paulyoong, line 180: I think this comment may be a copy-paste error?

control-plane/rest/src/versions/v0.rs
```diff
@@ -176,6 +177,37 @@ pub struct GetBlockDeviceQueryParams {
...
+
+impl TryFrom<&Watch> for RestWatch {
+    type Error = ();
+    fn try_from(value: &Watch) -> Result<Self, Self::Error> {
+        match &value.callback {
+            WatchCallback::Uri(uri) => Ok(Self {
+                resource: value.resource.to_string(),
+                callback: uri.to_string(),
+            }),
+            // _ => Err(()),
```
paulyoong, line 206: Does this need to be uncommented?

control-plane/store/src/etcd.rs
```diff
@@ -80,18 +97,69 @@ impl Store for Etcd {
...
     /// A receiver channel is returned which is signalled when the entry with
     /// the given key is changed.
-    async fn watch(
+    async fn watch_kv<K: StoreKey>(
         &mut self,
-        key: &Value,
+        key: &K,
     ) -> Result<Receiver<Result<WatchEvent, StoreError>>, StoreError> {
         let (sender, receiver) = channel(100);
+        println!("{}", key.to_string());
```
paulyoong, line 105: Should this be removed or changed to a `trace::info`?

control-plane/store/src/etcd.rs
```diff
@@ -80,18 +97,69 @@ impl Store for Etcd {
...
+                key: key.key(),
+            }),
+        }
+    }
+
+    async fn watch_obj<K: ObjectKey>(
+        &mut self,
+        key: &K,
+    ) -> Result<Receiver<Result<WatchEvent, StoreError>>, StoreError> {
+        println!("{}", key.key());
```
paulyoong, line 150: Should this be removed or changed to a `trace::info`?

control-plane/store/src/etcd.rs
```diff
@@ -111,16 +179,15 @@ fn watch(
...
+                            return;
+                        }
                     }
                 }
                 Err(e) => {
-                    // If we failed to get a message, continue to wait for the
-                    // next one.
                     tracing::error!("Failed to get message with error {}", e);
-                    continue;
+                    return;
```
paulyoong, line 190: Should we just return here or should we have a number of attempts before giving up?

control-plane/rest/src/versions/v0.rs
tiagolobocastro: No, because we only have one we can't other `_` for others. I could just deleted, just wanted to make it clear that everything else is not implemented yet, maybe I'll add such comment

control-plane/store/src/etcd.rs
tiagolobocastro: on the etcd-client examples they bail out immediately in case of an error. And from my testing once you get an error there's no recovery.

control-plane/rest/src/versions/v0.rs
paulyoong: A comment sounds good.

control-plane/store/src/etcd.rs
paulyoong: Ok, that sounds fine then.

tiagolobocastro:
bors try

paulyoong (approved):

tiagolobocastro:
bors try

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

PR #22, opened 2021-03-31
feat(store): define persistent data structs

paulyoong:
Initial definition of structures that need to be saved in the
persistent store.

Resolves: CAS-819


control-plane/store/src/lib.rs
```diff
@@ -1,2 +1,3 @@
 pub mod etcd;
 pub mod kv_store;
+pub mod v0;
```
tiagolobocastro, line 3: I wonder if this should be within another submodule, types::v0 or something

tiagolobocastro (approved):
The Cargo.lock changed so you'll have to change the cargo256

control-plane/store/src/lib.rs
paulyoong: Good suggestion - done.

paulyoong:
> The Cargo.lock changed so you'll have to change the cargo256

Good spot - done.

paulyoong:
bors merge

paulyoong:
bors merge

PR #28, opened 2021-04-19
fix(sharing): disambiguate sharing pre-conditions

tiagolobocastro:
Add explicit error codes for sharing pre-condition errors.


control-plane/agents/common/src/errors.rs
```diff
@@ -65,6 +65,14 @@ pub enum SvcError {
     VolumeNotFound { vol_id: String },
     #[snafu(display("Replica '{}' not found", replica_id))]
     ReplicaNotFound { replica_id: ReplicaId },
+    #[snafu(display("{} '{}' is already shared by {}", kind.to_string(), id, share))]
```
paulyoong, line 68: Should it say `...is already share **over** {}`? The share string is the protocol used.

paulyoong (approved):

control-plane/agents/common/src/errors.rs
tiagolobocastro: Indeed, done!

tiagolobocastro:
bors merge

gila (approved):

PR #29, opened 2021-04-19
chore(openapi): use openapi uuid format

tiagolobocastro:
Set the Replica, Nexus and Volume with a UUID format within openapi.
The pool/node are other candidates, which could either be a UUID or remain
 as a "string" but still unique.

Note this is only about openapi, there is no strict enforcement on the
data because how would we handle when a resource created under us in
mayastor is not on the UUI format?
In such case, we might want to validate these only in the rest frontend?


paulyoong (approved):

tiagolobocastro:
bors merge

PR #30, opened 2021-04-21
feat(store keys): define keys for store resources

paulyoong:
Define the keys that should be used for resources that can be saved in
the persistent store.

The keys are defined for the 'spec' and 'state' types for each resource.

Additional changes:
Reorganise the directory structure of the store types. Each type is
moved out of v0.rs and placed in its own source file.

Resolves: CAS-832


tiagolobocastro (approved):
:+1:

paulyoong:
bors merge

PR #35, opened 2021-04-30
chore: split the nexus and volume parts of the volume service

tiagolobocastro:
Given that the nexus is a "low level" api, split it from the volume api


control-plane/agents/core/src/nexus/specs.rs
```diff
@@ -0,0 +1,539 @@
...
+            }
+
+            let result = node.remove_child(request).await;
+            Self::nexus_complete_op(registry, result, nexus_spec, spec_clone).await
+        } else {
+            node.remove_child(request).await
+        }
+    }
+
+    /// Completes a nexus update operation by trying to update the spec in the persistent store.
```
paulyoong, line 429: Is it only "nexus update" operations?

paulyoong (approved):

control-plane/agents/core/src/nexus/specs.rs
tiagolobocastro: Done

tiagolobocastro:
bors merge

PR #39, opened 2021-06-15
chore: remove "allow fail" for test case

paulyoong:
The Jira item referenced for this test case has been marked as resolved
so the expectation is that this test case should now pass.


tiagolobocastro (approved):

paulyoong:
bors merge

PR #41, opened 2021-06-16
chore(swagger-ui): improve user xperience

tiagolobocastro:
chore(swagger-ui): improve user xperience
Add a filtering option.
Don't expand the tags and paths on load.
Auto enabled "TryItOut" so we don't need to click it everytime. Note: to see the schema now,
you'll have cancel the "TryItOut"

chore: add missing Serialize and Default traits 
This allows us to use the new auto generate example for openapi.

feat: add pool device uri type
Also update the Cargo lock file


paulyoong (approved):

tiagolobocastro:
bors merge

PR #42, opened 2021-06-22
feat(volume): add new volume operations

tiagolobocastro:
Add share/unshare and publish/unpublish volume operations.
The publish allows for a nexus to be created or a specified node or any suitable node. In
addition, a share protocol may also be specified.

Volume creation does not result in a nexus being created now. A nexus shall only exist
if the volume is published.

The spec update operations are now generic enough that the code can be shared making use of
the existing traits.
todo: spec creation should also be made generic!
todo: operation validation should also be made generic!


control-plane/agents/core/src/core/specs.rs
```diff
@@ -55,4 +60,77 @@ impl ResourceSpecsLocked {
...
+                }
+            }
+        }
+    }
+
+    /// Validates the outcome of a update operation .
+    /// In case of an error, it undoes the changes to the spec.
+    /// If the persistent store is unavailable the spec is marked as dirty and the dirty
+    /// spec reconciler will attempt to update the store when the store is back online.
+    pub(crate) async fn spec_step_op<R, O, S: SpecTransaction<O> + StorableObject>(
```
paulyoong, line 111: I'm not sure what "step" refers to here. Could we call it something like "validate_op"?

control-plane/agents/core/src/core/specs.rs
```diff
@@ -55,4 +60,77 @@ impl ResourceSpecsLocked {
...
+                    Err(error) => {
+                        spec.set_op_result(false);
+                        Err(error)
+                    }
+                }
+            }
+        }
+    }
+
+    /// Validates the outcome of a update operation .
```
paulyoong, line 107: Is it only intended for update operations?

control-plane/agents/core/src/nexus/specs.rs
```diff
@@ -227,28 +227,34 @@ impl ResourceSpecsLocked {
...
-                        share: spec.share.to_string(),
-                    });
+                } else {
+                    // validate the operation itself against the spec and status
+                    if spec.share != status.share {
+                        return Err(SvcError::NotReady {
+                            kind: ResourceKind::Nexus,
+                            id: request.uuid.to_string(),
+                        });
+                    } else if request.protocol == spec.share || spec.share.shared() {
```
paulyoong, line 237: Why do we check that the request protocol matches the spec? Can't we just check for `spec.share.shared()`?

control-plane/agents/core/src/pool/specs.rs
```diff
@@ -382,37 +382,43 @@ impl ResourceSpecsLocked {
...
-                        share: spec.share.to_string(),
-                    });
+                } else {
+                    // validate the operation itself against the spec and status
+                    if spec.share != status.share {
+                        return Err(SvcError::NotReady {
+                            kind: ResourceKind::Pool,
+                            id: request.uuid.to_string(),
+                        });
+                    } else if request.protocol == spec.share || spec.share.shared() {
```
paulyoong, line 392: Again, can we just check `spec.share.shared()`?

control-plane/agents/core/src/volume/specs.rs
```diff
@@ -134,7 +140,9 @@ async fn get_node_replicas(
 /// Implementation of the ResourceSpecs which is retrieved from the ResourceSpecsLocked
 /// During these calls, no other thread can add/remove elements from the list
 impl ResourceSpecs {
-    /// Get a list of all protected VolumeSpec's
+    /// Return a new blank VolumeSpec for the given request which has been committed to the
+    /// persistent store or returns an existing one if it the create operation has not
```
paulyoong, line 144: Can remove `it`; "..existing one if the create.."

control-plane/agents/core/src/volume/specs.rs
```diff
@@ -464,9 +436,416 @@ impl ResourceSpecsLocked {
...
+                    });
+                } else if request.protocol == spec.protocol || spec.protocol.shared() {
+                    return Err(SvcError::AlreadyShared {
+                        kind: ResourceKind::Volume,
+                        id: spec.uuid.to_string(),
+                        share: spec.protocol.to_string(),
+                    });
+                }
+                if status.children.len() != 1 {
+                    return Err(SvcError::NotReady {
```
paulyoong, line 481: I wonder if we should have an ANA not supported error?

control-plane/agents/core/src/volume/specs.rs
```diff
@@ -464,9 +436,416 @@ impl ResourceSpecsLocked {
...
+        }),
+    }
+}
+
+async fn get_volume_target_node(
+    registry: &Registry,
+    status: &Volume,
+    request: &PublishVolume,
+) -> Result<NodeId, SvcError> {
+    // If the nexus is already published, then it should first be unpublished
```
paulyoong, line 809: This comment mentions unpublishing, but we don't do that here. I guess that's a decision for the caller. Does the comment need updating?

paulyoong (approved):

control-plane/agents/core/src/core/specs.rs
tiagolobocastro: I guess I was looking at an operation as having multiple "steps". If each intermediate step was Ok(_) then we do nothing and move on to the next one; if it fails we fail the operation.

control-plane/agents/core/src/core/specs.rs
tiagolobocastro: Well no, not strictly true. "/// Validates the outcome of an intermediate step, part of a transaction operation." ?

control-plane/agents/core/src/nexus/specs.rs
tiagolobocastro: Ah yes indeed :+1:

control-plane/agents/core/src/pool/specs.rs
tiagolobocastro: Yep same here

control-plane/agents/core/src/volume/specs.rs
tiagolobocastro: Not sure, this is checking the status part of the volume. We can end up with N nexuses in we loose connectivity to a node, then decide to create the nexus on another node and then the first node is reachable again and reports the old nexus. Once we have the persistent reserves we can probably ignore this nexus as it won't be able to write to the children anymore.

control-plane/agents/core/src/volume/specs.rs
tiagolobocastro: Yep, updated to: "// We can't configure a new target_node if the volume is currently published"

control-plane/agents/core/src/core/specs.rs
paulyoong: :+1:

control-plane/agents/core/src/volume/specs.rs
paulyoong: Ok

control-plane/agents/core/src/volume/specs.rs
paulyoong: :+1:

tiagolobocastro:
bors merge

PR #45, opened 2021-06-28
refactor: similar operations should have the same behavior

tiagolobocastro:
As the code evolved different parts were using different validations rules and steps for each
resource type.
Try to address this by having a trait which runs the same validations and has the same error
handling logic for every resource type.
Additional specific behavior can still be added by implementing the traits.
Also, now that we have it, make use of the spec transactions for Create and Delete operations.


tiagolobocastro:
bors try

tiagolobocastro:
bors cancel

tiagolobocastro:
bors try

PR #47, opened 2021-06-28
refactor: use trait for common operations

tiagolobocastro:
As the code evolved different parts were using different validations rules and steps for each
resource type.
Try to address this by having a trait which runs the same validations and has the same error
handling logic for every resource type.
Additional specific behavior can still be added by implementing the traits.
Also, now that we have it, make use of the spec transactions for Create and Delete operations.


tiagolobocastro:
bors try

tiagolobocastro:
bors try

control-plane/agents/common/src/errors.rs
```diff
@@ -121,6 +125,14 @@ pub enum SvcError {
     WatchAlreadyExists {},
     #[snafu(display("Conflicts with existing operation - please retry"))]
     Conflict {},
+    #[snafu(display("Pending deletion - please retry"))]
+    Deleting {},
+    #[snafu(display(
```
paulyoong, line 130: Would it be useful to show both sets of parameters so that it is obvious what differs? Or maybe we could show the parameters that differ?

control-plane/agents/core/src/core/specs.rs
```diff
@@ -35,6 +37,416 @@ enum SpecError {
...
+    ) -> Result<(), SvcError>
+    where
+        Self: SpecTransaction<O>,
+        Self: StorableObject,
+    {
+        let mut spec = locked_spec.lock().await;
+        // we're busy with another request, try again later
+        let _ = spec.busy()?;
+        if spec.state().deleted() {
+            Ok(())
```
paulyoong, line 168: There's an asymmetry between create and destroy. The create returns an error if already created. The delete returns Ok if already deleted. It would be good if we could do the same thing in both cases.

control-plane/agents/core/src/core/specs.rs
```diff
@@ -35,6 +37,416 @@ enum SpecError {
...
+            let spec_clone = spec.clone();
+            drop(spec);
+
+            if let Err(error) = registry.store_obj(&spec_clone).await {
+                let mut spec = locked_spec.lock().await;
+                spec.clear_op();
+                Err(error)
+            } else {
+                Ok(())
+            }
```
paulyoong, line 199: This pattern repeats a number of times. We could split this out into its own "commit_spec" function.

control-plane/agents/core/src/core/specs.rs
```diff
@@ -35,6 +37,416 @@ enum SpecError {
...
+                    Err(error) => {
+                        spec.set_op_result(false);
+                        Err(error)
+                    }
+                }
+            }
+        }
+    }
+
+    /// Start a destroy operation and attempt to log the transaction to the store.
```
paulyoong, line 254: Copy paste error. This is an **update** operation.

control-plane/agents/core/src/core/specs.rs
```diff
@@ -35,6 +37,416 @@ enum SpecError {
...
+        reconciling: bool,
+    ) -> Result<Self, SvcError>
+    where
+        Self: PartialEq<Self::Status>,
+    {
+        // we're busy right now, try again later
+        let _ = self.busy()?;
+
+        if !self.state().created() {
+            Err(SvcError::PendingCreation {
```
paulyoong, line 294: If it's not created can we guarantee that creation is pending? Is it possible that it is not created and nothing is trying to create it?

control-plane/agents/core/src/core/specs.rs
```diff
@@ -35,6 +37,416 @@ enum SpecError {
...
+                id: self.uuid(),
+                kind: self.kind(),
+            })
+        } else {
+            // if it's not part of a reconcile effort
+            // validate the operation itself against the spec and status
```
paulyoong, line 300: It seems as though we are only validating against the status.

control-plane/agents/core/src/core/specs.rs
```diff
@@ -35,6 +37,416 @@ enum SpecError {
...
+                    kind: self.kind(),
+                })
+            } else {
+                self.start_update_op(status, operation)?;
+                Ok(self.clone())
+            }
+        }
+    }
+
+    /// Completes a volume update operation by trying to update the spec in the persistent store.
```
paulyoong, line 314: I think we should remove "volume" from the comment as this is a generic method.

control-plane/types/src/v0/store/replica.rs
```diff
@@ -97,8 +97,12 @@ impl SpecTransaction<ReplicaOperation> for ReplicaSpec {
     fn commit_op(&mut self) {
         if let Some(op) = self.operation.clone() {
             match op.operation {
-                ReplicaOperation::Unknown => {
-                    panic!("Unknown operation not supported");
+                ReplicaOperation::Unknown => unreachable!(),
```
paulyoong, line 100: For some of the resources you have used `unimplemented`. We should use the same for all of them.

paulyoong (approved):

control-plane/agents/common/src/errors.rs
tiagolobocastro: Yeah that'd be useful indeed :+1:

control-plane/agents/core/src/core/specs.rs
tiagolobocastro: Yeah, that was already there with the old code so I left it as such for now - we need to inspect the repo and make sure that for everywhere operation certain errors will not be treated as an error rather than just using ()?;

control-plane/agents/core/src/core/specs.rs
tiagolobocastro: ah yeah it does, I'll give it a go

control-plane/agents/core/src/core/specs.rs
tiagolobocastro: dohhhh

control-plane/agents/core/src/core/specs.rs
tiagolobocastro: the reconciler should be the one trying to do something with it, unless we want to define a state where we say we've given up on the resource and we'll stop trying to reconcile and the resource will never become usable again?

control-plane/agents/core/src/core/specs.rs
tiagolobocastro: we're validating the spec against the status?

control-plane/agents/common/src/errors.rs
tiagolobocastro: Using the Debug trait to generate the string for now, might want to think about using the Serialize or Display instead...

control-plane/agents/core/src/core/specs.rs
paulyoong: That's fine for now.

control-plane/agents/core/src/core/specs.rs
tiagolobocastro: Done, split up the creating/deleting

control-plane/agents/core/src/core/specs.rs
tiagolobocastro: I've improved it a bit by making the spec/status check explicit. For now it's still just a wrapper around the PartialEq but I think we'll probably move that directly to the impl instead, that will be more flexible.

tiagolobocastro:
bors merge

PR #59, opened 2021-07-22
feat: update openapi spec and server to include publish and setreplica

tiagolobocastro:
feat: update openapi spec and autogenerated code
Explicitly specify min/max for some of the integer types we use.

feat: update openapispec and model to include publish and setreplica
Add new openapi operations for creating a volume target (this means we create a nexus)
and new operation to change a volume's replica count.

feat: update rest server to include setreplica and publish
Refactor publish to return the volume.


common/src/mbus_api/message_bus/v0.rs
```diff
@@ -228,6 +229,24 @@ pub trait MessageBusTrait: Sized {
         Ok(())
     }
 
+    /// set volume replica count
```
paulyoong, line 232: Incorrect comment - this is the publish call

control-plane/rest/tests/v0_test.rs
```diff
@@ -335,6 +363,32 @@ async fn client_test(mayastor: &NodeId, test: &ComposeTest, auth: &bool) {
             .unwrap()
     );
 
+    let volume = client
+        .volumes_api()
+        .put_volume_target(&volume.uuid.to_string(), None, None)
```
paulyoong, line 368: Should we also test the case where the node is specified i.e. assert that it is published on the specified node?

paulyoong (approved):

common/src/mbus_api/message_bus/v0.rs
tiagolobocastro: dohh

control-plane/rest/tests/v0_test.rs
tiagolobocastro: well we didn't specify a node (None) - one is auto-selected

paulyoong: But should we also check the behaviour when one is explicitly specified?

tiagolobocastro:
bors merge

PR #60, opened 2021-07-22
chore: remove states from registry

paulyoong:
The state information for resources are store per node in the
NodeWrapper, therefore remove the states from the registry as this is no
longer needed.


tiagolobocastro (approved):

paulyoong:
bors merge

PR #66, opened 2021-08-03
chore: add InUse error when deleting a pool

paulyoong:
Return an InUse error message when attempting to delete a pool which has
replicas on it.

Resolves: CAS-996


tiagolobocastro (approved):

paulyoong:
bors merge

PR #74, opened 2021-08-18
feat: add observability stack to the deployer

tiagolobocastro:
Add Elasticsearch and Kibana to the deployer tool.
Fixup Jaegertracing to be able to connect to Elasticsearch. When using elastic with
jaeger we override the jaeger's entrypoint as the default one will crash on startup
if the elasticsearch is not ready, which is usually not as it's "slow"...

When using the wait option the kibana component will setup an index pattern!
Otherwise, a script has been provided for the effect.

Update the composer library to be able to reuse existing clusters without removing
all containers/networks.
This allows us to start different stacks independently, ex:
> deployer start -ejk -s -a "" --no-nats --no-rest --no-etcd -m 0
> deployer start -s -m 2

Updated the opentelemetry on the core agent and rest service to use a batch exporter
making use of the tokio runtime (otherwise the exports would be blocking)


tiagolobocastro:
bors try

gila (approved):

PR #75, opened 2021-08-19
feat: delete spec when volume creation fails

paulyoong:
Ensure that the volume spec is eventually deleted when volume creation
fails.

In the event that the persistent store cannot be reached, the
reconcile_dirty_volumes function will ensure that the spec will
eventually be removed from the store and registry.

Bug fix:
- Ensure that an error is returned when storing an object to the
persistent store fails.

Resolves: CAS-1059

### DIFF
--- a/control-plane/agents/core/src/core/specs.rs
+++ b/control-plane/agents/core/src/core/specs.rs
@@ -75,8 +75,13 @@ pub trait SpecOperations: Clone + Debug + Sized + StorableObject + OperationSequ
             spec.start_create_inner(request)?;
             spec.clone()
         };
-        Self::store_operation_log(registry, locked_spec, &spec_clone).await?;
-        Ok((spec_clone, guard))
+        match Self::store_operation_log(registry, locked_spec, &spec_clone).await {
+            Ok(_) => Ok((spec_clone, guard)),
+            Err(e) => {
+                Self::delete_spec(registry, locked_spec).await.ok();
+                Err(e)
+            }
+        }
     }
 
     /// When a create request is issued we need to validate by verifying that:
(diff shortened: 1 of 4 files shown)


control-plane/agents/core/src/core/specs.rs
```diff
@@ -141,20 +146,65 @@ pub trait SpecOperations: Clone + Debug + Sized + StorableObject + OperationSequ
...
-                    Ok(_) => {
-                        spec.clear_op();
-                        Err(error)
-                    }
-                    Err(error) => {
-                        spec.set_op_result(false);
-                        Err(error)
-                    }
-                }
+                spec.set_op_result(false);
```
tiagolobocastro, line 206: nit: could do it in one go now, ``locked_spec.lock().set_op_result(false);``

control-plane/agents/core/src/core/specs.rs
```diff
@@ -141,20 +146,65 @@ pub trait SpecOperations: Clone + Debug + Sized + StorableObject + OperationSequ
                 }
             }
             Err(error) => {
-                let mut spec_clone = locked_spec.lock().clone();
-                spec_clone.clear_op();
-                let stored = registry.store_obj(&spec_clone).await;
+                // The create failed so delete the spec.
+                Self::delete_spec(registry, locked_spec).await?;
```
tiagolobocastro, line 150: I think we want to return the original error rather than the etcd error?
```suggestion
                Self::delete_spec(registry, locked_spec).await.ok();
```

control-plane/agents/core/src/core/specs.rs
```diff
@@ -141,20 +146,65 @@ pub trait SpecOperations: Clone + Debug + Sized + StorableObject + OperationSequ
...
+        locked_spec: &Arc<Mutex<Self>>,
+    ) -> Result<R, SvcError>
+    where
+        Self: SpecTransaction<O>,
+        Self: StorableObject,
+    {
+        match result {
+            Ok(val) => Ok(val),
+            Err(error) => {
+                Self::delete_spec(registry, locked_spec).await?;
```
tiagolobocastro, line 171: same here, I guess

tests/bdd/test_volume_create.py
```diff
@@ -146,11 +146,10 @@ def volume_creation_should_fail_with_an_insufficient_storage_error(create_reques
...
         assert exception_info["status"] == requests.codes["insufficient_storage"]
-    # TODO: Uncomment the "finally" clause when CAS-1059 is completed.
-    # finally:
-    #     # Check that the volume wasn't created.
-    #     volumes = common.get_volumes_api().get_volumes()
-    #     assert len(volumes) == 0
+    finally:
+        # Check that the volume wasn't created.
+        volumes = common.get_volumes_api().get_volumes()
+        assert len(volumes) == 0
```
tiagolobocastro, line 152: nice :+1:

tiagolobocastro (approved):

control-plane/agents/core/src/core/specs.rs
paulyoong: :+1: Done.

control-plane/agents/core/src/core/specs.rs
paulyoong: Done.

control-plane/agents/core/src/core/specs.rs
paulyoong: Done.

paulyoong:
bors merge

paulyoong:
bors merge

paulyoong:
bors merge

PR #77, opened 2021-08-19
fix(openapi): return pool spec and state and simplify the spec

tiagolobocastro:
fix(openapi): simplify the responses

Use 4XX and 5XX for all error responses which simplifies the spec quite a bit.
We can specialize the errors later, if we want to...

Improve some of existing schemas to point to the refs and this way we get the
examples added as a bonus too

fix(openapi): no content should be returned as 204

fix(openapi): return pool spec and state

The openapi pool object now has a spec and a state. It can then represent both
control plane created pools and "unmanaged" pools created under it.


paulyoong (approved):

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

PR #81, opened 2021-08-24
feat(CI): build and push control plane images

paulyoong:
Enable CI to build and push control plane images for certain branches.

Build core and jsongrpc images separately to allow both to be loaded
from the resulting tar file. Attempting to build them both together was
resulting in only the core image being loaded from the tar file.

Provide deployment YAML files for the core agent and REST server. The
REST server is exposed through a NodePort service on port 30010.

Bug fix:

- Allow the etcd enpoint to be provided to the core agent without
the need for a port to be specified. Omitting the port results in the
default port being chosen (2379). This mimicks the behaviour of
Mayastor.

Resolves: CAS-636

### DIFF
--- a/control-plane/agents/core/src/core/registry.rs
+++ b/control-plane/agents/core/src/core/registry.rs
@@ -78,9 +78,11 @@ impl Registry {
         reconcile_period: std::time::Duration,
         reconcile_idle_period: std::time::Duration,
     ) -> Self {
-        let store = Etcd::new(&store_url)
+        let store_endpoint = Self::format_store_endpoint(&store_url);
+        let store = Etcd::new(&store_endpoint)
             .await
             .expect("Should connect to the persistent store");
+        tracing::info!("Connected to persistent store at {}", store_endpoint);
         let registry = Self {
             inner: Arc::new(RegistryInner {
                 nodes: Default::default(),
@@ -97,6 +99,15 @@ impl Registry {
         registry.init().await;
         registry
     }
+
+    /// Formats the store endpoint with a default port if one isn't supplied.
+    fn format_store_endpoint(endpoint: &str) -> String {
+        match endpoint.contains(':') {
+            true => endpoint.to_string(),
+            false => format!("{}:{}", endpoint, "2379"),
+        }
+    }
+
     /// Get the `CoreRegistryConfig` from etcd, if it exists, or use the default
     async fn get_config_or_panic<S: Store>(mut store: S) -> CoreRegistryConfig {
         let config = CoreRegistryConfig::new(NodeRegistration::Automatic);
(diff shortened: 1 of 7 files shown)


scripts/release.sh
```diff
@@ -119,14 +119,14 @@ cd $SCRIPTDIR/..
 
 if [ -z "$IMAGES" ]; then
   if [ -z "$DEBUG" ]; then
-    IMAGES="agents rest"
+    IMAGES="core jsongrpc rest"
   else
-    IMAGES="agents-dev rest-dev"
+    IMAGES="core-dev jsongrpc-dev rest-dev"
```
tiagolobocastro, line 124: btw you can use instead  `agents-dev.core` and `agents-dev.jsongrpc` and leave the nix file as is.

paulyoong: Thanks, that's good to know. I'll leave it as it is for now because I'll need to fix up the release script to work with the changed names. I'll raise a work item for this and circle back to it.

deploy/core-agents-deployment.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: core-agents
```
gila, line 4: agentS or agent?

control-plane/agents/core/src/core/registry.rs
```diff
@@ -97,6 +99,16 @@ impl Registry {
         registry.init().await;
         registry
     }
+
+    /// Adds the default port to the endpoint if one isn't already specified.
+    fn format_store_endpoint(endpoint: &str) -> String {
```
gila, line 104: Can we use the more common approaches? For example std::net::SocketAddrV4? That makes it in the end a whole lot easier. 

```
use std::net::{SocketAddrV4, Ipv4Addr};
let socket = SocketAddrV4::new(Ipv4Addr::new(127, 0, 0, 1), 8080);
```

https://doc.rust-lang.org/std/net/struct.SocketAddrV4.html

gila (approved):

deploy/core-agents-deployment.yaml
paulyoong: I used the plural because, although we only have a single container at the moment, as we create additional agents we could add additional containers to the deployment.

control-plane/agents/core/src/core/registry.rs
paulyoong: Thanks for the suggestion. I have changed this function to construct and return a SocketAddrV4 type.

paulyoong:
bors try

paulyoong:
bors try

control-plane/agents/core/src/core/registry.rs
paulyoong: The changes that I made for this ended up breaking the bdd & rust tests, so I'll raise a separate work item for this.

paulyoong:
bors merge

PR #84, opened 2021-09-09
feat(msp): add k8s operator for msp resources

gila:
This adds the MSP operator to the control plane including the
required changes to nix and deploy.

### DIFF
--- a/control-plane/msp-operator/Cargo.toml
+++ b/control-plane/msp-operator/Cargo.toml
@@ -1,0 +1,37 @@
+[package]
+name = "msp-operator"
+version = "0.1.0"
+edition = "2018"
+authors = ["Jeffry Molanus <jeffry.molanus@gmail.com>"]
+
+[dependencies]
+anyhow = "1.0.43"
+bytes = "1.1.0"
+chrono = "0.4.19"
+clap =  { version = "2.33.3", features = ["color"] }
+futures = "0.3.17"
+k8s-openapi = { version = "0.13.0", default-features = false, features = ["v1_20"] }
+kube = { version = "0.60.0", features = ["derive" ] }
+kube-runtime = "0.60.0"
+log = "0.4.14"
+prost = "0.8.0"
+prost-derive = "0.8.0"
+reqwest = { version = "0.11.4", features = ["json"] }
+schemars = "0.8.3"
+serde = "1.0.130"
+serde_json = "1.0.67"
+serde_yaml = "0.8.20"
+snafu = "0.6.10"
+tokio = { version = "1.11.0", features = ["full"] }
+tonic = "0.5.2"
+tracing = "0.1.26"
+tracing-subscriber = "0.2.20"
+url = "2.2.2"
+openapi = { path = "../../openapi"}
+
+# Tracing
+opentelemetry-jaeger = { version = "0.15.0", features = ["rt-tokio-current-thread"] }
+tracing-opentelemetry = "0.15.0"
+opentelemetry = { version = "0.16.0", features = ["rt-tokio-current-thread"] }
+[build-dependencies]
+tonic-build = { version = "0.5.2", features = ["prost" ] }
--- a/control-plane/msp-operator/src/main.rs
+++ b/control-plane/msp-operator/src/main.rs
@@ -1,0 +1,881 @@
+use chrono::Utc;
+/// Mayastor pool operator wachtes for pool CRDs and creates the pool on
+/// the given node. There is a maximum retry limit that will put the pool
+/// into a steady error state.
+///
+///
+/// Succesfully created pools are recreated by the control plane.
+use clap::{App, Arg, ArgMatches};
+use futures::StreamExt;
+use k8s_openapi::{
+    api::core::v1::{Event as k8Event, ObjectReference},
+    apimachinery::pkg::apis::meta::v1::MicroTime,
+};
+use kube::{
+    api::{Api, ListParams, ObjectMeta, Patch, PatchParams, PostParams},
+    Client, CustomResource, ResourceExt,
+};
+use kube_runtime::{
+    controller::{Context, Controller, ReconcilerAction},
+    finalizer::{finalizer, Event},
+};
+use openapi::models::{BlockDevice, Pool};
+use reqwest::RequestBuilder;
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+use serde_json::json;
+use snafu::Snafu;
+use std::{collections::HashMap, ops::Deref, sync::Arc, time::Duration};
+use tracing::{debug, error, info, trace, warn};
+use tracing_subscriber::{layer::SubscriberExt, util::SubscriberInitExt, EnvFilter, Registry};
+use url::Url;
+const WHO_AM_I: &str = "Mayastor pool operator";
+
+/// Our for Pool spec
+#[derive(CustomResource, Serialize, Deserialize, Default, Debug, PartialEq, Clone, JsonSchema)]
+#[kube(
+    group = "openebs.io",
+    version = "v1alpha1",
+    kind = "MayastorPool",
+    plural = "mayastorpools",
+    // The name of the struct that gets created that represents a resource
+    namespaced,
+    status = "MayastorPoolStatus",
+    derive = "PartialEq",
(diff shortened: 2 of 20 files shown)


deploy/msp-deployment.yaml
```diff
@@ -0,0 +1,38 @@
...
+              memory: "500Mi"
+            limits:
+              cpu: "1000m"
+              memory: "1Gi"
+          image: mayadata/mcp-msp-operator:aba4dd5
+          imagePullPolicy: Always
+          args:
+            - "-e http://$(REST_SERVICE_HOST):8081"
+            #imagePullSecrets:
+            #- name: regcred
```
paulyoong, line 33: Can this be removed? It is included further down.

control-plane/msp-operator/src/main.rs
```diff
@@ -0,0 +1,910 @@
...
+    #[snafu(display("Failed to connect to node: {}", source))]
+    /// gRPC transport error
+    RpcError {
+        source: tonic::transport::Error,
+    },
+    #[snafu(display("gRPC call failed {}", source))]
+    /// gRCP error status
+    RpcStatus {
+        source: tonic::Status,
+    },
```
paulyoong, line 146: Can we remove the RPC errors as we are no longer making RPC calls?

control-plane/msp-operator/src/main.rs
```diff
@@ -0,0 +1,910 @@
...
+impl From<PoolState> for String {
+    fn from(p: PoolState) -> Self {
+        p.to_string()
+    }
+}
+
+#[non_exhaustive]
+enum UrlPath {
+    /// GET for devices
+    BockDevices,
```
paulyoong, line 213: Typo: BlockDevices

control-plane/msp-operator/src/main.rs
```diff
@@ -0,0 +1,910 @@
...
+        // we only do puts for to one specific path
+        if matches!(n, UrlPath::Pool(_)) {
+            let url = self.as_url(n)?;
+            return Ok(self.ctx.http.put(url));
+        }
+
+        Err(Error::Noun {})
+    }
+
+    /// helper for setting the path of the URL we want to PUT operaiton on.
```
paulyoong, line 381: This is a DELETE operation.

control-plane/msp-operator/src/main.rs
```diff
@@ -0,0 +1,910 @@
...
+    }
+
+    /// helper function to set the path for the URL we want to perform the GET
+    /// operation on
+    fn get(&self, n: UrlPath) -> Result<RequestBuilder, Error> {
+        let url = self.as_url(n)?;
+        Ok(self.ctx.http.get(url))
+    }
+
+    /// helper for setting the path of the URL we want to PUT operaiton on.
```
paulyoong, line 370: Typo: operation

control-plane/msp-operator/src/main.rs
```diff
@@ -0,0 +1,910 @@
...
+        let _ = self.patch_status(MayastorPoolStatus::error()).await?;
+
+        error!(name = ?self.name(),"status set to error");
+        Ok(ReconcilerAction {
+            requeue_after: None,
+        })
+    }
+
+    /// patch the resource state to missing
+    async fn is_missing(&self) -> Result<ReconcilerAction, Error> {
```
paulyoong, line 433: Should this function be called something else because it isn't determining whether the resource is missing?

control-plane/msp-operator/src/main.rs
```diff
@@ -0,0 +1,910 @@
...
+        self.patch_status(MayastorPoolStatus::default()).await?;
```
paulyoong, line 434: We are using the default setting which will set it to "Creating" rather than "Missing".

paulyoong (approved):

control-plane/msp-operator/src/main.rs
```diff
@@ -0,0 +1,910 @@
...
+    let namespace = args.value_of("namespace").unwrap();
+    let msp: Api<MayastorPool> = Api::namespaced(k8s.clone(), namespace);
+    let lp = ListParams::default();
+
+    let context = Context::new(OperatorContext {
+        k8s,
+        inventory: tokio::sync::RwLock::new(HashMap::new()),
+        url: Url::parse(args.value_of("endpoint").unwrap()).expect("endpoint is not a valid URL"),
+        http: reqwest::ClientBuilder::new()
+            .timeout(Duration::from_secs(5))
```
tiagolobocastro, line 841: is 5s enough for deleting a pool? This is also something that we might have to tweak on the core agent.

control-plane/msp-operator/src/main.rs
```diff
@@ -0,0 +1,910 @@
...
+        .expect("failed to init tracing filter");
+
+    let subscriber = Registry::default()
+        .with(filter)
+        .with(tracing_subscriber::fmt::layer().pretty());
+
+    if let Some(jaeger) = matches.value_of("jaeger") {
+        let tracer = opentelemetry_jaeger::new_pipeline()
+            .with_agent_endpoint(jaeger)
+            .with_service_name("core-agent")
```
tiagolobocastro, line 937: s/core-agent/msp-operator

tiagolobocastro (approved):

control-plane/msp-operator/src/main.rs
gila: There is no missing state but good catch!. I've updated the comment. By setting the pool back into the creating state -- it means "we are creating the pool and driving it to... ". As we don't do anything here, we just wait for it to come back online.

openapi/Cargo.toml
```diff
@@ -5,15 +5,15 @@ authors = ["OpenAPI Generator team and contributors"]
 edition = "2018"
 
 [dependencies]
-serde = "^1.0"
-serde_derive = "^1.0"
-serde_json = "^1.0"
-url = { version = "^2.2", features = ["serde"] }
-async-trait = "0.1.42"
+serde = "1.0.130"
```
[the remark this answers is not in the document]
gila: :facepalm: yes you are right. reverted.

control-plane/msp-operator/src/main.rs
```diff
@@ -0,0 +1,910 @@
...
+
+        Ok(ReconcilerAction {
+            requeue_after: None,
+        })
+    }
+
+    /// Online the pool which is no-op from the dataplane point of view. However
+    /// it does provide us feedback from the k8s side of things which is
+    /// usefull when trouble shooting.
+    #[tracing::instrument(fields(name = ?self.name(), status = ?self.status) skip(self))]
```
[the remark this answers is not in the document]
gila: well, ... it seems to work...

control-plane/msp-operator/src/main.rs
gila: We are not deleting the pool, but the spec. So I would assume, we don't have to wait? If we do, then for sure, yes, we need to increase it!

control-plane/msp-operator/src/main.rs
```diff
@@ -0,0 +1,910 @@
+use chrono::Utc;
+/// Mayastor pool operator wachtes for pool CRDs and creates the pool on
```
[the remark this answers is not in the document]
gila: I think that is reserved for top-level module docs in rust-docs.

control-plane/msp-operator/Cargo.toml
```diff
@@ -0,0 +1,37 @@
+[package]
+name = "msp-operator"
```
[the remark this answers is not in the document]
gila: Yes, thats a good idea. It is not unlikely that we will have more of these things one way or the other.  However, we need to make several changes still (mostly to deploy) and likely also to some of the nix images (add curl and jq for example) which we can do then in one shot. Clearly, the files are not ready for usage yet.

control-plane/msp-operator/src/main.rs
gila: Yeah... naming :gun:

control-plane/msp-operator/src/main.rs
tiagolobocastro: Yes we do have to wait as things stand, unless we make the API async or something along those lines.

PR #85, opened 2021-09-10
chore: remove ana bits from existing non-ana volumes

tiagolobocastro:
chore: remove "ANA bits" from the current volume

A non-ANA volume is only meant to have one active nexus target at a time.
ANA adds a lot more complexity than than just a vector of nexuses, so it's better
to completely ignore multi volume nexuses for now and add it back later as a new
type of volume.
This way we can also make sure that we don't break non-ANA volumes when we start
adding ANA volumes..

--

chore: update the deployer docs and add block devices

Updates the readme.md to match the latest.
Add new option to add a host block device to the mayastor containers.

### DIFF
--- a/common/src/types/v0/message_bus/volume.rs
+++ b/common/src/types/v0/message_bus/volume.rs
@@ -1,6 +1,6 @@
 use super::*;
 
-use crate::{types::v0::store::volume::VolumeSpec, IntoOption, IntoVec};
+use crate::{types::v0::store::volume::VolumeSpec, IntoOption};
 use serde::{Deserialize, Serialize};
 use std::{convert::TryFrom, fmt::Debug};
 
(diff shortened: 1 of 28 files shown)


gila (approved):

common/src/types/v0/message_bus/volume.rs
```diff
@@ -119,7 +117,7 @@ impl From<(&VolumeId, &Nexus)> for VolumeState {
             size: nexus.size,
             status: nexus.status.clone(),
             protocol: nexus.share.clone(),
-            children: vec![nexus.clone()],
+            child: Some(nexus.clone()),
```
paulyoong, line 120: The comment on line 109 can now change so that it no longer refers to ANA.

paulyoong (approved):

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

PR #89, opened 2021-09-12
feat(kubectl-mayastor): consume REST endpoints using kubectl-plugin

Abhinandan-Purkait:
### What does this PR do?
- This PR adds consumption of the REST into user friendly(tabular), json and yaml ouputs.
- This can run as a binary in path and can consume the REST endpoints.


kubectl-plugin/src/resources/pool.rs
```diff
@@ -29,18 +50,41 @@ impl List for Pools {
...
+                    // Show the YAML form output if output format is YAML.
+                    let s = serde_yaml::to_string(&pool).unwrap();
+                    println!("{}", s);
+                }
+                utils::JSON_FORMAT => {
+                    // Show the JSON form output if output format is JSON.
+                    let s = serde_json::to_string(&pool).unwrap();
+                    println!("{}", s);
+                }
+                _ => {
```
mtzaurus, line 80: We could implement just one top-level check against provided output format and just use 100% correct value underneath.
A new enum can be introduced, which you can instantiate via 'try_from()' and then use directly in all your functions.

kubectl-plugin/README.md
```diff
@@ -18,22 +18,79 @@ The general command structure is `kubectl mayastor <operation> <resource>` where
...
-### Examples
 
-#### Volumes
-Getting all volumes:\
-`kubectl mayastor get volumes`
+### Examples and Outputs
 
-Getting a specific volume:\
-`kubectl mayastor get volume ec4e66fd-3b33-4439-b504-d49aba53da26`
+1. List Volumes
```
paulyoong, line 24: I think we want to change `List` to `Get` because the operation is now a `Get` operation.

kubectl-plugin/README.md
```diff
@@ -18,22 +18,79 @@ The general command structure is `kubectl mayastor <operation> <resource>` where
...
+``` 
+2. Get Volume by ID
+```
+❯ kubectl mayastor get volume 18e30e83-b106-4e0d-9fb6-2b04e761e18a
+ ID                                    PATHS  REPLICAS  PROTOCOL  STATUS  SIZE 
+ 18e30e83-b106-4e0d-9fb6-2b04e761e18a  1      4         none      Online  10485761 
 
-#### Pools
+```
+3. List Pools
```
paulyoong, line 39: I think we want to change `List` to `Get` because the operation is now a `Get` operation.

kubectl-plugin/README.md
```diff
@@ -18,22 +18,79 @@ The general command structure is `kubectl mayastor <operation> <resource>` where
...
+```
+5. Scale Volume by ID
+```
+❯ kubectl mayastor scale volume 0c08667c-8b59-4d11-9192-b54e27e0ce0f 5
+Volume 0c08667c-8b59-4d11-9192-b54e27e0ce0f Scaled Successfully 🚀
 
-Getting all pools:\
-`kubectl mayastor get pools`
+```
+6. List/Get Volume(s)/Pool(s) to a specific Output Format
```
paulyoong, line 59: `List` can be removed.

kubectl-plugin/src/resources/pool.rs
```diff
@@ -1,26 +1,47 @@
...
+                    // Incase of invalid output format, show error and gracefully end.
+                    println!("Output not supported for {} format", output);
+                }
+            },
+            Err(e) => {
+                println!("Failed to list pools. Error {}", e)
+            }
```
paulyoong, line 43: Suggestion: This pattern is repeated for both volumes and pools. A generic function could be implemented for this instead. I guess in this case you would require a "create_rows" trait that would be implemented by both volumes and pools.

kubectl-plugin/README.md
Abhinandan-Purkait: From the word `List` here I meant the operation and not the command, I have mentioned the command along with the output. I will change it to Get to avoid confusions.

kubectl-plugin/src/resources/pool.rs
Abhinandan-Purkait: Added a root check for the same. This would prevent an API call in case an output format is wrong.

kubectl-plugin/src/main.rs
```diff
@@ -24,6 +29,9 @@ struct CliArgs {
     /// The operation to be performed.
     #[structopt(subcommand)]
     operations: Operations,
+    /// Output Format
+    #[structopt(default_value = "", short = "o", long = "output")]
```
tiagolobocastro, line 33: Should option be an enum here?
ni: short and long seem to have their default values, you you can just leave them as short, long

kubectl-plugin/src/operations.rs
```diff
@@ -16,21 +15,24 @@ pub(crate) enum Operations {
 /// To be implemented by resources which support the 'list' operation.
 #[async_trait(?Send)]
 pub trait List {
-    async fn list() -> Result<()>;
+    type Format;
+    async fn list(output: &Self::Format);
```
tiagolobocastro, line 19: Why is the format a type specifier in these traits, we seem to always just set it to resources::OutputFormat? So could the parameter simply be: `output: &OutputFormat`?

kubectl-plugin/src/resources/mod.rs
```diff
@@ -1,11 +1,13 @@
 pub mod pool;
+pub mod utils;
 pub mod volume;
 
 use structopt::StructOpt;
 
 pub(crate) type VolumeId = String;
 pub(crate) type ReplicaCount = u8;
 pub(crate) type PoolId = String;
+pub(crate) type OutputFormat = String;
```
tiagolobocastro, line 10: OutputFormat could be defined in the utils as an enum OutputFormat { Wide,  Yaml, Json }, or something along those lines.

gila (commented):
Modules the comments from the others @Abhinandan-Purkait -- very nice work :+1:

kubectl-plugin/src/main.rs
Abhinandan-Purkait: Done. Thanks

kubectl-plugin/src/operations.rs
Abhinandan-Purkait: Yes, We could have just used the same. Done, Thanks.

kubectl-plugin/src/resources/utils.rs
```diff
@@ -32,48 +32,18 @@ pub fn table_printer(titles: Row, rows: Vec<Row>) {
...
-//     }
-//     rows
-// }
-
+// CreateRows trait to be implemented by Vec<Volume/Pool> to create the rows.
 pub trait CreateRows {
     fn create_rows(&self) -> Vec<Row>;
 }
 
+// CreateRows trait to be implemented by Volume/Pool to fetch the corresponding headers.
```
paulyoong, line 40: This is the "GetHeaderRow" trait.

paulyoong (approved):

kubectl-plugin/src/resources/utils.rs
Abhinandan-Purkait: Sorry, missed that.

Abhinandan-Purkait: Done.

kubectl-plugin/src/resources/mod.rs
Abhinandan-Purkait: OutputFormat has been defined as enum. The StructOpts args is still a string, as we can check the correctness and use it as an enum in the underlying code. Thank you.

kubectl-plugin/src/resources/pool.rs
Abhinandan-Purkait: Added a few more changes viz, GetHeaderRow, cleaned some code, added comments on top of your changes. Thank you.

kubectl-plugin/src/main.rs
```diff
@@ -34,24 +42,30 @@ async fn main() {
...
-
-    if let Err(e) = result {
-        println!("{}", e);
+    // Check whether the entered format is valid or not
+    if !&cli_args.output.is_empty()
+        && cli_args.output != utils::YAML_FORMAT
+        && cli_args.output != utils::JSON_FORMAT
+    {
+        // Incase of invalid output format, show error and gracefully end.
+        println!("Output not supported for {} format", &cli_args.output);
```
tiagolobocastro, line 51: You can avoid having to do this by making output a OutputFormat in the structOpt arguments btw.

kubectl-plugin/src/main.rs
```diff
@@ -24,6 +29,9 @@ struct CliArgs {
     /// The operation to be performed.
     #[structopt(subcommand)]
     operations: Operations,
+    /// Output Format
+    #[structopt(default_value = "", short, long)]
+    output: String,
```
tiagolobocastro, line 34: what do we get with --help? Do we need to add more comments above with the possible values?

kubectl-plugin/src/main.rs
Abhinandan-Purkait: Tried that, but it's not working that way. It needs fmt::Display, to_String and FromStr traits to be implemented. So, this way seemed straightforward.

kubectl-plugin/src/main.rs
Abhinandan-Purkait:
```
kubectl-mayastor 0.1.0
The types of operations that are supported

USAGE:
    kubectl-mayastor [OPTIONS] <SUBCOMMAND>

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information

OPTIONS:
    -o, --output <output>    Output Format [default: ]
    -r, --rest <rest>        Rest endpoint

SUBCOMMANDS:
    get      'Get' resources
    help     Prints this message or the help of the given subcommand(s)
    scale    'Scale' resources
```

Currently, it shows something like this. I would add the supported formats here as well.

tiagolobocastro: So, the default shows up as "empty". This is where adding the enum would perhaps make it clearer as it'd show the default value of the enum.
For the possible values you can actually add the `possible_values` option to the structopt argument.

Abhinandan-Purkait: Got your point. I was not aware of these functionalities of `StructOpts`. Changed output to `enum`. Added `possible_values`. Removed those redundant checks. The `--help` looks like this now. Thanks.

```
[nix-shell:~/github/Abhinandan-Purkait/mayastor-control-plane]$ ./target/debug/kubectl-mayastor --help                                                    
kubectl-mayastor 0.1.0
The types of operations that are supported

USAGE:
    kubectl-mayastor [OPTIONS] <SUBCOMMAND>

FLAGS:
    -h, --help       Prints help information
    -V, --version    Prints version information

OPTIONS:
    -o, --output <output>    The Output, viz yaml, json [default: none]  [possible values: yaml, json, none]
    -r, --rest <rest>        The rest endpoint, parsed from KUBECONFIG, if left empty 

SUBCOMMANDS:
    get      'Get' resources
    help     Prints this message or the help of the given subcommand(s)
    scale    'Scale' resources
```

kubectl-plugin/src/resources/utils.rs
```diff
@@ -0,0 +1,87 @@
...
+
+// OutputFormat to be used as an enum to match the output from args.
+#[derive(Debug)]
+pub enum OutputFormat {
+    Yaml,
+    Json,
+    NoFormat,
+}
+
+impl From<&str> for OutputFormat {
```
tiagolobocastro, line 53: btw you can implement TryFrom instead of From, if you want to use only "none" for NoFormat and make other things be an error

kubectl-plugin/src/main.rs
```diff
@@ -18,12 +23,15 @@ use yaml_rust::YamlLoader;
...
 #[derive(StructOpt, Debug)]
 struct CliArgs {
-    /// Rest endpoint.
+    /// The rest endpoint, parsed from KUBECONFIG, if left empty .
     #[structopt(long, short)]
     rest: Option<Url>,
     /// The operation to be performed.
     #[structopt(subcommand)]
     operations: Operations,
+    /// The Output, viz yaml, json.
```
tiagolobocastro, line 32: what is viz?

tiagolobocastro (approved):
:rocket:

kubectl-plugin/src/main.rs
Abhinandan-Purkait: viz, is acronym for namely. For example kind of.

tiagolobocastro: had not come across that, thanks :+1:

paulyoong:
bors merge

paulyoong:
bors merge

PR #90, opened 2021-09-13
Msp fixes

gila:


paulyoong (approved):

control-plane/msp-operator/src/main.rs
```diff
@@ -43,6 +43,9 @@ const WHO_AM_I: &str = "Mayastor pool operator";
     derive = "PartialEq",
     derive = "Default",
     shortname = "msp",
+    printcolumn = r#"{ "name":"node", "type":"string", "description":"node the pool is on", "jsonPath":".spec.node"}"#,
+    printcolumn = r#"{ "name":"status", "type":"string", "description":"pool status", "jsonPath":".status.state"}"#,
+    printcolumn = r#"{ "name":"used", "type":"integer", "description":"used bytes", "jsonPath":".status.used"}"#
```
tiagolobocastro, line 48: could we also show the free space? Avoids us having to do mental math :)

gila: We can yes, however -- the requires us to extend the status field of the status. And since we have the plugin I'd argue it's better to have a follow-up PR that addresses "nice to haves later".

PR #91, opened 2021-09-13
feat: add missing nexus and invalid protocol reconciler

tiagolobocastro:
feat: add missing nexus and invalid protocol reconciler

Add reconciler to recreate nexuses after mayastor restarts.
Add reconciler to fixup nexus share protocols.

As things stand, the nexus reconciler does not attempt to fixup nexuses which are
owned by a volume. Instead the volume reconciler calls the nexus reconciler...
We may change this but for the moment it feels "safer" to have the nexus reconciler
not touch nexus that belong to a volume directly.

----------
chore: add tmpfs disk api to the test cluster

Add with_tmpfs_pool to the cluster test library which allows to add tmpfs disk images
to each mayastor container, where pools can be created.
These images get automatically deleted on drop.

----------
test: test the volume nexus reconciler

Test that a nexus is recreated when missing and that the share protocol is correct.

--------
 chore: remove the old rest client

The "old" rest client is not needed now that we have the auto-generated openapi client.
We still keep the some parts of it as it facilitates the creation of the openapi client.

With this we remove a lot of From<models::*> To BUS that are not needed anymore.


tiagolobocastro:
bors try

tiagolobocastro:
bors try

control-plane/agents/core/src/nexus/scheduling.rs
```diff
@@ -0,0 +1,47 @@
...
+    registry: &Registry,
+) -> Result<HealthyChildItems, SvcError> {
+    let children = get_healthy_nexus_children(
+        &GetPersistedNexusChildren::new_recreate(nexus_spec),
+        registry,
+    )
+    .await?;
+
+    nexus_spec.trace(&format!(
+        "Healthy volume nexus replicas for volume: {:?}",
```
paulyoong, line 42: Could we simplify this message to just say "Healthy replicas for volume..."?

control-plane/agents/core/src/volume/registry.rs
```diff
@@ -35,13 +35,14 @@ impl Registry {
                 size: nexus_state.size,
                 status: match nexus_state.status {
                     NexusStatus::Online
-                        if nexus_state.children.len() != volume_spec.num_replicas as usize =>
+                        if nexus_state.children.len() != volume_spec.num_replicas as usize
+                            || nexus_state.share != volume_spec.protocol =>
```
paulyoong, line 39: Should the volume status be degraded just because the share protocol is different? The volume status is usually an indication of the health of the data (i.e. the number of replicas) but this would seem to conflate the meaning.

paulyoong (approved):

control-plane/agents/core/src/volume/tests.rs
```diff
@@ -83,6 +89,112 @@ async fn hotspare() {
...
+    let timeout = Duration::from_secs(RECONCILE_TIMEOUT_SECS);
+    let client = cluster.rest_v00();
+    let nexuses_api = client.nexuses_api();
+    let start = std::time::Instant::now();
+    loop {
+        let nexus = nexuses_api.get_nexus(nexus_id.to_string().as_str()).await;
+
+        match nexuses_api.get_nexus(nexus_id.to_string().as_str()).await {
+            Ok(nexus) => {
+                if let Some(state) = state {
```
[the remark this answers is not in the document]
tiagolobocastro: we check the whole content of the nexus when the function returns so everything must match, including the node.

control-plane/agents/core/src/core/reconciler/volume/nexus.rs
```diff
@@ -0,0 +1,78 @@
...
+};
+
+use parking_lot::Mutex;
+use std::sync::Arc;
+
+/// Volume nexus reconciler
+/// When mayastor instances restart they come up "empty" and so we need to recreate
+/// any previously created nexuses
+#[derive(Debug)]
+pub(super) struct NexusReconciler {}
```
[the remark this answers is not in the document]
tiagolobocastro: Yes indeed

control-plane/agents/core/src/volume/registry.rs
tiagolobocastro: I've removed this change.

tiagolobocastro:
bors merge

tiagolobocastro:
bors cancel

tiagolobocastro:
bors merge

PR #92, opened 2021-09-15
fix: use correct key when deleting from store

paulyoong:
Ensure the correct key is used when attempting to delete an entry from
the persistent store.

Add a BDD test to check that when a gRPC timeout occurs, associated
specs are correctly deleted from the persistent store.

Resolves: CAS-1103

### DIFF
--- a/tests/bdd/test_volume_create.py
+++ b/tests/bdd/test_volume_create.py
@@ -53,7 +53,12 @@ def create_request():
 @scenario(
     "features/volume/create.feature", "provisioning failure due to missing Mayastor"
 )
-def test_provisioning_failure():
+def test_provisioning_failure_due_to_missing_mayastor():
+    """provisioning failure."""
+
+
+@scenario("features/volume/create.feature", "provisioning failure due to gRPC timeout")
+def test_provisioning_failure_due_to_grpc_timeout():
     """provisioning failure."""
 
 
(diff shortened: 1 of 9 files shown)


paulyoong:
bors merge

tests/bdd/test_volume_create.py
```diff
@@ -137,6 +159,44 @@ def there_are_no_available_mayastor_instances():
...
+
+@then("volume creation should fail")
+def volume_creation_should_fail(create_request):
+    """volume creation should fail."""
+    request = create_request[CREATE_REQUEST_KEY]
+    try:
+        common.get_volumes_api().put_volume(VOLUME_UUID, request)
+    except Exception as e:
+        exception_info = e.__dict__
+        assert exception_info["status"] == requests.codes["precondition_failed"]
```
mtzaurus, line 193: Is 'precondition failed' here a strong indication of the timeout ?
Can we 100% be sure that we have experienced the operation timeout rather than a general failure due to some pre-conditions, like: we have no available mayastor nodes so the request completes immediately, hence no timeout in general.
Shall we measure time somehow in the test to detect the duration of the operation or so ?

paulyoong: I think for the time being this is the best we can do. The alternative is to check the exact error string but this isn't ideal because if the error message changes slightly all the tests would fall apart.

In this case this is a strong indication of the timeout because the previous step ensures that the mayastor instances are deleted and therefore cannot be reached.

paulyoong:
bors merge

PR #95, opened 2021-09-15
fix: improve how the control plane handles various timeouts

tiagolobocastro:
feat: add request specific minimum timeouts

Specific requests can now have a minimum timeout set. This can be disabled via
the cmdline.
The RestServer gets 1 extra second added to allow the core agent to timeout first.

------
chore(scripts): build the binaries and log alias tag

Build the workspace binaries on the bdd test script.
Add a log when pushing the containers with the branch alias.

------
feat: consider a node online even if the watchdog expired

If there is some kind of temporary nats issue the heartbeat will fail and the node
will be set as offline.
This means that offline may not always mean offline and we can't use this state
to avoid issuing client requests.

Now, when the HB expires we issue a liveness probe to assess is the node is online
or not, and if it is then rearm the watchdog and keep the node online.

------
fix: don't issue grpc requests if the node is not online

------
chore: handle different subscriptions concurrently

Previously, only different subscription channels were handled concurrently.
This meant that a GetPools and CreatePool would run sequentially.
This changes limits this to a specific subscription so GetPools and CreatePool
can run concurrently, but CreatePool runs sequentially.


tiagolobocastro:
bors try

gila (approved):

PR #96, opened 2021-09-16
fix: include kubectl-plugin in control plane build

paulyoong:
When building the control plane ensure that the kubectl plugin is also
built.

### DIFF
--- a/nix/pkgs/control-plane/cargo-project.nix
+++ b/nix/pkgs/control-plane/cargo-project.nix
@@ -39,6 +39,7 @@ let
       "composer"
       "control-plane"
       "deployer"
+      "kubectl-plugin"
       "openapi"
       "rpc"
       "tests"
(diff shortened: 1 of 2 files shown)


nix/pkgs/control-plane/cargo-project.nix
```diff
@@ -39,11 +39,12 @@ let
       "composer"
       "control-plane"
       "deployer"
+      "kubectl-plugin"
       "openapi"
       "rpc"
       "tests"
     ];
-    cargoBuildFlags = [ "-p rpc" "-p agents" "-p rest" "-p msp-operator" ];
+    cargoBuildFlags = [ "-p rpc" "-p agents" "-p rest" "-p msp-operator" "-p kubectl-mayastor" ];
```
tiagolobocastro, line 47: I don't think we need this

tiagolobocastro (approved):

nix/pkgs/control-plane/cargo-project.nix
paulyoong: You're right - I've removed it.

paulyoong:
bors merge

PR #97, opened 2021-09-16
fix(kubectl-mayastor): handle cases with empty state for volume and pools, add managed field

Abhinandan-Purkait:
### What does this PR do?
- This PR handles the cases when a pool is likely having its state as empty.
- This PR adds a field `MANAGED` to denote whether a pool is created by Control plane or not.

### DIFF
--- a/kubectl-plugin/src/resources/pool.rs
+++ b/kubectl-plugin/src/resources/pool.rs
@@ -28,8 +28,8 @@ impl CreateRows for Vec<openapi::models::Pool> {
             // The spec would be empty if it was not created using
             // control plane.
             let spec = pool.spec.clone().unwrap_or_default();
-            // In case the state is not coming as filled, either due to pool, node lost, fill in spec
-            // data and mark the status as Unknown.
+            // In case the state is not coming as filled, either due to pool, node lost, fill in
+            // spec data and mark the status as Unknown.
             let state = pool.state.clone().unwrap_or(openapi::models::PoolState {
                 capacity: 0,
                 disks: spec.disks,
--- a/kubectl-plugin/src/resources/pool.rs
+++ b/kubectl-plugin/src/resources/pool.rs
@@ -21,16 +21,32 @@ impl CreateRows for Vec<openapi::models::Pool> {
     fn create_rows(&self) -> Vec<Row> {
         let mut rows: Vec<Row> = Vec::new();
         for pool in self {
-            let state = pool.state.as_ref().unwrap();
-            // The disks are joined by a comma and shown in the table, ex /dev/vda, /dev/vdb
+            let mut managed = true;
+            if pool.spec.is_none() {
+                managed = false;
+            }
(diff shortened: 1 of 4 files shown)


gila (changes requested):
I think we can bump into the same problem in resource/volumes?

Abhinandan-Purkait:
> I think we can bump into the same problem in resource/volumes?

Added for the volume as well. Apparently, the VolumeSpec cannot be empty. Thus handled the case for VolumeState being empty.

paulyoong:
> > I think we can bump into the same problem in resource/volumes?
> 
> Added for the volume as well. Apparently, the VolumeSpec cannot be empty. Thus handled the case for VolumeState being empty.

Yes this is correct. The Volume is an abstract concept created by the control plane, therefore the Volume cannot be created beneath the control plane.

paulyoong (approved):

kubectl-plugin/src/resources/volume.rs
```diff
@@ -19,7 +19,17 @@ impl CreateRows for Vec<openapi::models::Volume> {
     fn create_rows(&self) -> Vec<Row> {
         let mut rows: Vec<Row> = Vec::new();
         for volume in self {
-            let state = volume.state.as_ref().unwrap();
+            let state = volume
+                .state
+                .clone()
+                // If the state comes as empty fill in the spec data and mark the status as Unknown.
+                .unwrap_or(openapi::models::VolumeState {
```
tiagolobocastro, line 26: @paulyoong as it stands looks like we always have a state for a volume too, maybe we should change the openapi spec to reflect this?

kubectl-plugin/src/resources/pool.rs
```diff
@@ -21,16 +21,32 @@ impl CreateRows for Vec<openapi::models::Pool> {
...
-            let state = pool.state.as_ref().unwrap();
-            // The disks are joined by a comma and shown in the table, ex /dev/vda, /dev/vdb
+            let mut managed = true;
+            if pool.spec.is_none() {
+                managed = false;
+            }
+            // The spec would be empty if it was not created using
+            // control plane.
+            let spec = pool.spec.clone().unwrap_or_default();
+            // Incase the state is not coming as filled, either due to pool, node lost, fill in spec
```
tiagolobocastro, line 31: s/Incase/In case

tiagolobocastro (approved):

kubectl-plugin/src/resources/volume.rs
paulyoong: If I remember correctly I think we copy things from the spec into the state if we don't have one and then return this.

kubectl-plugin/src/resources/pool.rs
Abhinandan-Purkait: Changed `Incase` --> `In case`

gila (approved):

paulyoong:
bors merge

PR #102, opened 2021-09-20
feat(csi): csi controller plugin

mtzaurus:
This commit introduces a new CSI Controller plugin, which uses Control Plane’s new REST API
and is deployed as a separate conainerised application.

Resolves: CAS-1065

### DIFF
--- a/control-plane/csi-controller/src/client.rs
+++ b/control-plane/csi-controller/src/client.rs
@@ -1,0 +1,377 @@
+use common_lib::types::v0::openapi::models::{
+    CreateVolumeBody, ExplicitTopology, Node, Pool, Topology, Volume, VolumePolicy,
+    VolumeShareProtocol,
+};
+
+use anyhow::{anyhow, Result};
+use once_cell::sync::OnceCell;
+use reqwest::{Client, Response, StatusCode, Url};
+use serde::{Deserialize, Serialize};
+use std::fmt::{Display, Formatter};
+use tracing::instrument;
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum ApiClientError {
+    // Error while communicating with the server.
+    ServerCommunication(String),
+    // Requested resource already exists. This error has a dedicated variant
+    // in order to handle resource idempotency properly.
+    ResourceAlreadyExists(String),
+    // No resource instance exists.
+    ResourceNotExists(String),
+    // Generic operation errors.
+    GenericOperation(String),
+    // Problems with parsing response body.
+    InvalidResponse(String),
+    /// URL is malformed.
+    MalformedUrl(String),
+}
+
+static REST_CLIENT: OnceCell<MayastorApiClient> = OnceCell::new();
+
+// REST API URI names for API objects.
+mod uri {
+    pub const VOLUMES: &str = "volumes";
+    pub const POOLS: &str = "pools";
+    pub const NODES: &str = "nodes";
+}
+
+/// Struct for representing URI.
+#[derive(Debug)]
+struct UrnType<'a>(&'a [&'a str]);
+
+impl UrnType<'_> {
+    /// Classifies URI as a tuple (resource type, resource id) based on URI.
+    pub fn classify(&self) -> (String, String) {
+        match self.0.len() {
+            0 | 1 => panic!("Resource URI must contain collection name and resource id"),
+            _ => {
+                let rtype = match self.0[0] {
+                    uri::VOLUMES => "volume",
+                    uri::POOLS => "pool",
+                    uri::NODES => "node",
+                    unknown => panic!("Unknown resource type: {}", unknown),
+                };
+
+                (rtype.to_string(), self.0[1].to_string())
+            }
+        }
+    }
+
+    /// Transform URI into a full URL based on the given base URL.
+    pub fn get_full_url(&self, base_url: &str) -> Result<Url, ApiClientError> {
+        let u = format!("{}/{}", base_url, self);
+        Url::parse(&u)
+            .map_err(|e| ApiClientError::MalformedUrl(format!("URL parsing error: {:?}", e)))
+    }
+}
+
+impl Display for UrnType<'_> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.0.join("/"))
+    }
+}
+
+/// Single instance API client for accessing REST API gateway.
+/// Encapsulates communication with REST API by exposing a set of
+/// high-level API functions, which perform (de)serialization
+/// of API request/response objects.
+#[derive(Debug)]
+pub struct MayastorApiClient {
+    base_url: String,
+    rest_client: Client,
+}
+
+impl MayastorApiClient {
+    /// Initialize API client instance. Must be called prior to
+    /// obtaining the client instance.
+    pub fn initialize(endpoint: String) -> Result<()> {
+        if REST_CLIENT.get().is_some() {
+            return Err(anyhow!("API client already initialized"));
+        }
+
+        // Make sure endpoint is a well-formed URL.
+        if let Err(u) = Url::parse(&endpoint) {
+            return Err(anyhow!("Invalid API endpoint URL {}: {:?}", endpoint, u));
+        }
+
+        let rest_client = reqwest::Client::builder()
+            .danger_accept_invalid_certs(true)
+            .build()
+            .expect("Failed to build REST client");
+
+        REST_CLIENT.get_or_init(|| Self {
+            base_url: format!("{}/v0", endpoint),
+            rest_client,
+        });
+
+        debug!("API client is initialized with endpoint {}", endpoint);
+        Ok(())
+    }
+
+    /// Obtain client instance. Panics if called before the client
+    /// has been initialized.
+    pub fn get_client() -> &'static MayastorApiClient {
+        REST_CLIENT.get().expect("Rest client is not initialized")
+    }
+}
+
+/// Generate a getter for a given collection URI.
+macro_rules! collection_getter {
+    ($name:ident, $t:ty, $urn:expr) => {
+        pub async fn $name(&self) -> Result<Vec<$t>, ApiClientError> {
+            self.get_collection::<$t>($urn).await
+        }
+    };
+}
+
+impl MayastorApiClient {
+    async fn get_collection_item<R>(&self, urn: UrnType<'_>) -> Result<R, ApiClientError>
+    where
+        for<'a> R: Deserialize<'a>,
+    {
+        let response = self.do_get(&urn).await?;
+
+        // Check HTTP status code.
+        match response.status() {
+            StatusCode::OK => {}
+            StatusCode::NOT_FOUND => {
+                let (rtype, rname) = urn.classify();
+                return Err(ApiClientError::ResourceNotExists(format!(
+                    "{} {} not found",
+                    rtype, rname
+                )));
+            }
+            http_status => {
+                return Err(ApiClientError::GenericOperation(format!(
+                    "Failed to GET {:?}, HTTP error = {}",
+                    urn, http_status,
+                )))
+            }
+        };
+
+        // Get response body if request succeeded.
+        let body = response.bytes().await.map_err(|e| {
+            ApiClientError::InvalidResponse(format!(
+                "Failed to obtain body from HTTP response while getting {}, error = {}",
+                urn, e,
+            ))
+        })?;
+
+        serde_json::from_slice::<R>(&body).map_err(|e| {
+            ApiClientError::InvalidResponse(format!(
+                "Failed to deserialize object {}, error = {}",
+                std::any::type_name::<R>(),
+                e
+            ))
+        })
+    }
+
+    // Get one resource instance.
+    async fn do_get(&self, urn: &UrnType<'_>) -> Result<Response, ApiClientError> {
+        self.rest_client
+            .get(urn.get_full_url(&self.base_url)?)
+            .send()
+            .await
+            .map_err(|e| {
+                ApiClientError::ServerCommunication(format!(
+                    "Failed to GET {:?}, error = {}",
+                    urn, e
+                ))
+            })
+    }
+
+    // Perform resource deletion, optionally idempotent.
+    async fn do_delete(&self, urn: &UrnType<'_>, idempotent: bool) -> Result<(), ApiClientError> {
+        let response = self
+            .rest_client
+            .delete(urn.get_full_url(&self.base_url)?)
+            .send()
+            .await
+            .map_err(|e| {
+                ApiClientError::ServerCommunication(format!(
+                    "DELETE {} request failed, error={}",
+                    urn, e
+                ))
+            })?;
+
+        // Check HTTP status code, handle DELETE idempotency transparently.
+        match response.status() {
+            StatusCode::OK => {
+                debug!("Resource {} successfully deleted", urn);
+                Ok(())
+            }
+            // Handle idempotency as requested by the caller.
+            StatusCode::NOT_FOUND | StatusCode::NO_CONTENT | StatusCode::PRECONDITION_FAILED => {
+                if idempotent {
+                    debug!("Resource {} successfully deleted", urn);
+                    Ok(())
+                } else {
+                    let (rtype, rname) = urn.classify();
+                    Err(ApiClientError::ResourceNotExists(format!(
+                        "{} {} not found",
+                        rtype, rname
+                    )))
+                }
+            }
+            code => Err(ApiClientError::GenericOperation(format!(
+                "DELETE {} failed, HTTP status code = {}",
+                urn, code
+            ))),
+        }
+    }
+
+    async fn do_put<I, O>(&self, urn: &UrnType<'_>, object: I) -> Result<O, ApiClientError>
+    where
+        I: Serialize + Sized,
+        for<'a> O: Deserialize<'a>,
+    {
+        let response = self
+            .rest_client
+            .put(urn.get_full_url(&self.base_url)?)
+            .json(&object)
+            .send()
+            .await
+            .map_err(|e| {
+                ApiClientError::ServerCommunication(format!(
+                    "PUT {} request failed, error={}",
+                    urn, e
+                ))
+            })?;
+
+        // Check HTTP status of the operation.
+        // TODO: Revisit status codes checks after improving REST API HTTP codes (CAS-1124).
+        match response.status() {
+            StatusCode::OK => {}
+            StatusCode::UNPROCESSABLE_ENTITY => {
+                return Err(ApiClientError::ResourceAlreadyExists(format!(
+                    "Resource {} already exists",
+                    urn
+                )));
+            }
+            _ => {
+                return Err(ApiClientError::GenericOperation(format!(
+                    "PUT {} failed, HTTP status = {}",
+                    urn,
+                    response.status()
+                )));
+            }
+        };
+
+        let body = response.bytes().await.map_err(|e| {
+            ApiClientError::InvalidResponse(format!(
+                "Failed to obtain body from HTTP PUT {} response, error = {}",
+                urn, e,
+            ))
+        })?;
+
+        serde_json::from_slice::<O>(&body).map_err(|e| {
+            ApiClientError::InvalidResponse(format!(
+                "Failed to deserialize object {}, error = {}",
+                std::any::type_name::<O>(),
+                e
+            ))
+        })
+    }
+
+    async fn get_collection<R>(&self, urn: UrnType<'_>) -> Result<Vec<R>, ApiClientError>
+    where
+        for<'a> R: Deserialize<'a>,
+    {
+        let body = self.do_get(&urn).await?.bytes().await.map_err(|e| {
+            ApiClientError::InvalidResponse(format!(
+                "Failed to obtain body from HTTP response while listing {:?}, error = {}",
+                urn, e,
+            ))
+        })?;
+
+        serde_json::from_slice::<Vec<R>>(&body).map_err(|e| {
+            ApiClientError::InvalidResponse(format!(
+                "Failed to deserialize objects {}, error = {}",
+                std::any::type_name::<R>(),
+                e
+            ))
+        })
+    }
+
+    // List all nodes available in Mayastor cluster.
+    collection_getter!(list_nodes, Node, UrnType(&[uri::NODES]));
+
+    // List all pools available in Mayastor cluster.
+    collection_getter!(list_pools, Pool, UrnType(&[uri::POOLS]));
+
+    // List all volumes available in Mayastor cluster.
+    collection_getter!(list_volumes, Volume, UrnType(&[uri::VOLUMES]));
+
+    // List pools available on target Mayastor node.
+    pub async fn get_node_pools(&self, node: &str) -> Result<Vec<Pool>, ApiClientError> {
+        self.get_collection(UrnType(&[uri::NODES, node, uri::POOLS]))
+            .await
+    }
+
+    #[instrument]
+    /// Create a volume of target size and provision storage resources for it.
+    /// This operation is not idempotent, so the caller is responsible for taking
+    /// all actions with regards to idempotency.
+    pub async fn create_volume(
+        &self,
+        volume_id: &str,
+        replicas: u8,
+        size: u64,
+        allowed_nodes: &[String],
+        preferred_nodes: &[String],
+    ) -> Result<Volume, ApiClientError> {
+        let topology = Topology::explicit(ExplicitTopology::new(
+            allowed_nodes.to_vec(),
+            preferred_nodes.to_vec(),
+        ));
+
+        let req = CreateVolumeBody {
+            replicas,
+            size,
+            topology: Some(topology),
+            policy: VolumePolicy::default(),
+            labels: None,
+        };
+
+        self.do_put(&UrnType(&[uri::VOLUMES, volume_id]), &req)
+            .await
+    }
+
+    #[instrument]
+    /// Delete volume and reclaim all storage resources associated with it.
+    /// This operation is idempotent, so the caller does not see errors indicating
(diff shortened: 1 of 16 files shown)


control-plane/csi-controller/src/client.rs
```diff
@@ -0,0 +1,373 @@
...
+                std::any::type_name::<R>(),
+                e
+            ))
+        })
+    }
+
+    // Get one resource instance.
+    async fn do_get(&self, urn: &UrnType<'_>) -> Result<Response, Error> {
+        let u = format!("{}/{}", self.base_url, urn);
+        let uri = Url::parse(&u).unwrap();
```
tiagolobocastro, line 164: Should this `unwrap` be a `?` ?
(same for other similar usages)

control-plane/csi-controller/src/client.rs
```diff
@@ -0,0 +1,373 @@
...
+                ApiClientError::ServerCommunicationError(format!(
+                    "PUT {} request failed, error={}",
+                    u, e
+                ))
+            })?;
+
+        // Check HTTP status of the operation.
+        match response.status() {
+            StatusCode::OK => {}
+            StatusCode::UNPROCESSABLE_ENTITY => {
```
tiagolobocastro, line 229: Did you end up creating a ticket to fix these error codes? If so could you please add a todo: CAS-XXX where it applies so that we know what to change when we fix the ticket?

control-plane/csi-controller/src/controller.rs
```diff
@@ -0,0 +1,726 @@
...
+
+        Ok(Self { nodes })
+    }
+
+    // Determine the list of nodes where the workload can be placed.
+    // If volume is created as pinned (i.e. local=true), then the nexus and the workload
+    // must be placed on the same node, which in fact means running workloads only on Mayastor
+    // daemonset nodes.
+    // For non-pinned volumes, workload can be put on any node in the Kubernetes cluster.
+    pub fn volume_accessible_topology(&self, pinned_volume: bool) -> Vec<CsiTopology> {
```
tiagolobocastro, line 177: Hmm not sure if we have any tests for this, @paulyoong ?

control-plane/csi-controller/src/controller.rs
```diff
@@ -0,0 +1,726 @@
...
+                }
+                Err(_) => return Err(Status::invalid_argument("Invalid replica count")),
+            },
+            None => 1,
+        };
+
+        let pinned_volume =
+            volume_opts::decode_local_volume_flag(args.parameters.get(volume_opts::LOCAL_VOLUME));
+
+        // For exaplanation of accessibilityRequirements refer to a table at
```
tiagolobocastro, line 283: s/exaplanation/explanation

control-plane/csi-controller/src/controller.rs
```diff
@@ -0,0 +1,726 @@
...
+                )
+                .await?;
+
+            debug!(
+                "Volume {} successfully created, pinned volume = {}",
+                volume_uuid, pinned_volume
+            );
+        }
+
+        let vt_mapper = VolumeTopologyMapper::init().await?;
```
tiagolobocastro, line 353: Should this be called before creating the volume? Otherwise if it fails we may "leak" the volume?

control-plane/csi-controller/src/controller.rs
```diff
@@ -0,0 +1,726 @@
...
+            .delete_volume(&args.volume_id)
+            .await
+            .map_err(|e| {
+                Status::internal(format!(
+                    "Failed to delete volume {}, error = {:?}",
+                    args.volume_id, e
+                ))
+            })?;
+
+        Ok(Response::new(DeleteVolumeResponse {}))
```
tiagolobocastro, line 387: do we need another debug! "deleted volume" ?

control-plane/csi-controller/src/controller.rs
```diff
@@ -0,0 +1,726 @@
...
+        // Check if the volume is already published.
+        let volume = MayastorApiClient::get_client()
+            .get_volume(&volume_id)
+            .await?;
+        let uri = if let Some(state) = volume.state.as_ref() {
+            let curr_proto = state.protocol.to_string();
+
+            // Volume is aready published, make sure the protocol matches and get URI.
+            if curr_proto != "none" {
+                if curr_proto != *args.volume_context.get("protocol").unwrap().to_string() {
```
tiagolobocastro, line 441: we might be able to use 'protocol' which we've constructed on 403?

control-plane/csi-controller/src/controller.rs
```diff
@@ -0,0 +1,726 @@
...
+
+        // Check if the volume is already published.
+        let volume = MayastorApiClient::get_client()
+            .get_volume(&volume_id)
+            .await?;
+        let uri = if let Some(state) = volume.state.as_ref() {
+            let curr_proto = state.protocol.to_string();
+
+            // Volume is aready published, make sure the protocol matches and get URI.
+            if curr_proto != "none" {
```
tiagolobocastro, line 440: I wonder if we should check for `volume.child` here, which is what signifies that the volume is published. The protocol should actually be part of that.

control-plane/csi-controller/src/controller.rs
```diff
@@ -0,0 +1,726 @@
...
+                if curr_proto != *args.volume_context.get("protocol").unwrap().to_string() {
+                    let m = format!(
+                        "Volume {} already shared via different protocol: {:?}",
+                        volume_id, state.protocol,
+                    );
+                    error!("{}", m);
+                    return Err(Status::failed_precondition(m));
+                }
+                let uri = get_volume_share_uri(&volume)?;
+                debug!("Volume {} already published at {}", volume_id, uri);
```
tiagolobocastro, line 450: Do we need to check if we're published on the right node?

tiagolobocastro (commented):
Nice :+1: I'll finish reviewing tomorrow

control-plane/csi-controller/src/controller.rs
```diff
@@ -0,0 +1,726 @@
...
+const K8S_HOSTNAME: &str = "kubernetes.io/hostname";
+const VOLUME_NAME_PATTERN: &str =
+    r"pvc-([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})";
+const PROTO_NVMF: &str = "nvmf";
+const MAYASTOR_NODE_PREFIX: &str = "mayastor://";
+
+#[derive(Debug, Default)]
+pub struct CsiControllerSvc {}
+
+mod volume_opts {
```
gila, line 24: I don't think we want to (and importantly need to) carry this over from the older implementation. We only support a little number of parameters, right? Some timeout settings, local true/false, etc. 

If we create a struct something like:

```rust
#[derive(Deserialize, Serialize)]
struct VolumeOptions {
   local: bool, 
   io_timeout: u64,
   ... 
}
```

We can then parse the whole parameters thing in one blow, and add nice impls as needed to make the code more readable i.e

```rust
impl VolumeOptions {
pub fn pinned(&self) -> bool {
  true 
}
```

It is more work initially but in the long run it will prevent making us go insane as most of it all is just "does you have this value set?"

deploy/csi-deployment.yaml
```diff
@@ -0,0 +1,62 @@
+apiVersion: apps/v1
```
blaisedias, line 1: Please add  deployment yaml files by creating a template file in `chart/templates/` and running `script/generate-deploy-yamls.sh`

blaisedias (changes requested):

deploy/csi-deployment.yaml
gila: @blaisedias We will leave that part for later, we are expecting a lot of churn in these files still.

control-plane/csi-controller/src/controller.rs
```diff
@@ -0,0 +1,726 @@
...
+            debug!(
+                "Volume {} successfully created, pinned volume = {}",
+                volume_uuid, pinned_volume
+            );
+        }
+
+        let vt_mapper = VolumeTopologyMapper::init().await?;
+
+        let volume = rpc::csi::Volume {
+            capacity_bytes: size as i64,
```
gila, line 356: :thinking: i64?

control-plane/csi-controller/src/controller.rs
```diff
@@ -0,0 +1,726 @@
...
+            }
+        }
+
+        let u = Uuid::parse_str(&volume_uuid).map_err(|_e| {
+            Status::invalid_argument(format!("Malformed volume UUID: {}", volume_uuid))
+        })?;
+
+        // First check if the volume already exists.
+        if let Some(existing_volume) = MayastorApiClient::get_client()
+            .list_volumes()
```
gila, line 326: This can list usize:max volumes... I've already mentioned that in the previous comment. Having that said, why do we not get the volume directly? i.e GET volume/{ID} ?

control-plane/csi-controller/src/controller.rs
tiagolobocastro: Yeah that's the spec... they even say "The value of this field MUST NOT be negative." -.-

control-plane/csi-controller/src/client.rs
```diff
@@ -0,0 +1,373 @@
...
+    /// Delete volume and reclaim all storage resources associated with it.
+    /// This operation is idempotent, so the caller does not see errors indicating
+    /// abscence of the resource.
+    pub async fn delete_volume(&self, volume_id: &str) -> Result<(), ApiClientError> {
+        self.do_delete(&UrnType(&[uri::VOLUMES, volume_id]), true)
+            .await
+    }
+
+    #[instrument]
+    /// Describe specific volume.
```
paulyoong, line 347: Nit: Should this be "Get specific volume".

control-plane/csi-controller/src/controller.rs
paulyoong: I don't think we do currently but we definitely should.

control-plane/csi-controller/src/controller.rs
```diff
@@ -0,0 +1,726 @@
...
+                .collect()
+        } else {
+            Vec::new()
+        }
+    }
+
+    /// Determines whether target volume is pinned.
+    /// TODO: as of now all volumes are assumed pinned.
+    pub fn is_volume_pinned(_volume: &Volume) -> bool {
+        true
```
paulyoong, line 195: Is it safe to just return true here? In the `volume_accessible_topology` function above we return an empty vector if pinned is false. Do these two function contradict themselves? (i.e. we report that the workload can be placed anywhere but we say that the volume is pinned). I don't know what the repercussions of this are, if any.

control-plane/csi-controller/src/client.rs
```diff
@@ -0,0 +1,373 @@
...
+}
+
+impl Display for UrnType<'_> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.0.join("/"))
+    }
+}
+
+/// Single instance API client for accessing REST API gateway.
+/// Incapsulates communication with REST API by exposing a set of
```
paulyoong, line 67: Nit: Incapsulates -> Encapsulates

control-plane/csi-controller/src/controller.rs
```diff
@@ -0,0 +1,726 @@
...
+
+    #[instrument]
+    async fn controller_unpublish_volume(
+        &self,
+        request: tonic::Request<ControllerUnpublishVolumeRequest>,
+    ) -> Result<tonic::Response<ControllerUnpublishVolumeResponse>, tonic::Status> {
+        let args = request.into_inner();
+
+        debug!("Request to unpublish volume: {:?}", args);
+        // Check if node exists.
```
paulyoong, line 492: We don't seem to be checking if the node exists. Should we be doing so?

paulyoong (commented):
Thanks @mtzaurus, looks awesome.

control-plane/csi-controller/src/controller.rs
```diff
@@ -0,0 +1,726 @@
...
+    // If volume is created as pinned (i.e. local=true), then the nexus and the workload
+    // must be placed on the same node, which in fact means running workloads only on Mayastor
+    // daemonset nodes.
+    // For non-pinned volumes, workload can be put on any node in the Kubernetes cluster.
+    pub fn volume_accessible_topology(&self, pinned_volume: bool) -> Vec<CsiTopology> {
+        if pinned_volume {
+            self.nodes
+                .iter()
+                .map(|n| {
+                    let mut segments = HashMap::new();
```
tiagolobocastro, line 182: these should probably be only nodes with the openebs.io/engine=mayastor I guess?

control-plane/csi-controller/src/controller.rs
```diff
@@ -0,0 +1,726 @@
...
+                    segments.insert(K8S_HOSTNAME.to_string(), n.id.to_string());
```
tiagolobocastro, line 183: these should probably be only nodes with the label `openebs.io/engine=mayastor` I guess?

control-plane/csi-controller/src/server.rs
```diff
@@ -0,0 +1,126 @@
...
+        Pin::new(&mut self.0).poll_shutdown(cx)
+    }
+}
+
+pub struct CsiServer {}
+
+async fn ping_rest_api() {
+    info!("Checking REST API endpoint accessibility ...");
+
+    match MayastorApiClient::get_client().list_nodes().await {
```
tiagolobocastro, line 74: We should probably add health endpoints to the rest api

tiagolobocastro (commented):
Are we adding any tests on this PR? IMHO we should at least add a few sanity checks to make sure the basic operations don't break with changes to the rest of the control plane.

control-plane/csi-controller/src/client.rs
```diff
@@ -0,0 +1,373 @@
...
+                u, e
+            ))
+        })?;
+
+        // Check HTTP status code, handle DELETE idempotency transparently.
+        let res = match response.status() {
+            StatusCode::OK => Ok(()),
+            // Handle idempotency as requested by the caller.
+            StatusCode::NOT_FOUND | StatusCode::NO_CONTENT | StatusCode::PRECONDITION_FAILED => {
+                if idempotent {
```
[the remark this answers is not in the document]
mtzaurus: The idea is that we handle operation as successful even if there are no resource exists (volume removal or so). In case of missing HTTP resource we will see 404, which is an error from HTTP server perspective, but not an error for us in case of volume removal, as per CSI spec.

control-plane/csi-controller/src/client.rs
mtzaurus: Good point.
We use URLs which are constructed by us, so parsing should always succeed. But as a defensive logic we could have it.

control-plane/csi-controller/src/client.rs
mtzaurus: Do we need to fix the errors returned by Control Plane ?

control-plane/csi-controller/src/controller.rs
mtzaurus: Yup, good point.

control-plane/csi-controller/src/controller.rs
mtzaurus: Yes, can be added here too.

control-plane/csi-controller/src/controller.rs
mtzaurus: Yes, we could probably use the existing Protocol enum, which simplifies things here.

control-plane/csi-controller/src/client.rs
```diff
@@ -0,0 +1,373 @@
...
+use once_cell::sync::OnceCell;
+use reqwest::{Client, Error, Response, StatusCode, Url};
+use serde::{Deserialize, Serialize};
+use std::fmt::{Display, Formatter};
+use tracing::instrument;
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum ApiClientError {
+    // Error while communicating with the server.
+    ServerCommunicationError(String),
```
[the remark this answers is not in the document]
mtzaurus: Good point, yes.

control-plane/csi-controller/src/client.rs
```diff
@@ -0,0 +1,373 @@
...
+#[derive(Debug)]
+pub struct MayastorApiClient {
+    base_url: String,
+    rest_client: Client,
+}
+
+impl MayastorApiClient {
+    /// Initialize API client instance. Must be called prior to
+    /// obtaining the client instance.
+    pub fn initialize(endpoint: String) -> Result<()> {
```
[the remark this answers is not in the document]
mtzaurus: IMHO new() is supposed to return a new instance. We obtain a single instance via 'get_instance()' and use initialize() as initialisation step, which does not yield any instances.

control-plane/csi-controller/src/client.rs
```diff
@@ -0,0 +1,373 @@
...
+            .build()
+            .expect("Failed to build REST client");
+
+        REST_CLIENT.get_or_init(|| Self {
+            base_url: format!("{}/v0", endpoint),
+            rest_client,
+        });
+
+        debug!("API client is initialized with endpoint {}", endpoint);
+        Ok(())
```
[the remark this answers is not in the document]
mtzaurus: I am going to add extra check for URL which we use for REST API endpoints, so we can have a malformed URL, which results in Err() returned.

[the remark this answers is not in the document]
mtzaurus: I am going to add extra check for URL which we use for REST API endpoints, so we can have a malformed URL, which results in Err() returned.

control-plane/csi-controller/src/client.rs
```diff
@@ -0,0 +1,373 @@
...
+        I: Serialize + Sized,
+        for<'a> O: Deserialize<'a>,
+    {
+        let u = format!("{}/{}", self.base_url, urn);
+        let uri = Url::parse(&u).unwrap();
+
+        let response = self
+            .rest_client
+            .put(uri)
+            .json(&object)
```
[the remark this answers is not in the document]
mtzaurus: Nope, 'json()' does not return any errors:

```
pub fn json<T: Serialize + ?Sized>(self, json: &T) -> RequestBuilder
```

control-plane/csi-controller/src/controller.rs
```diff
@@ -0,0 +1,726 @@
...
+
+        let entries = MayastorApiClient::get_client()
+            .list_volumes()
+            .await
+            .map_err(|e| Status::internal(format!("Failed to list volumes, error = {:?}", e)))?
+            .into_iter()
+            .take(if max_entries > 0 {
+                max_entries as usize
+            } else {
+                usize::MAX
```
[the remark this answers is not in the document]
mtzaurus: 1M volumes would suffice, I presume ?

control-plane/csi-controller/src/controller.rs
mtzaurus: Good point.

control-plane/csi-controller/src/controller.rs
mtzaurus: Exactly.

control-plane/csi-controller/src/controller.rs
mtzaurus: GETting volume directly by name will result in 404 every time create_volume() is called (in most of the cases, where we create a new volume), so we will see lots of false-positive errors in REST server logs.
Ideally, we could use GET /volumes?uuid=volume_id and get either 0 elements or 1.
Maybe we can add search options later to REST API.

control-plane/csi-controller/src/client.rs
```diff
@@ -0,0 +1,373 @@
...
+        replicas: u8,
+        size: u64,
+        allowed_nodes: &[String],
+        preferred_nodes: &[String],
+    ) -> Result<Volume, ApiClientError> {
+        let mut allowed = Vec::new();
+        let mut preferred = Vec::new();
+
+        allowed.extend_from_slice(allowed_nodes);
+        preferred.extend_from_slice(preferred_nodes);
```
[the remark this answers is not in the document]
mtzaurus: Good point, it looks indeed nicer.

control-plane/csi-controller/src/controller.rs
mtzaurus: The idea is that we should make judgements about pinned state based on some info that we pass to volume upon volume creation. Currently it's not possible to specify anything extra in VolumeCreatinBody, so we optimistically assume the volume is pinned.
Once Control Plane API supports extra info upon volume creation, volume locality will be determined.
properly.
For now all volumes are assumed to be pinned.

control-plane/csi-controller/src/controller.rs
mtzaurus: Yes, I might have overlooked this check - we also need to assure that node is published on the same node as in the request, so we need to add GET node info here.

control-plane/csi-controller/src/client.rs
```diff
@@ -0,0 +1,373 @@
...
+
+    async fn do_put<I, O>(&self, urn: &UrnType<'_>, object: I) -> Result<O, ApiClientError>
+    where
+        I: Serialize + Sized,
+        for<'a> O: Deserialize<'a>,
+    {
+        let u = format!("{}/{}", self.base_url, urn);
+        let uri = Url::parse(&u).unwrap();
+
+        let response = self
```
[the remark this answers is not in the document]
mtzaurus: We rely here on Reqwest, AFAIK it maintains a connection pool internally.

control-plane/csi-controller/src/client.rs
mtzaurus: I created CAS-1124 as a placeholder and a TODO.

control-plane/csi-controller/src/controller.rs
mtzaurus: Do we report any labels as part of node objects via REST API ?

```
[
   {
      "id" : "mayastor-2",
      "spec" : {
         "grpcEndpoint" : "10.1.0.6:10124",
         "id" : "mayastor-2"
      },
      "state" : {
         "grpcEndpoint" : "10.1.0.6:10124",
         "id" : "mayastor-2",
         "status" : "Online"
      }
   },
   {
      "id" : "mayastor-1",
      "spec" : {
         "grpcEndpoint" : "10.1.0.5:10124",
         "id" : "mayastor-1"
      },
      "state" : {
         "grpcEndpoint" : "10.1.0.5:10124",
         "id" : "mayastor-1",
         "status" : "Online"
      }
   }
]
```

Moreover, can Control Plane return any nodes other than Mayastor nodes ?

control-plane/csi-controller/src/server.rs
mtzaurus: I captured that in CAS-1125

control-plane/csi-controller/src/controller.rs
tiagolobocastro: We report nodes that showed up as mayastors nodes at some point in time. We never actually delete them so the list could be stale... at one point we talked about having a node operator which would "sync" with the kubernetes nodes.

deploy/csi-deployment.yaml
```diff
@@ -0,0 +1,62 @@
...
+            - "--csi-address=$(ADDRESS)"
+          env:
+            - name: ADDRESS
+              value: /var/lib/csi/sockets/pluginproxy/csi.sock
+          imagePullPolicy: "IfNotPresent"
+          volumeMounts:
+            - name: socket-dir
+              mountPath: /var/lib/csi/sockets/pluginproxy/
+        - name: csi-controller
+          image: mayadata/csi-controller:develop
```
paulyoong, line 50: I think you need to update `release.sh` so that the csi-controller is built and pushed to Docker Hub.

gila (approved):

paulyoong (approved):

control-plane/csi-controller/src/controller.rs
```diff
@@ -0,0 +1,774 @@
...
+            Err(ApiClientError::ResourceNotExists { .. }) => {
+                debug!("Volume {} does not exist, not unpublishing", args.volume_id);
+                return Ok(Response::new(ControllerUnpublishVolumeResponse {}));
+            }
+            Err(e) => return Err(Status::from(e)),
+        };
+
+        // Check if target volume is published and the node matches.
+        if let Some((node, _)) = get_volume_share_location(&volume) {
+            if !args.node_id.is_empty() && node != args.node_id {
```
tiagolobocastro, line 527: If we are published on a different node, should we not unpublish anyway? Otherwise we'll never be able to unpublish?

control-plane/csi-controller/src/controller.rs
```diff
@@ -0,0 +1,774 @@
...
+    async fn controller_unpublish_volume(
+        &self,
+        request: tonic::Request<ControllerUnpublishVolumeRequest>,
+    ) -> Result<tonic::Response<ControllerUnpublishVolumeResponse>, tonic::Status> {
+        let args = request.into_inner();
+
+        debug!("Request to unpublish volume: {:?}", args);
+        // Check if target volume exists.
+        let volume = match MayastorApiClient::get_client()
+            .get_volume(&args.volume_id)
```
tiagolobocastro, line 514: If I understood you reasoning with the 404, I think this is one of the cases where we'd want to use volumes/{id}, rather than volumes?id={id}, because the volume should indeed exist, and the 404 here is an error?

tiagolobocastro (approved):
I wonder if some of the debug!() at the beginning and end of the most important operations should be an info!

deploy/csi-deployment.yaml
blaisedias: If we do, it makes it difficult to automate running the tests with control plane 2.

blaisedias (approved):

blaisedias (approved):
LGTM

mtzaurus:
bors merge

mtzaurus:
bors merge

tiagolobocastro:
bors merge

PR #103, opened 2021-09-20
feat: add force volume unpublish

tiagolobocastro:
When a node is not reachable we cannot unpublish.
Add a force option which will ignore the error if the node is not online.
Should be used with care, only when the caller knows that it is safe to do so.


control-plane/agents/core/src/volume/specs.rs
```diff
@@ -592,8 +592,30 @@ impl ResourceSpecsLocked {
...
+                        let node_online = match registry.get_node_wrapper(&nexus_clone.node).await {
+                            Ok(node) => {
+                                let mut node = node.lock().await;
+                                node.is_online() && node.liveness_probe().await.is_ok()
+                            }
+                            _ => false,
+                        };
+                        if !node_online {
+                            nexus_clone.warn_span(|| {
+                                tracing::warn!("Force unpublish. Forgetting about the target nexus because the node is not online and it was requested")
```
paulyoong, line 611: Could we include the nexus ID and node?

control-plane/agents/core/src/volume/tests.rs
```diff
@@ -861,6 +856,26 @@ async fn publishing_test(cluster: &Cluster) {
...
+
+    UnpublishVolume::new(&volume_state.uuid, false)
+        .request()
+        .await
+        .expect_err("The node is not online...");
+
+    UnpublishVolume::new(&volume_state.uuid, true)
+        .request()
+        .await
+        .expect("With force comes great responsibility...");
```
paulyoong, line 870: :grin: May the force be with you!

paulyoong (approved):

control-plane/agents/core/src/volume/specs.rs
tiagolobocastro: It's already automatically included by the nexus :)

tiagolobocastro:
bors merge

PR #105, opened 2021-09-22
feat(tf): want terraform files

gila:
These set of `.tf` files will deploy mayastor and its control plane
components. The goal of these files is to easily deploy and remove
mayastor as a whole from the k8s cluster.


deploy/terraform/main.tf
```diff
@@ -0,0 +1,101 @@
...
+  source = "./mod/k8s-operator"
+  image  = var.msp_operator_image
+  depends_on = [
+    module.rbac,
+    module.core,
+    module.rest
+  ]
+}
+
+module "rest" {
```
paulyoong, line 65: Should rest depend on `module.nats` and `module.core` as that is what it is communicating with?

deploy/terraform/mod/rest/main.tf
```diff
@@ -0,0 +1,110 @@
+variable "image" {}
+
+resource "kubernetes_secret" "example" {
+  metadata {
+    name      = "docker-cfg"
+    namespace = "mayastor"
+  }
+  data = {
+    //".dockerconfigjson" = file("./docker-registry.json")
```
paulyoong, line 9: Can we remove this?

deploy/terraform/variables.tf
```diff
@@ -0,0 +1,77 @@
...
+
+variable "core_image" {
+  type    = string
+  default = "mayadata/mcp-core:develop"
+}
+
+variable "mayastor_image" {
+  type        = string
+  description = "mayastor image to use"
+  default     = "192.168.1.4:5000/mayadata/mayastor:latest"
```
paulyoong, line 34: Looks like you're using a local repo here.

paulyoong (approved):

deploy/terraform/mod/k8s-operator/main.tf
```diff
@@ -0,0 +1,57 @@
+variable "image" {}
+
+resource "kubernetes_deployment" "deployment_msp_operator" {
+  metadata {
```
tiagolobocastro, line 4: Is it possible to load the these parameters from the yamls files we already have? Otherwise we'll know need to make sure we modify both everytime something changes.

gila: The whole idea is that you don't use YAML. Anything you want to change should be changed in `variables.tf`. And so you don't need to have YAML. Note; this is just for me, initially, and we don't have to merge it. But if there is some variance in configs, and you bump into something I do not. You send over the variables.tf I apply it using that and ... we should have the same, that's the idea. 

Plus, its destroy/apply with ordering.

deploy/terraform/main.tf
gila: Yes, but good news; it works without it :) I did that purposely to test the dependency stuff. But, in principle: yes it should wait.

deploy/terraform/mod/k8s-operator/main.tf
tiagolobocastro: We don't need yaml if you install via terraform but for those who want to use yaml it would be nice to keep them in sync. Another possibility would be to generate yaml from this, rather than using helm for example. Is that possible at all?

deploy/terraform/README.md
```diff
@@ -0,0 +1,21 @@
+## Requirements
+
+
+The terraform scripts here assume that there is a k8s cluster with at least 3 nodes up and running. The nodes on which mayastor should run should be labeled as usual. Control plane components are to be scheduled on node(s) that have the label `openebs.io/engine=none`.
```
tiagolobocastro, line 4: Could we use a different label name. Could be useful if we want to run the mayaastor and control plane on the same node locally (to avoid having to deploy more VM's)

deploy/terraform/mod/mayastor/main.tf
```diff
@@ -0,0 +1,188 @@
...
+        }
+
+        container {
+          name  = "mayastor"
+          image = var.image
+          args = [
+            "-N$(MY_NODE_NAME)",
+            "-g$(MY_POD_IP)",
+            "-nnats",
+            "-l2,3",
```
tiagolobocastro, line 89: can this be made to be dependent on the cpu count?

deploy/terraform/variables.tf
tiagolobocastro: Can we customize it to take a local registry and alternative image tag as options?

tiagolobocastro (approved):
I wonder if it's possible to edit&remove the finalizers from the pools on destroy, otherwise the destroy gets stuck.
This is really cool :+1:

deploy/terraform/mod/k8s-operator/main.tf
gila: They don't seem to be interested in that, but here and there, there are some hacks and tools..

deploy/terraform/README.md
gila: Yes, we can add what we want, add it to the the variables maybe and have it be based on node names?

deploy/terraform/variables.tf
gila: Yeah we can do var.registry and var.tag and use format()

tiagolobocastro:
> > I wonder if it's possible to edit&remove the finalizers from the pools on destroy, otherwise the destroy gets stuck.
> > This is really cool +1
> 
> If the pools are created using tf, the order can be specified.. The key thing I want to achieve is; pop up cluster apply, experiment and likely destroy. So I think pools should be part of the mix, I will have a look at that later, it should be possible sweat_smile

One of the nicest things is being able to do modify parameters, and then tf apply, and it magically updates the cluster based on the changes

deploy/terraform/README.md
tiagolobocastro: Ah yes, that would better indeed, then we don't even need to manually label the nodes :+1:

deploy/terraform/mod/k8s-operator/main.tf
gila: Turns out there are external tools available to "from a running-config" generate YAML.

So you would apply config, and use a utility like kube-dump. You would lose, ofcourse, the ability to use variables and do "calculations".

gila:
bors merge

PR #106, opened 2021-09-22
chore(msp): add msp crd yaml file to the deploy

tiagolobocastro:
Update the msp with a flag to generates the crd yaml file and exit.
Update the msp to include pool capacity and free space.
Add script that generates the crd file when the msp src code changes.
Run the crd generation on pre-commit.
Run the helm-deploy generation on pre-commit.
Modify the msp poll interval to match the common cache update interval.


control-plane/msp-operator/Cargo.toml
```diff
@@ -28,10 +28,13 @@ tracing = "0.1.26"
 tracing-subscriber = "0.2.20"
 url = "2.2.2"
 openapi = { path = "../../openapi"}
+common-lib = { path = "../../common" }
```
gila, line 31: So this has some pretty big compile-time ramifications. Are you sure it is worth it?

control-plane/msp-operator/src/main.rs
```diff
@@ -859,6 +879,20 @@ async fn pool_controller(args: ArgMatches<'_>) -> anyhow::Result<()> {
     Ok(())
 }
 
+/// Generate the mayastor pool CRD file allowing users to register them with kubernetes
+/// before the pool operator starts running.
+/// Can also be used to unregister the CRDs on uninstall.
+fn write_msp_crd(name: Option<&str>) -> anyhow::Result<()> {
```
gila, line 885: We don't need the crd in yaml to delete the crd definition; a simple 

`k  delete crd mayastorpools.openebs.io` 

is sufficient, but I see what you are trying to say here :+1:

gila (approved):

control-plane/msp-operator/src/main.rs
tiagolobocastro: Right, good point!

control-plane/msp-operator/Cargo.toml
tiagolobocastro: Yeah that was bugging me too... Maybe we could have a tiny crate for shared constants to share between projects.

paulyoong (approved):

control-plane/msp-operator/Cargo.toml
gila: So we can "go ahead" and do this; the problem is it will get harder to unwind later. I mean, running `cargo build --bin msp-operator` I really think it's let us say; annoying, that then I have to "wait" for things to compile that I don't even use (smol, nats, what have you).  For this case, it is not that bad, but I spent a while back later trying to weed out some old depends and it was not a great experience:sweat:  So to just get a value of '30' is pretty harsh. 

So I was thinking, as the core agent is leading this in the sense that, it does not make it any better for you to poll me with an interval smaller than X. If we add variable to the core agents YAML as a label, we can then read it from the YAML to ensure it matches. 

This is what we want, right? Because if I change the core agent to 1min, it again is "too fast," see what I mean? We can then later one when we want to, at runtime ask for the value.  Using the metadata information we talked about yesterday for example. 

either way, I'm fine with it as it is :+1:

tiagolobocastro: So for the compilation times I removed the dependency and used a common file instead, so the compilation time did not increase at all.
We can read it from a yaml variable indeed, and use it to keep "everyone" in sync - or perhaps even a configmap that gets mounted into the pods, so it's common for all.

tiagolobocastro:
I've made a yaml file change,
bors cancel

tiagolobocastro:
bors merge

tiagolobocastro:
bors cancel

tiagolobocastro:
bors merge

PR #108, opened 2021-09-23
fix(kubectl-mayastor): return single element on get volumes or pools json or yaml output

Abhinandan-Purkait:
### Why do we need this PR?
- Previously to reuse the code, the get and list methods were returning a json array, which is not the correct type for get.
- This fix will identify each api call and return it corresponding format.

### DIFF
--- a/kubectl-plugin/src/resources/pool.rs
+++ b/kubectl-plugin/src/resources/pool.rs
@@ -15,47 +15,45 @@ use structopt::StructOpt;
 #[derive(StructOpt, Debug)]
 pub struct Pools {}
 
-// CreateRows being trait for Vec<Pool> would create the rows from the list of
+// CreateRows being trait for Pool would create the rows from the list of
 // Pools returned from REST call.
-impl CreateRows for Vec<openapi::models::Pool> {
+impl CreateRows for openapi::models::Pool {
     fn create_rows(&self) -> Vec<Row> {
         let mut rows: Vec<Row> = Vec::new();
-        for pool in self {
-            let mut managed = true;
-            if pool.spec.is_none() {
-                managed = false;
-            }
-            // The spec would be empty if it was not created using
-            // control plane.
-            let spec = pool.spec.clone().unwrap_or_default();
-            // In case the state is not coming as filled, either due to pool, node lost, fill in
-            // spec data and mark the status as Unknown.
-            let state = pool.state.clone().unwrap_or(openapi::models::PoolState {
-                capacity: 0,
-                disks: spec.disks,
-                id: spec.id,
-                node: spec.node,
-                status: openapi::models::PoolStatus::Unknown,
-                used: 0,
-            });
-            let disks = state.disks.join(", ");
-            rows.push(row![
-                pool.id,
-                state.capacity,
-                state.used,
-                disks,
-                state.node,
-                state.status,
-                managed
-            ]);
+        let mut managed = true;
+        if self.spec.is_none() {
+            managed = false;
         }
+        // The spec would be empty if it was not created using
+        // control plane.
+        let spec = self.spec.clone().unwrap_or_default();
+        // In case the state is not coming as filled, either due to pool, node lost, fill in
+        // spec data and mark the status as Unknown.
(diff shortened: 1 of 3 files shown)


kubectl-plugin/src/resources/pool.rs
```diff
@@ -67,7 +67,7 @@ impl List for Pools {
         match RestClient::client().pools_api().get_pools().await {
             Ok(pools) => {
                 // Print table, json or yaml based on output format.
-                utils::print_table::<openapi::models::Pool>(output, pools, "list");
+                utils::print_table::<openapi::models::Pool>(output, pools, utils::LIST);
```
tiagolobocastro, line 70: this is ok but what about changing things a bit and rather than implementing the traits to Vec<T> implement them for T. Then Vec<T> can also implement the same trait.

Abhinandan-Purkait: Yes. We can, but again the contents would be same for them, so I thought this would do the job with minimum code change. What do you suggest should I implement it again for T ?

tiagolobocastro: The contents would not be the same because the implementation for Vec<T> would output it correctly for a "list". I can show you what I mean on slack if you're interested.

Abhinandan-Purkait: Yep, got your point. That is a better approach.

kubectl-plugin/src/resources/pool.rs
```diff
@@ -15,47 +15,45 @@ use structopt::StructOpt;
...
+        let state = self.state.clone().unwrap_or(openapi::models::PoolState {
+            capacity: 0,
+            disks: spec.disks,
+            id: spec.id,
+            node: spec.node,
+            status: openapi::models::PoolStatus::Unknown,
+            used: 0,
+        });
+        let disks = state.disks.join(", ");
+        rows.push(row![
```
tiagolobocastro, line 41: you can just do vec![row![....]]

kubectl-plugin/src/resources/pool.rs
```diff
@@ -67,7 +65,7 @@ impl List for Pools {
         match RestClient::client().pools_api().get_pools().await {
             Ok(pools) => {
                 // Print table, json or yaml based on output format.
-                utils::print_table::<openapi::models::Pool>(output, pools);
+                utils::print_table::<Vec<openapi::models::Pool>>(output, pools);
```
tiagolobocastro, line 68: you probably don't need to specify the type you can just do `print_table(output, pools)` and the same for all the other calls of print_table.

kubectl-plugin/src/resources/utils.rs
```diff
@@ -60,11 +60,32 @@ impl From<&str> for OutputFormat {
     }
 }
 
-pub fn print_table<T>(output: &OutputFormat, obj: Vec<T>)
+impl<T> CreateRows for Vec<T>
+where
+    T: CreateRows,
+{
+    fn create_rows(&self) -> Vec<Row> {
+        let mut rows = vec![];
```
tiagolobocastro, line 68: nit: this might work: self.iter().map(|i| i.create_rows()).flatten().collect()

kubectl-plugin/src/resources/utils.rs
```diff
@@ -60,11 +60,32 @@ impl From<&str> for OutputFormat {
...
+    }
+}
+
+// GetHeaderRow trait to be implemented by Volume/Pool to fetch the corresponding headers.
+impl<T> GetHeaderRow for Vec<T>
+where
+    T: GetHeaderRow,
+{
+    fn get_header_row(&self) -> Row {
+        self[0].get_header_row()
```
tiagolobocastro, line 80: this would panic on empty lists, so I guess it could be something like this:
self.get(0).map(GetHeaderRow::get_header_row).unwrap_or_default()

tiagolobocastro (approved):

paulyoong:
bors d+

Abhinandan-Purkait:
bors merge

PR #109, opened 2021-09-24
csi fixes

gila:


control-plane/csi-controller/src/main.rs
```diff
@@ -1,55 +1,82 @@
...
-    #[structopt(short = "r", long = "rest-endpoint")]
-    pub rest_endpoint: String,
-    #[structopt(short="c", long="csi-socket", default_value=CSI_SOCKET)]
-    pub csi_socket: String,
-    #[structopt(short="l", long="log-level", default_value="info", possible_values=&["info", "debug", "trace"])]
-    pub log_level: String,
-}
+#[tokio::main(worker_threads = 2)]
+pub async fn main() -> Result<(), String> {
+    let args = App::new("Mayastor k8s pool operator")
```
tiagolobocastro, line 16: any particular reason for replacing structopt with clap, we tend to use struct on most of the control plane

deploy/terraform/mod/core/main.tf
```diff
@@ -18,7 +18,7 @@ resource "kubernetes_stateful_set" "core_stateful_set" {
   }
   spec {
     service_name = "core-agents"
-    replicas = 1
+    replicas     = 1
```
tiagolobocastro, line 21: what tool do you use to link these files? I installed a plugin for clion, but I guess it must not do linting.

tests/tests-mayastor/Cargo.toml
```diff
@@ -20,8 +20,8 @@ opentelemetry-jaeger = { version = "0.15.0", features = ["tokio"] }
 tracing-opentelemetry = "0.15.0"
 opentelemetry = "0.16.0"
 actix-web-opentelemetry = "0.11.0-beta.5"
-tracing = "0.1.26"
-anyhow = "1.0.43"
+tracing = "0.1.28"
+anyhow = "1.0.44"
```
tiagolobocastro, line 24: btw now you don't have to revert changes to the Cargo.toml from the openapi, it should work (well, I hope so).

tiagolobocastro (approved):

control-plane/csi-controller/src/main.rs
```diff
@@ -1,55 +1,82 @@
...
+                .default_value(CSI_SOCKET)
+                .help("CSI socket path"),
+        )
+        .arg(
+            Arg::with_name("jaeger")
+                .short("-j")
+                .env("JAEGER_ENDPOINT")
+                .help("enable open telemetry and forward to jaeger"),
+        )
+        .get_matches();
```
mtzaurus, line 45: The new code looks a bit more verbose than existing.

mtzaurus (approved):

control-plane/csi-controller/src/main.rs
gila: Oh, well -- to be honest, I compared it to the MSP bits. I prefer (and yes, its preference) to be more in and just explicitly type things when it comes to this. More often than not, I'm trying to unravel "what as the attribute again?" then simply typing it out. 

I think I was looking for  "How do say: use this or use env variable"... So no real reason other than that.

deploy/terraform/mod/core/main.tf
gila: terraform has a format command :) I have a nvim plugin that allows me to specify what formatted to use per file, and so for me it's a matter of `,fm` -- where `,` leader key

tests/tests-mayastor/Cargo.toml
gila: Oh okay, I did do that, I can try it out, would be nice to be able to upgrade it everywhere in one blow!

control-plane/csi-controller/src/main.rs
gila: It is mike, but my experience has shown that over time, you start to struggle a bit with the attributes and .. well.. 

This change does not make the code any better not any worse so, I just prefer to be explicit than macro magic. Its a limitation of my brain :) Not in all cases ofcourse.

PR #110, opened 2021-09-27
feat(pool-labels): add a unique label to all pools created by control plane on request from msp

Abhinandan-Purkait:
### What does this PR do?
- This PR changes the labels type to hashmap
- This adds changes to populate labels on request.
- This add changes in msp-operator, to add a unique label is the pool creation request comes to rest from msp-operator.

### DIFF
--- a/common/src/types/v0/message_bus/pool.rs
+++ b/common/src/types/v0/message_bus/pool.rs
@@ -247,15 +247,23 @@ pub struct CreatePool {
     pub id: PoolId,
     /// disk device paths or URIs to be claimed by the pool
     pub disks: Vec<PoolDeviceUri>,
+    /// labels to be set on the pool
+    pub labels: Option<::std::collections::HashMap<String, String>>,
 }
 
 impl CreatePool {
     /// Create new `Self` from the given parameters
-    pub fn new(node: &NodeId, id: &PoolId, disks: &[PoolDeviceUri]) -> Self {
+    pub fn new(
+        node: &NodeId,
+        id: &PoolId,
+        disks: &[PoolDeviceUri],
+        labels: &Option<::std::collections::HashMap<String, String>>,
+    ) -> Self {
         Self {
             node: node.clone(),
             id: id.clone(),
             disks: disks.to_vec(),
+            labels: labels.clone(),
         }
     }
 }
--- a/common/src/types/v0/message_bus/pool.rs
+++ b/common/src/types/v0/message_bus/pool.rs
@@ -1,6 +1,9 @@
 use super::*;
 
-use crate::{types::v0::store::pool::PoolSpec, IntoOption};
+use crate::{
+    types::v0::store::pool::{PoolLabel, PoolSpec},
+    IntoOption,
+};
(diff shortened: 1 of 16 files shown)


control-plane/msp-operator/src/main.rs
```diff
@@ -479,8 +479,16 @@ impl ResourceContext {
             });
         }
 
-        let mut body = HashMap::new();
-        body.insert("disks", self.spec.disks.clone());
+        let mut labels: HashMap<String, String> = HashMap::new();
+        labels.insert(
+            String::from("openebs.io/created-by"),
+            String::from("mayastor-control-plane"),
```
paulyoong, line 485: I wonder if this should be "msp-operator" as that is what is requesting that a pool should be created.

common/src/types/v0/message_bus/pool.rs
```diff
@@ -247,15 +247,23 @@ pub struct CreatePool {
     pub id: PoolId,
     /// disk device paths or URIs to be claimed by the pool
     pub disks: Vec<PoolDeviceUri>,
+    /// labels to be set on the pool
+    pub labels: Option<::std::collections::HashMap<String, String>>,
```
paulyoong, line 251: Rather than using the hash map directly, should we define a new type? i.e. `type PoolLabels = ::std::collections::HashMap<String, String>`. If we did this, we could easily change the type in one place rather than in every place where it is used.

common/src/types/v0/store/pool.rs
```diff
@@ -69,6 +67,9 @@ impl PartialEq<CreatePool> for PoolSpec {
     }
 }
 
+// PoolLabel is the type for the labels
+type PoolLabel = ::std::collections::HashMap<String, String>;
```
tiagolobocastro, line 71: any particular reason to redefine PoolLabel again? You can mark the other one as pub and then you can use it `message_bus::PoolLabel` or something like that

control-plane/msp-operator/src/main.rs
```diff
@@ -479,8 +479,16 @@ impl ResourceContext {
             });
         }
 
-        let mut body = HashMap::new();
-        body.insert("disks", self.spec.disks.clone());
+        let mut labels: HashMap<String, String> = HashMap::new();
+        labels.insert(
+            String::from("openebs.io/created-by"),
```
tiagolobocastro, line 484: can add these strings as const in `common/src/constants.rs` ?

tiagolobocastro (approved):

common/src/types/v0/store/pool.rs
Abhinandan-Purkait: No reason, I removed it and added back at a different place. I will make it pub. Thanks

control-plane/msp-operator/src/main.rs
Abhinandan-Purkait: Yeah. I would add them.

paulyoong (approved):

Abhinandan-Purkait:
bors merge

PR #112, opened 2021-09-27
feat(terraform): deploy jaegertracing via terraform

tiagolobocastro:
feat(terraform): deploy jaegertracing via terraform

Deploy own Jaeger CRD because the one from the helm chart is broken.
todo: skip helm and deploy the operator directly

---
fix(csi): enable volume self healing

### DIFF
--- a/control-plane/csi-controller/src/controller.rs
+++ b/control-plane/csi-controller/src/controller.rs
@@ -187,7 +187,7 @@ impl VolumeTopologyMapper {
 
 #[tonic::async_trait]
 impl rpc::csi::controller_server::Controller for CsiControllerSvc {
-    #[instrument]
+    #[instrument(error, fields(volume.uuid = tracing::field::Empty))]
     async fn create_volume(
         &self,
         request: tonic::Request<CreateVolumeRequest>,
@@ -212,6 +212,7 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
                 )))
             }
         };
+        tracing::Span::current().record("volume.uuid", &volume_uuid.as_str());
 
         check_volume_capabilities(&args.volume_capabilities)?;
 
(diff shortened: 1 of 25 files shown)


paulyoong (approved):

control-plane/csi-controller/src/controller.rs
```diff
@@ -212,6 +212,7 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
                 )))
             }
         };
+        tracing::Span::current().record("volume.uuid", &volume_uuid.as_str());
```
mtzaurus, line 215: Just out of my curiosity: why we use tracing::Span: only in create_volume() ?

mtzaurus (approved):

control-plane/csi-controller/src/controller.rs
tiagolobocastro: because the volume uuid is not trivially accessible from the request so it's set to Empty. Then here we already have the uuid so can finally record it in the span.

deploy/terraform/mod/rbac/main.tf
```diff
@@ -112,3 +116,16 @@ resource "kubernetes_cluster_role_binding" "mayastor" {
     name      = "mayastor-cluster-role"
   }
 }
+
+## When testing sometimes things get seriously broken and we can't easily use terraform destroy
```
gila, line 120: I dont understand what this brings. When we destroy everything gets 'removed' including the rbac roles? Why would adding a kubectl make things any better?

gila (approved):

deploy/terraform/mod/rbac/main.tf
tiagolobocastro: I'm not exactly sure how this happens, terraform gets into a bad state (probably from testing no doubt) but sometimes we can't use terraform destroy, and those CRD's are left and when applying again, it fails complaining they already exist.

tiagolobocastro:
bors merge

tiagolobocastro:
bors cancel

PR #114, opened 2021-09-28
fix: protocol should not be part of the volume

tiagolobocastro:
The share protocol should not be part of the volume as it's decided on publish.
In the future we may even do things such as publish the volume with a protocol
that is suited for the node where the application runs.

So, instead make the protocol part of the volume target (ie the nexus).

Also, make the state a non optional component of the volume so that we do not
need to "fake" it on multiple layers, instead the core agent will set it.


tiagolobocastro:
bors try

common/src/types/v0/message_bus/volume.rs
```diff
@@ -39,18 +36,18 @@ impl Volume {
...
     pub fn status(&self) -> Option<VolumeStatus> {
-        self.state.as_ref().map(|s| s.status.clone())
+        Some(self.state.status.clone())
     }
 }
 
 impl From<Volume> for models::Volume {
     fn from(volume: Volume) -> Self {
-        models::Volume::new_all(volume.spec(), volume.state().into_opt())
+        models::Volume::new_all(volume.spec(), volume.state)
```
mtzaurus, line 50: Shouldn't we have the same behaviour here: either move both members out of target volume or use getters ?

tiagolobocastro: Yes that should be state()

tiagolobocastro: Done

mtzaurus (approved):

paulyoong (approved):

tiagolobocastro:
bors merge

PR #115, opened 2021-09-29
fix(csi controller): prevent multiple restarts

paulyoong:
When deploying the CSI controller, it typically takes multiple restarts
before the pod is able to enter the running state. This is due to the
fact that it typically cannot communicate with the REST server
immediately.

On start up, in the event that the REST server cannot be reached return
"false" rather than an error. This communicates to the Container
Orchestrator that initialisation has not yet completed, but that the CSI
controller is otherwise healthy. This prevents the Container
Orchestrator from restarting the CSI controller (which is the case when
an error is returned).

Resolves: CAS-1148


paulyoong:
bors try

mtzaurus:
LGTM

mtzaurus (approved):

tiagolobocastro (approved):

paulyoong:
bors merge

PR #118, opened 2021-10-01
feat(replica-scheduling): filter by pool topology

Abhinandan-Purkait:
### What does this PR do?
- Splits Topology into Node and Pool Topology.
- Adds a new filter to filter the pools added through pool topology.
- Hard codes CSI-controller to create volumes with a common pool topology.
- Replicas would be scheduled on pools which are filtered using pool topology.
- If no pool topology is mentioned all pools would be used for replica scheduling.


control-plane/agents/core/src/volume/tests.rs
```diff
@@ -569,15 +569,20 @@ async fn nexus_persistence_test(cluster: &Cluster) {
 }
 async fn nexus_persistence_test_iteration(local: &NodeId, remote: &NodeId, fault: FaultTest) {
     tracing::debug!("arguments ({:?}, {:?}, {:?})", local, remote, fault);
-
+    let mut allowed_nodes: Vec<String> = Vec::new();
+    let preferred_nodes: Vec<String> = Vec::new();
+    allowed_nodes.push(String::from(local.clone()));
```
tiagolobocastro, line 574: you can simplify things here, to get a String from NodeId you can use .to_string()
you can also declare the vector in one go, eg: let allowed_nodes = vec![ local.to_string(), remote.to_string()]

control-plane/agents/core/src/core/scheduling/mod.rs
```diff
@@ -72,6 +75,42 @@ impl PoolFilters {
...
+        let pool_has_creation_label = match request.registry().specs().get_pool(&item.pool.id) {
+            Ok(spec) => match spec.labels {
+                None => false,
+                Some(label) => {
+                    label.contains_key(&*String::from(OPENEBS_CREATED_BY_KEY))
+                        && label.get(&*String::from(OPENEBS_CREATED_BY_KEY))
+                            == Some(&String::from(MSP_OPERATOR))
+                }
+            },
+            Err(_) => false,
```
paulyoong, line 90: If we get an error, it is because we failed to get the pool. Can't we just return false here rather than setting `pool_has_creation_label` to false and continuing?

control-plane/agents/core/src/core/scheduling/mod.rs
```diff
@@ -72,6 +75,42 @@ impl PoolFilters {
     pub(crate) fn usable(_: &GetSuitablePoolsContext, item: &PoolItem) -> bool {
         item.pool.status != PoolStatus::Faulted && item.pool.status != PoolStatus::Unknown
     }
+    /// Should only attempt to use pools having msp-operator creation label iff topology has it
```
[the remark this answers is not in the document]
Abhinandan-Purkait: `iff` is intentional here --> It means `if and only if` ref:- [iff](https://www.collinsdictionary.com/dictionary/english/iff)

control-plane/agents/core/src/core/scheduling/mod.rs
```diff
@@ -72,6 +75,42 @@ impl PoolFilters {
...
+                None => false,
+                Some(pool_topology) => match pool_topology {
+                    PoolTopology::Labelled(labelled_topology) => {
+                        if labelled_topology.inclusion.is_empty() {
+                            false
+                        } else {
+                            labelled_topology
+                                .inclusion
+                                .into_iter()
+                                .filter(|x| *x == creation_label)
```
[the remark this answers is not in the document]
Abhinandan-Purkait: I think we are supposed to compare the `pool labels` with the `pool topology inclusion`. But yeah we should compare the values of the creation key rather than hardcoded ones. Will make that change.

control-plane/agents/core/src/core/scheduling/mod.rs
Abhinandan-Purkait: Yes. We should end here.

control-plane/agents/core/src/core/scheduling/mod.rs
```diff
@@ -72,6 +75,42 @@ impl PoolFilters {
...
+                            .inclusion
+                            .contains_key(OPENEBS_CREATED_BY_KEY)
+                        {
+                            volume_creation_label = labelled_topology
+                                .inclusion
+                                .get(OPENEBS_CREATED_BY_KEY)
+                                .unwrap()
+                                .clone();
+                        } else {
+                            return true;
```
paulyoong, line 97: It might be a good idea to include a comment here stating that we don't support other keys for now. In other words, if any other label is set on the pool, we will ignore it and use any pool.

control-plane/csi-controller/src/client.rs
```diff
@@ -320,11 +323,19 @@ impl MayastorApiClient {
...
-            preferred_nodes.to_vec(),
-        ));
+        let exclusive_label_topology: HashMap<String, String> = HashMap::new();
+        let topology = Topology::new_all(
+            Some(NodeTopology::explicit(ExplicitNodeTopology::new(
+                allowed_nodes.to_vec(),
+                preferred_nodes.to_vec(),
+            ))),
+            Some(PoolTopology::labelled(LabelledTopology::new_all(
+                exclusive_label_topology.to_owned(),
```
paulyoong, line 335: Nit: You could just do `HashMap::new()` here and get rid of `exclusive_label_topology` altogether.

paulyoong (approved):

control-plane/agents/core/src/core/scheduling/mod.rs
Abhinandan-Purkait: Yep. Added the comment for the same.

control-plane/csi-controller/src/client.rs
Abhinandan-Purkait: Corrected. Thanks.

tiagolobocastro (approved):

Abhinandan-Purkait:
bors merge

PR #120, opened 2021-10-01
feat(Volume info): add replica topology to volumes

paulyoong:
Add replica topology information when returning volumes through the REST
API.

The topological information is represented as a hash map where the key
is the replica UUID and the value is a structure representing the
toplogy. The topology includes the node and pool where the replica
resides and the state of the replica.

The kubectl plugin now supports an additional "replica-topology" command
which returns the replica topology for a given volume.

Resolves: CAS-1145/CAS-1149


kubectl-plugin/src/main.rs
```diff
@@ -8,8 +8,10 @@ mod resources;
 mod rest_wrapper;
 
 use crate::{
-    operations::{Get, List, Scale},
-    resources::{node, pool, utils, volume, GetResources, ScaleResources},
+    operations::{Get, List, ReplicaTopology, Scale},
```
Abhinandan-Purkait, line 11: **[low-priority]**:- https://github.com/spf13/cobra#concepts, I have been using this and it says CLI commands are VERBS. Not sure about other best practices though.

Abhinandan-Purkait (approved):

kubectl-plugin/src/main.rs
paulyoong: That's a good point, thanks. Maybe it can be `GetReplicas` instead.

common/src/types/v0/message_bus/volume.rs
```diff
@@ -505,3 +513,25 @@ impl DestroyVolume {
         &self.uuid
     }
 }
+
+/// Replica topology information
+#[derive(Serialize, Deserialize, Default, Debug, Clone, Eq, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct ReplicaTopology {
+    /// id of the mayastor instance
+    pub node: Option<NodeId>,
```
tiagolobocastro, line 522: do these need to be pub?

common/src/types/v0/message_bus/volume.rs
```diff
@@ -505,3 +513,25 @@ impl DestroyVolume {
...
+    /// id of the pool
+    pub pool: Option<PoolId>,
+    /// status of the replica
+    pub status: ReplicaStatus,
+}
+
+impl From<&ReplicaTopology> for models::ReplicaTopology {
+    fn from(replica_topology: &ReplicaTopology) -> Self {
+        Self {
+            node: replica_topology.node.clone().map(Into::into),
```
tiagolobocastro, line 532: can use models::ReplicaTopology::new_all() instead

control-plane/rest/openapi-specs/v0_api_spec.yaml
```diff
@@ -2616,6 +2616,13 @@ components:
         size: 80241024
         status: Online
         uuid: 4be37dbd-4b60-44f3-b807-08f6693522ac
+        "replica_topology": {
```
tiagolobocastro, line 2619: this doesn't seem on the right format?

kubectl-plugin/src/resources/volume.rs
```diff
@@ -111,3 +115,40 @@ impl Scale for Volume {
...
+impl ReplicaTopology for Volume {
+    type ID = VolumeId;
+    async fn topology(id: &Self::ID, output: &OutputFormat) {
+        match RestClient::client().volumes_api().get_volume(id).await {
+            Ok(volume) => {
+                // Print table, json or yaml based on output format.
+                utils::print_table(output, volume.state.replica_topology);
+            }
+            Err(e) => {
+                println!("Failed to get volume {}. Error {}", id, e)
```
tiagolobocastro, line 129: on an error do we exit the process with an error code?

tiagolobocastro (approved):

kubectl-plugin/src/main.rs
paulyoong: I have changed this as you suggested.

common/src/types/v0/message_bus/volume.rs
paulyoong: No they don't - I've made them private and created a `new` method.

common/src/types/v0/message_bus/volume.rs
paulyoong: Thanks, done.

control-plane/rest/openapi-specs/v0_api_spec.yaml
paulyoong: Changed.

kubectl-plugin/src/resources/volume.rs
paulyoong: Not currently, we only print the error. I will raise a Jira for this.

paulyoong:
bors merge

paulyoong:
bors merge

PR #125, opened 2021-10-06
fix(csi): explicit timeout for rest api operatioins for csi controller

mtzaurus:
Explicit timeout is now applied to REST API client to prevent hangs
in case of not-accessible REST server / network issues.
Default I/O timeout is configured to 5 seconds.

Resolves: CAS-1140


control-plane/csi-controller/src/client.rs
```diff
@@ -108,7 +113,11 @@ impl MayastorApiClient {
             rest_client,
         });
 
-        debug!("API client is initialized with endpoint {}", endpoint);
+        debug!(
```
paulyoong, line 116: Would it be more appropriate to set this as an info message?

control-plane/csi-controller/src/main.rs
```diff
@@ -44,6 +46,12 @@ pub async fn main() -> Result<(), String> {
                 .env("JAEGER_ENDPOINT")
                 .help("enable open telemetry and forward to jaeger"),
         )
+        .arg(
+            Arg::with_name("timeout")
+                .short("-t")
+                .long("io-timeout")
```
paulyoong, line 52: Should we refer to this timeout as `rest-timeout` rather than `io-timeout`?

paulyoong (approved):

PR #127, opened 2021-10-06
refactor(msp): make use of the tower openapi client

tiagolobocastro:
This way we get the full tracing of the stack for pool operations
Remove seemingly unused crate dependencies


gila:
Thx for the cleanup with regards go unused dependencies🙏

gila (approved):

PR #128, opened 2021-10-07
test(replica scheduling): add bdd tests for volume topology based replica scheduling

Abhinandan-Purkait:
Add bdd tests to test the volume topology based scheduling of replicas on pools

Resolves: CAS-1156 and CAS-1116

### DIFF
--- a/tests/bdd/features/volume/topology.feature
+++ b/tests/bdd/features/volume/topology.feature
@@ -1,0 +1,47 @@
+Feature: Volume Topology
+  The volume topology feature is for selection of particular pools to schedule a replica on.
+  All the volume topology key and values should have an exact match with pool labels, although pools
+  can have extra labels, for the selection of that pool. No volume topology symbolizes the volume replicas
+  can be scheduled on any available pool.
+
+  Background:
+    Given a control plane, two Mayastor instances, two pools
+
+  Scenario: sufficient suitable pools which contain volume topology labels
+    Given a request for a volume with topology same as pool labels
+    When the number of volume replicas is less than or equal to the number of suitable pools
+    Then volume creation should succeed with a returned volume object with topology
+    And pool labels must contain all the volume request topology labels
+
+  Scenario: desired number of replicas cannot be created
+    Given a request for a volume with topology different from pools
+    When the number of suitable pools is less than the number of desired volume replicas
+    Then volume creation should fail with an insufficient storage error
+    And pool labels must not contain the volume request topology labels
+
+  Scenario: sufficient suitable pools which do not contain volume pool topology labels
+    Given a request for a volume without pool topology
+    When the number of volume replicas is less than or equal to the number of suitable pools
+    Then volume creation should succeed with a returned volume object without pool topology
+    And volume request should not contain any pool topology labels
+
+  Scenario: successfully adding a replica to a volume with pool topology
+    Given an existing published volume with a topology matching pool labels
+    And additional unused pools with labels containing volume topology
+    When a user attempts to increase the number of volume replicas
+    Then an additional replica should be added to the volume
+    And pool labels must contain all the volume topology labels
+
+  Scenario: cannot add a replica to a volume with pool topology labels
+    Given an existing published volume with a topology not matching pool labels
+    And a pool which does not contain the volume topology label
+    When a user attempts to increase the number of volume replicas
+    Then replica addition should fail with an insufficient storage error
+    And pool labels must not contain the volume topology labels
+
+  Scenario: successfully adding a replica to a volume without pool topology
+    Given an existing published volume without pool topology
+    And suitable available pools with labels
+    When a user attempts to increase the number of volume replicas
+    Then an additional replica should be added to the volume
+    And volume should not contain any pool topology labels
--- a/tests/bdd/test_volume_topology.py
+++ b/tests/bdd/test_volume_topology.py
@@ -1,0 +1,555 @@
+"""Volume Topology feature feature tests."""
+
+import docker
+import pytest
+import requests
+from openapi_client.model.volume_policy import VolumePolicy
+from pytest_bdd import (
+    given,
+    scenario,
+    then,
+    when,
+)
+
+import common
+from openapi.openapi_client.model.create_pool_body import CreatePoolBody
+from openapi.openapi_client.model.create_volume_body import CreateVolumeBody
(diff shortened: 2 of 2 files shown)


tests/bdd/test_volume_topology.py
```diff
@@ -0,0 +1,536 @@
...
+                }
+            }
+        },
+        _configuration=cfg,
+    )
+
+    # Check the volume object returned is as expected
+    request = create_request[CREATE_REQUEST_KEY]
+    volume = common.get_volumes_api().put_volume(VOLUME_UUID, request)
+    assert str(volume.spec) == str(expected_spec)
```
paulyoong, line 469: Because this function is testing that volume creation should succeed, it might be good to also add checks for the volume state i.e. is the volume online

tests/bdd/test_volume_topology.py
```diff
@@ -0,0 +1,536 @@
...
+    ) or common_labels(
+        create_request[CREATE_REQUEST_KEY]["topology"]["pool_topology"]["labelled"][
+            "inclusion"
+        ],
+        common.get_pools_api().get_pool(POOL_2_UUID),
+    ) == len(
+        create_request[CREATE_REQUEST_KEY]["topology"]["pool_topology"]["labelled"][
+            "inclusion"
+        ]
+    )
```
paulyoong, line 367: I think you might want to check all these complex assertions for parenthesis. At the moment it is of the form `a or b == len(c)` but I think it needs to be `(a or b) == len(c)`. The order of precedence matters here.

Additionally, in this statement I think the `or` is meant to be an `and` because we want to check that both pools contain the correct number of labels.

tests/bdd/test_volume_topology.py
```diff
@@ -0,0 +1,536 @@
...
+from openapi.openapi_client.model.create_pool_body import CreatePoolBody
+from openapi.openapi_client.model.create_volume_body import CreateVolumeBody
+from openapi.openapi_client.model.protocol import Protocol
+from openapi.openapi_client.model.spec_status import SpecStatus
+from openapi.openapi_client.model.volume_spec import VolumeSpec
+
+VOLUME_UUID = "5cd5378e-3f05-47f1-a830-a0f5873a1449"
+VOLUME_SIZE = 10485761
+NUM_VOLUME_REPLICAS = 1
+REST_SERVER = "http://localhost:8081/v0"
```
mtzaurus, line 24: It looks like this constant isn't used anywhere in the file.

tests/bdd/features/volume/topology.feature
```diff
@@ -0,0 +1,44 @@
...
+    Then volume creation should succeed with a returned volume object without pool topology
+    And volume request should not contain any pool topology labels
+
+  Scenario: successfully adding a replica to a volume with pool topology
+    Given a suitable available pool with labels containing volume topology
+    When a user attempts to increase the number of volume replicas
+    Then an additional replica should be added to the volume
+    And pool labels must contain all the volume topology labels
+
+  Scenario: cannot add a replica to a volume with pool topology labels
```
mtzaurus, line 34: Should this read 'cannot add a replica to a volume without pool topology labels' ?

tests/bdd/test_volume_topology.py
Abhinandan-Purkait: Yeah. This needs to be removed.

tests/bdd/test_volume_topology.py
paulyoong: Ah ok. Then we just need the parenthesis.

tests/bdd/test_volume_topology.py
```diff
@@ -0,0 +1,536 @@
...
+        common_labels(
+            volume["spec"]["topology"]["pool_topology"]["labelled"]["inclusion"],
+            common.get_pools_api().get_pool(POOL_2_UUID),
+        )
+        == 0
+    )
+
+
+@then("pool labels must not contain the volume request topology labels")
+def pool_labels_must_not_contain_the_volume_request_topology_labels(create_request):
```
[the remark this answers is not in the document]
Abhinandan-Purkait: Yeah, that would make it simpler, i.e no of suitable pools is 0 indirectly. I went by the wording and thus compared the no. of common labels.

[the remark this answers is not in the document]
Abhinandan-Purkait: Yeah. That would make it simpler, we can just check the no of suitable pool to be zero here. I went by the wording and checked the common labels.

tests/bdd/features/volume/topology.feature
```diff
@@ -0,0 +1,44 @@
...
+    Given pool which does not contain the volume topology label
```
paulyoong, line 35: Again, maybe we can split this clause to simplify the corresponding function.

tests/bdd/features/volume/topology.feature
```diff
@@ -0,0 +1,44 @@
...
+    And pool labels must not contain the volume request topology labels
+
+  Scenario: sufficient suitable pools which do not contain volume pool topology labels
+    Given a request for a volume without pool topology
+    When the number of volume replicas is less than or equal to the number of suitable pools
+    Then volume creation should succeed with a returned volume object without pool topology
+    And volume request should not contain any pool topology labels
+
+  Scenario: successfully adding a replica to a volume with pool topology
+    Given a suitable available pool with labels containing volume topology
```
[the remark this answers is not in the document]
Abhinandan-Purkait: Yep. Have split this.

tests/bdd/test_volume_topology.py
```diff
@@ -0,0 +1,555 @@
...
+
+@given("additional unused pools with labels containing volume topology")
+def additional_unused_pools_with_labels_containing_volume_topology():
+    """additional unused pools with labels containing volume topology."""
+    volume = common.get_volumes_api().get_volume(VOLUME_UUID)
+    assert (
+        no_of_suitable_pools(
+            volume["spec"]["topology"]["pool_topology"]["labelled"]["inclusion"]
+        )
+        != 0
```
paulyoong, line 364: Should we check that the number of labelled pools is greater than the current number of volume replicas? That way we know that there are additional pools that can be used. If we just check that the number of suitable pools is not zero, we could potentially already have replicas on them and so won't be able to reuse them.

paulyoong (approved):

PR #134, opened 2021-10-11
chore: generate openapi from nix preBuild

tiagolobocastro:
Explicitly regenerate the openapi from the nix pre build.
This allows us to use env for the generate-openapi script again.

Expands #129 to allow for the services to be built within the nix-build.


tiagolobocastro:
bors merge

PR #135, opened 2021-10-11
feat(replicas): garbage collector

paulyoong:
Add a garbage collector which destroys orphaned replicas. Orphaned
replicas are replicas that are managed (i.e. have been created through
the control plane) but are not owned by anything (and therefore are no
longer used).

Resolves: CAS-992, CAS-1159

### DIFF
--- a/control-plane/agents/core/src/core/reconciler/poller.rs
+++ b/control-plane/agents/core/src/core/reconciler/poller.rs
@@ -1,5 +1,5 @@
 use crate::core::{
-    reconciler::{nexus, persistent_store::PersistentStoreReconciler, pool, volume},
+    reconciler::{nexus, persistent_store::PersistentStoreReconciler, pool, replica, volume},
     registry::Registry,
     task_poller::{squash_results, PollContext, PollEvent, PollResult, PollerState, TaskPoller},
 };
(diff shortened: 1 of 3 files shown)


control-plane/agents/core/src/core/reconciler/poller.rs
```diff
@@ -28,6 +28,7 @@ impl ReconcilerWorker {
             Box::new(nexus::NexusReconciler::new()),
             Box::new(volume::VolumeReconciler::new()),
             Box::new(PersistentStoreReconciler::new()),
+            Box::new(replica::garbage_collector::GarbageCollector::new()),
```
tiagolobocastro, line 31: just a suggestion, to keep in line with the existing nexus and volumes, the garbage_collector could be made private within the replica as ReplicaReconciler, which in turns calls its own garbage collector

control-plane/agents/core/src/core/reconciler/replica/garbage_collector.rs
```diff
@@ -0,0 +1,73 @@
...
+        Self {
+            counter: PollTimer::from(1),
+        }
+    }
+}
+
+#[async_trait::async_trait]
+impl TaskPoller for GarbageCollector {
+    async fn poll(&mut self, context: &PollContext) -> PollResult {
+        let _ = destroy_orphaned_replicas(context).await;
```
tiagolobocastro, line 25: nit: you could just return `destroy_orphaned_replicas(context).await`

PR #137, opened 2021-10-11
feat(volume): add volume nexus/replica garbage collector

tiagolobocastro:
feat(nexus): add garbage collector

Nexuses which belong to a volume which no longer exists are disowned
Nexuses which are no longer used by a volume are disowned
Nexuses with no owner are deleted

--------
fix(replica): disown unused replicas

When a volume is published with offline replicas they are not used by the
nexus. In this case we should disown these replicas since we'll have to
do a full rebuild anyway.

Disowned replicas will then garbage collected when they reappear.

--------
chore(deployer): make the csi deployment optional

The csi is not usually needed when we use the deployer so disable it by default.
It has been explicitly enabled in the bdd tests.

--------
test: add volume garbage collection cargo tests

### DIFF
--- a/control-plane/agents/core/src/core/reconciler/nexus/garbage_collector.rs
+++ b/control-plane/agents/core/src/core/reconciler/nexus/garbage_collector.rs
@@ -1,0 +1,123 @@
+use crate::core::{
+    reconciler::{PollContext, TaskPoller},
+    specs::{OperationSequenceGuard, SpecOperations},
+    task_poller::{PollEvent, PollResult, PollTimer, PollerState},
+};
+use common_lib::types::v0::{
+    message_bus::DestroyNexus,
+    store::{nexus::NexusSpec, OperationMode, TraceSpan},
+};
+
+use parking_lot::Mutex;
+use std::sync::Arc;
+
+/// Nexus Garbage Collector reconciler
+#[derive(Debug)]
+pub(super) struct GarbageCollector {
+    counter: PollTimer,
+}
+impl GarbageCollector {
+    /// Return a new `Self`
+    pub(super) fn new() -> Self {
+        Self {
+            counter: PollTimer::from(5),
+        }
+    }
+}
+
+#[async_trait::async_trait]
+impl TaskPoller for GarbageCollector {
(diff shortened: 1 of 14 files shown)


control-plane/agents/core/src/core/reconciler/nexus/mod.rs
```diff
@@ -61,6 +66,11 @@ impl TaskPoller for NexusReconciler {
             };
             results.push(nexus_reconciler(&nexus, context, OperationMode::ReconcileStep).await);
         }
+        println!("AAA");
+        for target in &mut self.poll_targets {
+            println!("AAABBBB");
```
paulyoong, line 71: Can remove these prints.

control-plane/agents/core/src/core/reconciler/nexus/garbage_collector.rs
```diff
@@ -0,0 +1,113 @@
...
+    let mut results = vec![];
+    results.push(garbage_collector_reconcile1(nexus_spec, context, mode).await);
+    results.push(garbage_collector_reconcile2(nexus_spec, context, mode).await);
+    GarbageCollector::squash_results(results)
+}
+
+/// Given a control plane managed nexus
+/// When a nexus is not owned by a volume anymore
+/// Then it should eventually be destroyed
+async fn garbage_collector_reconcile1(
```
paulyoong, line 63: Naming suggestion: `destroy_disowned_nexus`?

control-plane/agents/core/src/core/reconciler/nexus/garbage_collector.rs
```diff
@@ -0,0 +1,113 @@
...
+            .await?;
+    }
+
+    PollResult::Ok(PollerState::Idle)
+}
+
+/// Given a control plane managed nexus
+/// When a nexus is owned by a volume which no longer exists
+/// Then it should eventually be destroyed
+async fn garbage_collector_reconcile2(
```
paulyoong, line 88: Naming suggestion: `destroy_orphaned_nexus`?

PR #142, opened 2021-10-13
fix(csi): determine volume locality based on volume config

mtzaurus:
Volume locality and accessible topology is now calculated based
on 'local' flag provided upon volume creation.

Resolves: CAS-1126, CAS-1162

### DIFF
--- a/control-plane/csi-controller/src/client.rs
+++ b/control-plane/csi-controller/src/client.rs
@@ -31,6 +31,28 @@ pub enum ApiClientError {
     MalformedUrl(String),
 }
 
+/// Placeholder for volume topology for volume creation operation.
+#[derive(Debug)]
+pub struct CreateVolumeTopology {
+    inclusive_label_topology: HashMap<String, String>,
+    allowed_nodes: Vec<String>,
+    preferred_nodes: Vec<String>,
+}
+
+impl CreateVolumeTopology {
+    pub fn new(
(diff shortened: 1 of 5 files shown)


tests/bdd/test_csi_controller.py
```diff
@@ -596,6 +663,13 @@ def _create_1_replica_nvmf_volume():
     csi_delete_1_replica_nvmf_volume1()
 
 
+@pytest.fixture
+def _create_1_replica_local_nvmf_volume():
+    csi_delete_1_replica_local_nvmf_volume1()
```
paulyoong, line 668: Why do we need to delete it first? We will do this as part of the cleanup after the `yield` below won't we?

control-plane/csi-controller/src/client.rs
```diff
@@ -31,6 +31,28 @@ pub enum ApiClientError {
     MalformedUrl(String),
 }
 
+/// Placeholder for volume topology for volume creation operation.
+#[derive(Debug)]
+pub struct CreateVolumeTopology {
+    inclusive_label_topology: HashMap<String, String>,
```
paulyoong, line 37: Would it be better to define a type rather than use `HashMap<String, String>` directly. That way we can change the key/value types in one place if we need to.

paulyoong (approved):

PR #145, opened 2021-10-15
test(BDD): replica garbage collection

paulyoong:
Add a BDD test to check that an orphaned replica is eventually garbage
collected.

Resolves: CAS-1172


tests/bdd/test_replicas_garbage_collection.py
```diff
@@ -0,0 +1,147 @@
...
+    except Exception as e:
+        # A Mayastor node is inaccessible, so deleting the volume will fail because the replica
+        # on this node cannot be destroyed. Attempting to do so results in a timeout. This is
+        # expected and results in a replica being orphaned.
+        exception_info = e.__dict__
+        assert exception_info["status"] == requests.codes["request_timeout"]
+        pass
+
+    # Allow some time for the replica to be disowned.
+    time.sleep(5)
```
tiagolobocastro, line 123: should this be a retry loop with small sleeps but a larger timeout to avoid failing if it ever takes longer for whatever reason?

paulyoong:
bors merge

PR #148, opened 2021-10-19
fix: allow volume destruction whilst node offline

paulyoong:
Volume destruction will now complete successfully even if a node with a
replica or nexus on it is offline. In the event that the node is
inaccessible the volume resources (replicas and nexus) are completely
disowned allowing them to be garbage collected at a later time.

Resolves: CAS-1180

### DIFF
--- a/control-plane/agents/core/src/volume/specs.rs
+++ b/control-plane/agents/core/src/volume/specs.rs
@@ -402,7 +402,10 @@ impl ResourceSpecsLocked {
         registry.get_volume(&request.uuid).await
     }
 
-    /// Destroy a volume based on the given `DestroyVolume` request
+    /// Destroy a volume based on the given `DestroyVolume` request.
+    /// Volume destruction will succeed even if the nexus or replicas cannot be destroyed (i.e. due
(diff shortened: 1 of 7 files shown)


control-plane/agents/core/src/volume/specs.rs
```diff
@@ -413,17 +413,18 @@ impl ResourceSpecsLocked {
         if let Some(volume) = &volume {
             SpecOperations::start_destroy(volume, registry, false, mode).await?;
 
-            let mut first_error = Ok(());
             let nexuses = self.get_volume_nexuses(&request.uuid);
             for nexus in nexuses {
                 let nexus = nexus.lock().deref().clone();
                 if let Err(error) = self
-                    .destroy_nexus(registry, &DestroyNexus::from(nexus), true, mode)
+                    .destroy_nexus(registry, &DestroyNexus::from(nexus.clone()), true, mode)
```
tiagolobocastro, line 420: Is it worth adding a note above the `destroy_volume` function that it success even if it can't delete the data replicas? (might be relevant for secure erase)

paulyoong: Good suggestion, done.

PR #150, opened 2021-10-21
fix(pool CRD): set to unknown when pool is missing 

paulyoong:
fix(pool CRD): set to unknown when pool is missing 

A pool CRD should appropriately reflect the state of a pool if the node
on which it resides becomes inaccessible.

When a node becomes inaccessible the control plane is unable to retrieve
any runtime state information for resources on that node. As a result,
the cached state information is invalidated (set to None). If a REST
request is made for a resource on an inaccessible node, the returned
object will not contain any state information.

If the MSP operator receives a pool object (via REST) without a state,
it cannot determine the health of the pool. Consequently, it will now
reflect this in the CRD as an 'Unknown' state.

The MSP operator will periodically check if the pool state is available.
Once this is received, the state of the CRD will be updated
appropriately (that is, it will be set to Online if the pool returns in
a healthy state).

-----------
fix(MSP operator): allow importing of pools

If a pool already exists in the control plane, the MSP operator should
import that pool.

Resolves: CAS-1183

### DIFF
--- a/control-plane/msp-operator/src/main.rs
+++ b/control-plane/msp-operator/src/main.rs
@@ -472,25 +472,32 @@ impl ResourceContext {
         );
 
         let body = CreatePoolBody::new_all(self.spec.disks.clone(), labels);
-        let res = self
+        match self
             .pools_api()
             .put_node_pool(&self.spec.node, &self.name(), body)
-            .await?;
+            .await
+        {
+            Ok(_) => {}
+            Err(clients::tower::Error::Response(response))
+                if response.status() == clients::tower::StatusCode::UNPROCESSABLE_ENTITY =>
+            {
+                // UNPROCESSABLE_ENTITY indicates that the pool spec already exists in the control
+                // plane. So we want to update the CRD to 'Created' to reflect this.
+            }
+            Err(e) => {
+                return Err(e.into());
+            }
+        };
 
-        if matches!(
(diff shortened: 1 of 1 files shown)


control-plane/msp-operator/src/main.rs
```diff
@@ -606,7 +637,10 @@ impl ResourceContext {
                 }
             }
         } else {
-            warn!("CRD does not contain the valid fields we except");
+            // There is no pool state, so we can't determine the health of the pool. Reflect this in
+            // the CRD as an 'Unknown' state.
+            debug!("Pool state is missing. Setting MSP state to 'Unknown'.");
```
tiagolobocastro, line 642: Should we check if we're already unknown to avoid logging this everytime?

tiagolobocastro (approved):

control-plane/msp-operator/src/main.rs
```diff
@@ -797,6 +831,13 @@ async fn reconcile(
             return msp.pool_check().await;
         }
 
+        Some(MayastorPoolStatus {
```
gila, line 834: We can move this to the above lines, where we check if the pool is Online. Something like:

```rust
match { 
Some(A) | Some(B) => { }
}
```

paulyoong:
bors merge

PR #152, opened 2021-10-21
fix(core): clear the states when a node goes offline

tiagolobocastro:
When a node goes offline, its states were not immediately cleared.
Make sure we always clear the states when the node is not online by making
it implicit when we set the node status.

Prevents us from returning Pools/Nexuses as online but the returning the node as offline, which is confusing at the very least..


tiagolobocastro:
what...
bors merge

tiagolobocastro:
bors merge

PR #156, opened 2021-10-28
feat(msp): normalize device schemas

gila:
As part of the MSP operator implementation support was added to validate
that devices are present. This provides more feedback to the user when
using the event system from k8s.

However, when device schemas are provided in the spec, we should
normalize before comparing them to the block devices on the targeted
nodes.


tiagolobocastro:
bors merge

tiagolobocastro:
bors doesn't like this PR :) let's try again...
bors merge

paulyoong:
bors merge

PR #159, opened 2021-10-29
fix(nats): set tcp read timeouts

tiagolobocastro:
chore(nats): use the nats with a stuck tcp stream fix

The tcp stream to the nats-server can get stuck if the server is not
shutdown gracefully. Even the connection lost callback does not get
triggered because the stream is somehow stuck.

----------------
fix(nats): set tcp read timeouts

The tcp stream to the nats-server can get stuck if the server is not
shutdown gracefully. As such Even the connection lost callback does not get
triggered because the stream is somehow stuck.

Adding the timeout if we do not receive a message with a timeout helps
address the issue: when the timeout expires the client now sends a ping
to the server and expects to receive a pong within the same timeout.
If a pong does not arrive in time then the connection is deemed lost.

As a side effect this may cause additional strain on the nats server when
too many clients are idle as the timeouts will be fired.

Don't cache connect_url's returned by the NATS INFO message as they may
get outdated and point to bogus IP addresses.

Also add connection handlers to highlight whenever the services connection
to the NATS server changes.

Resolves CAS-1192

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -255,7 +255,6 @@ dependencies = [
  "humantime",
  "itertools",
  "lazy_static",
- "nats 0.15.2",
  "once_cell",
  "opentelemetry",
  "opentelemetry-jaeger",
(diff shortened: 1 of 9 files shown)


tiagolobocastro:
> The fix looks good. Quick question about the fix in the local nats branch: is it worth opening a ticket against the mainstream source to make sure such a good feature is available as part of generic NATS source ?

Yep I have a PR open against the mainstream repo, but it seems this fix is not quite working - it fixes the connection when a node crashes, but when it's only the pod that crashes/restarts (which works without the fix) then it doesn't work and we end up in a connect/reconnect storm... back to debugging

paulyoong (approved):

tiagolobocastro:
bors merge

PR #161, opened 2021-11-02
feat(replica reconciler): missing replica owners

paulyoong:
Include a reconcile loop which removes any replica owners which no
longer exist.

Resolves: CAS-1203

### DIFF
--- a/control-plane/agents/core/src/core/reconciler/replica/mod.rs
+++ b/control-plane/agents/core/src/core/reconciler/replica/mod.rs
@@ -1,3 +1,6 @@
+#[cfg(test)]
+mod tests;
+
 use crate::core::{
     specs::{OperationSequenceGuard, SpecOperations},
     task_poller::{PollContext, PollEvent, PollResult, PollTimer, PollerState, TaskPoller},
@@ -15,7 +18,7 @@ impl ReplicaReconciler {
     /// Return a new `Self`
     pub fn new() -> Self {
         Self {
-            counter: PollTimer::from(1),
+            counter: PollTimer::from(5),
         }
     }
 }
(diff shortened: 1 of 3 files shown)


control-plane/agents/core/src/core/reconciler/replica/mod.rs
```diff
@@ -23,7 +26,10 @@ impl ReplicaReconciler {
 #[async_trait::async_trait]
 impl TaskPoller for ReplicaReconciler {
     async fn poll(&mut self, context: &PollContext) -> PollResult {
-        destroy_orphaned_replicas(context).await
+        let mut results = vec![];
+        results.push(disown_missing_owners(context).await);
+        results.push(destroy_orphaned_replicas(context).await);
+        Self::squash_results(results)
```
tiagolobocastro, line 32: I wonder if the pollTimer here should be set to 5 rather than 1, similar to what the volume does for garbage collection.

control-plane/agents/core/src/core/reconciler/replica/tests.rs
```diff
@@ -0,0 +1,86 @@
...
+    etcd.put_obj(&replica)
+        .await
+        .expect("Failed to store modified replica.");
+
+    // Restart the core agent so that it reloads the modified replica spec from the persistent
+    // store.
+    cluster.restart_core().await;
+
+    // Allow time for the core agent to restart.
+    sleep(Duration::from_secs(2));
```
tiagolobocastro, line 75: Should this be a tokio sleep? Though it probably doesn't matter here.

tiagolobocastro (approved):

control-plane/agents/core/src/core/reconciler/replica/mod.rs
paulyoong: Done.

control-plane/agents/core/src/core/reconciler/replica/tests.rs
paulyoong: I've left it as it is for now. As you say, I don't think it matters.

paulyoong:
bors merge

paulyoong:
bors merge

PR #162, opened 2021-11-02
fix(nexus_v2): use nexus create v2 api

tiagolobocastro:
This allows the control plane to create a nexus with a name and a uuid.
The name is set to the k8s volume uuid. This is required, otherwise the
csi node plugin cannot attach/detach the nexus properly.

Note: ignored the test "nexus_share_transaction_store" as the latest
mayastor now updates etcd so the test can no longer pause etcd as it is
as it blocks mayastor. TODO: fixme


paulyoong (approved):

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

PR #164, opened 2021-11-04
test(bdd): move the openapi_client package into openapi

tiagolobocastro:
test(bdd): move the openapi_client package into openapi

use the openapi client directly so python can determine we're using the
correct type and avoids having to specify discard_unknown_keys=true.

---------------
fix(test): check if replica is deleted by checking the spec

---------------
fix(bdd): use actual openapi Topology types


paulyoong (approved):

tiagolobocastro:
bors merge

PR #165, opened 2021-11-05
refactor(bdd): refactor bdd common

tiagolobocastro:
Add new common layout and separate functionality into separate classes

In preparation of the ANA bdd test


gila (approved):

tests/bdd/test_csi_identity.py
```diff
@@ -8,22 +8,21 @@
...
+from common.csi import CsiHandle
+from common.deployer import Deployer
+from common.apiclient import ApiClient
 
 
 @pytest.fixture(scope="module")
 def setup():
-    common.deployer_stop()
-    common.deployer_start(1)
+    Deployer.stop()
```
paulyoong, line 21: Nit: Do we need this stop? Will the start just restart it? Plus, if all goes well, we call stop after the yield anyway.

paulyoong (approved):

tests/bdd/test_csi_identity.py
tiagolobocastro: No, I don't think we don't, I'll remove it :+1:

tiagolobocastro:
bors merge

PR #166, opened 2021-11-08
test(bdd): add ana swap test

tiagolobocastro:
test(core): add env variable to setup random min ctrl id

Tidy up the nexus nvmf config a bit.
If the env var "TEST_NEXUS_NVMF_ANA_ENABLE" is set then the min controller
will be randomly generated.
NODE: This is only useful for the ana swap test as there is no other way
of influencing this.

----------------
test(bdd): add ana swap test

This tests the ana swap by:
1. enabling ana
2. publishing a volume to an ana host
3. running fio
4. killing the volume target (nexus)
5. republish the volume on another mayastor
6. connect to the new target
7. remove the old target
8. fio completes with no errors

### DIFF
--- a/common/src/types/v0/message_bus/nexus.rs
+++ b/common/src/types/v0/message_bus/nexus.rs
@@ -179,6 +179,89 @@ pub struct CreateNexus {
     pub managed: bool,
     /// Volume which owns this nexus, if any
     pub owner: Option<VolumeId>,
+    /// Nexus Nvmf Configuration
+    pub config: Option<NexusNvmfConfig>,
+}
+
+/// Nvmf Controller Id Range
+#[derive(Serialize, Deserialize, Debug, Clone, Eq, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct NvmfControllerIdRange(std::ops::RangeInclusive<u16>);
+impl NvmfControllerIdRange {
+    /// create `Self` with a random minimum controller id
+    pub fn random_min() -> Self {
+        let min = *Self::controller_id_range().start();
+        let max = *Self::controller_id_range().end();
+        let rand_min = u16::min(rand::random::<u16>() + min, max);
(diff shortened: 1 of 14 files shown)


common/src/types/v0/message_bus/nexus.rs
```diff
@@ -179,6 +179,89 @@ pub struct CreateNexus {
...
+    }
+    /// minimum controller id
+    pub fn min(&self) -> u16 {
+        self.0.start
+    }
+    /// maximum controller id
+    pub fn max(&self) -> u16 {
+        self.0.end
+    }
+    fn valid_controller_range() -> std::ops::Range<u16> {
```
paulyoong, line 206: Nit: This could just be "controller_id_range"

tests/bdd/test_ana_validate_nexus_swap.py
```diff
@@ -0,0 +1,221 @@
...
+@pytest.fixture
+def publish_to_node_2(background):
+    volume = background
+    device_uri = volume.state["target"]["deviceUri"]
+
+    try:
+        ApiClient.volumes_api().del_volume_target(VOLUME_UUID)
+    except ApiException as e:
+        # Time out or node not online
+        assert e.status == 408 or e.status == 412
```
paulyoong, line 166: Maybe we can define constants for the error codes so that it reads a little easier.

paulyoong (approved):

common/src/types/v0/message_bus/nexus.rs
tiagolobocastro: Done

tests/bdd/test_ana_validate_nexus_swap.py
tiagolobocastro: Done

tiagolobocastro:
bors merge

PR #167, opened 2021-11-09
refactor(store volume spec): add v0 versioning path to store

Abhinandan-Purkait:
Involves [CAS-1237](https://mayadata.atlassian.net/browse/CAS-1237)

### DIFF
--- a/common/src/constants.rs
+++ b/common/src/constants.rs
@@ -29,3 +29,6 @@ pub const OPENEBS_CREATED_BY_KEY: &str = "openebs.io/created-by";
 
 /// The value to mark the creation source of a pool to be msp-operator in labels
 pub const MSP_OPERATOR: &str = "msp-operator";
+
+/// The value to mark v0 version of resources to be stored in persistent store
+pub const V0: &str = "v0";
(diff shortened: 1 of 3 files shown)


common/src/constants.rs
```diff
@@ -29,3 +29,6 @@ pub const OPENEBS_CREATED_BY_KEY: &str = "openebs.io/created-by";
 
 /// The value to mark the creation source of a pool to be msp-operator in labels
 pub const MSP_OPERATOR: &str = "msp-operator";
+
+/// The value to mark v0 version of resources to be stored in persistent store
+pub const V0: &str = "v0";
```
gila, line 34: I suggest we think this through a bit. Why is for example are we not using an enum? This makes it a whole lot easier dealing with it later on, right?

common/src/types/v0/store/definitions.rs
```diff
@@ -159,16 +163,17 @@ pub enum StorableObjectType {
...
-        obj_type.to_string()
+        version,
+        obj_type.to_string(),
     )
 }
 
 /// create a key based on the object's key trait
 /// todo: version properly
 pub fn get_key<K: ObjectKey + ?Sized>(k: &K) -> String {
-    format!("{}/{}", key_prefix(k.key_type()), k.key_uuid())
```
gila, line 173: Can we move this into the trait ObjectKey ?

gila (changes requested):

PR #168, opened 2021-11-09
fix(bug): don't disown unused replicas

tiagolobocastro:
When the nexus is not in the created phase don't disown replicas, wait
until the nexus is created so that we have a clear picture of what's
unused.

Also, make sure we don't disown healthy replicas.


paulyoong (approved):

tiagolobocastro:
bors merge

PR #171, opened 2021-11-17
fix(msp cr status): add fix for msp cr going to error state on node down

Abhinandan-Purkait:
Fix for the bug which causes the CR status to go from CREATING to ERROR.
Added checks to handle PRECONDITION_FAILED errors.

Resolves [CAS-1244](https://mayadata.atlassian.net/browse/CAS-1244)

### DIFF
--- a/control-plane/msp-operator/src/main.rs
+++ b/control-plane/msp-operator/src/main.rs
@@ -356,6 +356,7 @@ impl ResourceContext {
     fn pools_api(&self) -> &dyn openapi::apis::pools_api::tower::client::Pools {
         self.ctx.http.pools_api()
     }
+
     fn block_devices_api(
         &self,
     ) -> &dyn openapi::apis::block_devices_api::tower::client::BlockDevices {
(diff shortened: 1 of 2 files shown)


paulyoong (approved):

control-plane/msp-operator/src/main.rs
```diff
@@ -435,75 +436,91 @@ impl ResourceContext {
...
+            Err(_) => match self.pools_api().get_pool(&self.name()).await {
+                Ok(response) => {
+                    let pool = response.into_body();
+                    // As pool exists, set the status based on the presence of pool state.
+                    self.set_status_or_unknown(pool).await
+                }
+                Err(_) => {
+                    // If we don't find the pool, i.e. its not present or not yet created
+                    // so, set the status to Creating to retry creation.
+                    return self.is_missing().await;
```
tiagolobocastro, line 520: In this case should we not go into the error state? Otherwise, if we never created the pool and the node is not online, do we stay in creating forever?

tiagolobocastro (approved):

control-plane/msp-operator/src/main.rs
Abhinandan-Purkait: Yeah, we should move to Error, as if the node is not online and the pool didnot exist previously, there is no point in staying in Creating forever. Thanks.

Abhinandan-Purkait:
bors merge

PR #173, opened 2021-11-23
fix(hotspare): schedule replicas individually

tiagolobocastro:
When >1 replicas fail at the same time, the core agent was running the
scheduling algorithm once and then using the result to create new replicas.

The fix is to run the scheduling algorithm for each new replica we're adding.
This way we always make sure that we respect the scheduling rules.


Abhinandan-Purkait (approved):

mtzaurus (approved):

tiagolobocastro:
bors merge

PR #174, opened 2021-11-24
fix(invalid sc parameters): csi-controller to not create volume on invalid sc parameters

Abhinandan-Purkait:
CSI-Controller will not create volume if the storage class parameters are invalid.
Resolves - [CAS-1211](https://mayadata.atlassian.net/browse/CAS-1211)

### DIFF
--- a/control-plane/csi-controller/src/controller.rs
+++ b/control-plane/csi-controller/src/controller.rs
@@ -16,7 +16,7 @@ use rpc::csi::Topology as CsiTopology;
 const K8S_HOSTNAME: &str = "kubernetes.io/hostname";
 const VOLUME_NAME_PATTERN: &str =
     r"pvc-([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})";
-const PROTO_NVMF: &str = "nvmf";
+const SUPPORTED_FS_TYPES: [&str; 2] = ["ext4", "xfs"];
 const MAYASTOR_NODE_PREFIX: &str = "mayastor://";
 const MAX_VOLUMES_TO_LIST: usize = 1024 * 1024;
 
@@ -41,6 +41,15 @@ mod volume_opts {
     }
 }
 
+/// Check whether the passed fs type is supported or not,
+/// if not provided we call it valid as fsType is an optional parameter
+pub fn valid_fs_type(fs_type: Option<&String>) -> bool {
+    match fs_type {
+        Some(fs) => SUPPORTED_FS_TYPES.iter().any(|p| p == fs),
+        None => true,
+    }
+}
+
 /// Check whether target volume capabilites are valid. As of now, only
 /// SingleNodeWriter capability is supported.
 fn check_volume_capabilities(capabilities: &[VolumeCapability]) -> Result<(), tonic::Status> {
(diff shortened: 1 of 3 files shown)


control-plane/csi-controller/src/controller.rs
```diff
@@ -233,10 +234,23 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
             }
         };
 
+        // Check filesystem type.
+        if let Some(fs) = args.parameters.get("fsType") {
+            if !SUPPORTED_FS_TYPES.iter().any(|f| f == &fs.to_string()) {
```
tiagolobocastro, line 239: could have a util function that checks this in case we need to reuse?

control-plane/csi-controller/src/controller.rs
```diff
@@ -233,10 +234,23 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
...
+                return Err(Status::invalid_argument("Invalid filesystem type"));
+            }
+        };
+
         // Check storage protocol.
         let protocol = match args.parameters.get("protocol") {
-            Some(p) => p.to_string(),
-            None => return Err(Status::invalid_argument("Missing storage protocol")),
+            Some(p) => {
```
tiagolobocastro, line 246: We already have a `parse_protocol` which is used on publish and still accepts icsi there.
Instead why not call that function here, maybe change it to receive an option an reuse it so the logic will be in a single function, eg:
```rust
let protocol = parse_protocol(args.parameters.get("protocol"))?;
```

tiagolobocastro (approved):

control-plane/csi-controller/src/controller.rs
```diff
@@ -233,10 +242,15 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
...
+        // Check filesystem type.
+        if !valid_fs_type(args.parameters.get("fsType")) {
+            return Err(Status::invalid_argument("Invalid filesystem type"));
+        }
+
         // Check storage protocol.
-        let protocol = match args.parameters.get("protocol") {
-            Some(p) => p.to_string(),
-            None => return Err(Status::invalid_argument("Missing storage protocol")),
+        let protocol = match parse_protocol(args.parameters.get("protocol")) {
```
tiagolobocastro, line 251: why not use the actual type rather than converting to string?

control-plane/csi-controller/src/controller.rs
```diff
@@ -394,13 +408,9 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
...
         }
 
-        let protocol = match args.volume_context.get("protocol") {
-            Some(p) => parse_protocol(p)?,
-            None => {
-                return Err(Status::invalid_argument(
-                    "No protocol specified for publish volume request",
-                ))
-            }
+        let protocol = match parse_protocol(args.volume_context.get("protocol")) {
```
tiagolobocastro, line 411: you don't need match, you can do:
```rust
let protocol = parse_protocol(args.volume_context.get("protocol"))?;
```

control-plane/csi-controller/src/controller.rs
Abhinandan-Purkait: The code below thats not changed uses this uses string comparison. Line 258. But yeah we can change it there as well.

mtzaurus: Yes, we could remove this as well, as it's not used anywhere else:

```
const PROTO_NVMF: &str = "nvmf";
```

mtzaurus (approved):

control-plane/csi-controller/src/controller.rs
Abhinandan-Purkait: Yes, I will remove this.

control-plane/csi-controller/src/controller.rs
```diff
@@ -233,15 +241,20 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
...
+            return Err(Status::invalid_argument("Invalid filesystem type"));
+        }
+
         // Check storage protocol.
-        let protocol = match args.parameters.get("protocol") {
-            Some(p) => p.to_string(),
-            None => return Err(Status::invalid_argument("Missing storage protocol")),
+        let protocol = match parse_protocol(args.parameters.get("protocol")) {
+            Ok(p) => p,
+            Err(e) => return Err(e),
```
tiagolobocastro, line 252: you can also use `?` here to avoid the match.

Abhinandan-Purkait: Done.

Abhinandan-Purkait:
bors merge

Abhinandan-Purkait:
bors merge

PR #177, opened 2021-12-01
fix(replica): decrementing volume replica count

paulyoong:
When decrementing the number of volume replicas, it is possible that the
runtime number of replicas already matches the desired number of
replicas. In this case no replica removal candidates will be identified
and the operation can simply be completed successfully (ensuring the
persistent store is updated).

Resolves: CAS-1251

### DIFF
--- a/control-plane/agents/common/src/errors.rs
+++ b/control-plane/agents/common/src/errors.rs
@@ -503,7 +503,7 @@ impl From<SvcError> for ReplyError {
                 extra: error.full_string(),
             },
             SvcError::ReplicaRemovalNoCandidates { .. } => ReplyError {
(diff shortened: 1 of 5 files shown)


tiagolobocastro (approved):
I wonder if `ChildSorters` should also take into account the ReplicaState, and with all else being equal should prefer to remove replicas with no state.

paulyoong:
> I wonder if `ChildSorters` should also take into account the `ReplicaState`, and with all else being equal should prefer to remove replicas with no state.

It's a good suggestion. We should probably think through the general strategy that we wish to take in these kinds of scenarios as they will likely be relevant to other resources too.

paulyoong:
bors merge

PR #178, opened 2021-12-01
refactor: tidy up the bdd and the scripts layout

tiagolobocastro:
Change the layout of the pytest bdd:
the features are now further broken into subfolders and the respective tests and moved together
with the feature files.

Rather than modifying the PYTHONPATH we now make use of *.pth files which works similarly but means
the IDE's like CLion will also benefit from it and allows for better code navigation.

Extend the bdd setup script to generate the csi, openapi and deployer binaries. This means on a
clean workspace, this script will setup all the prerequisites for testing the bdd features.

The bdd test script now uses the setup so it becomes simpler. Also allow for arguments to be passed
into the script which get sent directly to pytest.

Also refactor the scripts layout into folders.

CAS-1252


tiagolobocastro:
bors try

PR #179, opened 2021-12-02
refactor: tidy up the bdd and the scripts layout

tiagolobocastro:
Change the layout of the pytest bdd:
the features are now further broken into subfolders and the respective tests and moved together
with the feature files.

Rather than modifying the PYTHONPATH we now make use of *.pth files which works similarly but means
the IDE's like CLion will also benefit from it and allows for better code navigation.

Extend the bdd setup script to generate the csi, openapi and deployer binaries. This means on a
clean workspace, this script will setup all the prerequisites for testing the bdd features.

The bdd test script now uses the setup so it becomes simpler. Also allow for arguments to be passed
into the script which get sent directly to pytest.

Also refactor the scripts layout into folders.

Resolves: CAS-1252


openapi/build.rs
```diff
@@ -14,4 +14,9 @@ fn main() {
     }
 
     println!("cargo:rerun-if-changed=../nix/pkgs/openapi-generator");
+    println!("cargo:rerun-if-changed=../control-plane/rest/openapi-specs");
+    // seems the internal timestamp is taken before build.rs runs, so we can't set this
+    // directive against files created during the build of build.rs??
+    // https://doc.rust-lang.org/cargo/reference/build-scripts.html#rerun-if-changed
+    // println!("cargo:rerun-if-changed=.");
```
paulyoong, line 21: Can this line be deleted?

scripts/python/test.sh
```diff
@@ -0,0 +1,29 @@
...
+# Eg: FAST=1 scripts/python/test.sh tests/bdd/features/volume/create/test_feature.py -k test_sufficient_suitable_pools
+
+set -e
+
+SCRIPT_DIR="$(dirname "$0")"
+export ROOT_DIR="$SCRIPT_DIR/../.."
+
+# FAST mode to avoid rebuilding certain dependencies
+if [ -n "$FAST" ]; then
+  echo "FAST enabled - will not rebuild python csi, openapi and deployer. (Make sure they are built already)"
```
paulyoong, line 18: I think we can remove "python" from this sentence.

paulyoong (approved):
Nit: The feature files are generally named "feature.feature". I tend to prefer the naming used in ../tests/bdd/features/ana/validate where the feature file and test file have a more meaningful name, but this is a personal preference.

openapi/build.rs
tiagolobocastro: I thought of keeping it for reference and in case the cargo timestamp changes

tiagolobocastro:
> Nit: The feature files are generally named "feature.feature". I tend to prefer the naming used in ../tests/bdd/features/ana/validate where the feature file and test file have a more meaningful name, but this is a personal preference.

I kept the ana one as it is because the name was slightly different. The other ones were all called the same as the base, so I just wanted to avoid having duplicate names, eg: create / create.feature.
I'll merge it as it is now but yeah we can change it later easily.

tiagolobocastro:
bors merge

PR #186, opened 2021-12-08
refactor: composer library needs to be a separate crate

mtzaurus:
Composer is now used as a standalone crate from github.

Resolves: CAS-1255


mtzaurus:
bors try

mtzaurus:
bors try

paulyoong (approved):

common/Cargo.toml
```diff
@@ -39,6 +39,5 @@ opentelemetry-semantic-conventions = "0.8.0"
 tracing = "0.1.28"
 
 [dev-dependencies]
-composer = { path = "../composer" }
+composer = { git = "https://github.com/mayadata-io/composer", branch = "develop" }
```
tiagolobocastro, line 42: Should we use specific composer version, eg a release/0.1 branch or something? Also please make sure building the kubectl static binary image still works. nix-build -A utils.kubectl-binary

control-plane/agents/core/src/volume/tests.rs
```diff
@@ -19,7 +19,7 @@ use common_lib::{
     },
 };
 
-use rpc::mayastor::FaultNexusChildRequest;
+use composer::rpc::mayastor::FaultNexusChildRequest;
```
tiagolobocastro, line 22: We should remove the rpc from the composer library, the rpc should be separate now otherwise if we update the rpc we also may have to update the composer?

mtzaurus:
bors try

control-plane/agents/core/src/volume/tests.rs
mtzaurus: Composer currently exposes RPC access to containers (via gRPC handles or so). If we keep this functionality, then it's more useful to have rpc in this crate too, as for user it's more convenient to just use composer crate to access RPC. If we decide to completely remove RPC support from composer crate (in contrary to what we have now), than we can remove rpc, sure.

mtzaurus:
bors try

control-plane/agents/core/src/volume/tests.rs
tiagolobocastro: Yeah it's indeed convenient but also because it was in the same repo as the rpc which is no longer the case now as it's a "common" interface to docker.
We can have our own wrapper for it, as long as the composer exposes the endpoint information.

mtzaurus:
bors cancel

mtzaurus:
bors merge

PR #187, opened 2021-12-08
refactor(replace nats): add directory structure and proto for pool

Abhinandan-Purkait:
Add pool proto for the initial phase of implementation of grpc to replace nats.


control-plane/grpc/protos/pool.proto
```diff
@@ -0,0 +1,133 @@
...
+
+// Desired spec and the control plane related data
+message PoolDefinition {
+  // Client specification of the pool
+  PoolSpec spec = 1;
+  // Control plane related data for the pool
+  Metadata metadata = 2;
+}
+
+message Metadata {
```
paulyoong, line 37: I think we should include a comment for this message.

control-plane/grpc/protos/pool.proto
```diff
@@ -0,0 +1,133 @@
+syntax = "proto3";
+
+import "common.proto";
+import "google/protobuf/empty.proto";
+import "google/protobuf/wrappers.proto";
+
+package grpc;
```
paulyoong, line 7: Do we want the package to be generically named? Will other proto file (i.e. volume.proto) also be package grpc? Alternatively we could call it pool-grpc. What do you think?

control-plane/grpc/protos/pool.proto
```diff
@@ -0,0 +1,133 @@
...
+  optional PoolState state = 2;
+}
+
+// Multiple pools
+message Pools {
+  repeated Pool pools = 1;
+}
+
+// PoolLabel is the type for the labels
+message PoolLabel {
```
tiagolobocastro, line 25: Should this be a common type, eg: StringMapValue?

control-plane/grpc/protos/pool.proto
```diff
@@ -0,0 +1,133 @@
...
+  map<string, string> labels = 1;
```
tiagolobocastro, line 26: and then this could be called value for example

control-plane/grpc/protos/pool.proto
```diff
@@ -0,0 +1,133 @@
...
+// Destroy Pool Request
+message DestroyPoolRequest {
+  // id of the mayastor instance
+  string node_id = 1;
+  // id of the pool
+  string pool_id = 2;
+}
+
+// Service for managing storage pools
+service CoreGrpc {
```
tiagolobocastro, line 129: Should this be named PoolGrpc?

control-plane/grpc/protos/pool.proto
```diff
@@ -0,0 +1,133 @@
...
+  rpc CreatePool (CreatePoolRequest) returns (Pool) {}
```
tiagolobocastro, line 130: Do we need to define an error type? On the bus we currently return `ReplyError` in case of an error.
Then that be sent on the .details of the grpc `Status`?

control-plane/grpc/protos/common.proto
```diff
@@ -0,0 +1,11 @@
+syntax = "proto3";
+
+package grpc;
+
+// Enum defining the various states that a resource spec can be in.
```
tiagolobocastro, line 5: s/states/statuses ?

tiagolobocastro (approved):

control-plane/grpc/protos/pool.proto
Abhinandan-Purkait: Added comment.

control-plane/grpc/protos/pool.proto
Abhinandan-Purkait: Have split into directory versioned directories and changed it to `pool` package.

control-plane/grpc/protos/pool.proto
Abhinandan-Purkait: Added the type.

control-plane/grpc/protos/pool.proto
Abhinandan-Purkait: the code generator translates the return type to `Result<Pool, tonic::Status>`. 
`tonic::Status` -> ref: https://github.com/hyperium/tonic/blob/master/tonic/src/status.rs
After a discussion with @paulyoong,  we have added a composite type for ex :-
```
message CreatePoolReply {
  oneof reply {
    Pool pool = 1;
    ReplyError error = 2;
  }
}
```
with this, we can propagate internal error other than grpc error(denoted by `tonic::status`)
And the current code generated would look somewhat like this:- `Result<CreatePoolReply, tonic::Status>`

control-plane/grpc/proto/pool/v1/pool.proto
```diff
@@ -0,0 +1,157 @@
...
+  string pool_id = 1;
+}
+
+// Filter by Node and Pool id
+message NodePoolFilter{
+  string node_id = 1;
+  string pool_id = 2;
+}
+
+// Get all the pools from based on the args
```
paulyoong, line 102: Nit: Maybe this should say something like "Get all pools based on the filter criteria"

paulyoong (approved):

control-plane/grpc/src/lib.rs
```diff
@@ -0,0 +1,7 @@
+pub(crate) mod common {
```
tiagolobocastro, line 1: Should these be `pub`?
Should we put some doc comments above these?

Abhinandan-Purkait:
FYI: I have changed the structure a bit, as there was issue keeping `resource/version` structure. So currently I have moved everything under v1, and that way it seems to work, also I have added the errors. Needs one more round of review.

control-plane/grpc/proto/v1/misc/common.proto
```diff
@@ -0,0 +1,86 @@
...
+  Unavailable = 14;
+  Unauthenticated = 15;
+  Unauthorized = 16;
+  Conflict = 17;
+  FailedPersist = 18;
+  NotShared = 19;
+  AlreadyShared = 20;
+  NotPublished = 21;
+  AlreadyPublished = 22;
+  Deleting_ = 23;
```
paulyoong, line 51: If possible can we get rid of the underscore suffix. Maybe we could rename it to `IsDeleting`?

control-plane/grpc/proto/v1/pool/pool.proto
```diff
@@ -0,0 +1,151 @@
...
+
+// Desired spec and the control plane related data
+message PoolDefinition {
+  // Client specification of the pool
+  PoolSpec spec = 1;
+  // Control plane related data for the pool
+  Metadata metadata = 2;
+}
+
+// uuid and the status of pool for the control plane use
```
paulyoong, line 32: I would suggest rewording `status of pool` to `status of the pool spec` otherwise it may be read as the health of the pool (i.e. online, degraded etc.) rather than whether the pool is being created/deleted. It's the distinction between `SpecStatus` and `PoolStatus`.

control-plane/grpc/proto/v1/pool/pool.proto
```diff
@@ -0,0 +1,151 @@
...
+
+// status of the pool
+enum PoolStatus {
+  // unknown state
+  Unknown = 0;
+  // the pool is in normal working order
+  Online = 1;
+  // the pool has experienced a failure but can still function
+  Degraded = 2;
+  // the pool is completely inaccessible
```
paulyoong, line 76: Nit: I wonder if we should say that the pool is unable to be used rather than inaccessible. It may be accessible, i.e. we can connect to it, but the disks may be faulty.

paulyoong (approved):

Abhinandan-Purkait:
bors merge

PR #190, opened 2021-12-22
refactor(images): fix debug images and allow for faster incremental builds

tiagolobocastro:
chore(nix): format the existing nix files with nixpkgs-fmt

---------
refactor(images): tidy up building of images

There's a new build type at the root of the images which should simplify
adding new images (no need for -dev or -cov). Example, to build a debug
version of the core agent:
nix-build -A images.debug.agents.core

Add a build suffix to fix the debug builds which are broken. Example:
the core agent image will be named mcp-core and the debug mcp-core-dev

---------
chore(build): add allInOne and incremental arguments to the nix build

With allInOne set to true all components are built as part of the same "cargo build" derivation
this allows for a quicker build of all components but slower single components
with allInOne set to false each compoment gets its own "cargo build" derivation allowing for faster
individual builds but making the build of all components slower
EXPERIMENTAL incremental allows for faster incremental builds but as the build dependencies are
cached it might make the initial build sligtly slower as it's done in two steps
for this we use naersk which is not as fully fledged as the builtin rustPlatform so it should only
be used for development and not for CI.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -781,7 +781,7 @@ dependencies = [
 [[package]]
 name = "composer"
 version = "0.1.0"
-source = "git+https://github.com/mayadata-io/composer?branch=develop#d1d2f731c39f43b1ff7bd9bfcaf55db57ad755de"
+source = "git+https://github.com/mayadata-io/composer?branch=develop#9a5e3669f1e70d42c1d8ba33a4799e7c2d69a176"
 dependencies = [
  "bollard",
  "crossbeam",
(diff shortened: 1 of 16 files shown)


tiagolobocastro:
bors try

tiagolobocastro:
bors try

tiagolobocastro:
the bdd tests can't open /tmp/disk1.img for some reason...
bors try

Abhinandan-Purkait:
bors merge

Abhinandan-Purkait (approved):

Abhinandan-Purkait:
@tiagolobocastro Shall we merge this? The pipleline seems to be broken and would be fixed with this.

paulyoong (approved):

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

tiagolobocastro:
bors cancel

tiagolobocastro:
bors merge

PR #194, opened 2022-01-12
fix(node): node status overridden with stale status

tiagolobocastro:
Instead of using a copy of the node to reload, use the node itself with a grpc locked client to
update the nodes when the registry is polling mayastor.

Extra:
To clarity, in the client operations, pass the grpc client down to the update operations to that its
use is explicit rather than implicit. Added a TODO to allow us to use different timeouts for the
list operations without having to reconnect the gRPC client.
Use more getters rather than accessing fields directly

### DIFF
--- a/common/src/mbus_api/mod.rs
+++ b/common/src/mbus_api/mod.rs
@@ -163,6 +163,18 @@ impl MessageIdTimeout for MessageId {
     }
 }
 
+/// Generic Get Timeouts for Get* operations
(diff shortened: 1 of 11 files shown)


tiagolobocastro:
bors try

tiagolobocastro:
This is becoming a constant...
PermissionError: [Errno 13] Permission denied: '/tmp/disk2.img'
bors try

PR #198, opened 2022-01-17
chore: update example deployment definition files

GlennBullingham:
Update k8s definition files referenced in the user
docs as an example deployment

- Container image tags should reflect 1.0.0 release version
- Template used should be the 'Release' profile

### DIFF
--- a/deploy/core-agents-deployment.yaml
+++ b/deploy/core-agents-deployment.yaml
@@ -23,15 +23,13 @@ spec:
         - command:
           - sh
           - -c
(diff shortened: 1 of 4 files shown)


paulyoong (approved):

tiagolobocastro:
Hmm at the moment the pre-commit will reset the files to develop if there's another commit - might be worth changing that as well!  (`.pre-commit-config.yaml`)
For the future we probably should make this a bit smarter and pick up the profile from the branch name automatically.

GlennBullingham:
bors merge

PR #199, opened 2022-01-18
chore(version info): output crate version information

paulyoong:
chore(version info): use the utility function
    
    Make the various crates print out their version information on start up.
    
    Utilise the common utility function to print the versions so that the
    format is standardised across crates. This removes the need for each
    crate to import the git-version crate.

--------------------------------------------------

refactor(common lib): create new utils crate
    
    The constant definitions and building of the constants YAML file have
    been moved from the common-lib crate to a new utils crate. In addition
    to this the utils crate provides function and macro utilities that can
    be used by other crates.
    
    The purpose of splitting out the functionality is to improve build
    times of other crates if all they require are the constant definitions.
    Using the common-lib crate required pulling in a number of unnecessary
    dependencies. Now these crates can instead use the utils crate which is
    much lighter-weight (has fewer dependencies).


kubectl-plugin/src/operations.rs
```diff
@@ -9,6 +9,8 @@ pub(crate) enum Operations {
     Get(GetResources),
     /// 'Scale' resources.
     Scale(ScaleResources),
+    /// Version of the plugin
+    Version,
```
tiagolobocastro, line 13: This is a bit confusing now, we have a -V and a version?

tiagolobocastro (commented):
We seem to be missing some files in utils/?

kubectl-plugin/src/operations.rs
paulyoong: We don't have a `-V` option. It's use is only `kubectl-mayastor version`.

utils/src/functions.rs
```diff
@@ -0,0 +1,4 @@
+/// Return the git version.
+pub fn git_version() -> &'static str {
+    git_version::git_version!(args = ["--abbrev=12", "--always"], fallback = "unknown")
```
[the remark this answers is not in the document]
paulyoong: :+1:

control-plane/msp-operator/src/main.rs
```diff
@@ -32,13 +32,14 @@ use std::{collections::HashMap, io::Write, ops::Deref, sync::Arc, time::Duration
 use tracing::{debug, error, info, trace, warn};
 use tracing_subscriber::{layer::SubscriberExt, util::SubscriberInitExt, EnvFilter, Registry};
 
+extern crate utils as mayastor_utils;
```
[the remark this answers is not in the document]
paulyoong: Done

paulyoong:
> We seem to be missing some files in utils/?

Good spot, thanks - I've added the missing files.

utils/src/macros.rs
```diff
@@ -0,0 +1,33 @@
+/// Print package related information.
+/// This includes the package name and version. The git hash will also be included if supplied.
+#[macro_export]
+macro_rules! print_package_info {
```
tiagolobocastro, line 4: Should this make use of the package_info macro?

kubectl-plugin/src/main.rs
```diff
@@ -21,12 +21,15 @@ use structopt::StructOpt;
...
 
+extern crate utils as mayastor_utils;
+
 /// OpenTelemetry helpers for handling Processor Tags
 pub mod opentelemetry_helper {
     include!("../../common/src/opentelemetry.rs");
 }
 
 #[derive(StructOpt, Debug)]
+#[structopt(version = mayastor_utils::package_info!(mayastor_utils::git_version()))]
```
tiagolobocastro, line 32: Should we add this to the other binaries?

tiagolobocastro (approved):
I can see the missing files now :+1:

kubectl-plugin/src/main.rs
paulyoong: Done

utils/src/macros.rs
paulyoong: I chose not to do this because that is making use of heap allocations via boxing and I thought it would be better to avoid that.

kubectl-plugin/src/operations.rs
paulyoong: You are correct about the `-V` option. As we discussed, this has now been fixed up.

paulyoong:
bors merge

paulyoong:
bors merge

paulyoong:
bors merge

PR #203, opened 2022-01-20
fix: update nix derivations

tiagolobocastro:
fix: update nix derivations

Add new utils folder to the src derivation.
Remove debug breakpoint that should not have been left in the code.

--------------

chore: add cargo-expand to the nix-shell

### DIFF
--- a/nix/pkgs/control-plane/cargo-project.nix
+++ b/nix/pkgs/control-plane/cargo-project.nix
@@ -47,27 +47,28 @@ let
   LIBCLANG_PATH = "${llvmPackages.libclang}/lib";
   PROTOC = "${protobuf}/bin/protoc";
   PROTOC_INCLUDE = "${protobuf}/include";
-
+  src_list = [
+    ".git"
+    "Cargo.lock"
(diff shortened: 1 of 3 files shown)


nix/pkgs/control-plane/cargo-project.nix
```diff
@@ -47,27 +47,28 @@ let
   LIBCLANG_PATH = "${llvmPackages.libclang}/lib";
   PROTOC = "${protobuf}/bin/protoc";
   PROTOC_INCLUDE = "${protobuf}/include";
-
+  src_folders = [
```
paulyoong, line 50: Maybe we shouldn't use "folders". `Cargo.lock` & `Cargo.toml` aren't folders.

paulyoong (approved):

nix/pkgs/control-plane/cargo-project.nix
tiagolobocastro: Good point! `src_list` ?

tiagolobocastro:
bors merge

PR #205, opened 2022-01-21
fix(rust test): fix occasional CI test failures

paulyoong:
CI occasionally shows the "hotspare" test failing; specifically in the
"hotspare_replica_count_spread" function.

The problem appears to be in the "wait_till_volume" function where we
were waiting for the number of replicas to be correct. However, we were
checking the replica count in the spec, which defines the desired number
of replicas rather than the actual number of replicas in the system.

This function has been enhanced to check the replica topology field in
the volume state which will give the number of actual replicas.

### DIFF
--- a/control-plane/agents/core/src/volume/tests.rs
+++ b/control-plane/agents/core/src/volume/tests.rs
@@ -785,7 +785,6 @@ async fn wait_till_volume(volume: &VolumeId, replicas: usize) {
     let timeout = Duration::from_secs(RECONCILE_TIMEOUT_SECS);
     let start = std::time::Instant::now();
     loop {
-        // the volume state does not carry replica information, so inspect the replica spec instead.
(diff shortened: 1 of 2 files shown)


control-plane/agents/core/src/volume/tests.rs
```diff
@@ -806,6 +806,30 @@ async fn wait_till_volume(volume: &VolumeId, replicas: usize) {
     }
 }
 
+/// Return the number of replicas that exist.
```
tiagolobocastro, line 809: ...and are have a state?

tiagolobocastro (approved):

control-plane/agents/core/src/volume/tests.rs
paulyoong: I've included this in the comment.

paulyoong:
bors merge

PR #206, opened 2022-01-24
fix: sending node online event should not block

tiagolobocastro:
On startup if the cluster had more than 5 Nodes the send event notification per node would block.
This is because the queue has only 5 elements.
Simply increasing the queue size would not necessarily improve things so instead we drop the
notification if the queue is full which is acceptable as at the moment we have no way of
"flattening" the events or to avoid running the reconcilers too often so we probably don't want
a huge queue of notifications to be built. Losing one is not a problem, just may mean we'd be a
little slower.

### DIFF
--- a/control-plane/agents/core/src/core/reconciler/mod.rs
+++ b/control-plane/agents/core/src/core/reconciler/mod.rs
@@ -48,7 +48,10 @@ impl ReconcilerControl {
     }
 
     /// Send an event signal to the poller's main loop
+    /// todo: don't requeque duplicate events
     pub(crate) async fn notify(&self, event: PollEvent) {
-        self.event_channel.send(event).await.ok();
+        if let Err(error) = self.event_channel.try_send(event) {
+            tracing::warn!(error=?error, "Failed to send event to reconcile worker");
+        }
     }
 }
--- a/control-plane/agents/core/src/node/service.rs
+++ b/control-plane/agents/core/src/node/service.rs
@@ -53,17 +53,30 @@ impl NodeCommsTimeout {
 impl Service {
     /// New Node Service which uses the `registry` as its node cache and sets
     /// the `deadline` to each node's watchdog
-    pub(super) fn new(
+    pub(super) async fn new(
         registry: Registry,
         deadline: std::time::Duration,
         request: std::time::Duration,
         connect: std::time::Duration,
     ) -> Self {
-        Self {
+        let service = Self {
             registry,
             deadline,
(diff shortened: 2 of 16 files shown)


tiagolobocastro:
bors try

control-plane/agents/core/src/core/reconciler/mod.rs
```diff
@@ -48,7 +48,10 @@ impl ReconcilerControl {
     }
 
     /// Send an event signal to the poller's main loop
+    /// todo: don't requeque duplicate events
     pub(crate) async fn notify(&self, event: PollEvent) {
-        self.event_channel.send(event).await.ok();
+        if let Err(error) = self.event_channel.try_send(event) {
+            tracing::error!(error=?error, "Failed to send event to reconcile worker");
```
paulyoong, line 54: If it is legitimate that this can happen, should we emit a warning rather than an error?

tiagolobocastro:
Same old disk1 locked in tmp...
bors try

control-plane/agents/core/src/node/service.rs
```diff
@@ -84,11 +84,11 @@ impl Service {
     /// Register a new node through the register information
     pub(super) async fn register(&self, registration: &Register) {
         self.registry.register_node_spec(registration).await;
-        self.register_state(registration).await;
+        self.register_state(registration, false).await;
     }
 
     /// Attempt to Register a new node state through the register information
-    pub(super) async fn register_state(&self, registration: &Register) {
+    pub(super) async fn register_state(&self, registration: &Register, startup: bool) {
```
paulyoong, line 91: Could we do the initial state registration when we create the node service (i.e. in the `new` function)? We currently create the node service first and then immediately register the states.

paulyoong (approved):

control-plane/agents/core/src/core/reconciler/mod.rs
tiagolobocastro: Yep good point

control-plane/agents/core/src/node/service.rs
tiagolobocastro: Yes indeed, done

tiagolobocastro:
bors try

tiagolobocastro:
bors try

tiagolobocastro:
bors try

tiagolobocastro:
bors try

tiagolobocastro:
same disk1.img issue again...
bors try

tiagolobocastro:
bors merge

PR #209, opened 2022-01-27
refactor(replace nats): add support for pool and replica operations over grpc replacing nats

Abhinandan-Purkait:
### Changes
1. Removed the NATS subscription handlers for the pool and replica operations.
2. Add pool and replica operations over grpc.
3. Change tests to remove message bus usage for pool and replica operations.
4. Changes to cluster builder to use grpc.
5. Changes to the deployer to accept env-variable to rest containers.

### DIFF
--- a/common/src/mbus_api/mod.rs
+++ b/common/src/mbus_api/mod.rs
@@ -30,6 +30,7 @@ use std::{
     time::Duration,
 };
 use strum_macros::{AsRefStr, ToString};
+use tonic::Status;
 
 /// Result wrapper for send/receive
 pub type BusResult<T> = Result<T, Error>;
@@ -342,6 +343,18 @@ pub struct ReplyError {
     pub extra: String,
 }
 
+impl From<tonic::Status> for ReplyError {
+    fn from(status: Status) -> Self {
+        Self::tonic_reply_error(status.to_string(), status.full_string())
+    }
+}
+
+impl From<tonic::transport::Error> for ReplyError {
+    fn from(e: tonic::transport::Error) -> Self {
+        Self::tonic_reply_error(e.to_string(), e.full_string())
+    }
+}
+
 impl StdError for ReplyError {}
 impl ReplyError {
     /// extend error with source
@@ -350,6 +363,42 @@ impl ReplyError {
     pub fn extend(&mut self, source: &str, extra: &str) {
         self.source = format!("{}::{}", source, self.source);
         self.extra = format!("{}::{}", extra, self.extra);
+    }
+    /// useful when the grpc server is dropped due to panic
+    pub fn tonic_reply_error(source: String, extra: String) -> Self {
+        Self {
+            kind: ReplyErrorKind::Aborted,
+            resource: ResourceKind::Unknown,
+            source,
+            extra,
+        }
+    }
+    /// used only for testing, not used in code
+    pub fn invalid_reply_error(msg: String) -> Self {
+        Self {
+            kind: ReplyErrorKind::Aborted,
+            resource: ResourceKind::Unknown,
+            source: "Test Library".to_string(),
+            extra: msg,
+        }
+    }
+    /// used when we get an empty response from the grpc server
+    pub fn invalid_response(resource: ResourceKind) -> Self {
+        Self {
+            kind: ReplyErrorKind::Aborted,
+            resource,
+            source: "Empty response reply received from grpc server".to_string(),
+            extra: "".to_string(),
+        }
+    }
+    /// used when we error while unwrapping
+    pub fn unwrap_err(resource: ResourceKind) -> Self {
+        Self {
+            kind: ReplyErrorKind::Aborted,
+            resource,
+            source: "Error unwrapping to desired type".to_string(),
+            extra: "".to_string(),
+        }
     }
 }
 
@@ -510,6 +559,10 @@ impl TimeoutOptions {
     /// Get the tcp read timeout
     pub(crate) fn tcp_read_timeout(&self) -> Duration {
         self.tcp_read_timeout
+    }
+    /// Get the base timeout
+    pub fn base_timeout(&self) -> Duration {
+        self.timeout
     }
 }
 
--- a/control-plane/agents/core/src/lib.rs
+++ b/control-plane/agents/core/src/lib.rs
@@ -1,0 +1,85 @@
+#![warn(missing_docs)]
+
+use common::ServiceError;
+use futures::{future::join_all, FutureExt};
+use grpc::{pool::server::PoolServer, replica::server::ReplicaServer};
+use http::Uri;
+use tracing::error;
+
+/// the gprc service that encapsulates the base_service and the server for rpc
+pub struct Service {
+    base_service: common::Service,
+    tonic_grpc_server: tonic::transport::Server,
+}
+
+impl Service {
+    /// creates a new Service with the base_service and tonic server builder
+    pub fn new(base_service: common::Service) -> Self {
+        Self {
+            base_service,
+            tonic_grpc_server: tonic::transport::Server::builder(),
+        }
+    }
+
+    /// launch each of the services and the grpc server
+    pub async fn run(mut self) {
+        let grpc_addr = self.base_service.get_shared_state::<Uri>().clone();
+        let pool_service = self.base_service.get_shared_state::<PoolServer>().clone();
+        let replica_service = self
+            .base_service
+            .get_shared_state::<ReplicaServer>()
+            .clone();
+
+        let tonic_router = self
+            .tonic_grpc_server
+            .add_service(pool_service.into_grpc_server())
+            .add_service(replica_service.into_grpc_server());
+
+        let mut threads = self.base_service.mbus_handles().await;
+
+        let tonic_thread = tokio::spawn(async move {
+            tonic_router
+                .serve_with_shutdown(
+                    grpc_addr.authority().unwrap().to_string().parse().unwrap(),
+                    Self::shutdown_signal().map(|_| ()),
+                )
+                .await
+                .map_err(|source| ServiceError::GrpcServer { source })
+        });
+
+        threads.push(tonic_thread);
+
+        join_all(threads)
+            .await
+            .iter()
+            .for_each(|result| match result {
+                Err(error) => error!("Failed to wait for thread: {:?}", error),
+                Ok(Err(error)) => {
+                    error!(error=?error, "Error running service thread");
+                }
+                _ => {}
+            });
+    }
+
+    /// Get a shutdown_signal as a oneshot channel when the process receives either TERM or INT.
+    /// When received the opentel traces are also immediately flushed.
+    fn shutdown_signal() -> tokio::sync::oneshot::Receiver<()> {
+        let mut signal_term =
+            tokio::signal::unix::signal(tokio::signal::unix::SignalKind::terminate()).unwrap();
+        let mut signal_int =
+            tokio::signal::unix::signal(tokio::signal::unix::SignalKind::interrupt()).unwrap();
+        let (stop_sender, stop_receiver) = tokio::sync::oneshot::channel();
+        tokio::spawn(async move {
+            tokio::select! {
+                _term = signal_term.recv() => {tracing::info!("SIGTERM received")},
+                _int = signal_int.recv() => {tracing::info!("SIGINT received")},
+            }
+            opentelemetry::global::force_flush_tracer_provider();
+            if stop_sender.send(()).is_err() {
+                // should we panic here?
+                tracing::warn!("Failed to stop the tonic server");
+            }
+        });
+        stop_receiver
+    }
+}
--- a/control-plane/agents/core/src/pool/service.rs
+++ b/control-plane/agents/core/src/pool/service.rs
@@ -1,7 +1,10 @@
 use crate::core::{registry::Registry, specs::ResourceSpecsLocked, wrapper::GetterOps};
 use common::errors::{PoolNotFound, ReplicaNotFound, SvcError};
 use common_lib::{
-    mbus_api::message_bus::v0::{Pools, Replicas},
+    mbus_api::{
+        message_bus::v0::{Pools, Replicas},
+        ReplyError,
+    },
     types::v0::{
         message_bus::{
             CreatePool, CreateReplica, DestroyPool, DestroyReplica, Filter, GetPools, GetReplicas,
@@ -10,11 +13,97 @@ use common_lib::{
         store::OperationMode,
     },
 };
+use grpc::{
+    grpc_opts::Context,
+    pool::traits::{CreatePoolInfo, DestroyPoolInfo, PoolOperations},
+    replica::traits::{
+        CreateReplicaInfo, DestroyReplicaInfo, ReplicaOperations, ShareReplicaInfo,
+        UnshareReplicaInfo,
+    },
+};
 use snafu::OptionExt;
 
 #[derive(Debug, Clone)]
 pub(super) struct Service {
     registry: Registry,
+}
+
+#[tonic::async_trait]
+impl PoolOperations for Service {
+    async fn create(
+        &self,
+        pool: &dyn CreatePoolInfo,
+        _ctx: Option<Context>,
+    ) -> Result<Pool, ReplyError> {
+        let req = pool.into();
+        let pool = self.create_pool(&req).await?;
+        Ok(pool)
+    }
+
+    async fn destroy(
+        &self,
+        pool: &dyn DestroyPoolInfo,
+        _ctx: Option<Context>,
+    ) -> Result<(), ReplyError> {
+        let req = pool.into();
+        self.destroy_pool(&req).await?;
+        Ok(())
+    }
+
+    async fn get(&self, filter: Filter, _ctx: Option<Context>) -> Result<Pools, ReplyError> {
+        let req = GetPools { filter };
+        let pools = self.get_pools(&req).await?;
+        Ok(pools)
+    }
+}
+
+#[tonic::async_trait]
+impl ReplicaOperations for Service {
+    async fn create(
+        &self,
+        req: &dyn CreateReplicaInfo,
+        _ctx: Option<Context>,
+    ) -> Result<Replica, ReplyError> {
+        let create_replica = req.into();
+        let replica = self.create_replica(&create_replica).await?;
+        Ok(replica)
+    }
+
+    async fn get(&self, filter: Filter, _ctx: Option<Context>) -> Result<Replicas, ReplyError> {
+        let req = GetReplicas { filter };
+        let replicas = self.get_replicas(&req).await?;
+        Ok(replicas)
+    }
+
+    async fn destroy(
+        &self,
+        req: &dyn DestroyReplicaInfo,
+        _ctx: Option<Context>,
+    ) -> Result<(), ReplyError> {
+        let destroy_replica = req.into();
+        self.destroy_replica(&destroy_replica).await?;
+        Ok(())
+    }
+
+    async fn share(
+        &self,
+        req: &dyn ShareReplicaInfo,
+        _ctx: Option<Context>,
+    ) -> Result<String, ReplyError> {
+        let share_replica = req.into();
+        let response = self.share_replica(&share_replica).await?;
+        Ok(response)
+    }
+
+    async fn unshare(
+        &self,
+        req: &dyn UnshareReplicaInfo,
+        _ctx: Option<Context>,
+    ) -> Result<(), ReplyError> {
+        let unshare_replica = req.into();
+        self.unshare_replica(&unshare_replica).await?;
+        Ok(())
+    }
 }
 
 impl Service {
--- a/control-plane/agents/core/src/pool/tests.rs
+++ b/control-plane/agents/core/src/pool/tests.rs
@@ -2,12 +2,12 @@
 
 use super::*;
 use common_lib::{
-    mbus_api,
     mbus_api::{ReplyError, ReplyErrorKind, ResourceKind, TimeoutOptions},
     types::v0::{
         message_bus::{
-            GetNodes, GetSpecs, Protocol, Replica, ReplicaId, ReplicaName, ReplicaShareProtocol,
-            ReplicaStatus, VolumeId,
+            CreatePool, CreateReplica, DestroyPool, DestroyReplica, Filter, GetNodes, GetSpecs,
+            Protocol, Replica, ReplicaId, ReplicaName, ReplicaShareProtocol, ReplicaStatus,
+            ShareReplica, UnshareReplica, VolumeId,
         },
         openapi::{
             apis::StatusCode,
@@ -17,6 +17,7 @@ use common_lib::{
         store::replica::ReplicaSpec,
     },
 };
+use grpc::{grpc_opts::Context, pool::traits::PoolOperations, replica::traits::ReplicaOperations};
 use itertools::Itertools;
 use std::{convert::TryFrom, time::Duration};
 use testlib::{Cluster, ClusterBuilder};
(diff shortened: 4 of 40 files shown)


Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors try

control-plane/grpc/src/pool/traits.rs
```diff
@@ -0,0 +1,295 @@
...
+            kind: err.clone().kind.into(),
+            resource: err.clone().resource.into(),
+            source: err.clone().source,
+            extra: err.extra,
+        }
+    }
+}
+
+pub trait CreatePoolInfo {
+    fn pool_id(&self) -> String;
```
tiagolobocastro, line 194: These Should be NodeId and PoolId rather than String, same below

control-plane/agents/core/src/pool/service.rs
```diff
@@ -1,22 +1,171 @@
...
+}
+
+#[tonic::async_trait]
+impl ReplicaOperations for Service {
+    async fn create(
+        &self,
+        req: &(dyn CreateReplicaInfo + Sync + Send),
+        _ctx: Option<Context>,
+    ) -> Result<Replica, ReplyError> {
+        let create_replica = CreateReplica {
```
tiagolobocastro, line 80: please remove all the unwrap's where possible, non-test code should not have them as much as possible.

control-plane/agents/core/src/pool/service.rs
```diff
@@ -1,22 +1,171 @@
...
+            node: req.node().into(),
+            name: req.name().map(|name| name.into()),
+            uuid: ReplicaId::try_from(req.uuid()).unwrap(),
```
tiagolobocastro, line 83: I think most of these unwrap's are not needed if we change the traits to use specific types (eg: ReplicaId) rather than String

control-plane/grpc/Cargo.toml
```diff
@@ -10,6 +10,18 @@ authors = ["Abhinandan Purkait <abhinandan.purkait@mayadata.io>"]
 tonic = "0.5.2"
 prost = "0.8.0"
 prost-types = "0.8.0"
+tokio = { version = "1.12.0", features = ["full"] }
+common-lib = { path = "../../common" }
+dns-lookup = "1.0.8"
```
tiagolobocastro, line 15: is this used?

control-plane/grpc/src/client.rs
```diff
@@ -0,0 +1,29 @@
+use crate::{
+    pool::{client::PoolClient, traits::PoolOperations},
+    replica::{client::ReplicaClient, traits::ReplicaOperations},
+};
+use common_lib::mbus_api::TimeoutOptions;
+use tonic::transport::Uri;
+
+pub struct CoreClient {
```
tiagolobocastro, line 8: Please document the `pub`'s

control-plane/grpc/src/pool/server.rs
```diff
@@ -0,0 +1,108 @@
...
+            }))
+        })
+    }
+
+    async fn get_pools(
+        &self,
+        request: Request<GetPoolsRequest>,
+    ) -> Result<tonic::Response<pool_grpc::GetPoolsReply>, tonic::Status> {
+        let req: GetPoolsRequest = request.into_inner();
+        let filter = if req.filter.is_none() {
```
tiagolobocastro, line 83: can probably use something like `let filter = req.filter.map(Into::into).unwrap_or_default();`

control-plane/agents/core/src/pool/service.rs
Abhinandan-Purkait: We had a conversation on this, we wanted to keep the return types of methods in these traits to be native rust data types wherever possible that's the reason for it. cc @paulyoong

control-plane/grpc/Cargo.toml
Abhinandan-Purkait: Nope. Removing it.

common/src/types/v0/message_bus/replica.rs
```diff
@@ -373,10 +373,18 @@ impl From<i32> for ReplicaShareProtocol {
     fn from(src: i32) -> Self {
         match src {
             1 => Self::Nvmf,
+            0 => Self::Nvmf,
```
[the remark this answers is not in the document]
Abhinandan-Purkait: The reason for doing that is the grpc defaults in enum are to be 0, and the 1 was here from before. So to not disrupt anything I added both.

common/src/mbus_api/mod.rs
```diff
@@ -351,6 +398,24 @@ impl ReplyError {
...
+    pub fn tonic_reply_error() -> Self {
+        Self {
+            kind: ReplyErrorKind::Aborted,
+            resource: ResourceKind::Unknown,
+            source: "Core Grpc Server".to_string(),
+            extra: "Connection Dropped".to_string(),
+        }
+    }
+    /// used only for testing, not used in code
+    pub fn invalid_reply_error(msg: String) -> Self {
```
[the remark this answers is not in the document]
Abhinandan-Purkait: No its used in a method in test-lib, so we cannot add #[cfg(test)].

common/src/mbus_api/mod.rs
```diff
@@ -351,6 +398,24 @@ impl ReplyError {
         self.source = format!("{}::{}", source, self.source);
         self.extra = format!("{}::{}", extra, self.extra);
     }
+    /// useful when the grpc server is dropped due to panic
+    pub fn tonic_reply_error() -> Self {
+        Self {
+            kind: ReplyErrorKind::Aborted,
+            resource: ResourceKind::Unknown,
+            source: "Core Grpc Server".to_string(),
```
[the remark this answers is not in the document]
Abhinandan-Purkait: only `Grpc Server` would be good then ?

common/src/mbus_api/mod.rs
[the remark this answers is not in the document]
tiagolobocastro: Ah fair enough, then that's my fault for having a dual purpose test-lib, it'll be split into a deployer-lib and test-lib at some point.

common/src/mbus_api/mbus_nats.rs
```diff
@@ -37,6 +37,11 @@ pub fn bus() -> DynBus {
     )
 }
 
+/// Get the global timeout_opts
+pub fn timeout_opts() -> TimeoutOptions {
```
paulyoong, line 41: Where does this function get used?

common/src/types/v0/store/pool.rs
```diff
@@ -47,6 +47,17 @@ impl ResourceUuid for PoolState {
 /// Status of the Pool Spec
 pub type PoolSpecStatus = SpecStatus<message_bus::PoolStatus>;
 
+impl From<i32> for PoolSpecStatus {
+    fn from(i: i32) -> Self {
+        match i {
+            1 => PoolSpecStatus::Created(Default::default()),
+            2 => PoolSpecStatus::Deleting,
+            3 => PoolSpecStatus::Deleted,
+            _ => PoolSpecStatus::Creating,
```
paulyoong, line 56: Should we be setting the catch all to Creating? If it is possible to have an "unsupported" value, should we handle it differently?

common/src/mbus_api/mbus_nats.rs
Abhinandan-Purkait: We need this in the tests, but yeah I am working on removing it.

control-plane/agents/core/src/pool/service.rs
paulyoong: Yes, in retrospect maybe that isn't such a good idea. If it makes it cleaner to have a specific type, then I am happy with that.

Abhinandan-Purkait:
bors try

common/src/types/v0/message_bus/replica.rs
```diff
@@ -385,6 +393,17 @@ impl From<ReplicaShareProtocol> for Protocol {
     }
 }
 
+impl TryFrom<&str> for ReplicaShareProtocol {
+    type Error = String;
+
+    fn try_from(value: &str) -> Result<Self, Self::Error> {
+        Ok(match value {
+            "nvmf" => Self::Nvmf,
+            _ => panic!("Invalid replica share protocol {}", value),
```
tiagolobocastro, line 402: If this is a TryFrom we should return the Error here rather than panic

control-plane/agents/core/src/server.rs
```diff
@@ -75,6 +76,11 @@ pub(crate) struct CliArgs {
     /// Trace rest requests to the Jaeger endpoint agent
     #[structopt(long, short)]
     jaeger: Option<String>,
+    /// The GRPC Server URLs to connect to
+    /// (supports the http/https schema)
+    /// Default: 0.0.0.0:50051
+    #[structopt(long, short, default_value = "https://0.0.0.0:50051")]
```
tiagolobocastro, line 82: We can put this default value in utils rather than a string here, so it can be reused.

control-plane/grpc/src/client.rs
```diff
@@ -0,0 +1,29 @@
...
+    pub async fn init<O: Into<Option<TimeoutOptions>>>(addr: Uri, opts: O) -> Self {
+        let opts_clone = opts.into();
+        let pool_client = PoolClient::init(Some(addr.clone()), opts_clone.clone()).await;
+        let replica_client = ReplicaClient::init(Some(addr), opts_clone).await;
+        Self {
+            pool_client,
+            replica_client,
+        }
+    }
+    pub fn pool_client(&self) -> impl PoolOperations {
```
tiagolobocastro, line 23: nit: I would call these just pool and replica, since we're already in CoreClient.

control-plane/grpc/src/pool/server.rs
```diff
@@ -0,0 +1,108 @@
...
+        tokio::spawn(async move {
+            match service.destroy(&req, None).await {
+                Ok(()) => Ok(Response::new(DestroyPoolReply { error: None })),
+                Err(e) => Ok(Response::new(DestroyPoolReply {
+                    error: Some(e.into()),
+                })),
+            }
+        })
+        .await
+        .unwrap_or_else(|_| {
```
tiagolobocastro, line 45: we should probably capture the error here and pass it to tonic_reply_error

control-plane/grpc/src/pool/server.rs
```diff
@@ -0,0 +1,108 @@
...
+};
+use common_lib::mbus_api::ReplyError;
+use std::sync::Arc;
+use tonic::{Request, Response};
+
+use common_lib::types::v0::message_bus::Filter;
+// RPC Pool Server
+pub struct PoolServer {
+    // Service which executes the operations.
+    service: Arc<dyn PoolOperations + Send + Sync>,
```
tiagolobocastro, line 17: Could make PoolOperations Send + Sync to avoid typing it everytime

control-plane/grpc/src/pool/traits.rs
```diff
@@ -0,0 +1,295 @@
...
+};
+use std::collections::HashMap;
+
+/// Trait implemented by services which support pool operations.
+/// This trait can only be implemented on types which support the PoolInfo trait.
+#[tonic::async_trait]
+pub trait PoolOperations {
+    async fn create(
+        &self,
+        pool: &(dyn CreatePoolInfo + Sync + Send),
```
tiagolobocastro, line 21: Same as before, could make CreatePoolInfo et all Send + Sync to avoid adding it on the signatures.

control-plane/grpc/src/pool/traits.rs
```diff
@@ -0,0 +1,295 @@
...
+            pools: pools
+                .into_inner()
+                .iter()
+                .map(|pool| pool.clone().into())
+                .collect(),
+        }
+    }
+}
+
+impl From<i32> for crate::common::SpecStatus {
```
tiagolobocastro, line 130: If possible we should use a grpc type or the newtype pattern instead of i32

control-plane/grpc/src/pool/traits.rs
```diff
@@ -0,0 +1,295 @@
...
+    fn from(i: i32) -> Self {
+        match i {
+            0 => crate::common::SpecStatus::Creating,
+            1 => crate::common::SpecStatus::Created,
+            2 => crate::common::SpecStatus::Deleting,
+            3 => crate::common::SpecStatus::Deleted,
+            _ => crate::common::SpecStatus::Creating,
```
tiagolobocastro, line 137: This doesn't seem right? Should we have an unknown? or use a TryFrom ?
We can probably avoid this altogether if we use the grpc or newtype

control-plane/grpc/src/replica/client.rs
```diff
@@ -0,0 +1,198 @@
...
+    endpoint: Endpoint,
+}
+
+impl ReplicaClient {
+    pub async fn init<O: Into<Option<TimeoutOptions>>>(addr: Option<Uri>, opts: O) -> Self {
+        let opts = opts.into();
+        let timeout = opts
+            .map(|opt| opt.base_timeout())
+            .unwrap_or_else(|| humantime::parse_duration(DEFAULT_REQ_TIMEOUT).unwrap());
+        let endpoint = tonic::transport::Endpoint::from(addr.unwrap())
```
tiagolobocastro, line 41: if we're unwrapping addr, maybe it doesn't need to be an Option then?

control-plane/grpc/src/replica/client.rs
```diff
@@ -0,0 +1,198 @@
...
+        let ctx_timeout = ctx.map(|ctx| ctx.timeout_opts).flatten();
+        match ctx_timeout {
+            None => {
+                let timeout = timeout_grpc(op_id, self.base_timeout);
+                let endpoint = self
+                    .endpoint
+                    .clone()
+                    .connect_timeout(timeout)
+                    .timeout(timeout);
+                ReplicaGrpcClient::connect(endpoint.clone()).await.unwrap()
```
tiagolobocastro, line 63: We shouldn't unwrap here, reconnect should return a Result as it's possible that we may fail to connect.

Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors try

control-plane/agents/core/src/pool/service.rs
Abhinandan-Purkait: Done. Removed all possible unwraps.

common/src/types/v0/store/pool.rs
Abhinandan-Purkait: To handle this, I have used the corresponding grpc type to match each arm, so that we dont have a catch all thing.

Abhinandan-Purkait:
bors try

control-plane/grpc/src/replica/client.rs
```diff
@@ -0,0 +1,200 @@
...
+
+// RPC Replica Client
+#[derive(Clone)]
+pub struct ReplicaClient {
+    base_timeout: Duration,
+    endpoint: Endpoint,
+}
+
+impl ReplicaClient {
+    pub async fn init<O: Into<Option<TimeoutOptions>>>(addr: Uri, opts: O) -> Self {
```
paulyoong, line 36: Comments required on pub functions.

tests/tests-mayastor/tests/pools.rs
```diff
@@ -83,26 +84,34 @@ async fn create_pool_with_existing_disk() {
 #[tokio::test]
 async fn create_pool_idempotent() {
     let cluster = ClusterBuilder::builder().build().await.unwrap();
+    let pool_client = cluster.grpc_client().pool();
+    //cluster.grpc_client().pool_client();
```
paulyoong, line 88: Delete this line?

paulyoong (approved):
Nice work. Just got some minor comments.

control-plane/agents/core/src/lib.rs
```diff
@@ -0,0 +1,80 @@
+use common::ServiceError;
+use futures::{future::join_all, FutureExt};
+use grpc::{pool::server::PoolServer, replica::server::ReplicaServer};
+use http::Uri;
+use tracing::error;
+
+pub struct Service {
```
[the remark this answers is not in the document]
tiagolobocastro: Should we add #![warn(missing_docs)] to the top of lib.rs? So any missing docs should would be a warn and the build not pass CI (well, so I hope!)

control-plane/agents/core/src/lib.rs
```diff
@@ -0,0 +1,80 @@
...
+
+        threads.push(tonic_thread);
+
+        join_all(threads)
+            .await
+            .iter()
+            .for_each(|result| match result {
+                Err(error) => error!("Failed to wait for thread: {:?}", error),
+                Ok(Err(error)) => {
+                    error!("Error running channel thread: {:?}", error)
```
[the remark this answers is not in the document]
tiagolobocastro: This is where it's tricky now, now we have a mix of mbus channels and tonic routes. Maybe: 
```rust
error!(error=?error, "Error running service thread");
```
?

control-plane/agents/core/src/server.rs
```diff
@@ -75,6 +76,11 @@ pub(crate) struct CliArgs {
     /// Trace rest requests to the Jaeger endpoint agent
     #[structopt(long, short)]
     jaeger: Option<String>,
+    /// The GRPC Server URLs to connect to
+    /// (supports the http/https schema)
+    /// Default: 0.0.0.0:50051
```
tiagolobocastro, line 81: probably don't need this line, structopt will already show the default by itself

control-plane/rest/service/src/v0/pools.rs
```diff
@@ -1,19 +1,18 @@
...
     let destroy = match filter.clone() {
         Filter::NodePool(node_id, pool_id) => DestroyPool {
             node: node_id,
             id: pool_id,
         },
         Filter::Pool(pool_id) => {
-            let node_id = match MessageBus::get_pool(filter).await {
-                Ok(pool) => pool.node(),
+            let node_id = match client.get(filter, None).await {
+                Ok(pools) => pools.into_inner()[0].clone().node(),
```
tiagolobocastro, line 15: nit: I would use get(0) here as the slice will panic on an empty vec.

control-plane/rest/service/src/v0/pools.rs
```diff
@@ -50,27 +48,33 @@ impl apis::actix_server::Pools for RestApi {
     async fn get_node_pool(
         Path((node_id, pool_id)): Path<(String, String)>,
     ) -> Result<models::Pool, RestError<RestJsonError>> {
-        let pool = MessageBus::get_pool(Filter::NodePool(node_id.into(), pool_id.into())).await?;
-        Ok(pool.into())
+        let client = CORE_CLIENT.get().unwrap().pool();
+        let pool = client
+            .get(Filter::NodePool(node_id.into(), pool_id.into()), None)
+            .await?;
+        Ok(pool.into_inner()[0].clone().into())
```
tiagolobocastro, line 55: same thing here for all of these in regards to the [0]

tiagolobocastro (approved):
:rocket: 
Before merging please tidy up your commits, it's preferred to avoid things like "ack of comments" and squash those into the "original" commit.

control-plane/agents/core/src/lib.rs
```diff
@@ -0,0 +1,80 @@
...
+    fn shutdown_signal() -> tokio::sync::oneshot::Receiver<()> {
+        let mut signal_term =
+            tokio::signal::unix::signal(tokio::signal::unix::SignalKind::terminate()).unwrap();
+        let mut signal_int =
+            tokio::signal::unix::signal(tokio::signal::unix::SignalKind::interrupt()).unwrap();
+        let (stop_sender, stop_receiver) = tokio::sync::oneshot::channel();
+        tokio::spawn(async move {
+            tokio::select! {
+                _term = signal_term.recv() => {},
+                _int = signal_int.recv() => {},
```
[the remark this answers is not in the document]
Abhinandan-Purkait: Added these info
```
tracing::info!("SIGTERM received")
tracing::info!("SIGINT received")
```
Is that good enough ?

control-plane/agents/core/src/pool/tests.rs
```diff
@@ -167,7 +200,7 @@ async fn replica_spec(replica: &Replica) -> Option<ReplicaSpec> {
 #[tokio::test]
 async fn replica_transaction() {
     let cluster = ClusterBuilder::builder()
-        .with_rest(false)
+        .with_rest(true)
```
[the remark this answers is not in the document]
Abhinandan-Purkait: Sorry, I forgot to revert it back. We aren't using the rest.

control-plane/grpc/src/pool/client.rs
```diff
@@ -0,0 +1,130 @@
...
+
+// RPC Pool Client
+#[derive(Clone)]
+pub struct PoolClient {
+    base_timeout: Duration,
+    endpoint: Endpoint,
+}
+
+impl PoolClient {
+    pub async fn init<O: Into<Option<TimeoutOptions>>>(addr: Uri, opts: O) -> Self {
```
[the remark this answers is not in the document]
Abhinandan-Purkait: Yep, new sounds more reasonable.

control-plane/grpc/src/pool/traits.rs
```diff
@@ -0,0 +1,316 @@
...
+    fn from(data: &dyn DestroyPoolInfo) -> Self {
+        Self {
+            node: data.node_id(),
+            id: data.pool_id(),
+        }
+    }
+}
+
+// newtype to encapsulate the grpc poolstatus
+pub struct PoolStatus(pool_grpc::PoolStatus);
```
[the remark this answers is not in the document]
Abhinandan-Purkait: We actually don't need the newtype, I was able to use the grpc type and as per the `prost::Enumertation` we have `from_i32 ` method and `as i32`  for to and fro conversion.

control-plane/grpc/src/pool/traits.rs
```diff
@@ -0,0 +1,316 @@
...
+        message_bus,
+        message_bus::{
+            CreatePool, DestroyPool, Filter, NodeId, Pool, PoolDeviceUri, PoolId, PoolState,
+        },
+        store::pool::{PoolLabel, PoolSpec, PoolSpecStatus},
+    },
+};
+
+/// Trait implemented by services which support pool operations.
+/// This trait can only be implemented on types which support the PoolInfo trait.
```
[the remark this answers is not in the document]
Abhinandan-Purkait: Yep, removed this line.

control-plane/grpc/src/pool/client.rs
```diff
@@ -0,0 +1,130 @@
...
+            Filter::NodePool(node_id, pool_id) => GetPoolsRequest {
+                filter: Some(get_pools_request::Filter::NodePool(NodePoolFilter {
+                    node_id: node_id.into(),
+                    pool_id: pool_id.into(),
+                })),
+            },
+            _ => GetPoolsRequest { filter: None },
+        };
+        let response = client.clone().get_pools(req).await?.into_inner();
+        match response.reply.unwrap() {
```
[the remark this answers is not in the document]
Abhinandan-Purkait: Here the unwrap being handled using match. Is there a better way to handle this?

control-plane/grpc/src/misc/traits.rs
```diff
@@ -0,0 +1,142 @@
...
+            extra: err.extra,
+        }
+    }
+}
+
+impl From<crate::common::ReplyError> for ReplyError {
+    fn from(err: crate::common::ReplyError) -> Self {
+        ReplyError {
+            kind: common::ReplyErrorKind::from_i32(err.clone().kind)
+                .unwrap()
```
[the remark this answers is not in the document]
Abhinandan-Purkait: Okay, so instead of unwrap() we would do unwrap_or(UNKNOWN) something like that ?

control-plane/rest/service/src/v0/pools.rs
Abhinandan-Purkait: But then we have to handle the unwrap again right?

Abhinandan-Purkait:
```
match pool.into_inner().get(0) {
            None => Err(ReplyError {
                kind: ReplyErrorKind::NotFound,
                resource: ResourceKind::Pool,
                source: "Requested pool was not found".to_string(),
                extra: format!("Pool id : {}", pool_id),
            }
            .into()),
            Some(pool) => Ok(pool.clone().into()),
        }
```
Maybe something like this?

tiagolobocastro: Yeah, you can put that on a function, eg: fn pool(i: Option<&Pool>) -> Result<Pool, Error>, and then just pool(xxx)? on all the places where we need to do it

control-plane/agents/core/src/lib.rs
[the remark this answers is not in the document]
paulyoong: I think so.

control-plane/rest/service/src/v0/pools.rs
tiagolobocastro: I wonder if we should have a get that returns a single pool/replica.

control-plane/grpc/src/misc/traits.rs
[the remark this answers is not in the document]
tiagolobocastro: something like that yeah

control-plane/grpc/src/pool/client.rs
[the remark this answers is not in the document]
paulyoong: We are matching against the returned value of the unwrap. If the unwrap itself fails we will panic. Is it possible to return a suitable error in the event that we can't unwrap?

control-plane/rest/service/src/v0/pools.rs
Abhinandan-Purkait: even if we create a get that returns single, at some point be it at service or client we have to extract from a vector as grpc req to mayastor always returns a vector.

Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors merge

PR #211, opened 2022-01-31
chore(helm): add namespace value

paulyoong:
Add a namespace value which is used by the templates.

Update generate-yamls.sh to take a namespace argument. This defaults to
mayastor and is used when generating the YAML files using Helm.

### DIFF
--- a/chart/templates/core-agents-deployment.yaml
+++ b/chart/templates/core-agents-deployment.yaml
@@ -2,7 +2,7 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: core-agents
-  namespace: mayastor
+  namespace: {{ .Release.Namespace }}
   labels:
     app: core-agents
 spec:
--- a/scripts/deploy/generate-yamls.sh
+++ b/scripts/deploy/generate-yamls.sh
@@ -21,6 +21,7 @@ tag=
 helm_string=
 helm_file=
 helm_flags=
+namespace="mayastor"
 
 help() {
   cat <<EOF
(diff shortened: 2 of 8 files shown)


chart/templates/core-agents-deployment.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: core-agents
-  namespace: mayastor
+  namespace: {{.Values.namespace}}
```
tiagolobocastro, line 5: Should this be `{{ .Release.Namespace }}` ? As it stands this isn't actually picking up the -n provided to helm, it's using what you've defined in the yaml file?

chart/values.yaml
```diff
@@ -86,3 +86,5 @@ csi:
     requests:
       cpu: "16m"
       memory: "64Mi"
+
+namespace: "mayastor"
```
tiagolobocastro, line 90: do we need this, can't we use just use "{{ .Release.Namespace }}" ?

scripts/deploy/generate-yamls.sh
```diff
@@ -80,6 +81,10 @@ while [ "$#" -gt 0 ]; do
       shift
       tag=$1
       ;;
+    -n)
```
tiagolobocastro, line 84: we should add this to the help usage above

chart/templates/core-agents-deployment.yaml
paulyoong: You're correct. Thanks, I've changed it.

chart/values.yaml
paulyoong: I've removed this.

scripts/deploy/generate-yamls.sh
paulyoong: Done. Thanks.

tiagolobocastro (approved):

paulyoong:
bors merge

PR #212, opened 2022-02-01
fix: use dns name to reach rest

tiagolobocastro:
chore: remove helm generation from precommit

--------------

fix: use dns name to reach rest

Using the host name creates a race as rest's node port must start ahead of its consumers.
Instead, simply use the dns name, which will be updated whenever rest is ready.
Resolves: openebs/mayastor#1076
Resolves: CAS-1288


paulyoong (approved):

tiagolobocastro:
bors merge

PR #216, opened 2022-02-08
fix: don't disown replicas from faulted volumes and recreate faulted nexuses

tiagolobocastro:
fix: don't disown replicas from faulted volumes

Don't attempt to disown replicas if the replica volume is faulted.
Don't disown replicas if the volume target (nexus) contains the replica.
Update the replica owners on core-agent startup - we might get drop the replica nexus owners at
some point since we have obtain then from the nexus specs.

Resolves: CAS-1295

-------------

chore: don't trace reconciliation unless TRACE is set or there is work to do

For clusters with high number of volumes there are a lot of traces being sent that sometimes
seem to overwhelm the telemetry exporter.
During reconciliation we should not create spans needlessly unless we're on TRACE level or unless
the reconciler actually did work.
TODO: We need a generic strategy to cater for this

-------------

fix(nexus): recreate faulted nexuses when appropriate

When the healthy replicas of a faulted nexus are Online we should attempt to recreate said nexuses.
This is done by first destroying the faulted nexus. The missing nexus reconciler will then recreate
the nexus.

When no healthy children are available, the faulted nexus are left behind until such time the node
crashes or the volume is unpublished.

Resolves: CAS-1296

### DIFF
--- a/tests/bdd/test_volume_nexus.py
+++ b/tests/bdd/test_volume_nexus.py
@@ -1,0 +1,167 @@
+"""Managing a Volume Nexus Target feature tests."""
+import os
+
+from pytest_bdd import given, scenario, then, when
+import pytest
+from retrying import retry
+
+from common.deployer import Deployer
+from common.apiclient import ApiClient
+from common.docker import Docker
+
+from openapi.model.create_pool_body import CreatePoolBody
+from openapi.model.create_volume_body import CreateVolumeBody
+from openapi.model.protocol import Protocol
+from openapi.model.volume_policy import VolumePolicy
+from openapi.model.nexus_state import NexusState
+from openapi.model.topology import Topology
+from openapi.model.pool_topology import PoolTopology
+from openapi.model.labelled_topology import LabelledTopology
+
+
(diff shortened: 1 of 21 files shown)


tiagolobocastro:
bors try

tests/bdd/test_volume_nexus.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+@retry(wait_fixed=200, stop_max_delay=5000)
+def check_nexus_removed():
+    volume = ApiClient.volumes_api().get_volume(VOLUME_UUID)
+    assert (
+        not hasattr(volume.state, "target")
+        # or it might have been recreated if we lost the "race"...
+        or NexusState(volume.state.target["state"]) == NexusState("Online")
+        or NexusState(volume.state.target["state"]) == NexusState("Degraded")
```
paulyoong, line 154: I understand why this is necessary, but I wonder if we could make this check more robust somehow. I know it's unlikely, but what happens if, in the future, there was the possibility that the nexus got set from "Faulted" to another state without being recreated.

paulyoong (approved):

PR #218, opened 2022-02-09
refactor(replace nats): add proto for the volume transport

Abhinandan-Purkait:
1. Add Volume and Nexus Proto.
2. Move Protocol enum to common from replica as that is seemingly used in multiple protos.


gila (commented):
Why are the proto files not in the mayastor-api repo?

paulyoong:
> Why are the proto files not in the mayastor-api repo?

I was under the impression that this repo was just for Mayastor - making it convenient for other components to consume the API. The control plane proto files are only consumed by other control plane components, so it seemed simpler to keep them in the one place.

control-plane/grpc/proto/v1/nexus/nexus.proto
```diff
@@ -0,0 +1,60 @@
...
+  // protocol used for exposing the nexus
+  common.Protocol share = 9;
+}
+
+enum NexusStatus {
+  // Default Unknown state
+  Unknown = 0;
+  // healthy and working
+  Online = 1;
+  // not healthy but is able to serve IO (i.e. rebuild is in progress)
```
paulyoong, line 36: Nit: A rebuild is not necessarily in progress.

control-plane/grpc/proto/v1/nexus/nexus.proto
```diff
@@ -0,0 +1,60 @@
...
+
+enum ChildState {
+  // Default Unknown state
+  UnknownState = 0;
+  // healthy and contains the latest bits
+  OnlineState = 1;
+  // rebuild is in progress (or other recoverable error)
+  DegradedState = 2;
+  // unrecoverable error (control plane must act)
+  FaultedState = 3;
```
paulyoong, line 59: Do we need the "State" suffix because otherwise it will clash with the NexusStatus enum? If so, I would suggest to change the names to something like "ChildFaulted" so that the state is attributed to the resource specifically. That way, if we introduce another resource, with the same set of states we just change the prefix.

control-plane/grpc/proto/v1/volume/volume.proto
```diff
@@ -0,0 +1,270 @@
...
+  // Client specification of the volume
+  VolumeSpec spec = 1;
+  // Control plane related data for the volume
+  Metadata metadata = 2;
+}
+
+message VolumeSpec {
+  // uuid of the volume
+  google.protobuf.StringValue uuid = 1;
+  // Size that the volume should be.
```
paulyoong, line 36: Maybe mention the units of the size i.e. bytes.

control-plane/grpc/proto/v1/volume/volume.proto
```diff
@@ -0,0 +1,270 @@
...
+
+message Metadata {
+  // spec status of the volume
+  common.SpecStatus status = 4;
+}
+
+message VolumeTarget {
+  // The node where front-end IO will be sent to
+  string node_id = 1;
+  // The node where front-end IO will be sent to
```
paulyoong, line 60: I think "node" is meant to be "nexus ID"

control-plane/grpc/proto/v1/volume/volume.proto
```diff
@@ -0,0 +1,270 @@
...
+  // replicas can only be placed on these nodes
+  repeated string allowed_nodes = 1;
+  // preferred nodes to place the replicas
+  repeated string preferred_nodes = 2;
+}
+
+message VolumeState {
+  // uuid of the volume
+  google.protobuf.StringValue uuid = 1;
+  // Size that the volume should be.
```
paulyoong, line 110: This is the actual size of the volume as returned by Mayastor.

control-plane/grpc/proto/v1/volume/volume.proto
```diff
@@ -0,0 +1,270 @@
...
+
+// volume creation request
+message CreateVolumeRequest {
+  // uuid of the volume
+  google.protobuf.StringValue uuid = 1;
+  // Size that the volume should be.
+  uint64 size = 2;
+  // Volume labels.
+  optional common.StringMapValue labels = 3;
+  // Number of children the volume should have.
```
paulyoong, line 144: For consistency maybe "children" should be "replicas" here.

control-plane/grpc/proto/v1/nexus/nexus.proto
Abhinandan-Purkait: - Removing `(i.e. rebuild is in progress)`

control-plane/grpc/proto/v1/nexus/nexus.proto
Abhinandan-Purkait: Yeah, we cannot add same thing in two different enums. Okay then I will add the prefix `Child` and remove suffix `state`

control-plane/grpc/proto/v1/replica/replica.proto
```diff
@@ -71,7 +60,7 @@ message CreateReplicaRequest{
   // size of the replica in bytes
   uint64 size = 6;
   // protocol to expose the replica over
-  Protocol share = 7;
+  common.Protocol share = 7;
```
tiagolobocastro, line 63: maybe replicas need their own protocol, they can only be shared by Nvmf.

control-plane/grpc/proto/v1/volume/volume.proto
```diff
@@ -0,0 +1,270 @@
...
+import "v1/nexus/nexus.proto";
+import "v1/replica/replica.proto";
+import "google/protobuf/empty.proto";
+import "google/protobuf/wrappers.proto";
+
+package v1.volume;
+
+// A Mayastor Volume
+// It has a spec which is the specification provided by the creator
+// It has a state if such state is retrieved from a mayastor storage node
```
tiagolobocastro, line 13: ... from mayastor storage nodes (plural since the replicas are distributed)

control-plane/grpc/proto/v1/volume/volume.proto
```diff
@@ -0,0 +1,270 @@
...
+service VolumeGrpc {
+  rpc CreateVolume (CreateVolumeRequest) returns (CreateVolumeReply) {}
+  rpc DestroyVolume (DestroyVolumeRequest) returns (DestroyVolumeReply) {}
+  rpc GetVolumes (GetVolumesRequest) returns (GetVolumesReply) {}
+  rpc PublishVolume (PublishVolumeRequest) returns (PublishVolumeReply) {}
+  rpc UnpublishVolume (UnpublishVolumeRequest) returns (UnpublishVolumeReply) {}
+  rpc ShareVolume (ShareVolumeRequest) returns (ShareVolumeReply) {}
+  rpc UnshareVolume (UnshareVolumeRequest) returns (UnshareVolumeReply) {}
+  rpc SetVolumeReplica (SetVolumeReplicaRequest) returns (SetVolumeReplicaReply) {}
+}
```
tiagolobocastro, line 270: Should we start adding readiness probes? Though I wonder if that should be a separate service.

control-plane/grpc/src/lib.rs
```diff
@@ -18,3 +18,15 @@ pub(crate) mod pool {
 pub(crate) mod replica {
     tonic::include_proto!("v1.replica");
 }
+
+// Nexus GRPC module for the autogenerated pool code
```
tiagolobocastro, line 22: Missing '/' ?
Also on the other modules

control-plane/grpc/src/lib.rs
```diff
@@ -18,3 +18,15 @@ pub(crate) mod pool {
...
+#[allow(clippy::enum_variant_names)]
```
tiagolobocastro, line 23: Could you doc why we needed to add this ?

control-plane/grpc/src/operations/replica/traits.rs
```diff
@@ -71,7 +72,7 @@ impl From<replica::Replica> for Replica {
             pool: replica.pool_id.into(),
             thin: replica.thin,
             size: replica.size,
-            share: replica::Protocol::from_i32(replica.share).unwrap().into(),
+            share: common::Protocol::from_i32(replica.share).unwrap().into(),
```
tiagolobocastro, line 75: Maybe we need a tryFrom here to avoid unwrapping

tiagolobocastro (approved):

control-plane/grpc/proto/v1/replica/replica.proto
Abhinandan-Purkait: Its there, The ReplicaShareProtocol enum

control-plane/grpc/proto/v1/volume/volume.proto
Abhinandan-Purkait: We can probab;y once each of these features are ready.

control-plane/grpc/src/lib.rs
Abhinandan-Purkait: This is basically needed because Clippy wont allow same prefix and suffix to the enum members

control-plane/grpc/src/operations/replica/traits.rs
Abhinandan-Purkait: I will take this up on the next PR, it would include changes at several places.

control-plane/grpc/proto/v1/replica/replica.proto
Abhinandan-Purkait: I have used it according to its type in message_bus. https://github.com/openebs/mayastor-control-plane/blob/263a8222fb9fe632cb5a977df3f38f1ca95dbfb6/common/src/types/v0/message_bus/replica.rs#L150

tiagolobocastro: Ah so changing it will break things? Ok we can do it on another PR

paulyoong (approved):

Abhinandan-Purkait:
bors merge

PR #219, opened 2022-02-09
feat: enable long-lived grpc clients whilst supporting per request timeouts

tiagolobocastro:
feat: new base grpc client that supports adding timeouts per request

This can be used as a inner helper which provides common functionality of creating a new client
connection and setting specific timeouts for each request.
This allows us to use long-lived connections with keep-alives.

-------------

refactor(pool): use of the new generic gRPC client

-------------

refactor: keep the pool gRPC server handlers cancellable

The specific service implementations will be responsible for making it otherwise. This gives finer
grained control to decide how to implement each handler.

-------------

fix: spawn core pool create/destroy to avoid early drop on timeout

The existing code does not handle being dropped very well, as it will leave things in a
potentially "bad" state.
TODO: handle serive handlers drops gracefully. This will be tricky as eg ETCD client is async and so
we can't really flush things to ETCD in a nice way on drop.

For now, try to avoid dropping the futures when the client times out by spawning it.
NOTE: the future can still be dropped on reactor shutdown.

-------------

test: add grpc request timeout test

This validates that timeouts by modifying the grpc timeout header works.

------------

refactor(rest): use long lived gRPC connections

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1504,6 +1504,7 @@ version = "0.1.0"
 dependencies = [
  "common-lib",
  "humantime",
+ "once_cell",
  "opentelemetry",
  "opentelemetry-semantic-conventions",
  "prost 0.8.0",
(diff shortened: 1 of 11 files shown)


tiagolobocastro:
bors try

paulyoong (approved):

Abhinandan-Purkait (approved):

PR #221, opened 2022-02-10
fix(rest): enable http2 keep alive interval and timeout

tiagolobocastro:
Enables http2 keep alives - this allows us to failover to servers even when a node crashes and the
core agent is recreated on another node.


paulyoong (approved):

tiagolobocastro:
bors merge

PR #223, opened 2022-02-14
refactor(replace NATS): add volume rpc server, client, service impl and traits

Abhinandan-Purkait:
1. Add Volume rpc to common type conversion.
2. Add server, client and service implementations
3. Remove get volumes by node from openapi.
4. Add volume service liveness probe

### DIFF
--- a/common/src/mbus_api/mod.rs
+++ b/common/src/mbus_api/mod.rs
@@ -413,15 +413,6 @@ impl ReplyError {
             extra: "".to_string(),
         }
     }
-    /// used when we error while unwrapping
-    pub fn unwrap_err(resource: ResourceKind) -> Self {
-        Self {
-            kind: ReplyErrorKind::Aborted,
-            resource,
-            source: "Error unwrapping to desired type".to_string(),
-            extra: "".to_string(),
-        }
-    }
     /// used when we get an invalid argument
     pub fn invalid_argument(resource: ResourceKind, arg_name: &str, error: String) -> Self {
         Self {
--- a/control-plane/grpc/src/operations/volume/mod.rs
+++ b/control-plane/grpc/src/operations/volume/mod.rs
@@ -1,4 +1,3 @@
-#![warn(missing_docs)]
 /// Volume grpc Client related code
 pub mod client;
 
--- a/control-plane/grpc/src/operations/volume/server.rs
+++ b/control-plane/grpc/src/operations/volume/server.rs
@@ -1,5 +1,6 @@
 use crate::{
-    operations::volume::traits::{ValidateRequestTypes, VolumeOperations},
+    misc::traits::ValidateRequestTypes,
+    operations::volume::traits::VolumeOperations,
     volume::{
         create_volume_reply, get_volumes_reply, publish_volume_reply, set_volume_replica_reply,
         share_volume_reply, unpublish_volume_reply,
--- a/control-plane/grpc/src/operations/volume/traits.rs
+++ b/control-plane/grpc/src/operations/volume/traits.rs
@@ -1,6 +1,7 @@
 use crate::{
     common,
     grpc_opts::Context,
+    misc::traits::ValidateRequestTypes,
     nexus, replica, volume,
     volume::{
         get_volumes_request, CreateVolumeRequest, DestroyVolumeRequest, PublishVolumeRequest,
@@ -643,14 +644,6 @@ impl TryFrom<get_volumes_request::Filter> for Filter {
             }
         })
     }
-}
-
-/// Trait to validate the Grpc type by an intermediate conversion
-pub trait ValidateRequestTypes {
-    /// The validated type
-    type Validated;
-    /// The method that is needed to be implemented to ensure validation
-    fn validated(self) -> Result<Self::Validated, ReplyError>;
 }
 
 /// Trait to be implemented for CreateVolume operation
--- a/control-plane/grpc/src/lib.rs
+++ b/control-plane/grpc/src/lib.rs
@@ -1,6 +1,7 @@
 pub mod client;
 pub mod grpc_opts;
 pub mod misc;
+/// All server, client implementations and the traits
 pub mod operations;
 
 /// Common module for all the misc operations
--- a/control-plane/grpc/src/misc/traits.rs
+++ b/control-plane/grpc/src/misc/traits.rs
@@ -1,6 +1,14 @@
 use crate::common;
 use common_lib::mbus_api::{ReplyError, ReplyErrorKind, ResourceKind};
 
+/// Trait to validate the Grpc type by an intermediate conversion
+pub trait ValidateRequestTypes {
+    /// The validated type
+    type Validated;
+    /// The method that is needed to be implemented to ensure validation
+    fn validated(self) -> Result<Self::Validated, ReplyError>;
+}
+
 impl From<ResourceKind> for common::ResourceKind {
     fn from(kind: ResourceKind) -> Self {
         match kind {
--- a/control-plane/grpc/src/operations/mod.rs
+++ b/control-plane/grpc/src/operations/mod.rs
@@ -1,3 +1,4 @@
+#![warn(missing_docs)]
 /// module for all corresponding client, server, traits for nexus rpc transport
 pub mod nexus;
 
--- a/control-plane/grpc/src/operations/nexus/mod.rs
+++ b/control-plane/grpc/src/operations/nexus/mod.rs
@@ -1,1 +1,2 @@
+/// Nexus traits for the transport
 pub mod traits;
--- a/control-plane/grpc/src/operations/pool/traits.rs
+++ b/control-plane/grpc/src/operations/pool/traits.rs
@@ -19,16 +19,19 @@ use std::convert::TryFrom;
 /// Trait implemented by services which support pool operations.
 #[tonic::async_trait]
 pub trait PoolOperations: Send + Sync {
+    /// Create a pool
     async fn create(
         &self,
         pool: &dyn CreatePoolInfo,
         ctx: Option<Context>,
     ) -> Result<Pool, ReplyError>;
+    /// Destroy a pool
     async fn destroy(
         &self,
         pool: &dyn DestroyPoolInfo,
         ctx: Option<Context>,
     ) -> Result<(), ReplyError>;
+    /// Get pools based on the filters
     async fn get(&self, filter: Filter, ctx: Option<Context>) -> Result<Pools, ReplyError>;
 }
 
@@ -43,7 +46,13 @@ impl TryFrom<pool::Pool> for Pool {
                 disks: pool_state.disks_uri.iter().map(|i| i.into()).collect(),
                 status: match pool::PoolStatus::from_i32(pool_state.status) {
                     Some(status) => status.into(),
-                    None => return Err(ReplyError::unwrap_err(ResourceKind::Pool)),
+                    None => {
+                        return Err(ReplyError::invalid_argument(
+                            ResourceKind::Pool,
+                            "pool.state.status",
+                            "".to_string(),
+                        ))
+                    }
                 },
                 capacity: pool_state.capacity,
                 used: pool_state.used,
@@ -54,11 +63,21 @@ impl TryFrom<pool::Pool> for Pool {
             Some(pool_definition) => {
                 let pool_spec = match pool_definition.spec {
                     Some(spec) => spec,
-                    None => return Err(ReplyError::unwrap_err(ResourceKind::Pool)),
+                    None => {
+                        return Err(ReplyError::missing_argument(
+                            ResourceKind::Pool,
+                            "pool.definition.spec",
+                        ))
+                    }
                 };
                 let pool_meta = match pool_definition.metadata {
                     Some(meta) => meta,
-                    None => return Err(ReplyError::unwrap_err(ResourceKind::Pool)),
+                    None => {
+                        return Err(ReplyError::missing_argument(
+                            ResourceKind::Pool,
+                            "pool.definition.metadata",
+                        ))
+                    }
                 };
                 let pool_spec_status = match common::SpecStatus::from_i32(pool_meta.status) {
                     Some(status) => match status {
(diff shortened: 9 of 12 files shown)


Abhinandan-Purkait-Bolt:
bors try

common/src/mbus_api/mod.rs
```diff
@@ -665,6 +665,11 @@ impl TimeoutOptions {
     pub fn keep_alive_timeout(&self) -> Duration {
         self.keep_alive_timeout
     }
+
+    /// get the max retires
+    pub fn max_retires(&self) -> Option<u32> {
```
paulyoong, line 670: Typo: retries

common/src/types/v0/message_bus/nexus.rs
```diff
@@ -125,6 +125,7 @@ impl From<i32> for NexusShareProtocol {
     fn from(src: i32) -> Self {
         match src {
             1 => Self::Nvmf,
+            0 => Self::Nvmf,
```
paulyoong, line 128: Nit: Could we put this first so the match is in ascending order starting from 0? (Alternatively you could do `0 | 1 => Self::Nvmf,`)

control-plane/agents/core/src/volume/mod.rs
```diff
@@ -15,18 +9,11 @@ pub mod specs;
 
 pub(crate) fn configure(builder: common::Service) -> common::Service {
     let registry = builder.get_shared_state::<Registry>().clone();
+    let new_service = service::Service::new(registry);
```
paulyoong, line 12: Like we have done with the other services, can we just create the `Arc` here and then just pass it to `VolumeServer::new` below?

control-plane/agents/core/src/volume/mod.rs
```diff
@@ -15,18 +9,11 @@ pub mod specs;
...
-        .with_shared_state(service::Service::new(registry))
-        .with_subscription(handler!(GetVolumes))
-        .with_subscription(handler!(CreateVolume))
-        .with_subscription(handler!(DestroyVolume))
-        .with_subscription(handler!(ShareVolume))
-        .with_subscription(handler!(UnshareVolume))
-        .with_subscription(handler!(PublishVolume))
-        .with_subscription(handler!(UnpublishVolume))
-        .with_subscription(handler!(SetVolumeReplica))
+        .with_shared_state(new_service)
```
paulyoong, line 15: Do we need to pass this in? Aren't we just using the VolumeServer?

control-plane/grpc/src/operations/volume/client.rs
```diff
@@ -0,0 +1,203 @@
...
+    },
+};
+use common_lib::{
+    mbus_api::{v0::Volumes, ReplyError, ResourceKind, TimeoutOptions},
+    types::v0::message_bus::{Filter, MessageIdVs, Volume},
+};
+use std::{convert::TryFrom, ops::Deref};
+use tonic::transport::{Channel, Uri};
+
+/// RPC Pool Client
```
paulyoong, line 21: `Pool` -> `Volume`

control-plane/grpc/src/operations/volume/client.rs
```diff
@@ -0,0 +1,203 @@
...
+}
+
+impl Deref for VolumeClient {
+    type Target = Client<VolumeGrpcClient<Channel>>;
+    fn deref(&self) -> &Self::Target {
+        &self.inner
+    }
+}
+
+/// Implement pool operations supported by the Pool RPC client.
```
paulyoong, line 42: `pool` -> `volume`

control-plane/grpc/src/operations/volume/mod.rs
```diff
@@ -0,0 +1,9 @@
+#![warn(missing_docs)]
```
paulyoong, line 1: :+1: good idea

control-plane/grpc/src/operations/volume/server.rs
```diff
@@ -0,0 +1,174 @@
...
+    pub fn into_grpc_server(self) -> VolumeGrpcServer<VolumeServer> {
+        VolumeGrpcServer::new(self)
+    }
+}
+
+impl Drop for VolumeServer {
+    fn drop(&mut self) {
+        println!("Dropping volume server")
+    }
+}
```
paulyoong, line 40: Do we require this?

paulyoong (approved):
Nice work :+1:

control-plane/agents/core/src/volume/service.rs
```diff
@@ -10,12 +10,116 @@ use common_lib::{
...
+        let service = self.clone();
+        let volume =
+            tokio::spawn(async move { service.create_volume(&create_volume).await }).await??;
+        Ok(volume)
+    }
+
+    async fn get(&self, filter: Filter, _ctx: Option<Context>) -> Result<Volumes, ReplyError> {
+        let req = GetVolumes { filter };
+        let service = self.clone();
+        let volumes = tokio::spawn(async move { service.get_volumes(&req).await }).await??;
```
tiagolobocastro, line 43: don't need to spawn the get

control-plane/grpc/src/operations/volume/client.rs
```diff
@@ -0,0 +1,203 @@
...
+/// This converts the client side data into a RPC request.
+#[tonic::async_trait]
+impl VolumeOperations for VolumeClient {
+    async fn create(
+        &self,
+        create_volume_req: &dyn CreateVolumeInfo,
+        ctx: Option<Context>,
+    ) -> Result<Volume, ReplyError> {
+        let req = self.request(create_volume_req, ctx, MessageIdVs::CreateVolume);
+        let response = self.client().clone().create_volume(req).await?.into_inner();
```
tiagolobocastro, line 52: don't need to keep calling clone as client returns a cloned client already?

control-plane/grpc/src/operations/volume/mod.rs
tiagolobocastro: could we add this to the lib.rs instead? then it applies to whole package?

control-plane/grpc/src/operations/volume/server.rs
```diff
@@ -0,0 +1,174 @@
...
+    }
+    /// coverts the volumeserver to its corresponding grpc server type
+    pub fn into_grpc_server(self) -> VolumeGrpcServer<VolumeServer> {
+        VolumeGrpcServer::new(self)
+    }
+}
+
+impl Drop for VolumeServer {
+    fn drop(&mut self) {
+        println!("Dropping volume server")
```
tiagolobocastro, line 38: Don't think this should be checked in..

control-plane/grpc/src/operations/volume/server.rs
```diff
@@ -0,0 +1,174 @@
...
+                reply: Some(create_volume_reply::Reply::Error(err.into())),
+            })),
+        }
+    }
+    async fn destroy_volume(
+        &self,
+        request: tonic::Request<DestroyVolumeRequest>,
+    ) -> Result<tonic::Response<DestroyVolumeReply>, tonic::Status> {
+        let req = request.into_inner();
+        // Dispatch the destroy call to the registered service.
```
tiagolobocastro, line 64: do we need this comment?

control-plane/grpc/src/operations/volume/traits.rs
```diff
@@ -0,0 +1,908 @@
...
+            ReplicaTopology, SetVolumeReplica, ShareVolume, Topology, UnpublishVolume,
+            UnshareVolume, Volume, VolumeId, VolumeLabels, VolumePolicy, VolumeShareProtocol,
+            VolumeState,
+        },
+        store::volume::{VolumeSpec, VolumeSpecStatus, VolumeTarget},
+    },
+};
+use std::{collections::HashMap, convert::TryFrom};
+
+/// all volume crud operations to be a part of the VolumeOperations trait
```
tiagolobocastro, line 25: Should we start the doc-comments with upper case?

control-plane/grpc/src/operations/volume/traits.rs
```diff
@@ -0,0 +1,908 @@
...
+    }
+}
+
+impl From<Volumes> for volume::Volumes {
+    fn from(volumes: Volumes) -> Self {
+        volume::Volumes {
+            volumes: volumes
+                .into_inner()
+                .iter()
+                .map(|volume| volume.clone().into())
```
tiagolobocastro, line 232: Don't need to clone here as you're already consuming volumes with into_inner ?

control-plane/grpc/src/operations/volume/traits.rs
```diff
@@ -0,0 +1,908 @@
...
+        Ok(Volume::new(volume_spec, volume_state))
+    }
+}
+
+impl TryFrom<volume::Volumes> for Volumes {
+    type Error = ReplyError;
+    fn try_from(grpc_volume_type: volume::Volumes) -> Result<Self, Self::Error> {
+        let mut volumes: Vec<Volume> = vec![];
+        for volume in grpc_volume_type.volumes {
+            volumes.push(Volume::try_from(volume.clone())?)
```
tiagolobocastro, line 220: don't need to clone here if you consume grpc_volume_type? Also I'd consider renaming grpc_volume_type to grpc_volumes or something like that?

control-plane/grpc/src/operations/volume/traits.rs
```diff
@@ -0,0 +1,908 @@
...
+}
+
+/// trait to be implemented for CreateVolume operation
+pub trait CreateVolumeInfo: Send + Sync {
+    /// uuid of the volume
+    fn uuid(&self) -> VolumeId;
+    /// size in bytes of the volume
+    fn size(&self) -> u64;
+    /// no of replicas of the volume
+    fn replicas(&self) -> u64;
```
tiagolobocastro, line 509: nit: we tend to use u8 for the number of replicas IIRC but that's fine anyway

control-plane/grpc/src/operations/volume/traits.rs
```diff
@@ -0,0 +1,908 @@
...
+    }
+
+    fn labels(&self) -> Option<VolumeLabels> {
+        self.labels.clone()
+    }
+}
+
+impl CreateVolumeInfo for CreateVolumeRequest {
+    fn uuid(&self) -> VolumeId {
+        VolumeId::try_from(self.uuid.clone().unwrap()).unwrap()
```
tiagolobocastro, line 546: hmm, this is a bit annoying... How wonder if you need to have another CreateVolumeRequest that impl's a TryFrom and allows us to return errors? We have the same issue below on the TODO.
@paulyoong

tests/tests-mayastor/src/lib.rs
```diff
@@ -88,6 +92,34 @@ impl Cluster {
         self.grpc_client.as_ref().unwrap()
     }
 
+    pub async fn volume_service_liveness(
+        &self,
+        client: &dyn VolumeOperations,
+        timeout_opts: Option<TimeoutOptions>,
+    ) -> Result<bool, ReplyError> {
+        let timeout_opts = if timeout_opts.clone().is_none() {
```
tiagolobocastro, line 100: you can simply match and return a TimeoutOpts rather than an option.

control-plane/agents/core/src/volume/mod.rs
Abhinandan-Purkait-Bolt: Nope. Removing it.

control-plane/grpc/src/operations/volume/client.rs
Abhinandan-Purkait-Bolt: Right. Removing them.

control-plane/grpc/src/operations/volume/mod.rs
Abhinandan-Purkait-Bolt: Yeah. I wanted to do that in a separate PR as that would need addition of comments in other files as well, not related to volume ops.

control-plane/grpc/src/operations/volume/server.rs
Abhinandan-Purkait-Bolt: Yeah, added that for testing. Removing it.

control-plane/grpc/src/operations/volume/server.rs
Abhinandan-Purkait-Bolt: Removing it.

control-plane/grpc/src/operations/volume/traits.rs
Abhinandan-Purkait-Bolt: https://github.com/openebs/mayastor-control-plane/blob/05650c5cf673cf1433d3a68b4991674b87cfc214/common/src/types/v0/message_bus/volume.rs#L355

In the message bus type its like this, thats why I made it u64, so that I won't need to cast again.

control-plane/grpc/src/operations/volume/traits.rs
Abhinandan-Purkait-Bolt: For volumeId it maybe fine as uuid is system generated, so there shouldn't be anycase ideally where we would be able to convert. But yeah, for topology I am not really sure what to do.

control-plane/grpc/src/operations/volume/traits.rs
Abhinandan-Purkait-Bolt: We need the clone here because, `|volume|` is a `&Volume` type, so cloning would give us the `Volume` type

Abhinandan-Purkait:
bors try

common/src/mbus_api/mod.rs
```diff
@@ -665,6 +680,11 @@ impl TimeoutOptions {
     pub fn keep_alive_timeout(&self) -> Duration {
         self.keep_alive_timeout
     }
+
+    /// get the max retries
+    pub fn max_retires(&self) -> Option<u32> {
```
tiagolobocastro, line 685: max_retries?

control-plane/grpc/src/operations/volume/traits.rs
```diff
@@ -0,0 +1,1072 @@
...
+}
+
+impl ValidateRequestTypes for CreateVolumeRequest {
+    type Validated = ValidatedCreateVolumeRequest;
+    fn validated(self) -> Result<Self::Validated, ReplyError> {
+        Ok(ValidatedCreateVolumeRequest {
+            uuid: match self.uuid.clone() {
+                Some(uuid) => match VolumeId::try_from(uuid) {
+                    Ok(volumeid) => volumeid,
+                    Err(_) => return Err(ReplyError::unwrap_err(ResourceKind::Volume)),
```
tiagolobocastro, line 598: rather than unwrap_err, I think we want to pass the error string into invalid_argument_err

control-plane/grpc/src/operations/volume/traits.rs
```diff
@@ -305,7 +418,12 @@ impl TryFrom<volume::NodeTopology> for NodeTopology {
                     NodeTopology::Explicit(explicit.into())
                 }
             },
-            None => return Err(ReplyError::unwrap_err(ResourceKind::Volume)),
+            None => {
+                return Err(ReplyError::missing_argument(
+                    ResourceKind::Volume,
+                    "node_topology.topology",
```
Abhinandan-Purkait, line 424: FYI, this `.topology` refers to the grpc's node_topology type, which represents it in this way.

control-plane/grpc/src/operations/volume/traits.rs
Abhinandan-Purkait: Resolving this as we have handled this using the `ValidateRequestTypes` trait.

Abhinandan-Purkait:
bors merge

Abhinandan-Purkait:
bors merge

PR #227, opened 2022-02-17
feat(tracing): add opentelemetry tracing to the grpc interface

tiagolobocastro:
Added a tracing layer for the clients which injects tracing span context in the grpc headers.
Added a tracing layer for the servers which extracts what was injected by the client layer.

Added a new Context::spawn helper which spawns the future with the current spawn context.

Resolves: ER1-275

### DIFF
--- a/control-plane/grpc/src/tracing.rs
+++ b/control-plane/grpc/src/tracing.rs
@@ -1,0 +1,208 @@
+use opentelemetry::{
+    global,
+    trace::{FutureExt, SpanKind, TraceContextExt, Tracer},
+    KeyValue,
+};
+use opentelemetry_http::HeaderInjector;
+use opentelemetry_semantic_conventions::trace::{HTTP_STATUS_CODE, RPC_GRPC_STATUS_CODE};
+use std::{future::Future, pin::Pin};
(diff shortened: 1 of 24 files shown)


Abhinandan-Purkait (approved):

control-plane/grpc/src/tracing.rs
```diff
@@ -0,0 +1,199 @@
...
+        builder.parent_context = parent_context.clone();
+        builder.span_kind = Some(SpanKind::Server);
+
+        let span = tracer.build(builder);
+        let context = parent_context.with_span(span);
+
+        let service_clone = self.service.clone();
+        // we cannot simply use service_clone as self.service is ready but the service_clone may
+        // not be ready. Instead, swap them over.
+        let service = std::mem::replace(&mut self.service, service_clone);
```
paulyoong, line 150: This pattern is used a couple of times. Would it be beneficial to encapsulate it in a function like "clone_from_service"? It may avoid having to remember why we have to do the replace.

paulyoong (approved):

tiagolobocastro:
bors merge

PR #229, opened 2022-02-19
Fix Helm chart.

reitermarkus:
Previously, CRDs were always installed, since there was no conditional.

Also adds missing default values to `values.yaml` and conditionally passed flags for specified durations.

Analogous to https://github.com/openebs/mayastor/pull/1106.


tiagolobocastro:
Hey @reitermarkus. 
Some background information - the helm charts we're not yet "exposing" for direct consumption yet. For historical reasons, we've got a [script](https://github.com/openebs/mayastor-control-plane/blob/develop/scripts/deploy/generate-yamls.sh) on each repo that generates yamls based on the helm chart and some chosen parameters, eg tags and profiles.

chart/release/values.yaml
```diff
@@ -1,2 +0,0 @@
-mayastorCP:
```
tiagolobocastro, line 1: We need  to keep this file as it's used by the mentioned generate script.

chart/values.yaml
```diff
@@ -3,11 +3,15 @@
 # Declare variables to be passed into your templates.
 
 mayastorCP:
-  registry:
+  registry: ''
+  tag: latest
```
tiagolobocastro, line 7: I'm not sure if latest is the correct default here, as latest is the latest release, and the chart in develop might not be compatible with that version, so maybe this should be develop?

chart/values.yaml
```diff
@@ -3,11 +3,15 @@
...
   pullPolicy: Always
   logLevel: info
 
 base:
+  cache_poll_period: ''
```
tiagolobocastro, line 12: So these are actually read from [here](https://github.com/openebs/mayastor-control-plane/blob/develop/chart/constants.yaml) when using the script. By I guess it's fine to leave it as empty by default.

tiagolobocastro (commented):
Thanks @reitermarkus, if we can just restore the release profile yaml everything else seems good.

tiagolobocastro:
We use conventional commits, do you mind changing your commits? eg: fix(helm): this and that

chart/values.yaml
reitermarkus: Yes, I saw that `constants.yaml` gets generated, but `helm` doesn't know about that file. I think leaving it empty will use the same defaults since it will just reused the defaults from the executables themselves.

chart/values.yaml
reitermarkus: `latest` is the latest stable release, right? So I think the default value should be `latest` here. `develop/values.yaml` and `test/values.yaml` already override the `tag`.

reitermarkus:
> the helm charts we're not yet "exposing" for direct consumption yet

I am aware, that's why I'm trying to push it past that “yet” point. 😉

chart/values.yaml
tiagolobocastro: `latest` is the latest released version.
Let's say on develop we add a new cmdline argument - the `latest` will not be aware of this new argument, and install will likely fail. 
So maybe the master branch should usage tag `latest` and the develop branch should use the tag `develop`?

reitermarkus: > So maybe the master branch should usage tag `latest` and the develop branch should use the tag `develop`?

Yes, I guess so. Not sure what the best way is to accomplish that though.

tiagolobocastro: ugly suggestion - we could have a script that checks if the tag is "correct" - we could add it as a pre-commit hook and also to the jenkins pipeline.

reitermarkus: I think actually the master branch should have the exact version and `pullPolicy: IfNotPresent`. So basically what is contained in the generated files in `deploy` should already be present in `chart` as well.

reitermarkus:
@tiagolobocastro, I updated the `generate-yamls.sh` script to also generate the `values.yaml` file in addition to the files in `deploy`.

deploy/msp-deployment.yaml
```diff
@@ -56,3 +56,5 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.name
+      nodeSelector:
```
tiagolobocastro, line 59: Is the msp-operator using anything specify cpu specific?

tiagolobocastro:
thanks @reitermarkus I'll have a play with the chart tomorrow :+1:

tiagolobocastro:
I get an error running the generate-yaml script:
`❯ ./scripts/deploy/generate-yamls.sh develop                                                                                                                                                                                             ─╯
/nix/store/jgj50jymin33k7v1vzsl6njmah91nx1x-python3.9-yq-2.13.0/bin/yq
usage: yq [-h] [--yaml-output] [--yaml-roundtrip] [--width WIDTH]
          [--indentless-lists] [--in-place] [--version]
          [jq_filter] [files ...]
yq: error: argument files: can't open '. as $item ireduce ({}; . * $item )': [Errno 2] No such file or directory: '. as $item ireduce ({}; . * $item )'`

Also, could you please add `yq` to the shell.nix `buildInputs`, and since you're there we're probably missing `jq` there as well, if you don't mind.

reitermarkus:
I added them.

> `/nix/store/jgj50jymin33k7v1vzsl6njmah91nx1x-python3.9-yq-2.13.0/bin/yq`

I guess this (https://pypi.org/project/yq/) `yq` is not the same as https://mikefarah.gitbook.io/yq/.

tiagolobocastro:
> I added them.
> 
> > `/nix/store/jgj50jymin33k7v1vzsl6njmah91nx1x-python3.9-yq-2.13.0/bin/yq`
> 
> I guess this (https://pypi.org/project/yq/) `yq` is not the same as https://mikefarah.gitbook.io/yq/.
> 
> It should be the latter, v4.6.0+.

Ah... I'll have a look at nixpkgs to see if we can get that yq.

shell.nix
```diff
@@ -40,6 +41,7 @@ mkShell {
     utillinux
     which
     libudev
+    yq
```
tiagolobocastro, line 44: found the right one, please replace `yq` with `yq-go`.

tiagolobocastro:
Sorry about the delay @reitermarkus, we've moved the chart to https://github.com/openebs/mayastor-extensions/tree/develop/chart.

PR #231, opened 2022-02-22
refactor(replace NATS): add node proto and registration module

Abhinandan-Purkait:
- Adds the node proto
- Changes the mayastor-api submodule to point to master 
- Adds the registration module


control-plane/grpc/proto/v1/node/node.proto
```diff
@@ -0,0 +1,72 @@
...
+  string endpoint = 2;
+  // Node labels.
+  common.StringMapValue labels = 3;
+}
+
+message NodeState {
+  // id of the mayastor instance
+  string node_id = 1;
+  // grpc_endpoint of the mayastor instance
+  string endpoint = 2;
```
paulyoong, line 32: Do we need the endpoint in the state? We already have it in the spec. Or do we require it in order to update the spec (in case the endpoint changes)?

rpc/src/lib.rs
```diff
@@ -93,3 +93,7 @@ pub mod mayastor {
 pub mod csi {
     include!(concat!(env!("OUT_DIR"), "/csi.v1.rs"));
 }
+
+pub mod registration {
+    include!(concat!(env!("OUT_DIR"), "/v1.registration.rs"));
```
paulyoong, line 98: The CSI path has the version number after the module name. Should we keep them all consistent?

Abhinandan-Purkait: We surely can, I kept it the way its for control-plane. We would need to change first in `mayastor-api` for that.

rpc/build.rs
```diff
@@ -30,4 +30,12 @@ fn main() {
...
         )
         .unwrap_or_else(|e| panic!("CSI protobuf compilation failed: {}", e));
+
+    tonic_build::configure()
+        .build_server(true)
+        .compile(
+            &["mayastor-api/protobuf/v1/registration.proto"],
+            &["mayastor-api/protobuf/v1"],
+        )
+        .unwrap_or_else(|e| panic!("Registration protobuf compilation failed: {}", e));
```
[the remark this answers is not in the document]
Abhinandan-Purkait: No, we are not, but this directly interacts with the mayastor unlike the other services. Shall we add this in the `control-plane/grpc` ?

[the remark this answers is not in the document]
tiagolobocastro: We can leave it here for now then but we probably need to clear this at some point, having 2 *rpc crates is a bit confusing.

rpc/src/lib.rs
paulyoong: That's fine. Leave it as it is for now.

control-plane/grpc/proto/v1/node/node.proto
paulyoong: :+1:

paulyoong (approved):

tiagolobocastro (approved):

Abhinandan-Purkait:
bors merge

Abhinandan-Purkait:
bors merge

PR #236, opened 2022-02-23
feat(nexus info): delete from persistent store

paulyoong:
Ensure the NexusInfo structure is deleted from the persistent store when
a volume is destroyed. Also delete old NexusInfo structures when a
volume is re-published.

Also provide the ability for the control plane to pass the key that
should be use to persist the NexusInfo structure to the data plane. The
key itself has been modified to include the volume and nexus UUID.

------
Note: This PR shouldn't be merged until the corresponding data plane changes have been merged.

### DIFF
--- a/common/src/types/v0/store/nexus_persistence.rs
+++ b/common/src/types/v0/store/nexus_persistence.rs
@@ -1,6 +1,6 @@
 use crate::types::v0::{
-    message_bus::{NexusId, ReplicaId},
-    store::definitions::{ObjectKey, StorableObject, StorableObjectType},
+    message_bus::{NexusId, ReplicaId, VolumeId},
+    store::definitions::{key_namespace, ObjectKey, StorableObject, StorableObjectType},
 };
 use serde::{Deserialize, Serialize};
 use std::fmt::Debug;
@@ -12,6 +12,9 @@ pub struct NexusInfo {
     #[serde(skip)]
     /// uuid of the Nexus
     pub uuid: NexusId,
+    #[serde(skip)]
+    /// uuid of the Volume
+    pub volume_uuid: Option<VolumeId>,
     /// Nexus destroyed successfully.
     pub clean_shutdown: bool,
     /// Information about children.
@@ -43,26 +46,57 @@ pub struct ChildInfo {
 }
 
 /// Key used by the store to uniquely identify a NexusInfo structure.
-pub struct NexusInfoKey(NexusId);
+/// The volume is optional because a nexus can be created which is not associated with a volume.
+pub struct NexusInfoKey {
+    volume_id: Option<VolumeId>,
+    nexus_id: NexusId,
+}
 
-impl From<&NexusId> for NexusInfoKey {
-    fn from(id: &NexusId) -> Self {
-        Self(id.clone())
+impl NexusInfoKey {
+    /// Create a new NexusInfoKey.
+    pub fn new(volume_id: &Option<VolumeId>, nexus_id: &NexusId) -> Self {
+        Self {
+            volume_id: volume_id.clone(),
+            nexus_id: nexus_id.clone(),
+        }
+    }
+
+    /// Get the volume ID.
+    pub fn volume_id(&self) -> &Option<VolumeId> {
+        &self.volume_id
+    }
+
+    /// Get the nexus ID.
+    pub fn nexus_id(&self) -> &NexusId {
+        &self.nexus_id
     }
 }
 
 impl ObjectKey for NexusInfoKey {
     fn key(&self) -> String {
-        // no key prefix (as it's written by mayastor)
-        self.key_uuid()
+        let namespace = key_namespace();
+        let nexus_uuid = self.nexus_id.clone();
+        match &self.volume_id {
+            Some(volume_uuid) => {
+                format!(
+                    "{}/volume/{}/nexus/{}/info",
+                    namespace, volume_uuid, nexus_uuid
+                )
+            }
+            None => {
+                format!("{}/nexus/{}/info", namespace, nexus_uuid)
+            }
+        }
     }
 
     fn key_type(&self) -> StorableObjectType {
-        StorableObjectType::NexusInfo
+        // The key is generated directly from the `key()` function above.
+        unreachable!()
     }
 
     fn key_uuid(&self) -> String {
-        self.0.to_string()
+        // The key is generated directly from the `key()` function above.
+        unreachable!()
     }
 }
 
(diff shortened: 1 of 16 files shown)


tests/bdd/features/volume/nexus-info/feature.feature
```diff
@@ -0,0 +1,25 @@
...
+
+  Scenario: publishing a volume
+    Given a volume that is not published
+    When the volume is published
+    Then the nexus info structure should be present in the persistent store
+
+  Scenario: unpublishing a volume
+    Given a volume that is published
+    When the volume is unpublished
+    Then the nexus info structure should be present in the persistent store
```
Abhinandan-Purkait, line 14: Why do we not delete the nexus info when we unpublish, on re-publish we anyway delete it so can we not delete it at this point itself? Just for my understanding.

Abhinandan-Purkait (approved):

tests/bdd/features/volume/nexus-info/feature.feature
paulyoong: Good question. If we delete it when unpublishing, we lose the information about which children are healthy. Then, when we try to publish again we won't know which children should be used, so we end up adding all of them. This can lead to data inconsistency between children if one of them was previously faulted.

Abhinandan-Purkait: Okay. Thanks for explaining.

common/src/types/v0/store/nexus_persistence.rs
```diff
@@ -43,33 +46,69 @@ pub struct ChildInfo {
...
+pub struct NexusInfoKey {
+    volume_id: Option<VolumeId>,
+    nexus_id: NexusId,
+}
 
-impl From<&NexusId> for NexusInfoKey {
-    fn from(id: &NexusId) -> Self {
-        Self(id.clone())
+impl NexusInfoKey {
+    pub fn new(volume_id: &Option<VolumeId>, nexus_id: &NexusId) -> Self {
```
tiagolobocastro, line 56: nit: doc comment the pub?

common/src/types/v0/store/nexus_persistence.rs
```diff
@@ -43,33 +46,69 @@ pub struct ChildInfo {
...
 impl ObjectKey for NexusInfoKey {
     fn key(&self) -> String {
-        // no key prefix (as it's written by mayastor)
-        self.key_uuid()
+        let namespace = std::env::var("MY_POD_NAMESPACE").unwrap_or_else(|_| "default".into());
+        let nexus_uuid = self.nexus_id.clone();
+        match &self.volume_id {
+            Some(volume_uuid) => {
+                format!(
+                    "/namespace/{}/volume/{}/nexus/{}/info",
```
tiagolobocastro, line 84: can we keep the namespace logic in a common function that is then used by all types?

control-plane/agents/core/src/volume/specs.rs
```diff
@@ -570,6 +579,17 @@ impl ResourceSpecsLocked {
         let nexus =
             SpecOperations::validate_update_step(registry, result, &spec, &spec_clone).await?;
 
+        let (volume_id, last_nexus_id) = {
+            let volume_spec = spec.lock();
+            (volume_spec.uuid.clone(), volume_spec.last_nexus_id.clone())
+        };
+
+        // If there was a previous nexus we should delete the persisted NexusInfo structure.
+        if let Some(nexus_id) = last_nexus_id {
```
tiagolobocastro, line 588: we probably don't want to do this until after the `complete_update`?

control-plane/agents/core/src/volume/specs.rs
```diff
@@ -587,6 +607,19 @@ impl ResourceSpecsLocked {
...
 
+    // Delete the NexusInfo key from the persistent store.
+    // If deletion fails we just log it and continue.
+    async fn delete_nexus_info(key: &NexusInfoKey, registry: &Registry) {
+        match registry.delete_kv(&key.key()).await {
+            Ok(_) => {
+                tracing::info!(nexus_info_key=%key,"Deleted NexusInfo entry from persistent store")
+            }
+            Err(e) => {
+                tracing::error!(nexus_info_key=%key, error=%e, "Failed to delete NexusInfo entry from persistent store")
```
tiagolobocastro, line 618: Ah this is why you've implemented Display for NexusInfoKey?
That doesn't work nicely with tracing, it's probably better to add the individual fields by hand here; or uses the nexus and just emit a nexus.error which will contain the volume and nexus uuid.

control-plane/agents/core/src/volume/specs.rs
```diff
@@ -587,6 +607,19 @@ impl ResourceSpecsLocked {
         Ok(volume)
     }
 
+    // Delete the NexusInfo key from the persistent store.
+    // If deletion fails we just log it and continue.
+    async fn delete_nexus_info(key: &NexusInfoKey, registry: &Registry) {
+        match registry.delete_kv(&key.key()).await {
+            Ok(_) => {
+                tracing::info!(nexus_info_key=%key,"Deleted NexusInfo entry from persistent store")
```
tiagolobocastro, line 615: IIRC We don't usually log successful operations (exception being the reconcilers), so maybe we shouldn't do it here either.

tests/bdd/features/volume/nexus-info/test_feature.py
```diff
@@ -0,0 +1,168 @@
...
+
+@pytest.fixture
+def an_existing_volume():
+    volume = ApiClient.volumes_api().put_volume(
+        VOLUME_UUID, CreateVolumeBody(VolumePolicy(False), 1, VOLUME_SIZE)
+    )
+    yield
+    try:
+        ApiClient.volumes_api().del_volume(volume.spec.uuid)
+    except Exception:
```
tiagolobocastro, line 150: Should this be restricted to a NotFoundException?

common/src/types/v0/store/nexus_persistence.rs
paulyoong: Done.

common/src/types/v0/store/nexus_persistence.rs
paulyoong: Done.

control-plane/agents/core/src/volume/specs.rs
paulyoong: Good point. Done.

control-plane/agents/core/src/volume/specs.rs
paulyoong: :+1: Added them by hand.

control-plane/agents/core/src/volume/specs.rs
paulyoong: Changed this to a trace message as it might be useful to still have it.

tests/bdd/features/volume/nexus-info/test_feature.py
paulyoong: Yes, done.

paulyoong:
bors merge

paulyoong:
bors merge

paulyoong:
bors merge

PR #237, opened 2022-02-23
chore: replace mcp docker image prefix with mayastor

tiagolobocastro:


tiagolobocastro:
@blaisedias I don't know if this will cause troubles for e2e, please let me know and we can coordinate.

GlennBullingham:
I don't think that this works, at least not without changes to external infrastructure.

(for example)
changing 
`mayadata/mcp-core:`
to
`mayastor/mayastor-core:`

Actually changes the registry name and to my knowledge, we have `mayadata` as a registry name on DockerHub but not `mayastor`.  We *do* have `openebs`, so perhaps it is time to make that change, with the next release?

GlennBullingham (approved):

tiagolobocastro:
Replaced by #239

PR #238, opened 2022-02-25
chore: remove iscsi support from control plane

Abhinandan-Purkait:
Removes iscsi support from control-plane and fixes the tests around it.

### DIFF
--- a/common/src/types/v0/message_bus/nexus.rs
+++ b/common/src/types/v0/message_bus/nexus.rs
@@ -121,16 +121,6 @@ impl Default for NexusShareProtocol {
         Self::Nvmf
     }
 }
-impl From<i32> for NexusShareProtocol {
-    fn from(src: i32) -> Self {
-        match src {
-            0 | 1 => Self::Nvmf,
-            2 => Self::Iscsi,
-            _ => panic!("Invalid nexus share protocol {}", src),
-        }
-    }
-}
-
 impl From<NexusShareProtocol> for Protocol {
     fn from(src: NexusShareProtocol) -> Self {
         match src {
--- a/control-plane/agents/common/src/errors.rs
+++ b/control-plane/agents/common/src/errors.rs
@@ -80,6 +80,12 @@ pub enum SvcError {
     VolumeNotFound { vol_id: String },
     #[snafu(display("Volume '{}' not published", vol_id))]
     VolumeNotPublished { vol_id: String },
+    #[snafu(display("{} {} cannot be shared over invalid protocol '{}'", kind.to_string(), id, share))]
+    InvalidShareProtocol {
+        kind: ResourceKind,
+        id: String,
+        share: String,
+    },
     #[snafu(display(
         "Volume '{}' is already published on node '{}' with protocol '{}'",
         vol_id,
@@ -235,6 +241,12 @@ impl From<SvcError> for ReplyError {
             },
             SvcError::AlreadyShared { kind, .. } => ReplyError {
                 kind: ReplyErrorKind::AlreadyShared,
+                resource: kind,
+                source: desc.to_string(),
+                extra: error_str,
+            },
+            SvcError::InvalidShareProtocol { kind, .. } => ReplyError {
+                kind: ReplyErrorKind::InvalidArgument,
                 resource: kind,
                 source: desc.to_string(),
                 extra: error_str,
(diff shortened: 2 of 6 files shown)


Abhinandan-Purkait:
bors try

tiagolobocastro:
@Abhinandan-Purkait I don't think we actually want to remove iscsi from the protobuf, at least that's what what happened for mayastor. Can we just disable iscsi in the core agent?

Abhinandan-Purkait:
> @Abhinandan-Purkait I don't think we actually want to remove iscsi from the protobuf, at least that's what what happened for mayastor. Can we just disable iscsi in the core agent?

Ohh, then in that case we would not be able to change any of the enums which have the `iscsi` in it. All we can do is to fix the text to not use iscsi. Is that what we wanted to do?

tiagolobocastro:
> > @Abhinandan-Purkait I don't think we actually want to remove iscsi from the protobuf, at least that's what what happened for mayastor. Can we just disable iscsi in the core agent?
> 
> Ohh, then in that case we would not be able to change any of the enums which have the `iscsi` in it. All we can do is to fix the text to not use iscsi. Is that what we wanted to do?

We also want to remove iscsi from the core agent. So if you try to use the grpc to publish a volume via iscsi that should be rejected by the core agent.

control-plane/grpc/proto/v1/misc/common.proto
```diff
@@ -29,10 +29,8 @@ enum Protocol {
   None = 0;
   // shared as NVMe-oF TCP
   Nvmf = 1;
-  // shared as iSCSI
-  Iscsi = 2;
   // shared as NBD
-  Nbd = 3;
+  Nbd = 2;
```
akhilerm, line 33: Why do we still keep NBD here, ? IIUC, nbd support was removed.

akhilerm (commented):
@Abhinandan-Purkait Dont we need to remove Iscsi from the mayastor-api proto files also?

control-plane/grpc/proto/v1/misc/common.proto
```diff
@@ -29,10 +29,8 @@ enum Protocol {
   None = 0;
   // shared as NVMe-oF TCP
   Nvmf = 1;
-  // shared as iSCSI
-  Iscsi = 2;
```
jonathan-teh, line 33: This needs to remain for backward compatibility as 1.0 already shipped with this present.

control-plane/grpc/proto/v1/misc/common.proto
Abhinandan-Purkait: No @akhilerm we would not be removing from the protobufs, we will only add validations for now, we will reject anything other that `nvmf`. Please ignore the above changes, I will push the new changes in sometime.

jonathan-teh: This can be removed, it's not in the previous `mayastor.proto` file.

common/src/types/v0/message_bus/nexus.rs
```diff
@@ -124,8 +124,8 @@ impl Default for NexusShareProtocol {
 impl From<i32> for NexusShareProtocol {
     fn from(src: i32) -> Self {
         match src {
-            0 | 1 => Self::Nvmf,
-            2 => Self::Iscsi,
+            0 => Self::Nvmf,
+            1 => Self::Iscsi,
```
jonathan-teh, line 128: This no longer matches the protobuf file.

Abhinandan-Purkait: Without this the volume protobuf in the control-plane is having issues. Need to figure out a way,

jonathan-teh: What are the issues?

Abhinandan-Purkait: https://github.com/openebs/mayastor-control-plane/blob/e061a7ce68705113541bbf4cecb1b6b9352db0e9/common/src/types/v0/message_bus/nexus.rs#L107

Here we map in this way, we don't have a none equivalent.

Abhinandan-Purkait: https://github.com/openebs/mayastor-control-plane/blob/e061a7ce68705113541bbf4cecb1b6b9352db0e9/common/src/types/v0/message_bus/volume.rs#L119

Abhinandan-Purkait: Hi, seems like we weren't using that. I removed it. Its not going to affect the same.

tiagolobocastro (approved):

jonathan-teh (approved):

Abhinandan-Purkait:
bors merge

PR #240, opened 2022-02-25
fix(pool spec dirty): check for pending op

paulyoong:
It is possible for the pool spec to be in a "dirty" state so we should
ensure that we correctly check for this.

An issue relating to this was seen in production. It occurred when the
connection to etcd was lost in the middle of a pool creation operation.
As a consequence the pool could not subsequently be created. Manual
intervention was required to rectify this.


tiagolobocastro (approved):

paulyoong:
bors merge

PR #242, opened 2022-02-27
fix: for aio:/// devices, check that the file exists

neoaggelos:
### Summary

References https://github.com/openebs/mayastor/issues/1089#issuecomment-1042338434

When adding mayastorpool resources with `aio:///` devices, make sure that the normalized file exists on the disk instead of assuming that it has to be a block device.

Also, extend the `normalize_disk` function to strip any parameters from the filename.

My knowledge in Rust is quite limited, feel free to suggest how I can improve this PR.

### Changes

In `create_or_import`, if the device starts with `aio:///`, then do not poll the system block devices, but rather check that the device is a valid path instead.

### Testing

Tested in my own development setup, I have pushed the image as `neoaggelos/msp-operator:dev7`

Example with pool when the file exists:

```
Spec:
  Disks:
    aio:///data/a.img?blk_size=4096
  Node:  magnus
Status:
  Available:  5360320512
  Capacity:   5360320512
  State:      Online
  Used:       0
Events:
  Type    Reason   Age   From          Message
  ----    ------   ----  ----          -------
  Normal  Created  2s    msp-operator  Created or imported pool
  Normal  Online   2s    msp-operator  Pool online and ready to roll!
```

Example with pool when the file does not exist:

```
Spec:
  Disks:
    aio:///data/disk.img1?blk_size=4096
  Node:  magnus
Status:
  Available:  0
  Capacity:   0
  State:      Error
  Used:       0
Events:
  Type  Reason   Age   From          Message
  ----  ------   ----  ----          -------
  Warn  Missing  35s   msp-operator  No block device(s): aio:///data/disk.img1?blk_size=4096 can not be found
```

### DIFF
--- a/control-plane/msp-operator/src/main.rs
+++ b/control-plane/msp-operator/src/main.rs
@@ -429,93 +429,101 @@ impl ResourceContext {
             // we updated the resource as an error stop reconciliation
             return Err(Error::ReconcileError { name: self.name() });
         }
-        match self
+        let check_if_block_device = if self.spec.disks[0].starts_with("aio:///") {
+            let path = normalize_disk(&self.spec.disks[0][6..]);
+            !std::path::Path::new(&path).exists()
+        } else {
+            false
+        };
+        if check_if_block_device {
+            match self
             .block_devices_api()
             .get_node_block_devices(&self.spec.node, Some(true))
             .await
-        {
-            Ok(response) => {
-                if !response.into_body().into_iter().any(|b| {
-                    b.devname == normalize_disk(&self.spec.disks[0])
-                        || b.devlinks
-                            .iter()
-                            .any(|d| *d == normalize_disk(&self.spec.disks[0]))
-                }) {
-                    self.k8s_notify(
-                        "Create or import",
-                        "Missing",
-                        &format!(
-                            "The block device(s): {} can not be found",
-                            &self.spec.disks[0]
-                        ),
-                        "Warn",
-                    )
-                    .await;
-
-                    return Err(Error::SpecError {
-                        value: self.spec.disks[0].clone(),
-                        timeout: u32::pow(2, self.num_retries),
-                    });
-                }
-
-                let mut labels: HashMap<String, String> = HashMap::new();
-                labels.insert(
-                    String::from(utils::OPENEBS_CREATED_BY_KEY),
-                    String::from(utils::MSP_OPERATOR),
-                );
-
-                let body = CreatePoolBody::new_all(self.spec.disks.clone(), labels);
-                match self
-                    .pools_api()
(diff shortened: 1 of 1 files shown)


control-plane/msp-operator/src/main.rs
```diff
@@ -1060,13 +1060,18 @@ async fn main() -> anyhow::Result<()> {
...
+    let s = Url::parse(disk).map_or(disk.to_string(), |u| {
         u.to_file_path()
             .unwrap_or_else(|_| disk.into())
             .as_path()
             .display()
             .to_string()
-    })
+    });
+
+    match s.match_indices('?').next() {
```
gila, line 1071: The URL crate provides a nice interface to the whole URL; perhaps it is better/easier to use those?

control-plane/msp-operator/src/main.rs
```diff
@@ -435,93 +435,101 @@ impl ResourceContext {
             // we updated the resource as an error stop reconciliation
             return Err(Error::ReconcileError { name: self.name() });
         }
-        match self
+        let check_if_block_device = if self.spec.disks[0].starts_with("aio:///") {
+            let path = normalize_disk(&self.spec.disks[0][6..]);
+            !std::path::Path::new(&path).exists()
```
gila, line 440: It looks like you are checking if the file exists within the context of the MSP operator. Dont want to check that it's available on the target node?

control-plane/msp-operator/src/main.rs
```diff
@@ -435,93 +435,101 @@ impl ResourceContext {
...
+        } else {
+            false
+        };
+        if check_if_block_device {
```
gila, line 444: If you dont check if the block device exists (for whatever reason) then the only way to ever understand what's wrong here is to look in the logs of the node which was selected and read its error logs to understand what was wrong. IMHO the RPC methods in the mayastor node itself need to be extended to report files using AIO.

control-plane/msp-operator/src/main.rs
neoaggelos: Ah, yes, you are in fact correct.

control-plane/msp-operator/src/main.rs
neoaggelos: I understand that this makes sense, but it is a non-trivial change, and I highly doubt that my Rust skills are appropriate to take this on. :(

Would a different approach be acceptable, by which I mean a configuration option to disable validation on the side of `msp-operator`?

tiagolobocastro: Yeah as an interim solution I think an option to disable validation, would be appropriate, as long as by default we do validate.

PR #244, opened 2022-03-01
fix: delete resources pending deletion when the store is online

tiagolobocastro:
When the persistent store (etcd) is not online, we may not be able to delete resources. They remain
in the deleting state.
Mimic the volume dirty spec reconciler for other resources and simply delete them if the creation
was not successful.

Destroy resources in the deleting state from their garbage collection reconciler.

Resolves: CAS-1300

### DIFF
--- a/control-plane/agents/core/src/core/reconciler/replica/mod.rs
+++ b/control-plane/agents/core/src/core/reconciler/replica/mod.rs
@@ -2,13 +2,14 @@
 mod tests;
 
 use crate::core::{
-    specs::{OperationSequenceGuard, SpecOperations},
+    specs::{OperationSequenceGuard, ResourceSpecsLocked, SpecOperations},
     task_poller::{
         PollContext, PollEvent, PollResult, PollTimer, PollTriggerEvent, PollerState, TaskPoller,
     },
 };
-use common_lib::types::v0::{message_bus::ReplicaOwners, store::OperationMode};
-use std::ops::Deref;
+use common_lib::types::v0::store::{replica::ReplicaSpec, OperationMode};
+use parking_lot::Mutex;
+use std::sync::Arc;
 
 /// Replica reconciler
 #[derive(Debug)]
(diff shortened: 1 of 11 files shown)


control-plane/agents/core/src/pool/specs.rs
```diff
@@ -414,69 +413,104 @@ impl ResourceSpecsLocked {
...
+                        }
+                        ReplicaSpecStatus::Created(_) => {
+                            if replica_spec.lock().pending_op() {
+                                (replica_spec.lock().clone(), guard)
+                            } else {
+                                continue;
+                            }
+                        }
+                        ReplicaSpecStatus::Deleting => {
+                            // handled by the garbage collector
```
paulyoong, line 466: This match pattern seems to be the same for multiple types of specs (i.e. nexus and pool). I wonder if we can abstract it to a common function so that the behaviour is generic across types.

control-plane/agents/core/src/volume/specs.rs
```diff
@@ -587,10 +587,18 @@ impl ResourceSpecsLocked {
...
         if let Some(share) = request.share {
-            result = self
+            result = match self
                 .share_nexus(registry, &ShareNexus::from((&nexus, None, share)), mode)
                 .await
-                .map(|_| nexus);
+            {
+                Ok(_) => Ok(nexus),
+                Err(error) => {
+                    self.destroy_nexus(registry, &DestroyNexus::from(nexus), true, mode)
```
paulyoong, line 596: A comment just indicating why we want to destroy the nexus here would be nice.

control-plane/agents/core/src/volume/tests.rs
```diff
@@ -129,9 +129,88 @@ async fn garbage_collection() {
 
     unused_nexus_reconcile(&cluster).await;
     unused_reconcile(&cluster).await;
+    deleting_volume_reconcile(&cluster).await;
     offline_replicas_reconcile(&cluster, reconcile_period).await;
 }
 
+async fn deleting_volume_reconcile(cluster: &Cluster) {
+    let volumer = cluster.grpc_client().volume();
```
paulyoong, line 137: Nit: `volumer`?

paulyoong (approved):

control-plane/agents/core/src/core/reconciler/replica/mod.rs
```diff
@@ -49,45 +56,45 @@ impl TaskPoller for ReplicaReconciler {
 /// Remove replica owners who no longer exist.
 /// In the event that the replicas become orphaned (have no owners) they will be destroyed by the
 /// 'destroy_orphaned_replicas' reconcile loop.
-async fn disown_missing_owners(context: &PollContext) -> PollResult {
+async fn disown_missing_owners(
```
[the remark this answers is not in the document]
tiagolobocastro: Yeah good point :+1:

control-plane/agents/core/src/pool/specs.rs
tiagolobocastro: Yeah, I think we can make the whole process of reconciling dirty specs generic.

tiagolobocastro:
bors merge

PR #248, opened 2022-03-08
fix(csi): use k8s api server to get node name

mtzaurus:
Node plugin now retrieves hostname from node CRD via Kubernetes API
server, using kubernetes.io/hostname label.

Resolves: CAS-1095

### DIFF
--- a/control-plane/csi-driver/node/src/main.rs
+++ b/control-plane/csi-driver/node/src/main.rs
@@ -15,8 +15,11 @@ use crate::{identity::Identity, mount::probe_filesystems, node::Node, shutdown_e
 use clap::{App, Arg};
 use csi::{identity_server::IdentityServer, node_server::NodeServer};
 use futures::TryFutureExt;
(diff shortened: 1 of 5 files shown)


control-plane/csi-driver/node/src/main.rs
```diff
@@ -169,7 +195,7 @@ async fn main() -> Result<(), String> {
         )
         .get_matches();
 
-    let node_name = normalize_hostname(matches.value_of("node-name").unwrap());
+    let node_name = get_nodename(matches.value_of("node-name").unwrap()).await;
```
tiagolobocastro, line 198: We should only use get_nodename if we're running on k8s?

mtzaurus:
bors try

mtzaurus:
bors try

mtzaurus:
bors try

mtzaurus:
bors try

mtzaurus:
bors merge

PR #249, opened 2022-03-10
fix: fix run on arm64

xin3liang:
Mayastor can run on arm64 now, not just x86_64.
See verfied result here:
https://linaro.atlassian.net/browse/STOR-35

Signed-off-by: Xinliang Liu <xinliang.liu@linaro.org>

This PR is related to Mayastor arm64 support: https://github.com/openebs/mayastor/issues/1063

### DIFF
--- a/chart/templates/csi-node-daemonset.yaml
+++ b/chart/templates/csi-node-daemonset.yaml
@@ -21,8 +21,6 @@ spec:
     spec:
       serviceAccount: mayastor-service-account
       hostNetwork: true
-      nodeSelector:
-        kubernetes.io/arch: amd64
       # NOTE: Each container must have mem/cpu limits defined in order to
       # belong to Guaranteed QoS class, hence can never get evicted in case of
       # pressure unless they exceed those limits. limits and requests must be
--- a/deploy/csi-node-daemonset.yaml
+++ b/deploy/csi-node-daemonset.yaml
@@ -23,8 +23,6 @@ spec:
     spec:
       serviceAccount: mayastor-service-account
       hostNetwork: true
-      nodeSelector:
-        kubernetes.io/arch: amd64
       # NOTE: Each container must have mem/cpu limits defined in order to
       # belong to Guaranteed QoS class, hence can never get evicted in case of
       # pressure unless they exceed those limits. limits and requests must be


tiagolobocastro:
Hi @xin3liang, the images we push are x86 only so the default should be to keep the x86 nodeSelector.
Maybe we need an option on the charts that allows one to specify ARM on the node selector and or remove the nodeSelector completely?

xin3liang:
> Hi @xin3liang, the images we push are x86 only so the default should be to keep the x86 nodeSelector. Maybe we need an option on the charts that allows one to specify ARM on the node selector and or remove the nodeSelector completely?

Hi @tiagolobocastro, IMO I think we can remove nodeSelector completely. This make deployment yaml files consistent for both x86 and other CPU arches, especially if the images are multi-arch, so we will need to remove the CPU arch nodeSelector sooner or later. Now the user only need to change to the right images to run on other CPU arches. If the user run x86 images on wrong CPU arches, an exec format error will show the images are wrong.

xin3liang:
> > Hi @xin3liang, the images we push are x86 only so the default should be to keep the x86 nodeSelector. Maybe we need an option on the charts that allows one to specify ARM on the node selector and or remove the nodeSelector completely?
> 
> Hi @tiagolobocastro, IMO I think we can remove nodeSelector completely. This make deployment yaml files consistent for both x86 and other CPU arches, especially if the images are multi-arch, so we will need to remove the CPU arch nodeSelector sooner or later. Now the user only need to change to the right images to run on other CPU arches. If the user run x86 images on wrong CPU arches, an exec format error will show the images are wrong.

Looking at the image field "image: {{ .Values.mayastorCP.registry }}mayadata/mayastor-csi-node:{{ .Values.mayastorCP.tag }}", user only need to set to their registry to change to the wanted image. This is configurable already.

xin3liang:
Ping

tiagolobocastro:
Could you please raise this against the chart from https://github.com/openebs/mayastor-extensions ?

xin3liang:
> Could you please raise this against the chart from https://github.com/openebs/mayastor-extensions ?

Sure, current dataplane change a lot. I will verify it on arm again soon.

xin3liang:
> This PR is related to Mayastor arm64 support: [openebs/mayastor#1063](https://github.com/openebs/mayastor/issues/1063)

PR sent: https://github.com/openebs/mayastor-extensions/pull/3

PR #252, opened 2022-03-30
fix(delay): only lock nodes list when necessary

paulyoong:
Ensure that we don't hold on to the nodes list lock while waiting on
nodes.

While fetching node information we now don't hold the nodes list lock as they may prevent
other writers or readers (if a write is waiting) from proceeding which may cause GET delays.
Also, don't hold on to the read lock while fetching node information since we already have
the grpc lock for exclusivity.

Note: This change is to fix the occasional timeouts that we are seeing on CI.

### DIFF
--- a/control-plane/agents/core/src/core/registry.rs
+++ b/control-plane/agents/core/src/core/registry.rs
@@ -261,7 +261,9 @@ impl Registry {
     async fn poller(&self) {
         loop {
             {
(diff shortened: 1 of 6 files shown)


tiagolobocastro (approved):

paulyoong:
bors merge

paulyoong:
bors merge

paulyoong:
bors merge

PR #254, opened 2022-05-24
feat: add --request-timeout flag to MSP deployment

croomes:
Adds a `--request-timeout` flag to the MSP operator, and sets the `default_req_timeout` Helm value (also used by core and rest) to `5s`.  The default could probably be increased.

Fixes https://github.com/openebs/mayastor/issues/1138

### DIFF
--- a/control-plane/msp-operator/src/main.rs
+++ b/control-plane/msp-operator/src/main.rs
@@ -888,8 +888,16 @@ async fn pool_controller(args: ArgMatches<'_>) -> anyhow::Result<()> {
     let msp: Api<MayastorPool> = Api::namespaced(k8s.clone(), namespace);
     let lp = ListParams::default();
     let url = Url::parse(args.value_of("endpoint").unwrap()).expect("endpoint is not a valid URL");
-    let cfg = clients::tower::Configuration::new(url, Duration::from_secs(5), None, None, true)
-        .map_err(|error| {
+
+    let timeout: Duration = args
+        .value_of("request-timeout")
+        .unwrap()
+        .parse::<humantime::Duration>()
+        .expect("timeout value is invalid")
+        .into();
+
+    let cfg =
+        clients::tower::Configuration::new(url, timeout, None, None, true).map_err(|error| {
             anyhow::anyhow!(
                 "Failed to create openapi configuration, Error: '{:?}'",
                 error
@@ -965,6 +973,14 @@ async fn main() -> anyhow::Result<()> {
                 .env("INTERVAL")
                 .default_value(utils::CACHE_POLL_PERIOD)
                 .help("specify timer based reconciliation loop"),
+        )
+        .arg(
+            Arg::with_name("request-timeout")
+                .short("t")
+                .long("request-timeout")
+                .env("REQUEST_TIMEOUT")
+                .default_value(utils::DEFAULT_REQ_TIMEOUT)
+                .help("the timeout for remote requests"),
         )
         .arg(
             Arg::with_name("retries")
--- a/deploy/core-agents-deployment.yaml
+++ b/deploy/core-agents-deployment.yaml
@@ -23,8 +23,7 @@ spec:
         - command:
           - sh
           - -c
-          - trap "exit 1" TERM; until nc -vz mayastor-etcd 2379; do echo "Waiting for etcd...";
-            sleep 1; done;
+          - trap "exit 1" TERM; until nc -vz mayastor-etcd 2379; do echo "Waiting for etcd..."; sleep 1; done;
           image: busybox:latest
           name: etcd-probe
       containers:
--- a/chart/templates/msp-deployment.yaml
+++ b/chart/templates/msp-deployment.yaml
@@ -27,6 +27,7 @@ spec:
           imagePullPolicy: {{ .Values.mayastorCP.pullPolicy }}
           args:
             - "-e http://rest:8081"
+            - "--request-timeout={{ .Values.base.default_req_timeout }}"
             - "--interval={{ .Values.base.cache_poll_period }}"{{ if .Values.base.jaeger.enabled }}
             - "--jaeger={{ .Values.base.jaeger.agent.name }}:{{ .Values.base.jaeger.agent.port }}"{{ end }}
           env:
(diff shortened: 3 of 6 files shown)


deploy/core-agents-deployment.yaml
```diff
@@ -23,8 +23,7 @@ spec:
         - command:
           - sh
           - -c
-          - trap "exit 1" TERM; until nc -vz mayastor-etcd 2379; do echo "Waiting for etcd...";
-            sleep 1; done;
+          - trap "exit 1" TERM; until nc -vz mayastor-etcd 2379; do echo "Waiting for etcd..."; sleep 1; done;
```
croomes, line 26: I'm not sure why these lines changed - I didn't touch the helper that generates this?

control-plane/msp-operator/src/main.rs
```diff
@@ -888,7 +888,15 @@ async fn pool_controller(args: ArgMatches<'_>) -> anyhow::Result<()> {
     let msp: Api<MayastorPool> = Api::namespaced(k8s.clone(), namespace);
     let lp = ListParams::default();
     let url = Url::parse(args.value_of("endpoint").unwrap()).expect("endpoint is not a valid URL");
-    let cfg = clients::tower::Configuration::new(url, Duration::from_secs(5), None, None, true)
+
+    let timeout: Duration = args
+        .value_of("request-timeout")
+        .unwrap()
```
tiagolobocastro, line 894: can we unwrap_or() use this https://github.com/DataCoreSoftware/bolt-control-plane/blob/2f8b0fccbfac994747fec65a81c04dd45790e9f0/utils/utils-lib/src/constants.rs#L4 ?

tiagolobocastro:
bors try

control-plane/msp-operator/src/main.rs
croomes: Actually, I missed the bit where it's configuring allowed args: https://github.com/openebs/mayastor-control-plane/blob/develop/control-plane/msp-operator/src/main.rs#L954-L1008

I think it would be better to put this back to an `.unwrap()` here, and add:
```
        .arg(
            Arg::with_name("request-timeout")
                .short("t")
                .long("request-timeout")
                .env("REQUEST_TIMEOUT")
                .default_value(utils::DEFAULT_REQ_TIMEOUT)
                .help("the timeout for remote requests"),
        )
```
WDYT?

tiagolobocastro:
@croomes could you please run the linter and also squash your commits?
Please run the linter from the nix-shell or make sure you've got the same versions installed.

tiagolobocastro:
Thanks @croomes 
Btw what do you mean by "wasn't using the unstable package"?
bors try

croomes:
It looks like the nix config is trying to set features from the unstable packages (but I'm new to nix, so not sure):

```
$ nix-shell 
pre-commit installed at .git/hooks/pre-commit
pre-commit installed at .git/hooks/commit-msg

[nix-shell:~/src/mayastor-control-plane]$ cargo fmt --all -- --check
Warning: can't set `wrap_comments = true`, unstable features are only available in nightly channel.
Warning: can't set `comment_width = 100`, unstable features are only available in nightly channel.
Warning: can't set `struct_lit_single_line = true`, unstable features are only available in nightly channel.
Warning: can't set `imports_layout = Mixed`, unstable features are only available in nightly channel.
Warning: can't set `imports_granularity = Crate`, unstable features are only available in nightly channel.
Warning: can't set `spaces_around_ranges = true`, unstable features are only available in nightly channel.
Warning: can't set `ignore = IgnoreList { path_set: {"utils/mayastor-dependencies", "openapi"}, rustfmt_toml_path: "" }`, unstable features are only available in nightly channel.
Diff in /home/simon/src/mayastor-control-plane/common/src/mbus_api/mod.rs at line 195:
     fn from_str(source: &str) -> Result<Self, Self::Err> {
         match source.split('/').next() {
             Some(VERSION) => {
-                let id: MessageIdVs = source[VERSION.len() + 1 ..].parse()?;
+                let id: MessageIdVs = source[VERSION.len() + 1..].parse()?;
                 Ok(Self::v0(id))
             }
```

```
[nix-shell:~/src/mayastor-control-plane]$ cargo clippy --all-targets -- -D warnings
    Checking utils v0.1.0 (/home/simon/src/mayastor-control-plane/utils/utils-lib)
    Checking common-lib v1.0.0 (/home/simon/src/mayastor-control-plane/common)
    Checking msp-operator v1.0.0 (/home/simon/src/mayastor-control-plane/control-plane/msp-operator)
    Checking kubectl-plugin v1.0.0 (/home/simon/src/mayastor-control-plane/kubectl-plugin)
error: methods called `is_*` usually take `self` by reference or no `self`
   --> common/src/store/etcd_keep_alive.rs:204:26
    |
204 |     async fn is_replaced(&mut self) -> Option<bool> {
    |                          ^^^^^^^^^
    |
    = note: `-D clippy::wrong-self-convention` implied by `-D warnings`
    = help: consider choosing a less ambiguous name
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#wrong_self_convention

error: could not compile `common-lib` due to previous error
warning: build failed, waiting for other jobs to finish...
error: could not compile `common-lib` due to previous error
error: could not compile `common-lib` due to previous error
```

Not sure if this is right:
```
[nix-shell:~/src/mayastor-control-plane]$ nix-channel --list
nixos https://nixos.org/channels/nixos-unstable
nixpkgs https://nixos.org/channels/nixpkgs-unstable
```
Or if I need to do anything to switch to them.

I wasn't able to use `nix develop` either - I'm not sure if a target is defined:
```
$ nix develop -f . mayastor
error: attribute 'mayastor' in selection path 'mayastor' not found
```

tiagolobocastro:
We're not using flakes so the nix develop won't work.
We shouldn't be using the channels, but the pinned from https://github.com/openebs/mayastor-control-plane/blob/1ccffced79a1253e9dd75c47a608fec5d1d12542/nix/sources.json#L35, or so I thought.
Also, it might be that you're using your own cargo (if you've got it installed on your box) rather than the cargo provided by the nix-shell.
bors merge

PR #257, opened 2022-06-01
chore(terraform): public ssh key

paulyoong:
Update the public ssh key variable to take the file path rather than the
content of the file. This now behaves in the same way as the private ssh
key variable.


tiagolobocastro (approved):

paulyoong:
bors merge

PR #258, opened 2022-06-06
Terraform - Libvirt network fixes

tiagolobocastro:
fix(terraform): default ssh_user to current user (if not specified)
    
    Also, make terraform shell use the nixpkgs from the root shell.

---

chore(terraform): add network mode which defaults to NAT
    
    Creates a named (k8snet) network which is set to autostart when the host boots up.

---

fix(rust-src): compare rust sources recursively
    
    This is a WRITEABLE clone of the rust sources, copied from the nix output derivation.
    Useful when using rust-analyzer from JetBrains which wants a writeable rust source.

### DIFF
--- a/terraform/cluster/mod/libvirt/main.tf
+++ b/terraform/cluster/mod/libvirt/main.tf
@@ -11,6 +11,8 @@ variable "private_key_path" {}
 variable "disk_size" {}
 variable "pooldisk_size" {}
 variable "qcow2_image" {}
+variable "network_mode" {}
+variable "bridge_name" {}
 
 provider "libvirt" {
   uri = "qemu:///system"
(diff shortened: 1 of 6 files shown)


terraform/cluster/mod/libvirt/main.tf
```diff
@@ -64,6 +66,52 @@ data "template_file" "network_config" {
...
+  # the name used by libvirt
+  name = "k8snet"
+
+  # mode can be: "nat" (default), "none", "route", "bridge"
+  mode = var.network_mode
+
+  #  the domain used by the DNS server in this network
+  domain = "k8s.local"
+
+  #  list of subnets the addresses allowed for domains connected
```
paulyoong, line 81: Nit: additional space at beginning of line

paulyoong (approved):

PR #260, opened 2022-06-13
feat(replace NATS): enable jsongrpc ops over grpc as seperate service

Abhinandan-Purkait:
- JsonGrpc calls are now be initiated from rest with grpc as a transport down the flow.


Abhinandan-Purkait:
bors try

control-plane/grpc/src/operations/jsongrpc/client.rs
```diff
@@ -0,0 +1,84 @@
...
+        };
+        for attempt in 1 ..= timeout_opts.max_retries().unwrap_or_default() {
+            match self
+                .probe(Some(Context::new(Some(timeout_opts.clone()))))
+                .await
+            {
+                Ok(true) => return Ok(()),
+                _ => {
+                    let delay = std::time::Duration::from_millis(100);
+                    tracing::trace!(%attempt, delay=?delay, "Not available, retrying after...");
```
tiagolobocastro, line 42: this trace message is incorrect in the last iteration, as it won't be tried after..

control-plane/rest/service/src/main.rs
```diff
@@ -33,6 +33,10 @@ pub(crate) struct CliArgs {
     #[structopt(long, short = "z", default_value = DEFAULT_GRPC_CLIENT_ADDR)]
     core_grpc: Uri,
 
+    /// The json gRPC Server URL or address to connect to the service.
+    #[structopt(long, short = "J", default_value = DEFAULT_JSON_GRPC_CLIENT_ADDR)]
+    json_grpc: Uri,
```
tiagolobocastro, line 38: This should be optional, we don't deploy json  grpc in prod by default

control-plane/rest/service/src/v0/mod.rs
```diff
@@ -46,6 +48,13 @@ pub(crate) fn core_grpc<'a>() -> &'a CoreClient {
         .expect("gRPC Core Client should have been initialised")
 }
 
+/// Get Json gRPC Client
+pub(crate) fn json_grpc<'a>() -> &'a JsonGrpcClient {
+    JSON_GRPC_CLIENT
+        .get()
+        .expect("josn grpc Client should have been initialised")
```
tiagolobocastro, line 55: josn/json

control-plane/rest/service/src/v0/mod.rs
```diff
@@ -46,6 +48,13 @@ pub(crate) fn core_grpc<'a>() -> &'a CoreClient {
         .expect("gRPC Core Client should have been initialised")
 }
 
+/// Get Json gRPC Client
+pub(crate) fn json_grpc<'a>() -> &'a JsonGrpcClient {
```
tiagolobocastro, line 52: should return an option in case the server is not configured with the json grpc agent?

control-plane/grpc/src/operations/jsongrpc/client.rs
Abhinandan-Purkait: Right. Would skip it in the last iteration.

control-plane/rest/service/src/main.rs
Abhinandan-Purkait: Done. Made it an option.

control-plane/rest/service/src/v0/mod.rs
Abhinandan-Purkait: Yes. If a call is being made without the configuration, returning the below.
```
{
  "details": "",
  "message": "JsonGrpc Service is not configured/running.",
  "kind": "Unavailable"
}
```

control-plane/rest/service/src/v0/jsongrpc.rs
```diff
@@ -22,12 +22,29 @@ impl apis::actix_server::JsonGrpc for RestApi {
         Path((node, method)): Path<(String, String)>,
         Body(body): Body<Value>,
     ) -> Result<Value, RestError<RestJsonError>> {
-        let result = MessageBus::json_grpc_call(JsonGrpcRequest {
-            node: node.into(),
-            method: method.into(),
-            params: body.to_string().into(),
-        })
-        .await?;
+        let client = match json_grpc() {
```
tiagolobocastro, line 25: hmm, what about json_grpc returning a Result instead (with this error in case of not configured), and so here we'd just to:
```rust
json_grpc()?.call(...)...
```

common/src/mbus_api/mod.rs
```diff
@@ -381,6 +381,18 @@ impl From<JoinError> for ReplyError {
     }
 }
 
+/// for errors that can occur when serializing or deserializing JSON data
+impl From<serde_json::Error> for ReplyError {
```
tiagolobocastro, line 385: We've hardcoded Unknown which works for JsonGrpc, but might not for other cases, where the error might pertain to something specific. Also, is Aborted the right error kind here?
Maybe a helper method would be more appropriate?

tiagolobocastro (approved):

Abhinandan-Purkait:
bors merge

PR #261, opened 2022-06-13
fix(nodes): don't hold nodes list lock while updating 1 node

tiagolobocastro:
fix(nodes): don't hold nodes list lock while updating 1 node

This way registrations from other nodes don't get helf up by 1 flaky node

---

fix(pytest): check against lowercase state

Otherwise we'll always think the node is not running, which is wrong.
Also, make sure the node is offline in the control-plane before restarting it.


tiagolobocastro:
bors merge

PR #263, opened 2022-06-30
test(BDD): add inital BDD for node cordoning

paulyoong:
Describe the initial behaviours of the node cordoning feature.

### DIFF
--- a/tests/bdd/features/cordon/node/feature.feature
+++ b/tests/bdd/features/cordon/node/feature.feature
@@ -1,0 +1,47 @@
+Feature: Cordoning
+
+  Background:
+    Given Multiple uncordoned nodes
+
+  Scenario: Cordoning a node
+    Given an uncordoned node
+    When the node is cordoned
+    Then new resources of any type cannot be scheduled on the cordoned node
+
+  Scenario: Cordoning a cordoned node
+    Given a cordoned node
+    When the cordoned node is cordoned
+    Then the cordoned node should remain cordoned
+    And new resources of any type cannot be scheduled on the cordoned node
+
+  Scenario: Cordoning a node with existing resources
+    Given an uncordoned node with resources
+    When the node is cordoned
+    Then new resources of any type cannot be scheduled on the cordoned node
+    And existing resources remain unaffected
+
+  Scenario: Uncordoning a node
+    Given a cordoned node
+    When the node is uncordoned
+    Then new resources of any type can be scheduled on the uncordoned node
+
+  Scenario: Uncordoning an uncordoned node
+    Given an uncordoned node
+    When the uncordoned node is uncordoned
+    Then the uncordoned node should remain uncordoned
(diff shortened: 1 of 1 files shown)


tests/bdd/features/cordon/node/feature.feature
```diff
@@ -0,0 +1,48 @@
...
+    Given an uncordoned node
+    When the uncordoned node is uncordoned
+    Then the uncordoned node should remain uncordoned
+    And new resources of any type can be scheduled on the uncordoned node
+
+  Scenario: Uncordoning a node with existing resources
+    Given a cordoned node with resources
+    When the node is uncordoned
+    Then new resources of any type can be scheduled on the uncordoned node
+    And existing resources remain unaffected
```
tiagolobocastro, line 38: Do we need this use case, what does unaffected mean here?

tiagolobocastro (approved):

tests/bdd/features/cordon/node/feature.feature
```diff
@@ -0,0 +1,48 @@
...
+
+  Scenario: Deleting resources on a cordoned node
+    Given a cordoned node with resources
+    When the control plane deletes a resource on a cordoned node
```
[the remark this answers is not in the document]
GlennBullingham: @paulyoong Should this system level BDD go into Bolt-Specifications, rather than directly into the control plane, which I thought we were reserving for pytests?

tests/bdd/features/cordon/node/feature.feature
paulyoong: I was trying to highlight that resources that are on the cordoned node are unaffected by an uncordon operation. Do you think this is unnecessary?

tests/bdd/features/cordon/node/feature.feature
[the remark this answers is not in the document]
paulyoong: > nit: should this be when the control plane "attempts to delete"?

Done.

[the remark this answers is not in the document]
paulyoong: > @paulyoong Should this system level BDD go into Bolt-Specifications, rather than directly into the control plane, which I thought we were reserving for pytests?

@GlennBullingham I think this is where it becomes tricky. These scenarios can be tested by the control plane outside of the system level tests. However, they also hold true for the system level tests also. My thinking is that they are appropriate in both sets of tests, albeit the success criteria may be slightly different. However this is implementation specific.

tests/bdd/features/cordon/node/feature.feature
tiagolobocastro: Not sure, just seems a bit vague - eg: I wouldn't know how to write the fixture for this clause.

paulyoong: Fair enough - I've remove it.

paulyoong:
bors merge

PR #264, opened 2022-06-30
test(BDD): add initial BDD for node drain

paulyoong:
Describe the initial behaviours of the node drain feature.

### DIFF
--- a/tests/bdd/features/drain/node/feature.feature
+++ b/tests/bdd/features/drain/node/feature.feature
@@ -1,0 +1,36 @@
+Feature: Drain
+
+  Background:
+    Given Multiple uncordoned nodes
+
+  Scenario: Draining an uncordoned node
+    Given an uncordoned node
+    When the user issues a drain command to the node
+    Then the draining nodes nexuses should be moved to other suitable nodes
+    And the node should be cordoned
+
+  Scenario: Draining a cordoned node
+    Given a cordoned node
+    When the user issues a drain command to the node
+    Then the draining nodes nexuses should be moved to other suitable nodes
+    And the node should remain cordoned
+
+# The system is unable to move all resources from the draining node to another node
+  Scenario: Draining a node with insufficient system capacity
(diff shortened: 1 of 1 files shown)


tests/bdd/features/drain/node/feature.feature
```diff
@@ -0,0 +1,24 @@
+Feature: Drain
+
+  Background:
+    Given Multiple uncordoned nodes
+
+  Scenario: Draining a node
+    Given an uncordoned node
+    When the user issues a drain command to the node
+    Then the node should be cordoned
+    And nexuses should be moved to other suitable nodes
```
tiagolobocastro, line 10: nit: "its nexuses"?

tests/bdd/features/drain/node/feature.feature
```diff
@@ -0,0 +1,24 @@
...
+    When the user issues a drain command to the node
+    Then the node should remain cordoned
+    And nexuses should be moved to other suitable nodes
+
+# The system is unable to move all resources from the draining node to another node
+  Scenario: Draining a node with insufficient system capacity
+    Given an uncordoned node
+    And insufficient system-wide capacity on other nodes to accommodate all drained resources
+    When the user issues a drain command to the node
+    Then the drain should fail to complete in its entirety
```
tiagolobocastro, line 23: Do we need a use case for the system resources changing during the drain? Eg: initially we have enough capacity, but as we're draining, that changes for some reason (example: more volumes were created, or a node is lost, or even cordoned)

tiagolobocastro (approved):

tests/bdd/features/drain/node/feature.feature
paulyoong: Done.

tests/bdd/features/drain/node/feature.feature
paulyoong: Yes good idea. I have added an additional scenario. Can you take a look and let me know what you think?

paulyoong:
bors merge

PR #266, opened 2022-06-30
chore(bdd): add offline node deletion feature file

tiagolobocastro:


tests/bdd/features/delete/node/feature.feature
```diff
@@ -0,0 +1,28 @@
...
+  Scenario: Deleting an offline non-drained node
+    Given an offline node with provisioned resources
+    When the user issues a node delete command
+    Then the node should eventually be removed from the control-plane
+    And the node should not be reported by the control-plane
+
+  Scenario: Deleting a non-existing node
+    Given a non-existing node
+    When the user issues a node delete command
+    Then the command should succeed
```
paulyoong, line 27: Should the command succeed or should it return a "Not found" status code? How do we wish to handle idempotence?

paulyoong (approved):

tests/bdd/features/delete/node/feature.feature
tiagolobocastro: Because the **when** is "the user issues", I was thinking of the user side of things, and just succeeding for the user. Though internally via the rest api we'd return a NotFound error. 
Maybe we should change this feature file to be more focused on the openapi rather than the user which would better left for the system level bdd?

paulyoong: I think this relates to the comment @GlennBullingham has made - https://github.com/openebs/mayastor-control-plane/pull/263#discussion_r911405965

We need to decide how we handle the various success criteria.

tiagolobocastro:
bors merge

PR #267, opened 2022-07-04
fix(csi): csi node to register node name, correct topology logic

Abhinandan-Purkait:

### DIFF
--- a/control-plane/csi-driver/controller/src/controller.rs
+++ b/control-plane/csi-driver/controller/src/controller.rs
@@ -1,7 +1,7 @@
 use crate::{ApiClientError, CreateVolumeTopology, CsiControllerConfig, IoEngineApiClient};
 use regex::Regex;
 use rpc::csi::*;
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use tonic::{Response, Status};
 use tracing::{debug, error, instrument, warn};
 use uuid::Uuid;
@@ -13,11 +13,10 @@ use utils::{CREATED_BY_KEY, DSP_OPERATOR};
 
 use rpc::csi::Topology as CsiTopology;
 
-const K8S_HOSTNAME: &str = "kubernetes.io/hostname";
+const OPENEBS_TOPOLOGY_KEY: &str = "openebs.io/nodename";
 const VOLUME_NAME_PATTERN: &str =
     r"pvc-([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})";
 const SUPPORTED_FS_TYPES: [&str; 2] = ["ext4", "xfs"];
-const CSI_NODE_PREFIX: &str = "csi-node://";
 
 #[derive(Debug, Default)]
 pub struct CsiControllerSvc {}
@@ -73,15 +72,6 @@ fn parse_protocol(proto: Option<&String>) -> Result<VolumeShareProtocol, Status>
             "Invalid protocol: {:?}",
             proto
         ))),
-    }
-}
-
-/// Transform Kubernetes node ID into its real hostname.
-fn normalize_hostname(name: &str) -> String {
-    if let Some(hostname) = name.strip_prefix(CSI_NODE_PREFIX) {
-        hostname.to_string()
-    } else {
-        name.to_string()
     }
 }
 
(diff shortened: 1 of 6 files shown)


control-plane/csi-driver/controller/src/controller.rs
```diff
@@ -660,7 +656,7 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
 
         // Determine target node, if requested.
         let node: Option<&String> = if let Some(topology) = args.accessible_topology.as_ref() {
-            topology.segments.get(K8S_HOSTNAME)
+            topology.segments.get(OPENEBS_TOPOLOGY_KEY)
```
tiagolobocastro, line 659: Not related to your changes, but it doesn't seem like the logic in this fn is correct? maybe to look at in another task maybe..

control-plane/csi-driver/node/src/node.rs
```diff
@@ -22,11 +22,14 @@ use crate::{
 #[derive(Clone, Debug)]
 pub struct Node {
     pub node_name: String,
+    pub host_name: String,
     pub filesystems: Vec<String>,
 }
 
 const ATTACH_TIMEOUT_INTERVAL: Duration = Duration::from_millis(100);
 const ATTACH_RETRIES: u32 = 100;
+const K8S_HOSTNAME_KEY: &str = "kubernetes.io/hostname";
```
tiagolobocastro, line 31: Why do we need to host_name again? To keep it simple, could we do without it altogether and just use nodename?

tiagolobocastro (commented):
One question, with this in place, will upgrades work fine? Eg: if a volume was created with the old scheme, and now we have this in place, will that cause any issues? Ah, that is why we're keeping the hostname as well?

control-plane/csi-driver/node/src/node.rs
Abhinandan-Purkait: We need it for the topology key.

control-plane/csi-driver/controller/src/controller.rs
Abhinandan-Purkait: Yeah, will take a look, haven't explored this part.

control-plane/csi-driver/node/src/node.rs
tiagolobocastro: Can't the topology simply use node name?

control-plane/csi-driver/controller/src/controller.rs
```diff
@@ -266,19 +266,20 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
         let mut allowed_nodes: HashSet<String> = HashSet::new();
         let mut preferred_nodes: HashSet<String> = HashSet::new();
         let mut inclusive_label_topology: HashMap<String, String> = HashMap::new();
+        // List of supported topology keys are:-
+        // {
+        //     openebs.io/nodename
+        //     kubernetes.io/hostname
+        // }
```
tiagolobocastro, line 273: Do we need this, isn't it "self-documented" below?

Abhinandan-Purkait: Yeah we don't

control-plane/csi-driver/node/src/node.rs
Abhinandan-Purkait: Yes, for that we have the `openebs.io/nodename` key, but as per the other engines of openebs, we have both of these keys. So kept it like others.

control-plane/csi-driver/controller/src/controller.rs
```diff
@@ -272,31 +263,32 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
...
-                    if k != K8S_HOSTNAME {
-                        return Err(Status::invalid_argument(
-                            "Volume topology other than hostname not supported",
-                        ));
+                    // Reject all others than `supported_keys`
+                    if !supported_keys.iter().any(|supported| supported == k) {
+                        return Err(Status::invalid_argument(format!(
+                            "Volume topology keys other than {} and {} not supported",
+                            K8S_HOSTNAME_KEY, OPENEBS_TOPOLOGY_KEY
+                        )));
```
paulyoong, line 281: Nit: A suggestion for readability.
```suggestion
                   // Reject all others than `supported_keys`
                    if supported_keys.contains(&k.as_str()) {
                        allowed_nodes.insert(v.to_string());
                    } else {
                        return Err(Status::invalid_argument(format!(
                            "Volume topology keys other than {} and {} not supported",
                            K8S_HOSTNAME_KEY, OPENEBS_TOPOLOGY_KEY
                        )));
                    }
```

paulyoong (approved):

control-plane/csi-driver/node/src/node.rs
croomes: Is it possible to only use `kubernetes.io/hostname` for topology?  Admittedly I'm new to this code but I'm having trouble understanding why both are used from the context in this PR.  I agree it would be better to use the k8s native key wherever possible.

Are there cases where the node name is different from the k8s host name? In my cluster at least they're the same:

Node labels:
```yaml
    kubernetes.io/hostname: aks-storagepool-25014071-vmss000000
    openebs.io/engine: mayastor
    openebs.io/nodename: aks-storagepool-25014071-vmss000000
```
CSINode:
```
  - name: io.openebs.csi-mayastor
    nodeID: aks-storagepool-25014071-vmss000000
    topologyKeys:
    - kubernetes.io/hostname
    - openebs.io/nodename
```

Abhinandan-Purkait: Yes, we have found `aws managed clusters` where then nodename, differs from hostname. To handle that case as well we decided to go with this change.

croomes: Sorry, just a follow up as I'm trying to get my head around this.  I'm not clear why you would need to check both.

It's just a label so it shouldn't really matter as long as you set the values correctly when you register the node.  Previously it registered using the node name rather than the host name:

```rust
segments.insert("kubernetes.io/hostname".to_owned(), self.node_name.clone());
```
Couldn't you just fix this bit to use the host name?

It just seems over-the-top to check both:
```
accessibility requirements: topology map[kubernetes.io/hostname:aks-storagepool-25014071-vmss000001 openebs.io/nodename:aks-storagepool-25014071-vmss000001]
```

Abhinandan-Purkait: Previously it used to register with the hostname, rather than the node name. The variable name might be node_name but the value was hostname, if you go through the code. This was causing an issue when hostname differs from nodename.

Abhinandan-Purkait: With this change in the PR we are registering with the nodename and not the hostname.

Abhinandan-Purkait: Yeah, right. So as dediced we would go with the `nodename` only as that would potentially solve the issue for now, but having said that we need to revisit this later and find a proper way of doing it.

tiagolobocastro (approved):

Abhinandan-Purkait:
bors merge

Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors merge

PR #269, opened 2022-07-11
feat(kubectl plugin): add get block-devices command to list all/usable disks per node

Abhinandan-Purkait:
Usage/Output All, Usable

```
➜  mayastor-control-plane (device-inv) kubectl mayastor get block-devices <node-name> --all

 DEVNAME          DEVTYPE    SIZE       AVAILABLE  MODEL                       DEVPATH                                                         FSTYPE  FSUUID  MOUNTPOINT  PARTTYPE                              MAJOR            MINOR                                     DEVLINKS 
 /dev/nvme0n1     disk       167772160  no         Amazon Elastic Block Store  /devices/pci0000:00/0000:00:04.0/nvme/nvme0/nvme0n1             259     0                                                                                                                    "/dev/disk/by-path/pci-0000:00:04.0-nvme-1", "/dev/disk/by-id/nvme-nvme.1d0f-766f6c3038346136373931663365363866613164-416d617a6f6e20456c617374696320426c6f636b2053746f7265-00000001", "/dev/disk/by-id/nvme-Amazon_Elastic_Block_Store_vol084a6791f3e68fa1d" 
 /dev/nvme0n1p1   partition  167544799  no         Amazon Elastic Block Store  /devices/pci0000:00/0000:00:04.0/nvme/nvme0/nvme0n1/nvme0n1p1   259     1       ext4        4211f95f-755b-4c22-9740-17b12a609123  cloudimg-rootfs  gpt:0fc63daf-8483-4772-8e79-3d69d8477de4  "/dev/disk/by-uuid/4211f95f-755b-4c22-9740-17b12a609123", "/dev/disk/by-id/nvme-nvme.1d0f-766f6c3038346136373931663365363866613164-416d617a6f6e20456c617374696320426c6f636b2053746f7265-00000001-part1", "/dev/disk/by-partuuid/54ecbbe1-f829-4864-bd41-588e2ccc038b", "/dev/disk/by-label/cloudimg-rootfs", "/dev/disk/by-id/nvme-Amazon_Elastic_Block_Store_vol084a6791f3e68fa1d-part1", "/dev/disk/by-path/pci-0000:00:04.0-nvme-1-part1" 

```

```
➜  mayastor-control-plane (device-inv) kubectl mayastor get block-devices <node name>

 DEVNAME       DEVTYPE  SIZE       AVAILABLE  MODEL                       DEVPATH                                              MAJOR  MINOR  DEVLINKS 
 /dev/nvme3n2  disk     104857600  yes        Amazon Elastic Block Store  /devices/pci0000:00/0000:00:1e.0/nvme/nvme2/nvme3n2  259    7      "/dev/disk/by-path/pci-0000:00:1e.0-nvme-1", "/dev/disk/by-id/nvme-Amazon_Elastic_Block_Store_vol019d8523478a2a7cc", "/dev/disk/by-id/nvme-nvme.1d0f-766f6c3031396438353233343738613261376363-416d617a6f6e20456c617374696320426c6f636b2053746f7265-00000001" 

```

### DIFF
--- a/control-plane/plugin/src/operations.rs
+++ b/control-plane/plugin/src/operations.rs
@@ -42,3 +42,11 @@ pub trait ReplicaTopology {
     type ID;
     async fn topology(id: &Self::ID, output: &utils::OutputFormat);
 }
+
+/// GetBlockDevices trait.
+/// To be implemented by resources which support the 'get block-devices' operation
+#[async_trait(?Send)]
+pub trait GetBlockDevices {
+    type ID;
+    async fn get_blockdevices(id: &Self::ID, all: &bool, output: &utils::OutputFormat);
+}
--- a/control-plane/plugin/src/resources/blockdevice.rs
+++ b/control-plane/plugin/src/resources/blockdevice.rs
@@ -1,0 +1,214 @@
+use crate::{
+    operations::GetBlockDevices,
+    resources::{
+        utils::{
+            print_table, CreateRows, GetHeaderRow, OutputFormat, BLOCKDEVICE_HEADERS_ALL,
+            BLOCKDEVICE_HEADERS_USABLE,
+        },
+        NodeId,
+    },
+    rest_wrapper::RestClient,
+};
+use async_trait::async_trait;
+use openapi::apis::Url;
+use prettytable::Row;
+use serde::Serialize;
+
+/// Blockdevice resource.
+#[derive(clap::Args, Debug)]
+pub struct BlockDevice {}
+
+/// New type to wrap BD returned from REST call with all set to true
+#[derive(Clone, Debug, Serialize)]
+struct BlockDeviceAll(openapi::models::BlockDevice);
+
+/// New type to wrap BD returned from REST call with all set to false
+#[derive(Clone, Debug, Serialize)]
+struct BlockDeviceUsable(openapi::models::BlockDevice);
+
+#[derive(Debug, Clone, clap::Args)]
+/// BlockDevice args
+pub struct BlockDeviceArgs {
+    /// Id of the node
+    node_id: NodeId,
+    #[clap(long)]
+    /// Shows all devices if invoked, otherwise shows usable disks.
+    all: bool,
+}
+
+impl BlockDeviceArgs {
+    /// get the node id
+    pub fn node_id(&self) -> NodeId {
+        self.node_id.clone()
+    }
+    /// get the all value
+    pub fn all(&self) -> bool {
+        self.all
+    }
+}
+
+// CreateRows trait for BlockDeviceAll would create the row from the
+// BD returned from REST call with all set to true.
+impl CreateRows for BlockDeviceAll {
+    fn create_rows(&self) -> Vec<Row> {
+        vec![row![
+            self.0.devname.clone(),
+            self.0.devtype.clone(),
+            self.0.size.to_string(),
+            String::from(if self.0.available { "yes" } else { "no" }),
+            self.0.model.clone(),
+            self.0.devpath.clone(),
+            self.0.devmajor.to_string(),
+            self.0.devminor.to_string(),
+            self.0.filesystem.fstype.clone(),
+            self.0.filesystem.uuid.clone(),
+            self.0.filesystem.mountpoint.clone(),
+            get_partition_type(&self.0.partition),
+            self.0
+                .devlinks
+                .iter()
+                .map(|s| format!("\"{}\"", s))
+                .collect::<Vec<String>>()
+                .join(", "),
+        ]]
+    }
+}
+
+// CreateRows trait for BlockDeviceUsable would create row from the
+// BD returned from REST call with all set to false.
+impl CreateRows for BlockDeviceUsable {
+    fn create_rows(&self) -> Vec<Row> {
+        vec![row![
+            self.0.devname.clone(),
+            self.0.devtype.clone(),
+            self.0.size.to_string(),
+            String::from(if self.0.available { "yes" } else { "no" }),
+            self.0.model.clone(),
+            self.0.devpath.clone(),
+            self.0.devmajor.to_string(),
+            self.0.devminor.to_string(),
+            self.0
+                .devlinks
+                .iter()
+                .map(|s| format!("\"{}\"", s))
+                .collect::<Vec<String>>()
+                .join(", "),
+        ]]
+    }
+}
+
+fn get_partition_type(partition: &openapi::models::BlockDevicePartition) -> String {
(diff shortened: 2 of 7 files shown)


paulyoong (approved):

control-plane/plugin/src/operations.rs
```diff
@@ -42,3 +42,11 @@ pub trait ReplicaTopology {
     type ID;
     async fn topology(id: &Self::ID, output: &utils::OutputFormat);
 }
+
+/// GetBlockDevices trait.
+/// To be implemented by resources which support the 'get block-devices' operation
+#[async_trait(?Send)]
+pub trait GetBlockDevices {
```
tiagolobocastro, line 49: Do we need a command to get block devices for all nodes? Which would simply get all nodes, loop and for each issue get_blockdevices rest call?

k8s/plugin/src/main.rs
```diff
@@ -82,6 +82,14 @@ async fn execute(cli_args: CliArgs) {
...
             GetResources::Nodes => node::Nodes::list(&cli_args.output).await,
             GetResources::Node { id } => node::Node::get(&id, &cli_args.output).await,
+            GetResources::BlockDevices(bdargs) => {
+                blockdevice::BlockDevice::get_blockdevices(
+                    &bdargs.node_id(),
+                    &bdargs.all(),
+                    &cli_args.output,
+                )
+                .await
+            }
```
tiagolobocastro, line 92: Could we add examples to the README please?

tiagolobocastro:
Not related to these code changes but, out of curiosity, on your PR description example, after we create the pool, is the device still reported as `AVAILABLE` ?

control-plane/plugin/src/operations.rs
Abhinandan-Purkait: Should I add cluster wide as well? @tiagolobocastro @paulyoong

Abhinandan-Purkait:
> Not related to these code changes but, out of curiosity, on your PR description example, after we create the pool, is the device still reported as `AVAILABLE` ?

Yes, it does. Seems like the blockdevice api, doesn't mark the `disks` used by `mayastor` pool as `not available`. The identification factor a disk to be used is not present. Need to come up with some approach to handle this.

tiagolobocastro:
> 

Since this is the case, there's not much we could do...
we have a few options to think about:

IF we can tell the device IS in one of the `Pool`'s reported by the OpenApi, in such case we simply mark it as `NOT AVAILABLE` - we may even show the pool name as one of the properties?
But how do we know that pool is our pool? Should mayastor inspect each `AVAILABLE` device and check if it's a pool?
In any case this seems like a separate task so I propose that we simply document this on the help section of the plugin. And instead of `yes`, maybe it should read `yes ?` :) 
Thoughts?

control-plane/plugin/src/resources/blockdevice.rs
```diff
@@ -0,0 +1,210 @@
...
+                .expect("Expected to be initialised once");
+            Ok(())
+        }
+        Err(e) => Err(e),
+    }
+}
+
+fn check_available(possible_values: Vec<String>) -> bool {
+    for value in possible_values {
+        if USED_DISKS.get().unwrap().contains(&value) {
```
tiagolobocastro, line 205: This won't work if the diskpool is created with URI's?

control-plane/plugin/src/resources/blockdevice.rs
```diff
@@ -0,0 +1,210 @@
...
+    },
+    rest_wrapper::RestClient,
+};
+use async_trait::async_trait;
+use once_cell::sync::OnceCell;
+use openapi::{clients::tower::Error, models::RestJsonError};
+use prettytable::Row;
+use serde::Serialize;
+
+static USED_DISKS: OnceCell<Vec<String>> = OnceCell::new();
```
tiagolobocastro, line 18: Why do we need this? Can we not add the list of pools on the node to the new-types which we have below?

control-plane/plugin/src/operations.rs
Abhinandan-Purkait: Resolving for now.

control-plane/plugin/src/resources/blockdevice.rs
Abhinandan-Purkait: Yes, right. Thanks for pointing out. Added normalization.

control-plane/plugin/src/resources/blockdevice.rs
Abhinandan-Purkait: Yes, we don't need this. Thanks for pointing out. Will refactor it,

control-plane/plugin/src/resources/blockdevice.rs
```diff
@@ -0,0 +1,217 @@
...
+        let mut possible_values = vec![bd.devname.clone()];
+        for devlink in &bd.devlinks {
+            possible_values.push(devlink.clone());
+        }
+        if bd.available && check_available(used_disks, possible_values) {
+            bd.available = true;
+            transformed_bds.push(bd);
+        } else if all {
+            bd.available = false;
+            transformed_bds.push(bd);
```
tiagolobocastro, line 194: We don't need a new vector here right? we can return blockdevices if we use iter_mut on the for loop?

control-plane/plugin/src/resources/mod.rs
```diff
@@ -25,6 +28,8 @@ pub enum GetResources {
     Nodes,
     /// Get node with the given ID.
     Node { id: NodeId },
+    /// Get BlockDevices present on the Node. Lists usable devices by default.
```
tiagolobocastro, line 31: I think it's worth here mentioning here that the AVAILABLE field is not correct if the pool is not currently part of the control plane?

tiagolobocastro (approved):

control-plane/plugin/src/resources/mod.rs
Abhinandan-Purkait: I have added that in the args part. As an description to the --all flag. Do we need it here as well?

control-plane/plugin/src/resources/blockdevice.rs
Abhinandan-Purkait: Yeah.

control-plane/plugin/src/resources/mod.rs
tiagolobocastro: I'd say move it from all to here - no sure though, up to you.

Abhinandan-Purkait: Okay.

Abhinandan-Purkait:
bors merge

PR #272, opened 2022-07-19
feat(cordon): allow cordoning of a node

paulyoong:
Provide the ability to cordon a node. Applying a cordon prevents the
control plane from being able to schedule resources on that node.

When cordoning a node a label must be applied. This uniquely identifies
a particular cordon operation. When uncordoning a node the appropriate
label must also be supplied.

A particular cordon label can only be applied once. Attempting to apply
a cordon with a label that already exists will fail.

Cordoning a node can be achieved through the kubectl plugin using the
following command:
    kubectl-mayastor cordon node <node_id> <label>

Uncordoning a node can be achieved through the kubectl plugin using the
following command:
    kubectl-mayastor uncordon node <node_id> <label>

Cordon labels can be listed through the kubectl plugin using the
following command:
    kubectl-mayastor get cordon-labels <node_id>

### DIFF
--- a/common/src/types/v0/store/node.rs
+++ b/common/src/types/v0/store/node.rs
@@ -64,7 +64,7 @@ impl NodeSpec {
 
 impl From<NodeSpec> for models::NodeSpec {
     fn from(src: NodeSpec) -> Self {
-        Self::new(src.endpoint, src.id)
+        Self::new(src.endpoint, src.id, vec![""])
     }
 }
 
--- a/common/src/types/v0/store/node.rs
+++ b/common/src/types/v0/store/node.rs
@@ -34,14 +34,22 @@ pub struct NodeSpec {
     endpoint: String,
     /// Node labels.
     labels: NodeLabels,
+    /// Cordon labels.
+    cordon_labels: Vec<String>,
 }
 impl NodeSpec {
     /// Return a new `Self`
-    pub fn new(id: NodeId, endpoint: String, labels: NodeLabels) -> Self {
+    pub fn new(
+        id: NodeId,
+        endpoint: String,
+        labels: NodeLabels,
+        cordon_label: Option<Vec<String>>,
+    ) -> Self {
         Self {
             id,
             endpoint,
             labels,
+            cordon_labels: cordon_label.unwrap_or_default(),
         }
     }
     /// Node identification
@@ -60,11 +68,29 @@ impl NodeSpec {
     pub fn set_endpoint(&mut self, endpoint: String) {
         self.endpoint = endpoint
     }
+    /// Cordon node by applying the label.
+    pub fn cordon(&mut self, label: String) {
+        self.cordon_labels.push(label);
+    }
+    /// Uncordon node by removing the corresponding label.
+    pub fn uncordon(&mut self, label: String) {
+        if let Some(index) = self.cordon_labels.iter().position(|l| l == &label) {
+            self.cordon_labels.remove(index);
+        }
+    }
+    /// Returns whether or not the node is cordoned.
+    pub fn cordoned(&self) -> bool {
+        !self.cordon_labels.is_empty()
+    }
+    /// Returns the cordon labels
+    pub fn cordon_labels(&self) -> Vec<String> {
+        self.cordon_labels.clone()
+    }
 }
 
 impl From<NodeSpec> for models::NodeSpec {
     fn from(src: NodeSpec) -> Self {
-        Self::new(src.endpoint, src.id, vec![""])
+        Self::new(src.endpoint, src.id, src.cordon_labels)
     }
 }
 
--- a/control-plane/agents/core/src/core/registry.rs
+++ b/control-plane/agents/core/src/core/registry.rs
@@ -313,4 +313,9 @@ impl Registry {
             None => Ok(()),
         }
     }
+
+    /// Returns whether or not the node with the given ID is cordoned.
+    pub fn node_cordoned(&self, node_id: &NodeId) -> Result<bool, SvcError> {
+        Ok(self.specs.get_node(node_id)?.cordoned())
+    }
 }
--- a/control-plane/agents/core/src/node/service.rs
+++ b/control-plane/agents/core/src/node/service.rs
@@ -78,6 +78,14 @@ impl NodeOperations for Service {
         let blockdevices = self.get_block_devices(&req).await?;
         Ok(blockdevices)
     }
+
+    async fn cordon(&self, _id: NodeId, _label: String) -> Result<Node, ReplyError> {
+        todo!()
+    }
+
+    async fn uncordon(&self, _id: NodeId, _label: String) -> Result<Node, ReplyError> {
+        todo!()
+    }
 }
 
 #[tonic::async_trait]
--- a/control-plane/agents/core/src/node/service.rs
+++ b/control-plane/agents/core/src/node/service.rs
@@ -79,12 +79,14 @@ impl NodeOperations for Service {
         Ok(blockdevices)
     }
 
-    async fn cordon(&self, _id: NodeId, _label: String) -> Result<Node, ReplyError> {
-        todo!()
+    async fn cordon(&self, id: NodeId, label: String) -> Result<Node, ReplyError> {
+        let node = self.cordon(id, label).await?;
+        Ok(node)
     }
 
-    async fn uncordon(&self, _id: NodeId, _label: String) -> Result<Node, ReplyError> {
-        todo!()
+    async fn uncordon(&self, id: NodeId, label: String) -> Result<Node, ReplyError> {
+        let node = self.uncordon(id, label).await?;
+        Ok(node)
     }
 }
 
@@ -303,5 +305,25 @@ impl Service {
             .map(|rpc_bdev| rpc_bdev.to_mbus())
             .collect();
         Ok(BlockDevices(bdevs))
+    }
+
+    async fn cordon(&self, id: NodeId, label: String) -> Result<Node, SvcError> {
+        let spec = self
+            .registry
+            .specs()
+            .cordon_node(&self.registry, &id, label)
+            .await?;
+        let state = self.registry.get_node_state(&id).await.ok();
+        Ok(Node::new(id, Some(spec), state))
+    }
+
+    async fn uncordon(&self, id: NodeId, label: String) -> Result<Node, SvcError> {
+        let spec = self
+            .registry
+            .specs()
+            .uncordon_node(&self.registry, &id, label)
+            .await?;
+        let state = self.registry.get_node_state(&id).await.ok();
+        Ok(Node::new(id, Some(spec), state))
     }
 }
--- a/control-plane/agents/core/src/node/specs.rs
+++ b/control-plane/agents/core/src/node/specs.rs
@@ -30,6 +30,7 @@ impl ResourceSpecsLocked {
                         node.id.clone(),
                         node.grpc_endpoint.clone(),
                         NodeLabels::new(),
+                        None,
                     );
                     specs.nodes.insert(node.clone());
                     (true, node)
--- a/control-plane/agents/core/src/node/specs.rs
+++ b/control-plane/agents/core/src/node/specs.rs
@@ -63,7 +63,7 @@ impl ResourceSpecsLocked {
     }
 
     /// Get all locked node specs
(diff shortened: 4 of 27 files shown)


paulyoong:
bors try

common/src/types/v0/store/node.rs
```diff
@@ -60,11 +68,29 @@ impl NodeSpec {
     pub fn set_endpoint(&mut self, endpoint: String) {
         self.endpoint = endpoint
     }
+    /// Cordon node by applying the label.
+    pub fn cordon(&mut self, label: String) {
+        self.cordon_labels.push(label);
+    }
+    /// Uncordon node by removing the corresponding label.
+    pub fn uncordon(&mut self, label: String) {
+        if let Ok(index) = self.cordon_labels.binary_search(&label) {
```
Abhinandan-Purkait, line 77: Question: Does `binary_search()` method sort the array internally before searching? If not do we need to sort here?

control-plane/grpc/src/operations/node/client.rs
```diff
@@ -84,4 +84,34 @@ impl NodeOperations for NodeClient {
             None => Err(ReplyError::invalid_response(ResourceKind::Block)),
         }
     }
+
+    async fn cordon(&self, id: NodeId, label: String) -> Result<Node, ReplyError> {
```
Abhinandan-Purkait, line 88: Do we need to add the tracing tags here? ref: https://github.com/openebs/mayastor-control-plane/blob/08f07f949db2a852a13ea98bc56ee457f8ca222f/control-plane/grpc/src/operations/node/client.rs#L43

control-plane/grpc/src/operations/node/server.rs
```diff
@@ -71,4 +72,39 @@ impl NodeGrpc for NodeServer {
...
+        request: tonic::Request<CordonNodeRequest>,
+    ) -> Result<tonic::Response<CordonNodeReply>, tonic::Status> {
+        let req: CordonNodeRequest = request.into_inner();
+        match self.service.cordon(req.node_id.into(), req.label).await {
+            Ok(node) => {
+                let resp = Response::new(CordonNodeReply {
+                    reply: Some(cordon_node_reply::Reply::Node(node.into())),
+                });
+                Ok(resp)
+            }
```
Abhinandan-Purkait, line 86:
```suggestion
            Ok(node) => Ok(Response::new(CordonNodeReply {
                    reply: Some(cordon_node_reply::Reply::Node(node.into())),
            })),
```
Can we directly return from here?

control-plane/grpc/src/operations/node/server.rs
```diff
@@ -71,4 +72,39 @@ impl NodeGrpc for NodeServer {
...
+        request: tonic::Request<UncordonNodeRequest>,
+    ) -> Result<tonic::Response<UncordonNodeReply>, tonic::Status> {
+        let req: UncordonNodeRequest = request.into_inner();
+        match self.service.uncordon(req.node_id.into(), req.label).await {
+            Ok(node) => {
+                let resp = Response::new(UncordonNodeReply {
+                    reply: Some(uncordon_node_reply::Reply::Node(node.into())),
+                });
+                Ok(resp)
+            }
```
Abhinandan-Purkait, line 104:
```suggestion
            Ok(node) => Ok(Response::new(UncordonNodeReply {
                    reply: Some(uncordon_node_reply::Reply::Node(node.into())),
                })),
```

control-plane/grpc/src/operations/node/traits.rs
```diff
@@ -30,6 +33,10 @@ pub trait NodeOperations: Send + Sync {
         get_blockdevice: &dyn GetBlockDeviceInfo,
         ctx: Option<Context>,
     ) -> Result<BlockDevices, ReplyError>;
+    /// Cordon the node with the given ID and associate the label with the cordon.
```
Abhinandan-Purkait, line 36: ` associate the label with the cordon.` --> ` associate the label with the cordoned node.` ?

control-plane/plugin/src/bin/rest-plugin/main.rs
```diff
@@ -72,11 +72,22 @@ async fn execute(cli_args: CliArgs) {
                 )
                 .await
             }
+            GetResources::CordonLabels { id } => node::Node::get_labels(id).await,
```
Abhinandan-Purkait, line 75: Do we need a resource here? Otherwise if in future we have pool cordoning feature, then this probably won't support that. Rather should we just associate the command to a resource itself?

control-plane/plugin/src/resources/mod.rs
```diff
@@ -46,6 +48,13 @@ pub enum ScaleResources {
     },
 }
 
+/// The types of resources that support cordoning.
+#[derive(clap::Subcommand, Debug)]
+pub enum CordonResources {
+    /// Cordon the node with the given ID.
```
Abhinandan-Purkait, line 54: Should we also add a word or two about the `labels` as this probably would become the document in the `--help`.

Abhinandan-Purkait (commented):
Can we please have the README for the plugin updated with `usages` and `examples` for the new commands.

common/src/types/v0/store/node.rs
```diff
@@ -60,11 +68,29 @@ impl NodeSpec {
...
+            self.cordon_labels.remove(index);
+        }
```
chriswldenyer, line 79: Maybe this should return an error if the label cannot be found (rather than searching before calling this function)

control-plane/grpc/src/operations/node/server.rs
chriswldenyer: > Can we directly return from here?

What's the advantage? Just making the code denser often doesn't help - it makes it harder to unpick

chriswldenyer (commented):
Still reviewing but some initial comments.

common/src/types/v0/store/node.rs
paulyoong: I don't believe it sorts it internally. However, I don't think it matters. We disallow multiple instances of the same label being added, so we should only ever have one.

paulyoong: Yes you are correct. Let me find an alternative.

common/src/types/v0/store/node.rs
```diff
@@ -60,11 +68,29 @@ impl NodeSpec {
...
+    }
+    /// Returns whether or not the node is cordoned.
+    pub fn cordoned(&self) -> bool {
+        !self.cordon_labels.is_empty()
+    }
+    /// Returns the cordon labels
+    pub fn cordon_labels(&self) -> Vec<String> {
```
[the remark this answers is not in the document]
tiagolobocastro: AFAIK in Rust, it's "idiomatic", not to prefix get methods with get.

control-plane/agents/core/src/core/registry.rs
```diff
@@ -313,4 +313,12 @@ impl Registry {
             None => Ok(()),
         }
     }
+
+    /// Returns whether or not the node with the given ID is cordoned.
+    pub fn node_cordoned(&self, node_id: &NodeId) -> bool {
+        self.specs
+            .get_cordoned_nodes()
+            .into_iter()
+            .any(|node_spec| node_spec.id() == node_id)
```
tiagolobocastro, line 322: Would it be simpler to get just the node that has `node_id` and check if it's cordoned?

control-plane/agents/core/src/node/specs.rs
```diff
@@ -73,4 +74,70 @@ impl ResourceSpecsLocked {
...
+            if locked_node.cordon_labels().contains(&label) {
+                return Err(SvcError::CordonLabel {
+                    node_id: node_id.to_string(),
+                    label,
+                });
+            }
+            locked_node.cordon(label);
+            locked_node.clone()
+        };
+        registry.store_obj(&cordoned_node_spec).await?;
```
tiagolobocastro, line 99: I wonder if node specs should now follow the same pattern we have for the other resources?

control-plane/grpc/src/operations/node/server.rs
tiagolobocastro: With the new clippy version, IIRC it'll probably warn if we put it on a variable to simply return it on the next statement.

control-plane/plugin/src/resources/volume.rs
```diff
@@ -126,7 +126,7 @@ impl Scale for Volume {
                 }
                 OutputFormat::None => {
                     // In case the output format is not specified, show a success message.
-                    println!("Volume {} Scaled Successfully 🚀", id)
+                    println!("Volume {} scaled successfully 🚀", id)
```
tiagolobocastro, line 129: Tagging @blaisedias to check if a case change breaks the system tests.

tiagolobocastro (approved):
Could we break this into several logical commits? I can think of transport changes, openapi, kubectl and the core agent.

common/src/types/v0/store/node.rs
paulyoong: Good catch @Abhinandan-Purkait. I'm now using `.iter().position()` to get the index so that I can remove the label.

control-plane/grpc/src/operations/node/server.rs
paulyoong: Done.

common/src/types/v0/store/node.rs
```diff
@@ -60,11 +68,29 @@ impl NodeSpec {
     pub fn set_endpoint(&mut self, endpoint: String) {
         self.endpoint = endpoint
     }
+    /// Cordon node by applying the label.
+    pub fn cordon(&mut self, label: String) {
+        self.cordon_labels.push(label);
+    }
+    /// Uncordon node by removing the corresponding label.
+    pub fn uncordon(&mut self, label: String) {
```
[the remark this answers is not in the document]
paulyoong: This is handled by the caller. These functions were meant to be kept simple. If the caller doesn't care whether the label is there of not they can just carry on.

control-plane/agents/core/src/node/specs.rs
```diff
@@ -73,4 +74,70 @@ impl ResourceSpecsLocked {
...
+    /// Return the NodeSpec after uncordoning.
+    pub(crate) async fn uncordon_node(
+        &self,
+        registry: &Registry,
+        node_id: &NodeId,
+        label: String,
+    ) -> Result<NodeSpec, SvcError> {
+        let node = self.get_locked_node(node_id)?;
+        // Return an error if the uncordon label doesn't exist.
+        if !node.lock().cordon_labels().contains(&label) {
```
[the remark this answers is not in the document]
paulyoong: I think I prefer to keep it here. This caller happens to care if the label already exists, but others may not.

control-plane/agents/core/src/node/service.rs
```diff
@@ -78,6 +78,16 @@ impl NodeOperations for Service {
         let blockdevices = self.get_block_devices(&req).await?;
         Ok(blockdevices)
     }
+
+    async fn cordon(&self, id: NodeId, label: String) -> Result<Node, ReplyError> {
+        let node = self.cordon(id, label).await?;
```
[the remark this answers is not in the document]
paulyoong: I did this to be in keeping with the rest of the file. The reason we seem to do this currently is because it won't implicitly convert from a `SvcError` to a `ReplyError`.

control-plane/agents/core/src/node/specs.rs
paulyoong: I think you are probably correct but I'll leave this to a subsequent change.

control-plane/agents/core/src/core/registry.rs
paulyoong: Good suggestion. Done

control-plane/agents/core/src/node/specs.rs
tiagolobocastro: Yep, agreed!

paulyoong:
bors try

chriswldenyer (approved):
LGTM

Abhinandan-Purkait (approved):

control-plane/plugin/src/resources/mod.rs
paulyoong: Done.

paulyoong:
bors merge

PR #277, opened 2022-07-27
feat(ha-node-agent): adding initial code for HA node-agent

mynktl:
changes:
- HA node-agent supporting registration with HA cluster-agent
- modified deployer to run HA node-agent for testing


control-plane/grpc/src/operations/ha_node/client.rs
```diff
@@ -0,0 +1,53 @@
...
+        end_point: String,
+    ) -> Result<(), ReplyError> {
+        let response = self
+            .client()
+            .register_node_agent(HaNodeInfo {
+                nodename: node_name,
+                endpoint: end_point,
+            })
+            .await?;
+        println!("registering node agent {:?}", response);
```
tiagolobocastro, line 50: Should be a trace rather than a println.
How often do we register, do we need to log this message everytime we register?

deployer/src/lib.rs
```diff
@@ -261,6 +261,10 @@ pub struct StartOptions {
     /// Maximum number of concurrent rebuilds across the cluster.
     #[structopt(long)]
     max_rebuilds: Option<u32>,
+
+    /// Enable the HA node-agent
+    #[structopt(long)]
+    pub ha_node_agent: bool,
```
tiagolobocastro, line 267: If this is an agent, this should be added to the Agents "class" instead.

control-plane/agents/Cargo.toml
```diff
@@ -46,6 +50,7 @@ indexmap = "1.8.0"
 k8s-openapi = { version = "0.13.0", default-features = false, features = ["v1_20"] }
 kube = { version = "0.60.0", features = ["derive" ] }
 anyhow = "1.0.44"
+clap =  { version = "3.1.5", features = ["color", "derive"] }
```
tiagolobocastro, line 53: nit: I'd prefer to either use structopt since it's what the other agents use or instead change all agents to clap (latter would be my preference :))

control-plane/agents/Cargo.toml
```diff
@@ -17,6 +17,10 @@ path = "jsongrpc/src/server.rs"
 name = "common"
 path = "common/src/lib.rs"
 
+[[bin]]
+ name = "ha-node"
```
tiagolobocastro, line 21: We should start naming things in accordance to the new nomenclature? @paulyoong 
So the name here would be "agent-ha-node" ?

deployer/src/infra/hanodeagent.rs
```diff
@@ -0,0 +1,20 @@
...
+        })
+    }
+
+    async fn start(&self, options: &StartOptions, cfg: &ComposeTest) -> Result<(), Error> {
+        if options.ha_node_agent {
+            cfg.start("ha-node").await?;
+        }
+        Ok(())
+    }
+}
```
tiagolobocastro, line 20: Any reason for not implementing `wait_on` ?

rpc/build.rs
```diff
@@ -32,4 +32,9 @@ fn main() {
             &["api/protobuf/v1"],
         )
         .unwrap_or_else(|e| panic!("Registration protobuf compilation failed: {}", e));
+
+    tonic_build::configure()
+        .build_server(true)
+        .compile(&["api/protobuf/ha_cluster_agent.proto"], &["api/protobuf"])
```
tiagolobocastro, line 38: Should this follow the new protobuf versioning? in v1?
Also, is this a control-plane only api? If so the proto file should live instead in control-plane/grpc/proto/v1,

control-plane/agents/ha-node/src/main.rs
```diff
@@ -0,0 +1,45 @@
...
+    node_name: String,
+
+    /// IP address and port for the ha node-agent to listen on
+    #[clap(short, long, default_value = "0.0.0.0:11600")]
+    grpc_endpoint: String,
+}
+
+static CLUSTER_AGENT_CLIENT: OnceCell<ClusterAgentClient> = OnceCell::new();
+
+fn cluster_agent_client<'a>() -> &'a ClusterAgentClient {
```
mtzaurus, line 25: You can omit <'a> lifetime and use explicitly " -> &'static ClusterAgentClient", as you always deal with static instance.

control-plane/grpc/src/operations/ha_node/client.rs
mtzaurus: At this point, the agent is already registered, so can we please change the text to something like "node agent successfully registered" or so. Thanks.

control-plane/agents/ha-node/src/main.rs
```diff
@@ -0,0 +1,45 @@
...
+#[clap(author, version, about, long_about = None)]
+#[clap(propagate_version = true)]
+struct Cli {
+    /// HA Cluster Agent URL or address to connect to the services.
+    #[clap(short, long, default_value = "http://ha-cluster-agent:11500")]
+    cluster_agent: Uri,
+
+    /// Node name
+    #[clap(short, long)]
+    node_name: String,
```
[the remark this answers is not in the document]
mynktl: ok. I will also mention ref (spec.nodeName)

control-plane/agents/ha-node/src/main.rs
```diff
@@ -0,0 +1,45 @@
...
+    CLUSTER_AGENT_CLIENT
+        .set(ClusterAgentClient::new(cli.cluster_agent, None).await)
+        .ok()
+        .expect("Expect to be initialized only once");
+
+    cluster_agent_client()
+        .register_node_agent(cli.node_name.clone(), cli.grpc_endpoint.clone())
+        .await
+        .ok()
+        .expect("should have been registed");
```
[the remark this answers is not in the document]
mynktl: yes, I will use unwrap here since we don't want to continue if registration fails

deployer/src/infra/hanodeagent.rs
mynktl: for now, node-agent is not having any server. So skipped it. Will add with unimplemented.

rpc/build.rs
mynktl: It will be control-plane only. I will move it in control-plane.

mynktl:
Hi @mtzaurus @paulyoong @tiagolobocastro 
Thanks for the review comments.
I have updated the PR as per the review comments. PTAL.

mtzaurus (approved):

control-plane/grpc/src/operations/ha_node/client.rs
```diff
@@ -0,0 +1,49 @@
...
+    )]
+    async fn register(&self, node_name: String, end_point: String) -> Result<(), ReplyError> {
+        let _response = self
+            .client()
+            .register_node_agent(HaNodeInfo {
+                nodename: node_name,
+                endpoint: end_point,
+            })
+            .await?;
+        println!("node agent successfully registered");
```
tiagolobocastro, line 46: Should this be a trace! message?

control-plane/agents/Cargo.toml
tiagolobocastro: Please rename this to "agent-ha-node".

control-plane/grpc/src/operations/ha_node/traits.rs
```diff
@@ -0,0 +1,8 @@
+use common_lib::transport_api::ReplyError;
+
+/// ClusterAgentOperations trait implemented by client which supports cluster-agent operations
+#[tonic::async_trait]
+pub trait ClusterAgentOperations: Send + Sync {
+    /// Register node with cluster-agent
+    async fn register(&self, node_name: String, endpoint: String) -> Result<(), ReplyError>;
```
tiagolobocastro, line 7: endpoint should be a Uri type?

deployer/src/infra/hanodeagent.rs
tiagolobocastro: This was not what I meant with unimplemented, this was cause issues as it will panic if you start deployer with wait. I meant adding a:
`rpc Probe (ProbeRequest) returns (ProbeResponse) {}`
to the proto and for now we could simply add unimplemented on the probe impl of the trait.

deployer/src/infra/mod.rs
```diff
@@ -416,6 +417,7 @@ impl_component! {
     IoEngine,      4,
     CsiNode,       5,
     CsiController, 5,
+    HANodeAgent, 6,
```
tiagolobocastro, line 420: nit: Please align to the previous lines.
Should the HANodeAgent be 5 and run at the same time as the csi node, would it make sense to run them at the same time?

deployer/src/lib.rs
tiagolobocastro: This change is still missing

deployer/src/lib.rs
```diff
@@ -273,6 +273,10 @@ pub struct StartOptions {
     /// Maximum number of concurrent rebuilds across the cluster.
     #[structopt(long)]
     max_rebuilds: Option<u32>,
+
+    /// Enable HA node-agent
+    #[structopt(long)]
+    pub ha_node_agent: bool,
```
tiagolobocastro, line 279: This change is still missing

tiagolobocastro (changes requested):

mynktl:
> This was not what I meant with unimplemented, this was cause issues as it will panic if you start deployer with wait. I meant adding a:
rpc Probe (ProbeRequest) returns (ProbeResponse) {}
to the proto and for now we could simply add unimplemented on the probe impl of the trait.

For now, node-agent is not running the server. So I will remove the `wait_on`. We will add it once we have the server/check in node-agent.

tiagolobocastro (approved):

mynktl:
bors merge

mynktl:
bors merge

PR #287, opened 2022-08-03
feature: drain

chriswldenyer:
Add APIs to drain a node.
Drain is not yet fully implemented - what is missing is the reconciler that moves nexuses to another node, which is dependent on HA being implemented.
With this PR, drain labels are now stored in the core agent and perform the same cordoning function as cordon labels but also flag that a drain is to be undertaken.
The Node object returned from the core-agent now includes a DrainStatus field which is one of:
NotDraining
Draining
Drained
dependent on the presence of nexuses and drain labels on the node.
    kubectl-plugin drain node <node id> <drain label>
and cancels it via:
    kubectl-plugin uncordon node <node id> <drain label>

A new option --show-drain when used with "get node" or "get nodes" will display the drain labels and the state.
When used with "get nodes" only nodes that are in Draining or Drained state are listed.

### DIFF
--- a/common/src/types/v0/store/node.rs
+++ b/common/src/types/v0/store/node.rs
@@ -37,7 +37,11 @@ pub struct NodeSpec {
     /// Cordon labels.
     #[serde(default)] // Ensure backwards compatibility in etcd when upgrading.
     cordon_labels: Vec<String>,
+    /// Drain labels.
+    #[serde(default)] // Ensure backwards compatibility in etcd when upgrading.
+    drain_labels: Vec<String>,
 }
+
 impl NodeSpec {
     /// Return a new `Self`
     pub fn new(
@@ -45,12 +49,14 @@ impl NodeSpec {
         endpoint: String,
         labels: NodeLabels,
         cordon_label: Option<Vec<String>>,
+        drain_label: Option<Vec<String>>,
     ) -> Self {
         Self {
             id,
             endpoint,
             labels,
             cordon_labels: cordon_label.unwrap_or_default(),
+            drain_labels: drain_label.unwrap_or_default(),
         }
     }
     /// Node identification
@@ -73,25 +79,49 @@ impl NodeSpec {
     pub fn cordon(&mut self, label: String) {
         self.cordon_labels.push(label);
     }
+    /// Drain node by applying the drain label.
+    pub fn drain(&mut self, label: String) {
+        self.drain_labels.push(label);
+    }
     /// Uncordon node by removing the corresponding label.
     pub fn uncordon(&mut self, label: String) {
         if let Some(index) = self.cordon_labels.iter().position(|l| l == &label) {
             self.cordon_labels.remove(index);
         }
+        if let Some(index) = self.drain_labels.iter().position(|l| l == &label) {
+            self.drain_labels.remove(index);
+        }
     }
     /// Returns whether or not the node is cordoned.
     pub fn cordoned(&self) -> bool {
-        !self.cordon_labels.is_empty()
+        !self.cordon_labels.is_empty() || !self.drain_labels.is_empty()
+    }
+    /// Returns whether or not the node is cordoned for drain.
+    pub fn cordoned_for_drain(&self) -> bool {
+        !self.drain_labels.is_empty()
     }
     /// Returns the cordon labels
     pub fn cordon_labels(&self) -> Vec<String> {
         self.cordon_labels.clone()
     }
+    /// Returns the drain labels
+    pub fn drain_labels(&self) -> Vec<String> {
+        self.drain_labels.clone()
+    }
+    pub fn has_cordon_label(&self, label: String) -> bool {
+        if self.cordon_labels.contains(&label) {
+            return true;
+        }
+        if self.drain_labels.contains(&label) {
+            return true;
+        }
+        false
+    }
 }
 
 impl From<NodeSpec> for models::NodeSpec {
     fn from(src: NodeSpec) -> Self {
-        Self::new(src.endpoint, src.id, src.cordon_labels)
+        Self::new(src.endpoint, src.id, src.cordon_labels, src.drain_labels)
     }
 }
 
--- a/common/src/types/v0/transport/node.rs
+++ b/common/src/types/v0/transport/node.rs
@@ -105,6 +105,17 @@ impl Default for NodeStatus {
     }
 }
 
+#[derive(Serialize, Deserialize, Debug, Default, Clone, EnumString, ToString, Eq, PartialEq)]
+pub enum DrainStatus {
+    /// Node has no drain cordon labels
+    #[default]
+    NotDraining,
+    /// Node is draining
+    Draining,
+    /// Node is drained
+    Drained,
+}
+
 /// Node State information
 #[derive(Serialize, Deserialize, Default, Debug, Clone, Eq, PartialEq)]
 #[serde(rename_all = "camelCase")]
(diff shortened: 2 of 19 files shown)


common/src/types/v0/store/node.rs
```diff
@@ -73,25 +79,49 @@ impl NodeSpec {
...
             self.cordon_labels.remove(index);
         }
+        if let Some(index) = self.drain_labels.iter().position(|l| l == &label) {
+            self.drain_labels.remove(index);
+        }
     }
     /// Returns whether or not the node is cordoned.
     pub fn cordoned(&self) -> bool {
-        !self.cordon_labels.is_empty()
+        !self.cordon_labels.is_empty() || !self.drain_labels.is_empty()
```
tiagolobocastro, line 97: How do we expose these 2 label fields to the user?
A user cancels a drain by uncordoning right?

common/src/types/v0/transport/node.rs
```diff
@@ -105,6 +105,21 @@ impl Default for NodeStatus {
...
+    /// Node has no drain cordon labels
+    NotDraining,
+    /// Node is draining
+    Draining,
+    /// Node is drained
+    Drained,
+}
+impl Default for DrainStatus {
+    fn default() -> Self {
+        Self::NotDraining
```
tiagolobocastro, line 119: I think with the new rust version you can simply add #[default] above the NodeDraining field in the enum itself - I've not tested this myself

control-plane/agents/core/src/node/registry.rs
```diff
@@ -17,7 +17,10 @@ impl Registry {
         let nodes = self.nodes().read().await;
         let mut nodes_vec = vec![];
         for node in nodes.values() {
-            nodes_vec.push(node.read().await.node_state().clone());
+            let node_id = node.read().await.id().clone();
```
tiagolobocastro, line 20: can you get the `node_id` from the "copy_node_state" ?

control-plane/agents/core/src/node/registry.rs
```diff
@@ -17,7 +17,10 @@ impl Registry {
...
+            let mut copy_node_state = node.read().await.node_state().clone();
+            copy_node_state.drain_status = self.get_node_drain_status(&node_id).await;
+            nodes_vec.push(copy_node_state.clone());
```
tiagolobocastro, line 23: redundant clone?

control-plane/agents/core/src/node/registry.rs
```diff
@@ -57,8 +60,27 @@ impl Registry {
...
+            Some(node) => {
+                let mut copy_node_state = node.read().await.node_state().clone();
+                copy_node_state.drain_status = self.get_node_drain_status(node_id).await;
+                Ok(copy_node_state)
+            }
+        }
+    }
+
+    // deduce the drain state from the presence of drain labels and nexus instances
+    async fn get_node_drain_status(&self, node_id: &NodeId) -> DrainStatus {
```
tiagolobocastro, line 72: Probably should return a Result<> to handle the unwraps below

control-plane/agents/core/src/node/registry.rs
```diff
@@ -57,8 +60,27 @@ impl Registry {
...
+        let cordoned_for_drain = self.specs().get_node(node_id).unwrap().cordoned_for_drain();
+        let mut drain_status = DrainStatus::NotDraining;
+        if cordoned_for_drain {
+            let nexuses = self.get_node_nexuses(node_id).await;
```
tiagolobocastro, line 76: hmm should we be inferring the drain status from the nexus specs instead?

control-plane/agents/core/src/node/specs.rs
```diff
@@ -140,4 +141,29 @@ impl ResourceSpecsLocked {
...
+            if locked_node.has_cordon_label(label.clone()) {
+                return Err(SvcError::CordonLabel {
+                    node_id: node_id.to_string(),
+                    label,
+                });
+            }
+            locked_node.drain(label);
+            locked_node.clone()
+        };
+        registry.store_obj(&drained_node_spec).await?;
```
tiagolobocastro, line 166: now that we're modifying the node spec for cordon and drain we should probably impl the SpecOperations trait for the Node, and update the spec through that interface.
This should be a separate task though.

control-plane/grpc/proto/v1/node/node.proto
```diff
@@ -32,7 +38,8 @@ message NodeSpec {
   common.StringMapValue labels = 3;
   // Cordon information.
   NodeCordon cordon = 4;
-
+  // Drain information.
+  NodeDrain drain = 5;
```
tiagolobocastro, line 42: Interesting that we've laid out the cordon and drain in separate "containers" at the grpc layer, but we've added them to the parent type directly on the openapi, I wonder if the openapi should do the same, wdyt?

control-plane/grpc/proto/v1/node/node.proto
```diff
@@ -61,6 +70,16 @@ enum NodeStatus {
   Offline = 2;
 }
 
+// Status of the Node
+enum DrainStatus {
+    // Node has no drain labels
+    NotDraining = 0;
+    // Node has at least 1 drain label and at least 1 nexus
```
tiagolobocastro, line 77: In future releases, we'll do full drain, at which point the semantics of this variable will change.
I wonder what we can do about this, maybe having different drain types?

control-plane/plugin/src/bin/rest-plugin/main.rs
```diff
@@ -62,10 +78,18 @@ async fn execute(cli_args: CliArgs) {
             }
             GetResources::Pools => pool::Pools::list(&cli_args.output).await,
             GetResources::Pool { id } => pool::Pool::get(id, &cli_args.output).await,
-            GetResources::Nodes => node::Nodes::list(&cli_args.output).await,
+            GetResources::Nodes(args) => {
+                if args.show_drain() {
```
tiagolobocastro, line 82: would be good to move this logic into node::Nodes somehow

common/src/types/v0/store/node.rs
chriswldenyer: kubectl-plugin get node <node id> --show-drain
 lists the drain labels

    kubectl-plugin get node <node id> --show-cordon-labels
lists both sets of labels in a single comma-separated list

```
./kubectl-plugin/bin/kubectl-mayastor get node node-1-10196 --show-cordon-labels
 ID            GRPC ENDPOINT         STATUS  CORDONED  CORDON LABELS 
 node-1-10196  157.90.227.228:10124  Online  true      cordon1, drain1 
```
```
./kubectl-plugin/bin/kubectl-mayastor get node node-1-10196 --show-drain
 ID            GRPC ENDPOINT         STATUS  CORDONED  DRAIN STATE  DRAIN LABELS 
 node-1-10196  157.90.227.228:10124  Online  true      Drained      drain1
 ```
 Cancelling a drain is done with the uncordon command

control-plane/agents/core/src/node/registry.rs
chriswldenyer: You're right. Fixing.

common/src/types/v0/transport/node.rs
chriswldenyer: seems to work :+1:

control-plane/agents/core/src/node/registry.rs
chriswldenyer: fixed

control-plane/agents/core/src/node/registry.rs
chriswldenyer: Added error handling to this and to get_node_states() as a consequence.

control-plane/plugin/src/bin/rest-plugin/main.rs
chriswldenyer: Agreed, I'll see what I can do.

control-plane/grpc/proto/v1/node/node.proto
chriswldenyer: TBH I was following Paul's pattern, not knowing any better.
In the interest of simplicity I'd prefer to go the other way and have NodeCordon and NodeDrain as repeated strings in grpc, if we can do that.

PR #289, opened 2022-08-05
chore: add worker thread knobs to the api-rest service

tiagolobocastro:
By default set the max number of worker threads is set to 8.
This will unblock running on servers with large numbers of cpus and using more memory
than what our defaults in the helm charts are prepared to dish out.

Adds 2 new args to the api-rest, workers and max workers.
By default workers is the number of physical cpus and the default max is 8.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2988,6 +2988,7 @@ dependencies = [
  "http",
  "humantime",
  "jsonwebtoken",
+ "num_cpus",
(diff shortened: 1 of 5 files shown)


blaisedias (approved):
Looks reasonable.
Would be nicer if it were configurable at install time.
But we can do that later.

tiagolobocastro:
bors merge

PR #290, opened 2022-08-05
feat: drain alternative implementation

chriswldenyer:
Signed-off-by: chriswldenyer <christopher.denyer@datacore.com>

Add APIs to drain a node.
Drain is not yet fully implemented - what is missing is the reconciler
that moves nexuses to another node, which is dependent on HA being
implemented.
With this PR, drain labels are now stored in the core agent and perform
the same cordoning function as cordon labels but also flag that a drain
is to be undertaken.
The Node object returned from the core-agent now includes a boolean
is_draining field.
This is set via the drain api command but will be cleared by the
reconciler, when implemented.

    kubectl-plugin drain node
and cancels it via:
    kubectl-plugin uncordon node

Two new get objects cordon and drain are implemented for the plugin.

    get cordon <node id> # lists cordon information concerning the node
    get cordons          # lists cordon information for all cordoned nodes
    get drain <node id>  # lists drain information concerning the node
    get drains           # as above but for all nodes with drain labels

Note: the get node <node id> --show-cordon-labels option has been removed.

### DIFF
--- a/common/src/types/v0/store/node.rs
+++ b/common/src/types/v0/store/node.rs
@@ -37,7 +37,14 @@ pub struct NodeSpec {
     /// Cordon labels.
     #[serde(default)] // Ensure backwards compatibility in etcd when upgrading.
     cordon_labels: Vec<String>,
+    /// Drain labels.
+    #[serde(default)] // Ensure backwards compatibility in etcd when upgrading.
+    drain_labels: Vec<String>,
+    /// Drain state
+    #[serde(default)] // Ensure backwards compatibility in etcd when upgrading.
+    is_draining: bool,
 }
+
 impl NodeSpec {
     /// Return a new `Self`
     pub fn new(
@@ -45,12 +52,16 @@ impl NodeSpec {
         endpoint: String,
         labels: NodeLabels,
         cordon_label: Option<Vec<String>>,
+        drain_label: Option<Vec<String>>,
+        is_draining: Option<bool>,
     ) -> Self {
         Self {
             id,
             endpoint,
             labels,
             cordon_labels: cordon_label.unwrap_or_default(),
+            drain_labels: drain_label.unwrap_or_default(),
+            is_draining: is_draining.unwrap_or_default(),
         }
     }
     /// Node identification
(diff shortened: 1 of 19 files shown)


common/src/types/v0/store/node.rs
```diff
@@ -37,20 +37,31 @@ pub struct NodeSpec {
...
+
 impl NodeSpec {
     /// Return a new `Self`
     pub fn new(
         id: NodeId,
         endpoint: String,
         labels: NodeLabels,
         cordon_label: Option<Vec<String>>,
+        drain_label: Option<Vec<String>>,
+        is_draining: Option<bool>,
```
Abhinandan-Purkait, line 56: Why is this field needed? It seems like the will be true as long as there is a drain label, so can we not use that directly?

control-plane/plugin/src/operations.rs
```diff
@@ -64,5 +80,4 @@ pub trait Cordoning {
     type ID;
     async fn cordon(id: &Self::ID, label: &str, output: &utils::OutputFormat);
     async fn uncordon(id: &Self::ID, label: &str, output: &utils::OutputFormat);
-    async fn get_labels(id: &Self::ID, output: &utils::OutputFormat);
```
Abhinandan-Purkait, line 67: Why do we want to remove this? Can you please update the [README](https://github.com/openebs/mayastor-control-plane/blob/develop/k8s/plugin/README.md) with the examples and usages.

Abhinandan-Purkait (approved):

common/src/types/v0/store/node.rs
chriswldenyer: is_draining is needed to distinguish between a node which is draining and one in which the drain is completed.
Both will still have at least one drain label.
When the reconciler is implemented, it will set the is_draining flag to false.

Abhinandan-Purkait: Okay. Thanks

control-plane/plugin/src/operations.rs
chriswldenyer: After talking with Tiago, it seemed better to implement a new "get cordon" command in the kubectl plugin to replace the --get-cordon-labels flag associated with get node. It makes the code more elegant and more obvious for the user.
I will update the readme in a final PR but I think there may be other changes before the final PR (which is why this is still a draft PR) e.g. Tiago would like to replace cordon_labels + drain_labels + is_draining with an enum variant. I am still looking into that.

Abhinandan-Purkait: Okay.

PR #291, opened 2022-08-08
refactor: cleaning up spec operations api

tiagolobocastro:
refactor: add reconciling traits to the poller worker tasks

Some reconciler operations are common amond various resources types, so it makes sense to create
a common interface for them.

todo: make the operation guard part of the SpecOperations api using the type system, this will
make it clearer, and less error-prone.

---

refactor: make operation guard usage explicit

Enforce use of the operation guard for SpecOperations.
Previously it would have been possible to drop the guard before
continuing or completing an operation.

---

refactor: use resource spec as argument to operations

This is in preparation of the next commit where we'll change the SpecOps trait impl from
Spec to Arc<Mutex<Spec>> and thus make the operation mode implicit, by passing a
ResourceGuard which is either Exclusive or Reconcile and from which we can retrieve
the Arc<Mutex<Spec>> thus simplifying the calls and reducing errors.

---

refactor: split SpecOps into separate traits

GuardedSpecOperations now is implemented only for the Guards.
A reduced SpecOperations is implemented only for the inner Specs.

todo: change all operations to receive the Guard as part of the call - this should
make the api more obvious and also make sure there are no modifications in case
someone forgets to setup the guard at the top level.

### DIFF
--- a/common/src/types/v0/store/mod.rs
+++ b/common/src/types/v0/store/mod.rs
@@ -154,12 +154,17 @@ pub type OperationGuardArc<T> = OperationGuard<Arc<Mutex<T>>>;
 /// It unlocks the sequence lock on drop.
 #[derive(Debug)]
 pub struct OperationGuard<T: OperationSequencer> {
-    locked: Option<(OperationSequenceState, T)>,
+    inner: T,
+    locked: Option<OperationSequenceState>,
 }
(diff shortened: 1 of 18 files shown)


Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

PR #300, opened 2022-08-12
feat(v1 host api): add support for v1 host api calls

Abhinandan-Purkait:
- Restructure `msg_translation` module.
- Add conversion trait for v1 Blockdevice, Filesystem, partition types
- Add v1 blockdevice api call support


control-plane/agents/common/src/msg_translation/mod.rs
```diff
@@ -0,0 +1,5 @@
+/// translate between message bus and v0 gRPC type
```
tiagolobocastro: there's no more message bus right?

tiagolobocastro (approved):

control-plane/agents/common/src/msg_translation/mod.rs
Abhinandan-Purkait: Ahh right, I will change it to `transport` type.

Ab-hishek (approved):

Abhinandan-Purkait:
bors merge

PR #303, opened 2022-08-15
feat(ha-agent): dynamic path failure detection for nvme paths

mtzaurus:
Implemented NVMe path detector which detects failed path and reports
them to HA Cluster agent.
UDEV-based cache keeps in sync cached NVMe paths with system device tree
and updates locally cached device paths in response to device
addition/removal events.

Signed-off-by: Mikhail Tcymbaliuk <mtzaurus@gmail.com>

### DIFF
--- a/common/src/transport_api/mod.rs
+++ b/common/src/transport_api/mod.rs
@@ -264,6 +264,15 @@ impl ReplyError {
             extra: "".to_string(),
         }
     }
+    /// for errors that represent unimplemented functionality.
+    pub fn unimplemented(msg: String) -> Self {
+        Self {
+            kind: ReplyErrorKind::Unimplemented,
+            resource: ResourceKind::Unknown,
+            source: "Test Library".to_string(),
+            extra: msg,
+        }
+    }
 }
 
 impl std::fmt::Display for ReplyError {
--- a/common/src/types/v0/transport/cluster_agent.rs
+++ b/common/src/types/v0/transport/cluster_agent.rs
@@ -26,3 +26,42 @@ impl NodeAgentInfo {
         self.endpoint
     }
 }
+
+/// Failed NVMe path.
+#[derive(Debug, Clone)]
+pub struct FailedPath {
+    target_nqn: String,
+}
+
+impl FailedPath {
+    /// Create a new instance of FailedPath for a given NVMe target NQN.
+    pub fn new(target_nqn: String) -> Self {
+        Self { target_nqn }
+    }
+
+    /// Get target NQN.
+    pub fn target_nqn(&self) -> &str {
+        &self.target_nqn
+    }
+}
+
+/// Report failed NVMe paths.
+#[derive(Debug)]
+pub struct ReportFailedPaths {
+    node: String,
+    failed_paths: Vec<FailedPath>,
+}
+
+impl ReportFailedPaths {
+    pub fn new(node: String, failed_paths: Vec<FailedPath>) -> Self {
+        Self { node, failed_paths }
+    }
+
+    pub fn node_name(&self) -> &str {
+        &self.node
+    }
+
+    pub fn failed_paths(&self) -> &Vec<FailedPath> {
+        &self.failed_paths
+    }
+}
--- a/control-plane/agents/ha-node/src/detector.rs
+++ b/control-plane/agents/ha-node/src/detector.rs
@@ -1,0 +1,208 @@
+use crate::{
+    path_provider::{CachedNvmePathProvider, NvmePathNameCollection},
+    reporter::PathReporter,
+    Cli,
+};
+use nvmeadm::nvmf_subsystem::Subsystem;
+use std::{collections::HashMap, rc::Rc};
+use tokio::time::{sleep, Duration};
+
+/// Possible states of every path record.
+#[derive(Debug, Clone)]
+enum PathState {
+    Good,
+    Suspected,
+    Failed,
+}
+
+impl std::fmt::Display for PathState {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> Result<(), std::fmt::Error> {
+        write!(f, "{:?}", *self)
+    }
+}
+
+/// Object that represents a broken/suspected NVMe path.
+#[derive(Debug, Clone)]
+struct PathRecord {
+    /// Used to allow detection of outdated records.
+    epoch: u64,
+    nqn: String,
+    state: PathState,
+    reporter: Rc<PathReporter>,
+}
+
+impl PathRecord {
+    fn new(nqn: String, epoch: u64, reporter: Rc<PathReporter>) -> Self {
+        Self {
+            nqn,
+            epoch,
+            state: PathState::Good,
+            reporter,
+        }
+    }
+
+    fn get_nqn(&self) -> &str {
+        &self.nqn
+    }
+
+    #[inline]
+    fn get_epoch(&self) -> u64 {
+        self.epoch
+    }
+
+    #[inline]
+    fn set_epoch(&mut self, epoch: u64) {
+        self.epoch = epoch;
+    }
+
+    // Trigger state transition based on 'connecting' state of the underlying NVMe controller.
+    fn report_connecting(&mut self) {
+        match self.state {
+            PathState::Good => {
+                self.state = PathState::Suspected;
+                tracing::info!(
+                    state=%PathState::Suspected,
+                    target=self.nqn,
+                    "Target state transition"
+                );
+            }
+            PathState::Suspected => {
+                self.state = PathState::Failed;
+                tracing::error!(
+                    state=%PathState::Failed,
+                    target=self.nqn,
+                    "Target state transition",
+                );
+                self.reporter.report_failed_path(self.nqn.clone());
+            }
+            PathState::Failed => {} // Multiple failures don't cause any state transitions.
+        }
+    }
+
+    fn report_live(&mut self) {
+        self.state = PathState::Good;
+    }
+}
+
+/// Path failure detector for NVMe paths.
+/// All known NVMe paths on system are periodically checked for liveness and get classified
+/// as:
+///  Good  - path is fully functional.
+///  Suspected - path experiences connectivity problems for the first time.
+///  Failed - path has experienced connectivity problems two times in a row.
+/// Once a path is classified as Failed, it's reported to PathReporter and gets sent to
+/// HA Cluster agent.
+#[derive(Debug)]
+pub struct PathFailureDetector {
+    epoch: u64,
+    detection_period: Duration,
+    suspected_paths: HashMap<String, PathRecord>,
+    reporter: Rc<PathReporter>,
+}
+
+impl PathFailureDetector {
+    pub(crate) fn new(args: &Cli) -> anyhow::Result<Self> {
+        let reporter = PathReporter::new(
+            args.node_name.clone(),
+            *args.retransmission_period,
+            *args.aggregation_period,
+        );
+
+        Ok(Self {
+            epoch: 0,
+            detection_period: *args.detection_period,
+            suspected_paths: HashMap::new(),
+            reporter: Rc::new(reporter),
+        })
+    }
+
+    fn rescan_paths(&mut self, path_collection: &mut NvmePathNameCollection) {
+        // Update epoch before scanning controllers.
+        self.epoch += 1;
+
+        // Scan all reported NVMe paths on system and check for connectivity.
+        for ctrlr in path_collection.get_entries() {
+            match Subsystem::new(ctrlr.path()) {
+                Ok(subsystem) => {
+                    let existing_record = match subsystem.state.as_str() {
+                        "connecting" => {
+                            // Add a new record in case no record exists for target NQN.
+                            if !self.suspected_paths.contains_key(&subsystem.nqn) {
+                                self.suspected_paths.insert(
+                                    subsystem.nqn.clone(),
+                                    PathRecord::new(
+                                        subsystem.nqn.clone(),
+                                        self.epoch,
+                                        self.reporter.clone(),
+                                    ),
+                                );
+                            }
+
+                            let rec = self.suspected_paths.get_mut(&subsystem.nqn).unwrap();
+                            rec.report_connecting();
+                            Some(rec)
+                        }
+                        "live" => self.suspected_paths.get_mut(&subsystem.nqn).map(|rec| {
+                            rec.report_live();
+                            rec
+                        }),
+                        _ => None,
+                    };
+
+                    // Update epoch for the existing record.
+                    if let Some(rec) = existing_record {
+                        rec.set_epoch(self.epoch);
+                    }
+                }
+                Err(e) => {
+                    tracing::error!("Failed to get status for NVMe path: {}", e);
+                }
+            }
+        }
+
+        // Remove all existing records that don't have underlying NVMe controllers:
+        // can happen in case controller was removed after it had been identified as
+        // broken/suspected. Stalled/outdated records have a different (old) epoch number
+        // since they have not been touched during the current iteration.
+        let mut to_remove = vec![];
+
+        for v in self.suspected_paths.values() {
+            if v.get_epoch() != self.epoch {
+                to_remove.push(v.get_nqn().to_owned());
+            }
+        }
+
+        if !to_remove.is_empty() {
+            for k in to_remove {
+                self.suspected_paths.remove(&k);
+                tracing::debug!("Removing stalled path record for NQN {}", k);
+            }
+        }
+    }
+
+    /// Start NVMe path error detection loop.
+    pub async fn start(mut self) -> anyhow::Result<()> {
+        let mut path_provider = CachedNvmePathProvider::new();
+        let mut path_collection = path_provider.get_path_collection().unwrap();
+        let start = path_provider.start()?;
+        tokio::pin!(start);
+
+        tracing::info!(
+            "Starting NVMe path error detection loop, path detection interval: {:?}",
+            self.detection_period,
+        );
+
+        loop {
+            tokio::select! {
+                _ = &mut start => {
+                    tracing::warn!("NVMe path provider completed, stopping error detection");
+                    break;
+                },
+                _ = sleep(self.detection_period) => self.rescan_paths(&mut path_collection),
+            }
+        }
+
+        tracing::info!("Stopping NVMe path error detection loop");
+        Ok(())
+    }
+}
--- a/control-plane/agents/ha-node/src/main.rs
+++ b/control-plane/agents/ha-node/src/main.rs
@@ -2,15 +2,24 @@ use common_lib::types::v0::transport::cluster_agent::NodeAgentInfo;
 use grpc::operations::ha_node::{client::ClusterAgentClient, traits::ClusterAgentOperations};
 use http::Uri;
 use once_cell::sync::OnceCell;
+use opentelemetry::KeyValue;
 use structopt::StructOpt;
 use utils::{
     package_description, version_info_str, DEFAULT_CLUSTER_AGENT_CLIENT_ADDR,
-    DEFAULT_NODE_AGENT_SERVER_ADDR,
+    DEFAULT_NODE_AGENT_SERVER_ADDR, NVME_PATH_AGGREGATION_PERIOD, NVME_PATH_CHECK_PERIOD,
+    NVME_PATH_RETRANSMISSION_PERIOD,
 };
 
+mod detector;
+mod path_provider;
+mod reporter;
+
+use detector::PathFailureDetector;
+
+/// TODO
 #[derive(Debug, StructOpt)]
 #[structopt(name = package_description!(), version = version_info_str!())]
-struct Cli {
+pub(crate) struct Cli {
     /// HA Cluster Agent URL or address to connect to the services.
     #[structopt(long, short, default_value = DEFAULT_CLUSTER_AGENT_CLIENT_ADDR)]
     cluster_agent: Uri,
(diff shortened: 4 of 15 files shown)


mtzaurus:
bors try

common/src/types/v0/transport/cluster_agent.rs
```diff
@@ -26,3 +26,40 @@ impl NodeAgentInfo {
...
+    pub fn new(node: String, failed_paths: Vec<FailedPath>) -> Self {
+        Self { node, failed_paths }
+    }
+
+    pub fn node_name(&self) -> &str {
+        &self.node
+    }
+
+    pub fn failed_paths(&self) -> Vec<FailedPath> {
+        self.failed_paths.clone()
```
dsavitskiy: Why cloning Vec here? You could return &[FailedPath].

control-plane/agents/ha-node/src/detector.rs
```diff
@@ -0,0 +1,193 @@
...
+}
+
+/// Object that represents a broken/suspected NVMe path.
+#[derive(Debug, Clone)]
+struct PathRecord {
+    /// Used to allow detection of outdated records.
+    epoch: u64,
+    nqn: String,
+    state: PathState,
+    reporter: Arc<PathReporter>,
```
dsavitskiy: Why Arc? If it is supposed to be shared across thread, shouldn't PathReporter be protected by Mutex?

control-plane/agents/ha-node/src/detector.rs
```diff
@@ -0,0 +1,193 @@
...
+        self.state = PathState::Good;
+    }
+}
+
+#[derive(Debug)]
+pub struct PathFailureDetector {
+    epoch: u64,
+    detection_period: Duration,
+    suspected_paths: HashMap<String, PathRecord>,
+    reporter: Arc<PathReporter>,
```
dsavitskiy: Why Arc? If it is supposed to be shared across thread, shouldn't PathReporter be protected by Mutex?

control-plane/agents/ha-node/src/path_provider.rs
```diff
@@ -0,0 +1,230 @@
...
+/// and signalling cache modification in case there are suitable UDEV events related
+/// to NVMe subsystem.
+/// Frontend: (NvmePathNameCollection), which represents a cache for all known NVMe
+/// paths in the system. Note: we only cache paths that represent NVMe targets with product's
+/// In order to eliminate cache inconsistency, there must be only one front-end path collection.
+/// NQN, so cache acts as a filter for 'unknown' NQNs.
+/// To eliminate locking/contention, backend communicates with frontend via a lockless queue.
+#[derive(Debug)]
+pub struct CachedNvmePathProvider {
+    udev_queue: Arc<SegQueue<CacheOp>>,
```
dsavitskiy, line 71: Shouldn't SegQueue be protected by Mutex?

dsavitskiy (approved):

mtzaurus:
bors try

control-plane/agents/ha-node/src/path_provider.rs
mtzaurus: It's a lockless queue which doesn't require locks.

common/src/transport_api/mod.rs
```diff
@@ -264,6 +264,15 @@ impl ReplyError {
             extra: "".to_string(),
         }
     }
+    /// for errors that represent unimplemented functionality.
+    pub fn unimplemented(msg: String) -> Self {
+        Self {
+            kind: ReplyErrorKind::Unimplemented,
+            resource: ResourceKind::Unknown,
+            source: "Test Library".to_string(),
```
tiagolobocastro, line 272: can we pass the source as an argument here?

common/src/types/v0/transport/cluster_agent.rs
```diff
@@ -26,3 +26,40 @@ impl NodeAgentInfo {
...
 }
+
+/// Failed NVMe path.
+#[derive(Debug, Clone)]
+pub struct FailedPath {
+    target_nqn: String,
+}
+
+impl FailedPath {
+    pub fn new(target_nqn: String) -> Self {
```
tiagolobocastro, line 38: missing doc comments

control-plane/agents/ha-node/src/detector.rs
mtzaurus: Mutex is not required, as PathReporter does not require mutable references to work properly.
However, we can replace Arc with Rc (since one PathReporter is shared between multiple instances of PathRecord), as we don't need Send semantics since no async context exists.

control-plane/agents/ha-node/src/detector.rs
mtzaurus: Explained in my comment below.

control-plane/agents/ha-node/src/detector.rs
```diff
@@ -0,0 +1,193 @@
...
+    fn set_epoch(&mut self, epoch: u64) {
+        self.epoch = epoch;
+    }
+
+    // Trigger state transition based on 'connecting' state of the underlying NVMe controller.
+    fn report_connecting(&mut self) {
+        match self.state {
+            PathState::Good => {
+                self.state = PathState::Suspected;
+                info!("Target {} transitioned into SUSPECTED state", self.nqn);
```
tiagolobocastro:
```suggestion
                info!(state=PathState::Suspected, target.nqn=self.nqn, "Target state transition", self.nqn);
```
might need to impl ToString for PathState

control-plane/agents/ha-node/src/detector.rs
```diff
@@ -0,0 +1,193 @@
...
+            }
+            PathState::Suspected => {
+                self.state = PathState::Failed;
+                error!(
```
tiagolobocastro: please change this in the style of the comment above.
This will make it a lot nicer to use with Jaeger

control-plane/agents/ha-node/src/detector.rs
```diff
@@ -0,0 +1,193 @@
...
+#[derive(Debug)]
+pub struct PathFailureDetector {
+    epoch: u64,
+    detection_period: Duration,
+    suspected_paths: HashMap<String, PathRecord>,
+    reporter: Arc<PathReporter>,
+}
+
+impl PathFailureDetector {
+    pub async fn new(args: &Cli) -> anyhow::Result<Self> {
```
tiagolobocastro: nit: doc comment

control-plane/agents/ha-node/src/detector.rs
```diff
@@ -0,0 +1,193 @@
...
+                    let existing_record = match subsystem.state.as_str() {
+                        "connecting" => {
+                            // Add a new record in case no record exists for target NQN.
+                            if !self.suspected_paths.contains_key(&subsystem.nqn) {
+                                self.suspected_paths.insert(
+                                    subsystem.nqn.clone(),
+                                    PathRecord::new(
+                                        subsystem.nqn.clone(),
+                                        self.epoch,
+                                        Arc::clone(&self.reporter),
```
tiagolobocastro: can simply be self.reporter.clone()

control-plane/agents/ha-node/src/detector.rs
```diff
@@ -0,0 +1,193 @@
...
+        self.epoch += 1;
+
+        // Scan all reported NVMe paths on system and check for connectivity.
+        for ctrlr in path_collection.get_entries() {
+            match Subsystem::new(ctrlr.path()) {
+                Ok(subsystem) => {
+                    let existing_record = match subsystem.state.as_str() {
+                        "connecting" => {
+                            // Add a new record in case no record exists for target NQN.
+                            if !self.suspected_paths.contains_key(&subsystem.nqn) {
```
tiagolobocastro, line 130: nit: you can make use of `HashMap::get_mut` here to avoid having to do another get with unwrap on L123

control-plane/agents/ha-node/src/detector.rs
```diff
@@ -0,0 +1,193 @@
...
+    fn rescan_paths(&mut self, path_collection: &mut NvmePathNameCollection) {
+        // Update epoch before scanning controllers.
+        self.epoch += 1;
+
+        // Scan all reported NVMe paths on system and check for connectivity.
+        for ctrlr in path_collection.get_entries() {
+            match Subsystem::new(ctrlr.path()) {
+                Ok(subsystem) => {
+                    let existing_record = match subsystem.state.as_str() {
+                        "connecting" => {
```
tiagolobocastro, line 128: Is state an enum? if so can we match on the enum variant instead of str?

control-plane/agents/ha-node/src/detector.rs
```diff
@@ -0,0 +1,193 @@
...
+                        rec.set_epoch(self.epoch);
+                    }
+                }
+                Err(e) => {
+                    error!("Failed to get status for NVMe path: {}", e);
+                }
+            }
+        }
+
+        // Remove all existing records that don't have underlaying NVMe controllers:
```
tiagolobocastro: did you mean underlying?

control-plane/agents/ha-node/src/detector.rs
```diff
@@ -0,0 +1,193 @@
...
+        if !to_remove.is_empty() {
+            for k in to_remove {
+                self.suspected_paths.remove(&k);
+                debug!("Removing stalled path record for NQN {}", k);
+            }
+        }
+    }
+
+    /// Start NVMe path error detection loop.
+    pub async fn start(mut self) -> anyhow::Result<()> {
```
tiagolobocastro, line 184: nit: should this be run? since it's running forever here it seems

control-plane/agents/ha-node/src/main.rs
```diff
@@ -2,15 +2,27 @@ use common_lib::types::v0::transport::cluster_agent::NodeAgentInfo;
...
 use structopt::StructOpt;
 use utils::{
     package_description, version_info_str, DEFAULT_CLUSTER_AGENT_CLIENT_ADDR,
-    DEFAULT_NODE_AGENT_SERVER_ADDR,
+    DEFAULT_NODE_AGENT_SERVER_ADDR, NVME_PATH_AGGREGATION_PERIOD, NVME_PATH_CHECK_PERIOD,
+    NVME_PATH_RETRANSMISSION_PERIOD,
 };
 
+#[macro_use]
+extern crate tracing;
```
tiagolobocastro: this is not needed right?

control-plane/agents/ha-node/src/main.rs
```diff
@@ -2,15 +2,27 @@ use common_lib::types::v0::transport::cluster_agent::NodeAgentInfo;
...
+mod path_provider;
+mod reporter;
+
+use detector::PathFailureDetector;
+
+/// TODO
 #[derive(Debug, StructOpt)]
 #[structopt(name = package_description!(), version = version_info_str!())]
-struct Cli {
+pub struct Cli {
```
tiagolobocastro: should be pub(crate) ?

control-plane/agents/ha-node/src/main.rs
```diff
@@ -40,23 +72,47 @@ impl Cli {
...
+    );
 
     CLUSTER_AGENT_CLIENT
-        .set(ClusterAgentClient::new(cli.cluster_agent, None).await)
+        .set(ClusterAgentClient::new(cli_args.cluster_agent.clone(), None).await)
         .ok()
         .expect("Expect to be initialized only once");
 
-    cluster_agent_client()
+    if let Err(e) = cluster_agent_client()
```
tiagolobocastro, line 87: This component only registers itself on start?

control-plane/agents/ha-node/src/path_provider.rs
```diff
@@ -0,0 +1,230 @@
...
+};
+use tokio_udev::{AsyncMonitorSocket, EventType, MonitorBuilder};
+use utils::NVME_TARGET_NQN_PREFIX;
+
+const SYSFS_PREFIX: &str = "/sys/devices/virtual/nvme-fabrics/ctl/";
+/// Maximum number of events to process upon cache update (to prevent event flooding).
+const CACHE_EVENTS_BATCH_SIZE: usize = 128;
+
+#[derive(Debug, Clone)]
+pub struct NvmePath {
```
tiagolobocastro, line 19: missing doc comments

control-plane/agents/ha-node/src/path_provider.rs
```diff
@@ -0,0 +1,230 @@
...
+enum CacheOp {
+    Add(String),
+    Remove(String),
+    InvalidateCache,
+}
+
+/// UDEV-based NVMe path provider that uses cache to keep all known NVMe targets
+/// in memory and avoid wildcard Sysfs enumeration when there is a need to get
+/// a list of all existing NVMe path in the system.
+/// This name provider sonsists of 2 parts: backend and frontend.
```
tiagolobocastro: consists

control-plane/agents/ha-node/src/path_provider.rs
```diff
@@ -0,0 +1,230 @@
...
+/// Backend: (CachedNvmePathProvider itself) is responsible for monitoring UDEV events
+/// and signalling cache modification in case there are suitable UDEV events related
+/// to NVMe subsystem.
+/// Frontend: (NvmePathNameCollection), which represents a cache for all known NVMe
+/// paths in the system. Note: we only cache paths that represent NVMe targets with product's
+/// In order to eliminate cache inconsistency, there must be only one front-end path collection.
+/// NQN, so cache acts as a filter for 'unknown' NQNs.
```
tiagolobocastro, line 66: is this in the wrong line, it fine after 54 "with product's" :)

control-plane/agents/ha-node/src/path_provider.rs
```diff
@@ -0,0 +1,230 @@
...
+/// To eliminate locking/contention, backend communicates with frontend via a lockless queue.
+#[derive(Debug)]
+pub struct CachedNvmePathProvider {
+    udev_queue: Arc<SegQueue<CacheOp>>,
+    frontend: Option<NvmePathNameCollection>,
+}
+
+impl CachedNvmePathProvider {
+    /// Start UDEV monitoring loop.
+    pub async fn start(&mut self) {
```
tiagolobocastro: Can we please have start return a Result so we can use `?` below?

control-plane/agents/ha-node/src/path_provider.rs
```diff
@@ -0,0 +1,230 @@
...
+            .await;
+    }
+
+    /// Get a front-end path collection for this name provider.
+    pub fn get_path_collection(&mut self) -> Option<NvmePathNameCollection> {
+        self.frontend.take()
+    }
+
+    /// Create a new cached name provider.
+    pub async fn new() -> anyhow::Result<Self> {
```
tiagolobocastro: we don't seem to ever return an error here?

control-plane/agents/ha-node/src/reporter.rs
```diff
@@ -0,0 +1,185 @@
...
+                    }
+                });
+            }
+        });
+    }
+
+    pub fn report_failed_path(&self, nqn: String) {
+        self.channel
+            .send(nqn)
+            .expect("Reporter channel disappeared");
```
tiagolobocastro, line 95: should this error be handled?

control-plane/agents/ha-node/src/reporter.rs
```diff
@@ -0,0 +1,185 @@
...
+            loop {
+                // Phase 1: wait for the first path to trigger batch aggregation.
+                let p = path_receiver.recv().await.expect("Path sender disappeared");
+
+                // Phase 2: add all subsequent reported paths to the batch.
+                let mut batch = PathBatch::new();
+                batch.add_path(p);
+
+                let timeout = sleep(aggregation_period);
+                tokio::pin!(timeout);
```
tiagolobocastro: why do we need this pin, we can just add the tokio timeout to the select?
https://github.com/openebs/mayastor-control-plane/blob/3fa7845457d3cfb114a078501a7c0e08105b36ac/control-plane/agents/core/src/core/reconciler/poller.rs#L87

control-plane/grpc/src/operations/ha_node/traits.rs
```diff
@@ -36,3 +45,39 @@ impl NodeInfo for HaNodeInfo {
         self.endpoint.clone()
     }
 }
+
+/// Trait to be implemented for ReportFailedNvmePaths operation.
+pub trait ReportFailedPathsInfo: Send + Sync + std::fmt::Debug {
+    /// Id of the io-engine instance
```
tiagolobocastro: hmm shouldn't this be the "application" node?

control-plane/grpc/src/operations/ha_node/traits.rs
```diff
@@ -36,3 +45,39 @@ impl NodeInfo for HaNodeInfo {
...
+    fn failed_paths(&self) -> Vec<FailedPath> {
+        self.failed_paths()
+    }
+}
+
+impl From<&dyn ReportFailedPathsInfo> for ReportFailedNvmePathsRequest {
+    fn from(info: &dyn ReportFailedPathsInfo) -> Self {
+        Self {
+            nodename: info.node(),
+            failed_paths: info.failed_paths().into_iter().map(|p| p.into()).collect(),
```
tiagolobocastro:
```suggestion
            failed_paths: info.failed_paths().into_vec(),
```

utils/utils-lib/src/constants.rs
```diff
@@ -83,3 +83,14 @@ pub const DEFAULT_REST_MAX_WORKER_THREADS: &str = "8";
...
+pub const NVME_TARGET_NQN_PREFIX: &str = "nqn.2019-05.io.openebs:";
+
+/// NVMe path check period.
+pub const NVME_PATH_CHECK_PERIOD: &str = "3s";
+
+/// The default retransmission interval for reporting failed paths in case of network issues.
+pub const NVME_PATH_RETRANSMISSION_PERIOD: &str = "3s";
+
+/// Period for aggregating multiple failed paths before reporting them.
+pub const NVME_PATH_AGGREGATION_PERIOD: &str = "1s";
```
tiagolobocastro: 3 seconds seems quite often, was this intentional?

mtzaurus: What do you suggest ? 5 secs ? 10 secs ?

control-plane/grpc/src/operations/ha_node/traits.rs
mtzaurus: Good point, yes.

control-plane/grpc/src/operations/ha_node/traits.rs
mtzaurus: In case id id nodename, it's OK. But I can fix the comment, not a problem.

control-plane/agents/ha-node/src/reporter.rs
mtzaurus: Agreed.

control-plane/agents/ha-node/src/reporter.rs
mtzaurus: Currently there is no possibility for the listener to disappear (it's eternal loop), so this is just to properly consume Result.

control-plane/agents/ha-node/src/detector.rs
mtzaurus: This version looks nicer, agreed.

common/src/transport_api/mod.rs
mtzaurus: Hmmm....
What's the semantics for 'source' for something that it's not implemented ?

utils/utils-lib/src/constants.rs
tiagolobocastro: I don't know :D My punt would be 10s, but I don't really know what the best thing to do here is, @GlennBullingham any suggestion?

mtzaurus: I will change it to 10 seconds for now, we can adjust the value later, if needed.

tiagolobocastro: And it's a cmdline argument anyway so we can adjust on the fly if needed which is nice.

tiagolobocastro: Ah that's interesting, if this agent restarts, can the path ever disappear completely from the node while the agent is restarting? And in such case, the agent will not report the path as broken?

control-plane/agents/ha-node/src/path_provider.rs
mtzaurus: Can just return Self, yes.

control-plane/agents/ha-node/src/path_provider.rs
mtzaurus: OK, we can use Result instead of just panicking.

control-plane/agents/ha-node/src/path_provider.rs
mtzaurus: Good catch, thanks.

control-plane/agents/ha-node/src/path_provider.rs
mtzaurus: Nasty typo. :-)

control-plane/agents/ha-node/src/path_provider.rs
mtzaurus: Doc was placed close to 'impl', so I move it to NvmePath structure.

control-plane/agents/ha-node/src/main.rs
mtzaurus: registration logic will change for sure (more to 'heartbeat' semantics).
For now it's just to avoid panics upon agent start in case cluster agent is not running.

control-plane/agents/ha-node/src/main.rs
mtzaurus: Can be.

control-plane/agents/ha-node/src/main.rs
mtzaurus: OK.

control-plane/agents/ha-node/src/detector.rs
mtzaurus: Indeed. :-)

control-plane/agents/ha-node/src/detector.rs
mtzaurus: Unfortunately, it's a string literal in our 'nvmeadm' library.

control-plane/agents/ha-node/src/detector.rs
mtzaurus: Yup, added comments for PathFailureDetector structure to clarify how failure detection works.

control-plane/agents/ha-node/src/detector.rs
mtzaurus: Done.

common/src/types/v0/transport/cluster_agent.rs
mtzaurus: Added function comments.

mtzaurus:
bors try

control-plane/agents/ha-node/src/detector.rs
```diff
@@ -0,0 +1,212 @@
...
+        let mut path_provider = CachedNvmePathProvider::new();
+        let mut path_collection = path_provider.get_path_collection().unwrap();
+        let start = path_provider.start()?;
+        tokio::pin!(start);
```
tiagolobocastro, line 188: is this required?

control-plane/agents/ha-node/src/detector.rs
```diff
@@ -0,0 +1,212 @@
...
+
+        tracing::info!(
+            "Starting NVMe path error detection loop, path detection interval: {:?}",
+            self.detection_period,
+        );
+
+        loop {
+            let timeout = sleep(self.detection_period).fuse();
+            tokio::pin!(timeout);
```
tiagolobocastro: can use the timeout directly?

control-plane/agents/ha-node/src/reporter.rs
tiagolobocastro: What about on shutdown? When we start dropping things?
Ah, are we missing a graceful shutdown?

tiagolobocastro (approved):

common/src/types/v0/transport/cluster_agent.rs
mtzaurus: Yup, indeed.

control-plane/agents/ha-node/src/detector.rs
mtzaurus: Yup.

mtzaurus:
bors merge

PR #304, opened 2022-08-16
chore: seperate AgentToEngine trait for conversion for v0,v1 apis

Abhinandan-Purkait:

### DIFF
--- a/control-plane/agents/common/src/msg_translation/v0.rs
+++ b/control-plane/agents/common/src/msg_translation/v0.rs
@@ -11,9 +11,17 @@ use common_lib::{
     },
 };
 
-use crate::msg_translation::{AgentToIoEngine, IoEngineToAgent, TryIoEngineToAgent};
+use crate::msg_translation::{IoEngineToAgent, TryIoEngineToAgent};
 use rpc::io_engine as v0_rpc;
 use std::convert::TryFrom;
+
+/// Trait for converting agent messages to io-engine messages.
+pub trait AgentToIoEngine {
+    /// RpcIoEngine message type.
+    type IoEngineMessage;
+    /// Conversion of agent message to io-engine message.
+    fn to_rpc(&self) -> Self::IoEngineMessage;
+}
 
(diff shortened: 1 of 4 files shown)


Ab-hishek (approved):

control-plane/agents/common/src/msg_translation/v0.rs
```diff
@@ -11,10 +11,18 @@ use common_lib::{
     },
 };
 
-use crate::msg_translation::{AgentToIoEngine, IoEngineToAgent, TryIoEngineToAgent};
+use crate::msg_translation::{IoEngineToAgent, TryIoEngineToAgent};
 use rpc::io_engine as v0_rpc;
 use std::convert::TryFrom;
 
+/// Trait for converting agent messages to io-engine messages.
```
tiagolobocastro, line 18: I don't understand this move, why do we need separate traits, which are essentially the same?

Abhinandan-Purkait: Since `AgentToIoEngine` has the `IoEngineMessage` type embeded in it, the conversion to v0 is in conflict with v1, because the trait cannot be implemented on the same transport type twice.

tiagolobocastro: Ah right... Could instead have?
```rust
/// Trait for converting agent messages to io-engine messages.
pub trait AgentToIoEngine<IoEngineMessage> {
    /// Conversion of agent message to io-engine message.
    fn to_rpc(&self) -> IoEngineMessage;
}
```
But that would require a lot of changes, so yeah let's keep it the way you have it now

tiagolobocastro (approved):

control-plane/agents/common/src/msg_translation/v0.rs
Abhinandan-Purkait: Yeah, but yeah, this would be a good refactor. Will take it up later, as this is blocking @Ab-hishek for now.

Abhinandan-Purkait:
bors merge

PR #305, opened 2022-08-16
feat: drain initial implementation

chriswldenyer:
Signed-off-by: chriswldenyer <christopher.denyer@datacore.com>

Add APIs to drain a node.
Drain is not yet fully implemented - what is missing is the reconciler
that moves nexuses to another node, which is dependent on HA being
implemented.
With this PR, drain labels are now stored in the core agent and perform
the same cordoning function as cordon labels but also flag that a drain
is to be undertaken.
The Node object within the core-agent now includes a DrainState
enum variant which contains both cordon labels and drain labels.
This is controlled via the REST API but also by the reconciler,
when implemented.

    kubectl-plugin drain node <node name> <drain label>
and cancels it via:
    kubectl-plugin uncordon node <node name> <drain label>

Two new get objects cordon and drain are implemented for the plugin.

    get cordon node <node id> # lists cordon information concerning the node
    get cordon nodes          # lists cordon information for all cordoned nodes
    get drain node <node id>  # lists drain information concerning the node
    get drain nodes           # as above but for all nodes with drain labels

Note: the "get node <node id> --show-cordon-labels" variant has been removed.

### DIFF
--- a/common/src/types/v0/store/node.rs
+++ b/common/src/types/v0/store/node.rs
@@ -1,95 +1,324 @@
 //! Definition of node types that can be saved to the persistent store.
-
-use crate::types::v0::{
-    openapi::models,
-    store::definitions::{ObjectKey, StorableObject, StorableObjectType},
-    transport::{self, NodeId},
+use crate::{
+    types::v0::{
+        openapi::models,
+        store::definitions::{ObjectKey, StorableObject, StorableObjectType},
+        transport::{self, NodeId},
+    },
+    IntoOption,
 };
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 
+/// Generic labels associated with the node.
 pub type NodeLabels = HashMap<String, String>;
 
+/// Data relating to the cordoning of a node.
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
+pub struct CordonedState {
+    /// Labels used to enforce a node cordon.
+    pub cordonlabels: Vec<String>,
+}
+
+impl CordonedState {
+    /// Create a new CordonedState object.
+    pub fn new(cordonlabels: Vec<String>) -> Self {
+        Self { cordonlabels }
+    }
+    /// Create a DrainState from a CordonedState.
+    pub fn into_drain(&self, drainlabel: &str) -> DrainState {
+        DrainState::new(self.cordonlabels.clone(), vec![String::from(drainlabel)])
+    }
+    /// Remove a cordon label from a CordonedState object.
+    pub fn remove_label(&mut self, label: &str) {
+        self.cordonlabels.retain(|l| l != label)
+    }
+}
+
+/// Data relating to a draining or drained node, including non-drain cordon labels.
+#[derive(Clone, Serialize, Deserialize, Debug, PartialEq)]
+pub struct DrainState {
+    /// Labels used to enforce a node cordon.
+    pub cordonlabels: Vec<String>,
+    /// Labels used to cordon a node specifically for a drain.
+    pub drainlabels: Vec<String>,
+}
+
+impl DrainState {
+    /// Create a new DrainState object.
+    pub fn new(cordonlabels: Vec<String>, drainlabels: Vec<String>) -> Self {
+        Self {
+            cordonlabels,
+            drainlabels,
+        }
+    }
+    /// Remove a label from a DrainState object.
+    pub fn remove_label(&mut self, label: &str) {
+        self.cordonlabels.retain(|l| l != label);
+        self.drainlabels.retain(|l| l != label);
+    }
+    /// Add a drain label to a DrainState object.
+    pub fn add_drain_label(&mut self, label: &str) {
+        self.drainlabels.push(label.to_string());
+    }
+}
+
+/// Enum variant encompassing data related to a cordoned or draining node.
+#[derive(Clone, Serialize, Deserialize, Debug, PartialEq)]
+pub enum CordonDrainState {
+    /// The node is being cordoned.
+    Cordoned(CordonedState),
+    /// The node is being drained.
+    Draining(DrainState),
+    /// The drain has completed.
+    Drained(DrainState),
+}
+
+impl CordonDrainState {
+    /// Add the given label to the cordon labels.
+    pub fn add_cordon_label(&mut self, label: &str) {
+        match self {
+            CordonDrainState::Draining(state) => state.cordonlabels.push(label.to_string()),
+            CordonDrainState::Drained(state) => state.cordonlabels.push(label.to_string()),
+            CordonDrainState::Cordoned(state) => state.cordonlabels.push(label.to_string()),
+        }
+    }
+    /// Returns a new Cordoned enum with the given cordon label.
+    pub fn cordon(label: &str) -> Self {
+        CordonDrainState::Cordoned(CordonedState::new(vec![String::from(label)]))
+    }
+
+    /// Returns whether the state has the specified cordon label.
+    pub fn has_cordon_only_label(&self, label: &str) -> bool {
+        match self {
+            CordonDrainState::Draining(state) => state.cordonlabels.iter().any(|i| i == label),
+            CordonDrainState::Drained(state) => state.cordonlabels.iter().any(|i| i == label),
+            CordonDrainState::Cordoned(state) => state.cordonlabels.iter().any(|i| i == label),
+        }
+    }
+    /// Returns whether the state has the specified drain label.
+    pub fn has_drain_label(&self, label: &str) -> bool {
+        match self {
+            CordonDrainState::Draining(state) => state.drainlabels.iter().any(|i| i == label),
+            CordonDrainState::Drained(state) => state.drainlabels.iter().any(|i| i == label),
+            CordonDrainState::Cordoned(_) => false,
+        }
+    }
+    /// Returns whether the state corresponds to draining.
+    pub fn is_draining(&self) -> bool {
+        match self {
+            CordonDrainState::Cordoned(_) => false,
+            CordonDrainState::Draining(_) => true,
+            CordonDrainState::Drained(_) => false,
+        }
+    }
+}
+
+/// Node information.
 #[derive(Serialize, Deserialize, Debug, PartialEq)]
 pub struct Node {
-    /// Node information.
+    /// Node state information.
     node: transport::NodeState,
     /// Node labels.
     labels: NodeLabels,
 }
 
+/// Node state information.
 pub struct NodeState {
-    /// Node information
+    /// Node information.
     pub node: transport::NodeState,
 }
 
+/// Node spec data, including the cordon/drain state.
 #[derive(Serialize, Deserialize, Debug, PartialEq, Default, Clone)]
 pub struct NodeSpec {
     /// Node identification.
     id: NodeId,
-    /// Endpoint of the io-engine instance (gRPC)
+    /// Endpoint of the io-engine instance (gRPC).
     endpoint: String,
     /// Node labels.
     labels: NodeLabels,
-    /// Cordon labels.
+    /// Cordon/drain state.
+    #[serde(skip_serializing_if = "Option::is_none")]
     #[serde(default)] // Ensure backwards compatibility in etcd when upgrading.
-    cordon_labels: Vec<String>,
+    cordon_drain_state: Option<CordonDrainState>,
 }
 
 impl NodeSpec {
-    /// Return a new `Self`
+    /// Return a new `Self`.
     pub fn new(
         id: NodeId,
         endpoint: String,
         labels: NodeLabels,
-        cordon_label: Option<Vec<String>>,
+        cordon_drain_state: Option<CordonDrainState>,
     ) -> Self {
         Self {
             id,
             endpoint,
             labels,
-            cordon_labels: cordon_label.unwrap_or_default(),
+            cordon_drain_state,
         }
     }
-    /// Node identification
+    /// Node identification.
     pub fn id(&self) -> &NodeId {
         &self.id
     }
-    /// Node gRPC endpoint
+    /// Node gRPC endpoint.
     pub fn endpoint(&self) -> &str {
         &self.endpoint
     }
-    /// Node labels
+    /// Node labels.
     pub fn labels(&self) -> &NodeLabels {
         &self.labels
     }
-    /// Node gRPC endpoint
+    /// Node labels.
+    pub fn cordon_drain_state(&self) -> &Option<CordonDrainState> {
+        &self.cordon_drain_state
+    }
+
+    /// Node gRPC endpoint.
     pub fn set_endpoint(&mut self, endpoint: String) {
         self.endpoint = endpoint
     }
+
     /// Cordon node by applying the label.
     pub fn cordon(&mut self, label: String) {
-        self.cordon_labels.push(label);
+        match &mut self.cordon_drain_state {
+            Some(ds) => ds.add_cordon_label(&label),
+            None => {
+                //add the label and set the state to cordoned
+                self.cordon_drain_state = Some(CordonDrainState::cordon(&label));
+            }
+        }
+    }
+
+    /// Drain node by applying the drain label.
+    pub fn set_drain(&mut self, label: String) {
+        // the the node has the label, return with an error
+        match &mut self.cordon_drain_state {
+            Some(ds) => match ds {
+                CordonDrainState::Cordoned(state) => {
+                    // set state to draining and add label
+                    self.cordon_drain_state =
+                        Some(CordonDrainState::Draining(state.into_drain(&label)))
+                }
+                CordonDrainState::Draining(state) => {
+                    // add the label
+                    state.add_drain_label(&label);
+                }
+                CordonDrainState::Drained(state) => {
+                    // add the label
+                    state.add_drain_label(&label);
+                }
+            },
+            None => {
+                //add the label and set the state to draining
+                let cordonlabels = Vec::<String>::new();
+                let drainlabels = vec![label];
+                self.cordon_drain_state = Some(CordonDrainState::Draining(DrainState::new(
+                    cordonlabels,
+                    drainlabels,
+                )));
+            }
+        }
     }
     /// Uncordon node by removing the corresponding label.
     pub fn uncordon(&mut self, label: String) {
-        if let Some(index) = self.cordon_labels.iter().position(|l| l == &label) {
-            self.cordon_labels.remove(index);
+        match &mut self.cordon_drain_state {
+            Some(ds) => match ds {
+                CordonDrainState::Cordoned(state) => {
+                    // remove the cordon label
+                    // if none left set ds to none
+                    state.remove_label(&label);
+                    if state.cordonlabels.is_empty() {
+                        self.cordon_drain_state = None;
+                    }
+                }
+                CordonDrainState::Draining(state) | CordonDrainState::Drained(state) => {
+                    // if no drain labels left, set to cordoned or none
+                    state.remove_label(&label);
+                    if state.drainlabels.is_empty() {
+                        if state.cordonlabels.is_empty() {
+                            self.cordon_drain_state = None;
+                        } else {
+                            self.cordon_drain_state = Some(CordonDrainState::Cordoned(
+                                CordonedState::new(state.cordonlabels.clone()),
+                            ));
+                        }
+                    }
+                }
+            },
+            None => {
+                // should not be possible
+            }
         }
     }
+
     /// Returns whether or not the node is cordoned.
     pub fn cordoned(&self) -> bool {
-        !self.cordon_labels.is_empty()
+        self.cordon_drain_state.is_some()
     }
-    /// Returns the cordon labels
-    pub fn cordon_labels(&self) -> Vec<String> {
-        self.cordon_labels.clone()
+
+    /// Returns true if it has the label in either of the lists.
+    pub fn has_cordon_label(&self, label: &str) -> bool {
+        self.has_cordon_only_label(label) || self.has_drain_label(label)
+    }
+
+    /// Returns true if it has the label in the cordon list.
+    pub fn has_cordon_only_label(&self, label: &str) -> bool {
+        match &self.cordon_drain_state {
+            Some(ds) => ds.has_cordon_only_label(label),
+            None => false,
+        }
+    }
+
+    /// Returns true if it has the label in the drain list.
+    pub fn has_drain_label(&self, label: &str) -> bool {
+        match &self.cordon_drain_state {
+            Some(ds) => ds.has_drain_label(label),
+            None => false,
+        }
+    }
+
+    /// Returns true if the node is in the draining state.
+    pub fn is_draining(&self) -> bool {
+        match &self.cordon_drain_state {
+            Some(ds) => ds.is_draining(),
+            None => false,
+        }
     }
 }
 
 impl From<NodeSpec> for models::NodeSpec {
     fn from(src: NodeSpec) -> Self {
-        Self::new(src.endpoint, src.id, src.cordon_labels)
+        Self::new_all(src.endpoint, src.id, src.cordon_drain_state.into_opt())
+    }
+}
+
+impl From<CordonDrainState> for models::CordonDrainState {
+    fn from(node_ds: CordonDrainState) -> Self {
+        match node_ds {
+            CordonDrainState::Cordoned(state) => {
+                let cs = models::CordonedState {
+                    cordonlabels: state.cordonlabels,
+                };
+                models::CordonDrainState::cordonedstate(cs)
+            }
+            CordonDrainState::Draining(state) => {
+                let ds = models::DrainState {
+                    cordonlabels: state.cordonlabels,
+                    drainlabels: state.drainlabels,
+                };
+                models::CordonDrainState::drainingstate(ds)
+            }
+            CordonDrainState::Drained(state) => {
+                let ds = models::DrainState {
+                    cordonlabels: state.cordonlabels,
+                    drainlabels: state.drainlabels,
+                };
+                models::CordonDrainState::drainedstate(ds)
+            }
+        }
     }
 }
 
--- a/control-plane/grpc/proto/v1/node/node.proto
+++ b/control-plane/grpc/proto/v1/node/node.proto
@@ -17,12 +17,6 @@ message Node {
   optional NodeState state = 3;
 }
 
-// Node cordon information
-message NodeCordon {
-  // Node cordon label.
-  repeated string label = 2;
-}
-
 message NodeSpec {
   // Node identification
   string node_id = 1;
@@ -30,8 +24,8 @@ message NodeSpec {
   string endpoint = 2;
   // Node labels.
   common.StringMapValue labels = 3;
-  // Cordon information.
-  NodeCordon cordon = 4;
+  // Drain state
+  optional CordonDrainState cordon_drain_state = 4;
 }
 
 message NodeState {
@@ -112,10 +106,45 @@ message UncordonNodeReply {
   }
 }
 
+message DrainNodeRequest {
+  // Node identification
+  string node_id = 1;
+  // Node cordon label
+  string label = 2;
+}
+
+message DrainNodeReply {
+  oneof reply {
+    Node node = 1;
+    common.ReplyError error = 2;
+  }
+}
+
+message CordonDrainState {
+    oneof cordondrainstate {
+        CordonedState cordoned = 1;
+        DrainState draining = 2;
+        DrainState drained = 3;
+      }
+}
+
+message CordonedState {
+  // Cordon information.
+  repeated string cordon_labels = 1;
+}
+
+message DrainState {
+  // Cordon information.
+  repeated string cordon_labels = 1;
+  // Drain information.
+  repeated string drain_labels = 2;
+}
+
 service NodeGrpc {
   rpc GetNodes (GetNodesRequest) returns (GetNodesReply) {}
   rpc GetBlockDevices (blockdevice.GetBlockDevicesRequest) returns (blockdevice.GetBlockDevicesReply) {}
   rpc Probe (ProbeRequest) returns (ProbeResponse) {}
   rpc CordonNode (CordonNodeRequest) returns (CordonNodeReply) {}
   rpc UncordonNode (UncordonNodeRequest) returns (UncordonNodeReply) {}
+  rpc DrainNode (DrainNodeRequest) returns (DrainNodeReply) {}
 }
--- a/control-plane/plugin/src/resources/cordon.rs
+++ b/control-plane/plugin/src/resources/cordon.rs
@@ -1,0 +1,60 @@
+pub struct NodeCordon {}
+pub struct NodeCordons {}
+
+use async_trait::async_trait;
+use openapi::models::CordonDrainState;
+
+use crate::{
+    operations::{Get, List},
+    resources::{
+        node::{node_display_print, node_display_print_one, NodeDisplayFormat},
+        utils::OutputFormat,
+        NodeId,
+    },
+    rest_wrapper::RestClient,
+};
+
+#[async_trait(?Send)]
+impl Get for NodeCordon {
+    type ID = NodeId;
+    async fn get(id: &Self::ID, output: &OutputFormat) {
+        match RestClient::client().nodes_api().get_node(id).await {
+            Ok(node) => {
+                node_display_print_one(node.into_body(), output, NodeDisplayFormat::CordonLabels)
+            }
+            Err(e) => {
+                println!("Failed to get node {}. Error {}", id, e)
+            }
+        }
+    }
+}
+
+#[async_trait(?Send)]
+impl List for NodeCordons {
+    async fn list(output: &OutputFormat) {
+        match RestClient::client().nodes_api().get_nodes().await {
+            Ok(nodes) => {
+                // iterate through the nodes and filter for only those that have cordon or drain
+                // labels
+                let nodelist = nodes.into_body();
+                let mut filteredlist = nodelist;
+                // remove nodes with no cordon or drain labels
+                filteredlist.retain(|i| {
+                    i.spec.is_some()
+                        && match &i.spec.as_ref().unwrap().cordondrainstate {
+                            Some(ds) => match ds {
+                                CordonDrainState::cordonedstate(_) => true,
+                                CordonDrainState::drainingstate(_) => true,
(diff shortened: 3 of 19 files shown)


common/src/types/v0/store/node.rs
```diff
@@ -13,6 +13,104 @@ use std::collections::HashMap;
 
 pub type NodeLabels = HashMap<String, String>;
 
+#[derive(Clone, Serialize, Deserialize, Debug, PartialEq)]
```
tiagolobocastro: missing doc comments

tiagolobocastro: please add Eq as well

common/src/types/v0/store/node.rs
```diff
@@ -13,6 +13,104 @@ use std::collections::HashMap;
...
+    pub fn remove_cordon_label(mut self, label: String) -> Self {
+        if let Some(index) = self.cordonlabels.iter().position(|l| l == &label.clone()) {
+            self.cordonlabels.remove(index);
+        }
+        self
+    }
+}
+
+#[derive(Clone, Serialize, Deserialize, Debug, PartialEq)]
+pub struct DrainingState {
```
tiagolobocastro: We could have a single struct for both draining and drained right?
If need be you can always create type aliases, eg:
```rust
struct DrainLabels { ... }
type DrainState = DrainLabels;
type DrainedState = DrainLabels;
```

common/src/types/v0/store/node.rs
```diff
@@ -66,33 +165,242 @@ impl NodeSpec {
...
     /// Node gRPC endpoint
     pub fn set_endpoint(&mut self, endpoint: String) {
         self.endpoint = endpoint
     }
+
     /// Cordon node by applying the label.
     pub fn cordon(&mut self, label: String) {
-        self.cordon_labels.push(label);
+        // the the node has the label, return with an error
+        match self.drain_state.clone() {
```
tiagolobocastro: nit: I think this can be made simpler by getting a mutable reference  and changing it, rather than resetting the drain_state in its entirety if add_cordon_label takes a mut reference instead of moving.

common/src/types/v0/store/node.rs
```diff
@@ -66,33 +165,242 @@ impl NodeSpec {
...
+            }
+        } else {
+            // return error
         }
     }
+
     /// Returns whether or not the node is cordoned.
     pub fn cordoned(&self) -> bool {
-        !self.cordon_labels.is_empty()
+        match self.drain_state.clone() {
```
tiagolobocastro: can use a reference here to avoid cloning

common/src/types/v0/store/node.rs
```diff
@@ -66,33 +165,242 @@ impl NodeSpec {
...
 }
 
 impl From<NodeSpec> for models::NodeSpec {
     fn from(src: NodeSpec) -> Self {
-        Self::new(src.endpoint, src.id, src.cordon_labels)
+        if src.drain_state.is_none() {
+            Self::new(src.endpoint, src.id)
+        } else {
+            let openapi_ds: Option<models::DrainState> = src.drain_state.map(|ds| ds.into());
+            Self::new_all(src.endpoint, src.id, openapi_ds)
```
tiagolobocastro: might be able to just use `new_all(..., src.drain_state.into_opt())`

control-plane/agents/core/src/node/specs.rs
```diff
@@ -140,4 +140,29 @@ impl ResourceSpecsLocked {
...
+                return Err(SvcError::CordonLabel {
+                    node_id: node_id.to_string(),
+                    label,
+                });
+            }
+            locked_node.drain(label);
+            locked_node.clone()
+        };
+        registry.store_obj(&drained_node_spec).await?;
+        Ok(drained_node_spec.clone())
```
tiagolobocastro: can just return `drained_node_spec` ?

control-plane/grpc/proto/v1/node/node.proto
```diff
@@ -17,21 +17,15 @@ message Node {
...
   // Node identification
   string node_id = 1;
   // Endpoint of the io-engine instance (gRPC)
   string endpoint = 2;
   // Node labels.
   common.StringMapValue labels = 3;
-  // Cordon information.
-  NodeCordon cordon = 4;
+  // Drain state
+  optional DrainState drainstate = 4;
```
tiagolobocastro: can we improve the `DrainState` naming, maybe `CordonDrainState`, or `SchedulingState`, something else?

tiagolobocastro (commented):
Seems you have some changes on the submodules causing changes on the Cargo.lock that need to be resolved

common/src/types/v0/store/node.rs
chriswldenyer: Oh right, I didn't think of that, will do.

common/src/types/v0/store/node.rs
chriswldenyer: Doc comments now added.

chriswldenyer: Now has Eq.

common/src/types/v0/store/node.rs
chriswldenyer: Now done, I opted for one type for the two drain enum variants.

common/src/types/v0/store/node.rs
chriswldenyer: Changed to updating the state instead of replacing, wherever the enum doesn't change, here and in other places.

common/src/types/v0/store/node.rs
chriswldenyer: fixed

control-plane/agents/core/src/node/specs.rs
chriswldenyer: fixed

control-plane/grpc/proto/v1/node/node.proto
chriswldenyer: Changed to CordonDrainState throughout (I considered CordonOrDrainState, but that's getting too long)

common/src/types/v0/store/node.rs
chriswldenyer: fixed

chriswldenyer:
bors try

chriswldenyer:
bors try

chriswldenyer:
bors try

common/src/types/v0/store/node.rs
```diff
@@ -66,33 +140,237 @@ impl NodeSpec {
...
+            }
+        }
     }
     /// Uncordon node by removing the corresponding label.
     pub fn uncordon(&mut self, label: String) {
-        if let Some(index) = self.cordon_labels.iter().position(|l| l == &label) {
-            self.cordon_labels.remove(index);
+        // if the label is a cordon label
+        if self.has_cordon_only_label(label.clone()) {
+            match &mut self.cordon_drain_state {
```
Abhinandan-Purkait: We could have used ```if self.cordon_drain_state.is_some()``` to avoid the empty `None` arm in match.

common/src/types/v0/store/node.rs
```diff
@@ -66,33 +140,237 @@ impl NodeSpec {
...
+                                ));
+                            }
+                        }
+                    }
+                },
+                None => {
+                    // should not be possible
+                }
+            }
+        } else {
```
Abhinandan-Purkait: `else` block is not needed, it can be removed.

common/src/types/v0/store/node.rs
```diff
@@ -66,33 +140,237 @@ impl NodeSpec {
...
+                CordonDrainState::Draining(_) => true,
+                CordonDrainState::Drained(_) => true,
+            },
+            None => false,
+        }
+    }
+
+    /// Returns true if it has the label in either of the lists
+    pub fn has_cordon_label(&self, label: String) -> bool {
+        if self.has_cordon_only_label(label.clone()) || self.has_drain_label(label) {
```
Abhinandan-Purkait: we could have jiust returned `self.has_cordon_only_label(label.clone()) || self.has_drain_label(label)` here instead of extra condition checking

control-plane/plugin/src/resources/node.rs
```diff
@@ -142,17 +135,36 @@ impl Cordoning for Node {
...
-                        .unwrap_or_default();
-                    if cordon_labels.is_empty() {
+                    let mut cordon_labels: Vec<String> = vec![];
+                    let mut drain_labels: Vec<String> = vec![];
+                    match node.clone().into_body().spec.unwrap().cordondrainstate {
+                        Some(ds) => {
+                            cordon_labels = match ds {
+                                CordonDrainState::cordonedstate(state) => state.cordonlabels,
+                                CordonDrainState::drainingstate(state) => state.cordonlabels,
+                                CordonDrainState::drainedstate(state) => state.cordonlabels,
```
Abhinandan-Purkait: Can we move this to a function to which takes a `CordonDrainState` and gives the cordon labels, similarly for drain labels. Seems to be repeated at quite a few places.

control-plane/plugin/src/resources/node.rs
```diff
@@ -211,24 +258,74 @@ impl CreateRows for NodeDisplay {
...
+                    row[0].add_cell(Cell::new(&labelstring));
+                    rows.push(row[0].clone());
+                }
                 rows
             }
         }
     }
 }
 
+pub fn node_display_print(
```
Abhinandan-Purkait: Can we make this `pub(crate)` also doc comments are needed.

control-plane/plugin/src/resources/node.rs
```diff
@@ -211,24 +258,74 @@ impl CreateRows for NodeDisplay {
...
+        OutputFormat::Yaml | OutputFormat::Json => {
+            print_table(output, node_display.inner);
+        }
+        OutputFormat::None => {
+            print_table(output, node_display);
+        }
+    }
+}
+
+pub fn node_display_print_one(
```
Abhinandan-Purkait: `pub(crate)`,  `doc comments`

control-plane/plugin/src/resources/node.rs
```diff
@@ -239,6 +336,126 @@ impl GetHeaderRow for NodeDisplay {
...
+    type ID = NodeId;
+    async fn drain(
+        id: &Self::ID,
+        label: String,
+        drain_timeout: Option<humantime::Duration>,
+        output: &utils::OutputFormat,
+    ) {
+        let mut timeout_instant: Option<time::Instant> = None;
+        if let Some(dt) = drain_timeout {
+            //let duration: std::time::Duration = drain_timeout.unwrap().into();
```
Abhinandan-Purkait: remove this line?

control-plane/plugin/src/resources/node.rs
```diff
@@ -239,6 +336,126 @@ impl GetHeaderRow for NodeDisplay {
...
+    /// Return the node ID.
+    pub fn node_id(&self) -> NodeId {
+        self.node_id.clone()
+    }
+    /// Return the drain label.
+    pub fn label(&self) -> String {
+        self.label.clone()
+    }
+    /// Return whether or not we should show the drain labels.
+    pub fn drain_timeout(&self) -> Option<humantime::Duration> {
```
Abhinandan-Purkait, line 363: The comment doesn't seem to go with the actual operation

control-plane/plugin/src/resources/node.rs
```diff
@@ -239,6 +336,126 @@ impl GetHeaderRow for NodeDisplay {
...
+                        Err(e) => {
+                            println!("Failed to get node {}. Error {}", id, e);
+                            break;
+                        }
+                    }
+                    if timeout_instant.is_some() && time::Instant::now() > timeout_instant.unwrap()
+                    {
+                        println!("Drain command timed out");
+                        break;
+                    }
```
Abhinandan-Purkait: Question: Is this the time until which the user wants to look at the drain command's output and when time is up we just say the time is up and come out of the loop, but the drain may not be complete by then? What's the significance of this?

tiagolobocastro:
bors try

tiagolobocastro:
bors cancel

tiagolobocastro:
bors try-

common/src/types/v0/store/node.rs
```diff
@@ -1,31 +1,104 @@
...
 
+/// Data relating to the cordoning of a node
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
+pub struct CordonedState {
+    /// labels used to enforce a node cordon
+    pub cordonlabels: Vec<String>,
+}
+
+impl CordonedState {
+    pub fn new(cordonlabels: Vec<String>) -> Self {
```
tiagolobocastro, line 25: missing doc comments

common/src/types/v0/store/node.rs
```diff
@@ -1,31 +1,104 @@
...
+        },
+        transport::{self, NodeId},
     },
-    transport::{self, NodeId},
+    IntoOption,
 };
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 
+/// Generic labels associated with the node
```
tiagolobocastro: we've been trying to tidy up the doc comments now, start with upper case (already done here) and end them with a period `.`

common/src/types/v0/store/node.rs
```diff
@@ -1,31 +1,104 @@
...
+
+impl CordonedState {
+    pub fn new(cordonlabels: Vec<String>) -> Self {
+        Self { cordonlabels }
+    }
+    pub fn add_cordon_label(&mut self, label: String) {
+        self.cordonlabels.insert(0, label);
+    }
+    pub fn remove_cordon_label(&mut self, label: String) {
+        if let Some(index) = self.cordonlabels.iter().position(|l| l == &label.clone()) {
```
tiagolobocastro: IIRC you can do something like:
```rust
self.cordonlabels.retain(|l| l == label);
```

common/src/types/v0/store/node.rs
```diff
@@ -1,31 +1,104 @@
...
 use std::collections::HashMap;
 
+/// Generic labels associated with the node
 pub type NodeLabels = HashMap<String, String>;
 
+/// Data relating to the cordoning of a node
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
+pub struct CordonedState {
+    /// labels used to enforce a node cordon
+    pub cordonlabels: Vec<String>,
```
tiagolobocastro, line 20: nit: cordon`_`labels ?

control-plane/plugin/src/resources/node.rs
chriswldenyer: That's correct. AIUI we want the drain command to block until the drain is completed, but have some option of timing-out the command if drain takes too long.

common/src/types/v0/store/node.rs
```diff
@@ -1,31 +1,104 @@
...
+}
+
+impl CordonedState {
+    pub fn new(cordonlabels: Vec<String>) -> Self {
+        Self { cordonlabels }
+    }
+    pub fn add_cordon_label(&mut self, label: String) {
+        self.cordonlabels.insert(0, label);
```
tiagolobocastro:
```suggestion
        self.cordonlabels.push(label);
```

common/src/types/v0/store/node.rs
```diff
@@ -66,33 +140,237 @@ impl NodeSpec {
...
+
     /// Node gRPC endpoint
     pub fn set_endpoint(&mut self, endpoint: String) {
         self.endpoint = endpoint
     }
+
     /// Cordon node by applying the label.
     pub fn cordon(&mut self, label: String) {
-        self.cordon_labels.push(label);
+        // the the node has the label, return with an error
```
tiagolobocastro: one extra `the` :)

common/src/types/v0/store/node.rs
```diff
@@ -66,33 +140,237 @@ impl NodeSpec {
...
+            None => {
+                //add the label and set the state to cordoned
+                self.cordon_drain_state =
+                    Some(CordonDrainState::Cordoned(CordonedState::new(vec![label])));
+            }
+        }
+    }
+
+    /// Drain node by applying the drain label.
+    pub fn drain(&mut self, label: String) {
```
tiagolobocastro: nit: could drain be misinterpreted as draining the node `right now`? What about `set_drain`, or something like that?

common/src/types/v0/store/node.rs
```diff
@@ -66,33 +140,237 @@ impl NodeSpec {
...
+                    state.add_drain_label(label);
+                }
+                CordonDrainState::Drained(state) => {
+                    // add the label
+                    state.add_drain_label(label);
+                }
+            },
+            None => {
+                //add the label and set the state to draining
+                let cordonlabels: Vec<String> = vec![];
```
tiagolobocastro: nit:
```rust
let cordonlabels = Vec::<String>::new();
```

common/src/types/v0/store/node.rs
```diff
@@ -66,33 +140,237 @@ impl NodeSpec {
...
+                )));
+            }
+        }
     }
     /// Uncordon node by removing the corresponding label.
     pub fn uncordon(&mut self, label: String) {
-        if let Some(index) = self.cordon_labels.iter().position(|l| l == &label) {
-            self.cordon_labels.remove(index);
+        // if the label is a cordon label
+        if self.has_cordon_only_label(label.clone()) {
```
tiagolobocastro: `has_cordon_only_label` can take a `&str` to avoid cloning.

control-plane/plugin/src/resources/node.rs
Abhinandan-Purkait: Okay. That makes sense, but I wonder from a user perspective whether a we would set a timeout or rather do a `Ctrl+c` to terminate the command.

common/src/types/v0/store/node.rs
tiagolobocastro: Do we need to check at all though? Can't we just do
```rust
match &mut self.cordon_drain_state {
Some(ds) => match ds {
    CordonDrainState::Cordoned(state) => {
        state.remove_label(label);
    }
    CordonDrainState::Draining(state) => {
        state.remove_label(label);
        ...
    }
    CordonDrainState::Drained(state) => {
        state.remove_label(label);
        ...
    }
},
None => {
   // Not cordoned or drained
}
```

common/src/types/v0/store/node.rs
```diff
@@ -1,31 +1,104 @@
...
+    }
+}
+
+/// Data relating to a draining or drained node, including non-drain cordon labels
+#[derive(Clone, Serialize, Deserialize, Debug, PartialEq)]
+pub struct DrainState {
+    /// labels used to enforce a node cordon
+    pub cordonlabels: Vec<String>,
+    /// labels used to cordon a node specifically for a drain
+    pub drainlabels: Vec<String>,
```
tiagolobocastro, line 44: Why do we need to distinguish the labels again? Once we're in drain, we can only move to down to uncordon and not cordoned?

common/src/types/v0/store/node.rs
```diff
@@ -66,33 +140,237 @@ impl NodeSpec {
...
+            }
+        } else {
+            // return error
         }
     }
+
     /// Returns whether or not the node is cordoned.
     pub fn cordoned(&self) -> bool {
-        !self.cordon_labels.is_empty()
+        match &self.cordon_drain_state {
```
tiagolobocastro: nit: self.cordon_drain_state.is_some() ?

common/src/types/v0/store/node.rs
```diff
@@ -66,33 +140,237 @@ impl NodeSpec {
...
 }
 
 impl From<NodeSpec> for models::NodeSpec {
     fn from(src: NodeSpec) -> Self {
-        Self::new(src.endpoint, src.id, src.cordon_labels)
+        if src.cordon_drain_state.is_none() {
+            Self::new(src.endpoint, src.id)
+        } else {
+            Self::new_all(src.endpoint, src.id, src.cordon_drain_state.into_opt())
+        }
```
tiagolobocastro:
```suggestion
        Self::new_all(src.endpoint, src.id, src.cordon_drain_state.into_opt())
```

control-plane/agents/core/src/node/specs.rs
```diff
@@ -141,4 +141,29 @@ impl ResourceSpecsLocked {
             })
             .collect()
     }
+
+    /// Drain the node with the given ID.
```
tiagolobocastro: We're not actually draining the node here?

common/src/types/v0/store/node.rs
chriswldenyer: It could be drained (e.g. by the update process) and cordoned (by a user for whatever reason) at the same time.
If a drain is cancelled, the drain label is removed but the user would still want their cordon in place.
Well, that's how it was explained to me...

common/src/types/v0/store/node.rs
chriswldenyer: doh!

common/src/types/v0/store/node.rs
chriswldenyer: fair enough, will fix

common/src/types/v0/store/node.rs
chriswldenyer: agreed

common/src/types/v0/store/node.rs
chriswldenyer: Arguably you should do that only for complete sentences, but ok fixed.

control-plane/plugin/src/resources/node.rs
chriswldenyer: fixed

control-plane/plugin/src/resources/node.rs
chriswldenyer: fixed

control-plane/plugin/src/resources/node.rs
chriswldenyer: fixed

common/src/types/v0/store/node.rs
chriswldenyer: I'm trying to avoid the hyphens vs underscores issue in identifiers/paths by using neither unless necessary.
I think here cordonlabels is legible enough not to need it IMO.

common/src/types/v0/store/node.rs
chriswldenyer: Oh, yeah :-)

common/src/types/v0/store/node.rs
chriswldenyer: fixed

common/src/types/v0/store/node.rs
chriswldenyer: fixed

common/src/types/v0/store/node.rs
chriswldenyer: fixed

common/src/types/v0/store/node.rs
chriswldenyer: fixed

control-plane/plugin/src/resources/node.rs
chriswldenyer: fixed

control-plane/agents/core/src/node/specs.rs
chriswldenyer: I've amended the comments.

common/src/types/v0/store/node.rs
chriswldenyer: fixed

common/src/types/v0/store/node.rs
chriswldenyer: Yes, that works 👍

common/src/types/v0/store/node.rs
chriswldenyer: Wouldn't you also then have to unwrap() (IIUIC) ?
Having the empty None arm seems to result in less code, but I may be missing a trick.

common/src/types/v0/store/node.rs
chriswldenyer: Cool. But with:
l != label
I think

tiagolobocastro: Ah yes right!

common/src/types/v0/store/node.rs
tiagolobocastro: I have a slight preference for the "_" :) but it's fine with me as it is :+1: 
q: what about removing "cordon" altogether since we're in the CordonedState?
So just "labels", even simpler.

common/src/types/v0/store/node.rs
tiagolobocastro: Did you forget to push? Still can't see them.

common/src/types/v0/store/node.rs
```diff
@@ -1,31 +1,104 @@
...
+        if let Some(index) = self.drainlabels.iter().position(|l| l == &label.clone()) {
+            self.drainlabels.remove(index);
+        }
+    }
+}
+
+/// Enum variant encompassing data related to a cordoned or draining node
+#[derive(Clone, Serialize, Deserialize, Debug, PartialEq)]
+pub enum CordonDrainState {
+    /// the node is being cordoned
```
tiagolobocastro: nit: Should Capitalize and end with a period.

common/src/types/v0/store/node.rs
chriswldenyer: You could, but if you call them just "labels" it's less obvious that they become (or are copied from) the same set of labels as the "cordonlabels" in the DrainState when the state changes. That would make harder for developers to follow IMO.

common/src/types/v0/store/node.rs
chriswldenyer: fixed

common/src/types/v0/store/node.rs
Abhinandan-Purkait: Yes, you would still have to unwrap but since you are already checking using is_some() you can staightaway unwrap() inside the check without thinking of error handling.

chriswldenyer: That function is now refactored from a different review comment, so the suggested change no longer applies.

chriswldenyer:
bors try

chriswldenyer:
bors try

chriswldenyer:
bors try-

chriswldenyer:
bors try

common/src/types/v0/store/node.rs
```diff
@@ -1,95 +1,318 @@
...
+    pub cordonlabels: Vec<String>,
+}
+
+impl CordonedState {
+    /// Create a new CordonedState object.
+    pub fn new(cordonlabels: Vec<String>) -> Self {
+        Self { cordonlabels }
+    }
+    /// Add a cordon label to a CordonedState object.
+    pub fn add_cordon_label(&mut self, label: &String) {
```
tiagolobocastro: It's usually preferred to have `&str` instead of `&String` can be coerced into `&str`.

common/src/types/v0/store/node.rs
```diff
@@ -1,95 +1,318 @@
...
+    }
+
+    /// Drain node by applying the drain label.
+    pub fn set_drain(&mut self, label: String) {
+        // the the node has the label, return with an error
+        match &mut self.cordon_drain_state {
+            Some(ds) => match ds {
+                CordonDrainState::Cordoned(state) => {
+                    // set state to draining and add label
+                    let drainlabels = vec![label];
```
tiagolobocastro: what about moving this logic to a function drain in CordonState, eg:
```rust
impl CordonState {
  fn into_draining(self, drainlabel: &str) -> DrainState {
     DrainState::new(self.cordonlabels, vec![drainlabel])
  }
}
...
self.cordon_drain_state = state.into_draining(label);
```

common/src/types/v0/store/node.rs
```diff
@@ -1,95 +1,318 @@
...
         self.endpoint = endpoint
     }
+
     /// Cordon node by applying the label.
     pub fn cordon(&mut self, label: String) {
-        self.cordon_labels.push(label);
+        // the node has the label, return with an error
+        match &mut self.cordon_drain_state {
+            Some(ds) => match ds {
+                CordonDrainState::Cordoned(state) => {
```
tiagolobocastro: would it be tidier to add a add_cordon_label fn to CordonDrainState and handle this logic there?

common/src/types/v0/store/node.rs
```diff
@@ -1,95 +1,318 @@
...
+                }
+                CordonDrainState::Drained(state) => {
+                    // just add the label
+                    state.add_cordon_label(&label);
+                }
+            },
+            None => {
+                //add the label and set the state to cordoned
+                self.cordon_drain_state =
+                    Some(CordonDrainState::Cordoned(CordonedState::new(vec![label])));
```
tiagolobocastro: nit: what about having a CordonDrainState::cordon(label: &str) -> Self ?

common/src/types/v0/store/node.rs
```diff
@@ -1,95 +1,318 @@
...
+
+    /// Returns true if it has the label in either of the lists.
+    pub fn has_cordon_label(&self, label: &String) -> bool {
+        self.has_cordon_only_label(label) || self.has_drain_label(label)
+    }
+
+    /// Returns true if it has the label in the cordon list.
+    pub fn has_cordon_only_label(&self, label: &String) -> bool {
+        match &self.cordon_drain_state {
+            Some(ds) => match ds {
```
tiagolobocastro: can this logic be moved into `CordonDrainState`?

control-plane/grpc/proto/v1/node/node.proto
```diff
@@ -17,21 +17,15 @@ message Node {
...
   // Node identification
   string node_id = 1;
   // Endpoint of the io-engine instance (gRPC)
   string endpoint = 2;
   // Node labels.
   common.StringMapValue labels = 3;
-  // Cordon information.
-  NodeCordon cordon = 4;
+  // Drain state
+  optional CordonDrainState cordondrainstate = 4;
```
tiagolobocastro: Should this be cordon_drain_state?

control-plane/plugin/src/resources/cordon.rs
```diff
@@ -0,0 +1,61 @@
...
+            Ok(nodes) => {
+                // iterate through the nodes and filter for only those that have cordon or drain
+                // labels
+                let nodelist = nodes.into_body();
+                let mut filteredlist = nodelist;
+                // remove nodes with no cordon or drain labels
+                filteredlist.retain(|i| {
+                    i.spec.is_some()
+                        && i.spec.is_some()
+                        && match &i.spec.as_ref().unwrap().cordondrainstate {
```
tiagolobocastro: Is there a duplicate `i.spec.is_some()` ?
```rust
i.spec.as_ref().map(|s| s.cordondrainstate.is_some()).unwrap_or_default()
```
?

control-plane/plugin/src/resources/node.rs
```diff
@@ -142,17 +152,28 @@ impl Cordoning for Node {
...
                     // In case the output format is not specified, show a success message.
-                    let cordon_labels = node
-                        .into_body()
-                        .spec
-                        .map(|node_spec| node_spec.cordon_labels)
-                        .unwrap_or_default();
-                    if cordon_labels.is_empty() {
+                    let mut cordon_labels: Vec<String> = vec![];
+                    let mut drain_labels: Vec<String> = vec![];
+                    match node.clone().into_body().spec.unwrap().cordondrainstate {
```
tiagolobocastro: need to handle case when spec is None?

control-plane/plugin/src/resources/node.rs
```diff
@@ -162,47 +183,70 @@ impl Cordoning for Node {
...
-            inner: node,
+            inner: nodes,
             format,
         }
     }
+    pub(crate) fn get_label_list(node: &openapi::models::Node) -> Vec<String> {
+        let mut cordon_labels: Vec<String> = vec![];
+        let mut drain_labels: Vec<String> = vec![];
+        if node.spec.is_some() {
+            match &node.spec.as_ref().unwrap().cordondrainstate {
```
```suggestion
        if let Some(spec) = &node.spec {
          match spec.cordondrainstate
```
or 
```suggestion
        match node.spec.as_ref().flat_map(|s| s.cordondrainstate) {
         Some(v) =>
         None =>
        }
```

Abhinandan-Purkait (approved):

common/src/types/v0/store/node.rs
chriswldenyer: fixed

common/src/types/v0/store/node.rs
chriswldenyer: fixed

common/src/types/v0/store/node.rs
chriswldenyer: fixed

common/src/types/v0/store/node.rs
chriswldenyer: fixed (and did the same for some other related functions)

control-plane/grpc/proto/v1/node/node.proto
chriswldenyer: fixed

control-plane/plugin/src/resources/cordon.rs
chriswldenyer: fixed, not sure how that got in.

control-plane/plugin/src/resources/node.rs
chriswldenyer: fixed, and took the opportunity to simplify the code

control-plane/plugin/src/resources/node.rs
chriswldenyer: I meant to use your 1st suggestion but ended up with 2 layers of 'match'es.

chriswldenyer:
Note, I also changed some of the plugin output to specify the node being drained, to more-closely match the output of "cordon".

common/src/types/v0/store/node.rs
```diff
@@ -1,95 +1,319 @@
...
+    /// The node is being cordoned.
+    Cordoned(CordonedState),
+    /// The node is being drained.
+    Draining(DrainState),
+    /// The drain has completed.
+    Drained(DrainState),
+}
+
+impl CordonDrainState {
+    pub fn add_cordon_label(&mut self, label: &str) {
```
tiagolobocastro, line 79: missing doc comments

tiagolobocastro (approved):

control-plane/plugin/src/resources/node.rs
chriswldenyer: fixed

common/src/types/v0/store/node.rs
chriswldenyer: fixed

chriswldenyer:
bors try

chriswldenyer:
bors merge

PR #313, opened 2022-08-21
feat(replica): add support for v1 replica api calls

Ab-hishek:
- Add v1 replica api calls(create/destroy/share/unshare/list)

Signed-off-by: Abhishek Agarwal <abhishek.agarwal@mayadata.io>


common/src/types/v0/transport/replica.rs
```diff
@@ -21,6 +22,103 @@ impl GetReplicas {
...
+    #[test]
+    fn test_replica_pool_deserializer() {
+        let inputs: Vec<&str> = vec![
+            r#"{"name":"pool-1", "uuid":null}"#,
+            r#"{"name":"pool-1", "uuid":"22ca10d3-4f2b-4b95-9814-9181c025cc1a"}"#,
+        ];
+
+        for input in &inputs {
+            let replica_pool: ReplicaPool = serde_json::from_str(input).unwrap();
+            println!("{:?}", replica_pool);
```
vharsh: Instead of printing out objects can we instead assert-compare it with expected object?

control-plane/agents/core/src/controller/wrapper.rs
```diff
@@ -400,7 +404,7 @@ impl NodeWrapper {
                     .get_replica_states()
                     .filter_map(|r| {
                         let replica = r.lock();
-                        if replica.replica.pool == pool_state.uuid() {
+                        if replica.replica.pool.name == pool_state.uuid() {
```
vharsh: A pool's name & pool_state's uuid are expected to be same? this doesn't feel right in a first glance

common/src/types/v0/store/replica.rs
```diff
@@ -75,7 +75,7 @@ pub struct ReplicaSpec {
     /// The size that the replica should be.
     pub size: u64,
     /// The pool that the replica should live on.
-    pub pool: PoolId,
+    pub pool: ReplicaPool,
```
Abhinandan-Purkait: Should this be called `PoolRef`?

common/src/types/v0/store/replica.rs
```diff
@@ -93,6 +93,25 @@ pub struct ReplicaSpec {
     pub operation: Option<ReplicaOperationState>,
 }
 
+#[cfg(test)]
+mod tests_de {
```
Abhinandan-Purkait: `test_deserializer` ?

common/src/types/v0/transport/pool.rs
```diff
@@ -87,6 +88,32 @@ impl From<PoolState> for models::PoolState {
 }
 
 bus_impl_string_id!(PoolId, "ID of a pool");
+bus_impl_string_uuid!(PoolUuid, "UUID of a pool");
+
+impl TryFrom<&Value> for PoolId {
```
Abhinandan-Purkait: Should this be `impl TryFrom<Value> for PoolId ..`

common/src/types/v0/transport/replica.rs
```diff
@@ -21,6 +22,103 @@ impl GetReplicas {
...
+
+impl ReplicaPool {
+    /// Return a new `Self` from the provided arguments
+    pub fn new(name: &PoolId, uuid: Option<PoolUuid>) -> Self {
+        Self {
+            name: name.clone(),
+            uuid,
+        }
+    }
+    /// get the uuid of the pool
```
Abhinandan-Purkait: We have been following this  :- `writing doc comments starting with Capital and ending with period`.

common/src/types/v0/transport/replica.rs
```diff
@@ -21,6 +22,103 @@ impl GetReplicas {
...
+    }
+}
+
+impl<'de> Deserialize<'de> for ReplicaPool {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        let json: Value = Value::deserialize(deserializer)?;
+        println!("json: {:?}", json);
```
Abhinandan-Purkait: Can we be remove this?

common/src/types/v0/transport/replica.rs
```diff
@@ -21,6 +22,103 @@ impl GetReplicas {
...
+            Value::String(name) => ReplicaPool::new(&name.into(), None),
+            _ => return Err(serde::de::Error::custom("Invalid replica pool".to_string())),
+        };
+
+        Ok(replica_pool)
+    }
+}
+
+#[cfg(test)]
+mod tests_de {
```
Abhinandan-Purkait: `test_deserializer` ?

common/src/types/v0/transport/replica.rs
```diff
@@ -31,8 +129,8 @@ pub struct Replica {
     pub name: ReplicaName,
     /// uuid of the replica
     pub uuid: ReplicaId,
-    /// id of the pool
-    pub pool: PoolId,
+    /// pool on which the replica resides
+    pub pool: ReplicaPool,
```
Abhinandan-Purkait: `PoolRef` ?

control-plane/grpc/proto/v1/replica/replica.proto
```diff
@@ -13,8 +13,8 @@ message Replica{
   string name = 2;
   // uuid of the replica
   google.protobuf.StringValue replica_id = 3;
-  // id of the pool
-  string pool_id = 4;
+  // pool on which replica resides
+  Pool pool = 4;
```
Abhinandan-Purkait: `PoolRef`? This would need changes are several places, commenting here.

common/src/types/v0/transport/pool.rs
tiagolobocastro: These are not needed, we should use `Deserialize::deserialize` instead.

common/src/types/v0/transport/replica.rs
```diff
@@ -21,6 +22,103 @@ impl GetReplicas {
...
+    where
+        D: Deserializer<'de>,
+    {
+        let json: Value = Value::deserialize(deserializer)?;
+        println!("json: {:?}", json);
+
+        let replica_pool = match json {
+            Value::Object(pool) => {
+                let name = match pool.get(&"name".to_string()) {
+                    Some(name) => match PoolId::try_from(name) {
```
tiagolobocastro: Can use instead the deser trait, example:
Some(name) => PoolId::deserialize(name).unwrap(),
But as pointed out above, I don't think we even need this if we move to an untagged. enum.

common/src/types/v0/transport/replica.rs
```diff
@@ -21,6 +22,103 @@ impl GetReplicas {
...
+        }
+    }
+
+    /// get name of the pool from uuid
+    pub fn get_name_from_uuid(_uuid: &PoolUuid) -> PoolId {
+        unimplemented!()
+    }
+}
+
+impl<'de> Deserialize<'de> for ReplicaPool {
```
tiagolobocastro: I think we don't need to implement this manually by making ReplicaPool as such:
```rust
pub enum PoolRef {
    Named { name: PoolId },
    Uuid { name: PoolId, uuid: PoolUuid },
}
```
And then on the replicaSpec use #[serde(untagged)]

common/src/types/v0/transport/replica.rs
tiagolobocastro: would be good to use asserts to make sure the values are expected.
We're also missing a test case for the string case.

Cargo.lock
```diff
@@ -4796,4 +4796,4 @@ checksum = "9fd07cbbc53846d9145dbffdf6dd09a7a0aa52be46741825f5c97bdd4f73f12b"
 dependencies = [
  "cc",
  "libc",
-]
+]
```
tiagolobocastro: Do not attempt to resolve conflicts manually, simply reset file and run cargo build.

common/src/types/v0/store/replica.rs
Ab-hishek: Added this new type for `pool` field inside ReplicaSpec.

common/src/types/v0/transport/replica.rs
Ab-hishek: Removed this test.

common/src/types/v0/transport/replica.rs
Ab-hishek: `PoolRef` used inside ReplicaSpec for now.

control-plane/grpc/proto/v1/replica/replica.proto
Ab-hishek: PoolRef used inside ReplicaSpec for now. As a modification if we feel to change these as well to PoolRef, will take up as a refactor task.

common/src/types/v0/transport/replica.rs
Ab-hishek: Removed this custom deserializer.

common/src/types/v0/transport/replica.rs
Ab-hishek: Done.

common/src/types/v0/transport/replica.rs
Ab-hishek: Custom deserializer for ReplicaPool has been removed hence this test also.

Cargo.lock
Ab-hishek: How do I resolve it now? When I try to rebase with the develop branch(or reset it), it doesn't show me any changes for `Cargo.lock` and hence not able to commit it.

common/src/types/v0/store/replica.rs
```diff
@@ -93,6 +93,92 @@ pub struct ReplicaSpec {
...
+    Named(PoolId),
+    /// Uuid and name referenced pool.
+    Uuid(PoolId, PoolUuid),
+}
+
+impl Default for PoolRef {
+    fn default() -> Self {
+        Self::Named("".into())
+    }
+}
```
tiagolobocastro, line 113: Instead you might be able to add #[default] on top of named in the enum definition.

common/src/types/v0/store/replica.rs
```diff
@@ -93,6 +93,92 @@ pub struct ReplicaSpec {
...
+
+/// Returns pool id from a given pool reference.
+pub fn get_pool_id_from_pool_ref(pool_ref: PoolRef) -> PoolId {
```
tiagolobocastro: Instead, why not enhance PoolRef, example:
```rust
impl PoolRef {
  pub fn pool_name(&self) -> &PoolId {
      match &pool_ref {
        PoolRef::Named(name) => name,
        PoolRef::Uuid(name, _) => name,
    }
  }
}
```

control-plane/grpc/proto/v1/replica/replica.proto
tiagolobocastro: What if we keep this as is (pool_id as a string), and add a new optional uuid separately, would that simplify things?

common/src/types/v0/store/replica.rs
Ab-hishek: For enum, it errors out. We have you define our own for enum definition.

tiagolobocastro: Ah, looks like it only works for unit enum variants.

control-plane/grpc/proto/v1/replica/replica.proto
Ab-hishek: I guess so, that should simply things to a great extend.

common/src/types/v0/store/replica.rs
Ab-hishek: yes.

control-plane/agents/core/src/controller/wrapper.rs
Ab-hishek: It's changed now.

common/src/types/v0/store/replica.rs
```diff
@@ -119,10 +207,20 @@ impl ResourceUuid for ReplicaSpec {
...
+            PoolRef::Named(_) => None,
+            PoolRef::Uuid(_, uuid) => Some(uuid),
+        };
+        Self::new_all(
             src.managed,
+            None,
             src.owners,
-            src.pool,
+            pool_id,
+            pool_uuid.map(|uuid| openapi::apis::Uuid::try_from(uuid).unwrap()),
```
tiagolobocastro: we should not unwrap here, if need need we can use a TryFrom.
but in this case I don't think we need to, I think we can simply do pool_uuid.into_opt() here?

common/src/types/v0/store/replica.rs
```diff
@@ -119,10 +207,20 @@ impl ResourceUuid for ReplicaSpec {
 
 impl From<ReplicaSpec> for models::ReplicaSpec {
     fn from(src: ReplicaSpec) -> Self {
-        Self::new(
+        let pool_id = match src.pool.clone() {
```
tiagolobocastro: We have a lot of this code repetition here, we should add methods to PoolRef, example:
fn name(&self) -> PoolId { ... }
fn uuid(&self) -> Option<PoolUuid> { ... }

common/src/types/v0/transport/replica.rs
```diff
@@ -107,9 +109,11 @@ impl From<ReplicaName> for String {
 
 impl From<Replica> for models::Replica {
     fn from(src: Replica) -> Self {
-        Self::new(
+        Self::new_all(
             src.node,
-            src.pool,
+            src.pool_id,
+            src.pool_uuid
+                .map(|uuid| openapi::apis::Uuid::try_from(uuid).unwrap()),
```
tiagolobocastro: same here as above, with perhaps into_opt().

control-plane/grpc/proto/v1/replica/replica.proto
```diff
@@ -25,6 +25,8 @@ message Replica{
   string uri = 8;
   // status of the replica
   ReplicaStatus status = 9;
+  // uuid of the pool
+  optional google.protobuf.StringValue pool_uuid = 10;
```
tiagolobocastro, line 29: Doesn't StringValue already make it optional?

tiagolobocastro (approved):

Abhinandan-Purkait (approved):

control-plane/grpc/proto/v1/replica/replica.proto
Ab-hishek: Yes, it does but just for readability purpose, I added it as optional. Adding optional here doesn't have any affect though on the generated code.

Ab-hishek:
bors try

Ab-hishek:
bors merge

PR #314, opened 2022-08-24
feat(csi): decouple volume target from application

Abhinandan-Purkait:
As of now volume targets are always created on the same node where the application lives.
With the addition of HA w’ll be able to switch the nvmf target at the application node level, so we can now allow the target to live on any io-engine node in the cluster. This PR enables that change.

### DIFF
--- a/control-plane/agents/core/tests/pool/mod.rs
+++ b/control-plane/agents/core/tests/pool/mod.rs
@@ -770,7 +770,11 @@ async fn disown_unused_replicas() {
         .unwrap();
 
     let volume = volumes_api
-        .put_volume_target(&volume.spec.uuid, &node, models::VolumeShareProtocol::Nvmf)
+        .put_volume_target(
+            &volume.spec.uuid,
+            models::VolumeShareProtocol::Nvmf,
+            Some(&node),
+        )
         .await
         .unwrap();
 
--- a/control-plane/agents/core/tests/volume/mod.rs
+++ b/control-plane/agents/core/tests/volume/mod.rs
@@ -32,6 +32,7 @@ use std::{
     str::FromStr,
     time::Duration,
 };
+use tokio::time::sleep;
 
 #[tokio::test]
 async fn volume() {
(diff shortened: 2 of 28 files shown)


control-plane/csi-driver/controller/src/config.rs
```diff
@@ -12,7 +12,10 @@ pub(crate) struct CsiControllerConfig {
     /// I/O timeout for REST API operations.
     io_timeout: Duration,
     /// IO-Engine DaemonSet selector labels.
+    #[allow(dead_code)]
     io_engine_selector: HashMap<String, String>,
```
Abhinandan-Purkait: Question: Should we removed this?

control-plane/csi-driver/controller/src/controller.rs
```diff
@@ -133,13 +132,9 @@ impl VolumeTopologyMapper {
...
 
     /// Determine the list of nodes where the workload can be placed.
-    /// If volume is created as pinned (i.e. local=true), then the nexus and the workload
-    /// must be placed on the same node, which in fact means running workloads only on IO Engine
-    /// daemonset nodes.
-    /// For non-pinned volumes, workload can be put on any node in the cluster.
     fn volume_accessible_topology(&self) -> Vec<CsiTopology> {
         vec![rpc::csi::Topology {
-            segments: CsiControllerConfig::get_config().io_engine_selector(),
+            segments: CsiControllerConfig::get_config().csi_node_selector(),
```
Abhinandan-Purkait: Question: Is it a good idea to bind csi-nodes with some lables as in here, or let it be anything?

control-plane/agents/core/tests/pool/mod.rs
```diff
@@ -770,7 +770,11 @@ async fn disown_unused_replicas() {
         .unwrap();
 
     let volume = volumes_api
-        .put_volume_target(&volume.spec.uuid, &node, models::VolumeShareProtocol::Nvmf)
+        .put_volume_target(
+            &volume.spec.uuid,
+            models::VolumeShareProtocol::Nvmf,
+            Some(&node),
```
tiagolobocastro, line 776: why is the order different now?

control-plane/agents/core/tests/volume/mod.rs
```diff
@@ -1916,3 +1916,264 @@ async fn smoke_test(cluster: &Cluster) {
...
+                target_node: None,
+                ..Default::default()
+            },
+            None,
+        )
+        .await
+        .expect("Should not fail");
+
+    // determine in which node the volume1 was published
+    let kill_node = if pub_vol1.state().target.unwrap().node.as_str() == "io-engine-1" {
```
tiagolobocastro: should not use hardcoded strings, the names can be retrieved from the Cluster using cluster.node IIRC

control-plane/agents/core/tests/volume/mod.rs
```diff
@@ -1916,3 +1916,264 @@ async fn smoke_test(cluster: &Cluster) {
...
+    };
+
+    // kill the node where the volume was published
+    cluster
+        .composer()
+        .stop(kill_node)
+        .await
+        .unwrap_or_else(|_| panic!("The {} container should stop", kill_node));
+
+    sleep(Duration::from_secs(4)).await;
```
tiagolobocastro: Do we have to wait so long here?

control-plane/agents/core/tests/volume/mod.rs
```diff
@@ -1916,3 +1916,264 @@ async fn smoke_test(cluster: &Cluster) {
...
+        .composer()
+        .kill("io-engine-1")
+        .await
+        .expect("The io-engine-1 container should stop");
+    cluster
+        .composer()
+        .kill("io-engine-2")
+        .await
+        .expect("The io-engine-2 container should stop");
+    sleep(Duration::from_secs(4)).await;
```
tiagolobocastro: If we reduce the node deadline timeout we shouldn't have to wait so long.

control-plane/csi-driver/controller/src/config.rs
tiagolobocastro: yes probably

control-plane/csi-driver/controller/src/controller.rs
tiagolobocastro: We can't bind this otherwise e2e will likely break - this will have to be discussed with a wider audience.
I would make it optional via cmdline and by default it should not bind.

control-plane/csi-driver/controller/src/controller.rs
```diff
@@ -266,20 +258,10 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
...
-                        thin,
-                    )
+                    .create_volume(&u, replica_count, size, volume_topology, thin)
                     .await?;
 
-                debug!(
-                    "Volume {} successfully created, pinned volume = {}",
-                    volume_uuid, pinned_volume
-                );
+                debug!("Volume {} successfully created", volume_uuid);
```
tiagolobocastro:
```suggestion
                debug!(volume.uuid=volume_uuid, "Volume successfully created");
```

control-plane/csi-driver/controller/src/controller.rs
```diff
@@ -398,9 +380,16 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
                     }
                 },
             _ => {
+                // if the csi-node happens to be a data-plane node, use that for nexus creation, otherwise
+                // let the control-plane select the target node.
+                let target_node = match IoEngineApiClient::get_client().get_node(&node_id).await {
+                    Ok(_) => Some(node_id.as_str()),
```
tiagolobocastro: this is not quite accurate, depends on the type of error? Eg: we could simply fail to issue the api call but that doesn't mean the node is not a data-plane node.

I think we should also discuss what to do here with a wider audience:
Options:
1. Always defer target placement to the control-plane.
2. Run on app node is possible.

control-plane/agents/core/tests/pool/mod.rs
Abhinandan-Purkait: Yeah, this is what I even noticed, I am not sure, somehow the openapi generator generated code changed the order after making the node field not required.

control-plane/csi-driver/controller/src/controller.rs
Abhinandan-Purkait: Sure.

control-plane/csi-driver/controller/src/config.rs
Abhinandan-Purkait: Okay. Removing it.

control-plane/agents/core/tests/volume/mod.rs
Abhinandan-Purkait: Yeah, right. I will do that.

control-plane/agents/core/tests/volume/mod.rs
Abhinandan-Purkait: Well I added `4s` as the node takes `2s` to change the state from Online, so 4 to be on safer side.

tiagolobocastro: But we're stopping io-engine gracefully with stop, so the core agent will change the state when it receives the stop message?

Abhinandan-Purkait: Ahh right. It will deregister. Now that you have pointed, should it have been a `kill` here rather than `stop`?

tiagolobocastro: I don't think it should make any difference for what we're testing for here?

control-plane/csi-driver/controller/src/controller.rs
Abhinandan-Purkait: That's right, but anyhow if we are not able to determine it is a `dataplane` node or not we will always end up with Option 1. So probably option 1 is the right way.

control-plane/agents/core/tests/volume/mod.rs
Abhinandan-Purkait: We are testing, if a node was previously being not selected due to equal distribution, will now be selected if there are no other nodes to select.

control-plane/csi-driver/controller/src/controller.rs
tiagolobocastro: We can determine if it's a `dataplane` by inspecting checking the exact error we receive? If it's 404 then it must not be a `dataplane` node?

control-plane/agents/core/tests/volume/mod.rs
tiagolobocastro: but why does `stop` or `kill` make a difference here?

control-plane/csi-driver/controller/src/controller.rs
Abhinandan-Purkait: Done.

control-plane/agents/core/tests/volume/mod.rs
Abhinandan-Purkait: Difference if graceful shutdown and abrupt shutdown. But, yeah checking for any of these two should suffice.

control-plane/csi-driver/controller/src/controller.rs
Abhinandan-Purkait: Removed it for now.

control-plane/agents/core/tests/volume/mod.rs
Abhinandan-Purkait: Done.

control-plane/agents/core/tests/volume/mod.rs
Abhinandan-Purkait: Done.

Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors try

control-plane/csi-driver/controller/src/controller.rs
```diff
@@ -398,9 +379,16 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
                     }
                 },
             _ => {
+                // if the csi-node happens to be a data-plane node, use that for nexus creation, otherwise
+                // let the control-plane select the target node.
+                let target_node = match IoEngineApiClient::get_client().get_node(&node_id).await {
+                    Err(ApiClientError::ResourceNotExists(_)) => None,
+                    _ => Some(node_id.as_str()),
+                };
```
tiagolobocastro:
```suggestion
                let target_node = match IoEngineApiClient::get_client().get_node(&node_id).await {
                    Err(ApiClientError::ResourceNotExists(_)) => Ok(None),
                    Err(error) => Err(error),
                    Ok(_) => Ok(Some(node_id.as_str())),
                }?;
```

tiagolobocastro (approved):

control-plane/csi-driver/controller/src/controller.rs
Abhinandan-Purkait: Done.

Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors merge

Abhinandan-Purkait:
bors merge

Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors merge

PR #315, opened 2022-08-24
refactor: tidy up several components

tiagolobocastro:
refactor: split up the core volume tests into modules

refactor: rename unused node service and remove extern usage

feat(node-service): add device path to find_volume


tiagolobocastro:
bors try

control-plane/csi-driver/node/src/main.rs
```diff
@@ -224,6 +208,17 @@ impl CsiServer {
             let uds = UnixListener::bind(csi_socket).unwrap();
             info!("CSI plugin bound to {}", csi_socket);
 
+            // Change permissions on CSI socket to allow non-privileged clients to access it
+            // to simplify testing.
+            if let Err(e) = fs::set_permissions(
```
Abhinandan-Purkait, line 213: Question: Why do we need to change the permission? How was it working without it?

control-plane/csi-driver/node/src/nodeplugin_svc.rs
```diff
@@ -31,6 +32,7 @@ pub(crate) enum ServiceError {
     BlockDeviceMount { volid: String },
```
Abhinandan-Purkait: `volid` -> `vol_id` ?

control-plane/csi-driver/node/src/main.rs
tiagolobocastro: To make testing easier, otherwise we can't connect to the socket without "chmoding" it.

control-plane/csi-driver/node/src/nodeplugin_svc.rs
tiagolobocastro: This is existing code, but it should have an _ indeed, I'll change it.

control-plane/csi-driver/node/src/main.rs
Abhinandan-Purkait: Okay.

control-plane/csi-driver/node/src/nodeplugin_svc.rs
Abhinandan-Purkait: Thanks

Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

PR #319, opened 2022-08-29
fix(nexus): refactor v0 grpc nexus create call response

Ab-hishek:
This PR moves the handling of the response for `create_nexus` call from the caller function to be called function to be able to handle them according to the versions. Since, in v0 grpc there were two different nexus structs: `Nexus` and `NexusV2` and based on the returned struct some info(name & uuid) needed to be appropriately filled on the older nexus struct. Hence this should be handled in version specific calls. Therefore, this part of the code is moved specifically into the v0 version call block.

Signed-off-by: Abhishek Agarwal <abhishek.agarwal@mayadata.io>


tiagolobocastro (approved):

Abhinandan-Purkait (approved):

Ab-hishek:
bors merge

PR #322, opened 2022-08-31
refactor: tidy up agents library and deployer

tiagolobocastro:
refactor(deployer): tidy up the component folder and reduce macro usage

Organize the components a bit better.
Reduce some of the code generation within macros to make the code more readable and
easier to navigate with an IDE.

---

refactor(agents): use socket type for local endpoints

---

refactor(agents): agents library should handle tonic services

When we moved away from message bus this library became underutilized with the tonic logic
being moved to the core service.
This changes moves the server logic back to the library so it may be reused.


tiagolobocastro:
bors try

control-plane/agents/common/src/lib.rs
```diff
@@ -22,30 +23,93 @@ pub enum ServiceError {
     GrpcServer { source: tonic::transport::Error },
 }
 
+type LayerStack = tower::layer::util::Stack<OpenTelServer, tower::layer::util::Identity>;
 /// Runnable service with N subscriptions which listen on a given
 /// message bus channel on a specific ID.
```
Abhinandan-Purkait: Should this doc comment change?

utils/utils-lib/src/constants.rs
```diff
@@ -65,13 +65,13 @@ pub const RUST_LOG_SILENCE_DEFAULTS: &str =
...
 /// The default value to be assigned as cluster agent GRPC server addr if not overridden
-pub const DEFAULT_CLUSTER_AGENT_SERVER_ADDR: &str = "https://0.0.0.0:11500";
+pub const DEFAULT_CLUSTER_AGENT_SERVER_ADDR: &str = "0.0.0.0:11500";
 
 /// The default value to be assigned as cluster agent GRPC client addr if not overridden
 pub const DEFAULT_CLUSTER_AGENT_CLIENT_ADDR: &str = "https://ha-cluster-agent:11500";
 
 /// The default value to be assigned as node-agent GRPC server addr if not overridden
-pub const DEFAULT_NODE_AGENT_SERVER_ADDR: &str = "https://0.0.0.0:11600";
+pub const DEFAULT_NODE_AGENT_SERVER_ADDR: &str = "0.0.0.0:11600";
```
Abhinandan-Purkait, line 74: By default it uses `https` scheme?

Abhinandan-Purkait (approved):
LGTM. Thanks

utils/utils-lib/src/constants.rs
tiagolobocastro: Actually we don't use the scheme, we had Uri on the cli but then we strip it out and use only the authority which is the socket address in this case. And this is because that's how the tonic server works, it binds to a socket. Tls is provided as a separate configuration.
```rust
            cli_args.grpc_endpoint,
                .grpc_endpoint
                .authority()
                .unwrap()
                .to_string()
                .parse()
                .unwrap(),
```

control-plane/agents/common/src/lib.rs
tiagolobocastro: Doh :)

utils/utils-lib/src/constants.rs
Abhinandan-Purkait: Ahh right. Thanks for explaining.

control-plane/agents/common/src/lib.rs
tiagolobocastro: Fixed and updated other comments throughout the code.

tiagolobocastro:
bors merge

tiagolobocastro:
bors cancel

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

PR #324, opened 2022-09-02
chore: update conversion from v1 replica grpc to transport replica struct

Ab-hishek:
Some REST replica operations like `get_node_pool_replicas`, `del_pool_replica` were failing because these apply some filters based on poolid(or poolname). But since in v1 replica we only get pooluuid in response from mayastor, the filters failed to recognise the replicas on respective pools.

Hence, this PR updates the rpc submodule dependency to have the poolname also embedded in the response and the corresponding type conversions needed for it so that the above operations pass successfully.

Signed-off-by: Abhishek Agarwal <abhiagarwal.agarwal2@gmail.com>


Abhinandan-Purkait (approved):

tiagolobocastro (approved):

Ab-hishek:
bors merge

PR #325, opened 2022-09-02
fix: add default tag value for image version

vharsh:
Signed-off-by: Harsh Vardhan <harsh.vardhan@datacore.com>

### DIFF
--- a/chart/values.yaml
+++ b/chart/values.yaml
@@ -6,6 +6,7 @@ mayastorCP:
   registry:
   pullPolicy: Always
   logLevel: info
+  tag: develop
 
 base:
   initContainers:


chart/values.yaml
```diff
@@ -6,6 +6,7 @@ mayastorCP:
   registry:
   pullPolicy: Always
   logLevel: info
+  tag: develop
```
vharsh, line 9: @rohan2794 Please share the appropriate tag for `v1.0.2` image-tag.

GlennBullingham (changes requested):
This cannot be merged directly into master.  Merging into master, in our current "GitFlow"-based branching scheme creates a new release, which by definition can't be 1.0.2 (which is already released).

What are we trying to achieve here?  v1.0.2 already has Helm charts and a documented installation process which flows from them.

tiagolobocastro:
The helm chart on 1* doesn't work on its own; it was used as means to generate yaml through a script which sets some other parameters; the resulting yaml is then checked into the deploy folder [here](https://github.com/openebs/mayastor-control-plane/tree/master/deploy).

vharsh:
@GlennBullingham Agreed, but the current `1.0.2` chart cannot be used right now and the error from `helm` binary isn't super indicative of this missing value.
```bash
$ helm template .....
Error: YAML parse error on mayastor-control-plane/templates/core-agents-deployment.yaml: error converting YAML to JSON: yaml: line 44: mapping values are not allowed in this context
helm.go:84: [debug] error converting YAML to JSON: yaml: line 44: mapping values are not allowed in this context
YAML parse error on mayastor-control-plane/templates/core-agents-deployment.yaml
```

The file has 43 lines and the error is on a non-existent line.

GlennBullingham:
@vharsh  @tiagolobocastro 

I think it boils down to, we've never documented or supported our "internal" Helm charts as a deployment method for any version prior to v2.0.  As Tiago said, rightly or wrongly, they've been there historically to support the scripted generation of static definition files, which are then provided for users in the `/deploy` folder of their respective repo (and documented).  There is therefore no need, in my opinion, to fix them for direct consumption.  Particularly given that the 2.0 release, with its new unified Helm chart for user consumption, is being readied for release in the next 2-3 weeks.

That's my _opinion_.  If we still collectively feel there is value in doing this, then we should make it a patch release (v1.0.3), rather than revise an existing published release.

vharsh:
Right, I was considering a sub-release, but I'll close this PR for now as we aren't planning to edit an already done release.

PR #329, opened 2022-09-06
fix(replica-openapi): revert back changes for poolid in Replica & ReplicaSpec openapi

Ab-hishek:
This change is needed so that no breaking change is experienced in terms of system tests.

Signed-off-by: Abhishek Agarwal <abhiagarwal.agarwal2@gmail.com>

### DIFF
--- a/control-plane/rest/openapi-specs/v0_api_spec.yaml
+++ b/control-plane/rest/openapi-specs/v0_api_spec.yaml
@@ -2226,7 +2226,7 @@ components:
       properties:
         node:
           $ref: '#/components/schemas/NodeId'
-        poolId:
+        pool:
           $ref: '#/components/schemas/PoolId'
         poolUuid:
           $ref: '#/components/schemas/PoolUuid'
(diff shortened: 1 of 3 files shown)


Abhinandan-Purkait (approved):

tiagolobocastro (approved):

tiagolobocastro:
Aren't we missing changes on the `From<xxx> To <Replica>` to use the changed field name pool?

Ab-hishek:
That is not required coz the models for openapi are auto generated and regarding conversion from any type to the `models::Replica` is done via a function where we pass parameters and have a `From` trait implemented for it. And for `poolid` since its of a type macro rule, the required conversions are already implemented.

tiagolobocastro:
> That is not required coz the models for openapi are auto generated and regarding conversion from any type to the `models::Replica` is done via a function where we pass parameters and have a `From` trait implemented for it. And for `poolid` since its of a type macro rule, the required conversions are already implemented.

No, it's because openapi::models::Replica is created via a function so the parameters are unnamed; as longer as the order is kept no changes should be required.

Ab-hishek:
Yeah, I meant the sequence of parameters are not changed.

Ab-hishek:
bors merge

PR #331, opened 2022-09-07
feat(ha): add apis for volume republish and targets destroy for switchover 

Abhinandan-Purkait:
1. Add volume republish api.
2. Add volume destroy shutdown targets api.
3. Add changes to the reconcilers to not act on shutdown states.

### DIFF
--- a/common/src/types/v0/transport/nexus.rs
+++ b/common/src/types/v0/transport/nexus.rs
@@ -70,14 +70,16 @@ rpc_impl_string_uuid!(NexusId, "UUID of a nexus");
 /// Nexus State information
 #[derive(Serialize, Deserialize, Debug, Clone, EnumString, ToString, Eq, PartialEq)]
 pub enum NexusStatus {
-    /// Default Unknown state
+    /// Default Unknown state.
     Unknown = 0,
-    /// healthy and working
+    /// Healthy and working.
     Online = 1,
-    /// not healthy but is able to serve IO (i.e. rebuild is in progress)
+    /// Not healthy but is able to serve IO (i.e. rebuild is in progress).
     Degraded = 2,
-    /// broken and unable to serve IO
+    /// Broken and unable to serve IO.
     Faulted = 3,
+    /// Shutdown state, i.e. blocked from serving IO.
+    Shutdown = 4,
 }
 impl Default for NexusStatus {
     fn default() -> Self {
@@ -90,6 +92,7 @@ impl From<i32> for NexusStatus {
             1 => Self::Online,
             2 => Self::Degraded,
             3 => Self::Faulted,
+            4 => Self::Shutdown,
             _ => Self::Unknown,
         }
     }
@@ -97,10 +100,11 @@ impl From<i32> for NexusStatus {
 impl From<NexusStatus> for models::NexusState {
     fn from(src: NexusStatus) -> Self {
         match src {
-            NexusStatus::Unknown => Self::Unknown,
             NexusStatus::Online => Self::Online,
             NexusStatus::Degraded => Self::Degraded,
             NexusStatus::Faulted => Self::Faulted,
+            NexusStatus::Unknown => Self::Unknown,
+            NexusStatus::Shutdown => Self::Shutdown,
         }
     }
 }
@@ -371,6 +375,25 @@ impl DestroyNexus {
     /// Return a reference to the disowners.
     pub fn disowners(&self) -> &NexusOwners {
         &self.disowners
+    }
+}
+
+/// Shutdown Nexus Request.
+#[derive(Serialize, Deserialize, Default, Debug, Clone, Eq, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct ShutdownNexus {
+    /// The uuid of the nexus.
+    uuid: NexusId,
+}
+
+impl ShutdownNexus {
+    /// Create a new `ShutdownNexus` using `uuid`.
+    pub fn new(uuid: NexusId) -> Self {
+        Self { uuid }
+    }
+    /// Get uuid of the nexus.
+    pub fn uuid(&self) -> NexusId {
+        self.uuid.clone()
     }
 }
 
--- a/control-plane/agents/src/bin/core/controller/reconciler/volume/hot_spare.rs
+++ b/control-plane/agents/src/bin/core/controller/reconciler/volume/hot_spare.rs
@@ -64,6 +64,7 @@ async fn hot_spare_reconcile(
             hot_spare_nexus_reconcile(&mut volume, &volume_state, context).await
         }
         VolumeStatus::Faulted => PollResult::Ok(PollerState::Idle),
+        VolumeStatus::Shutdown => PollResult::Ok(PollerState::Idle),
     }
 }
 
--- a/control-plane/agents/src/bin/core/nexus/operations.rs
+++ b/control-plane/agents/src/bin/core/nexus/operations.rs
@@ -1,11 +1,13 @@
 use crate::controller::{
     registry::Registry,
     resources::{
-        operations::{ResourceLifecycle, ResourceOffspring, ResourceSharing},
+        operations::{
+            ResourceLifecycle, ResourceOffspring, ResourceSharing, ResourceShutdownOperations,
+        },
         operations_helper::{GuardedOperationsHelper, OperationSequenceGuard},
         OperationGuardArc,
     },
-    wrapper::ClientOps,
+    wrapper::{ClientOps, GetterOps},
 };
 use agents::errors::{SvcError, SvcError::CordonedNode};
 use common_lib::types::v0::{
@@ -16,7 +18,7 @@ use common_lib::types::v0::{
     transport::{
         child::Child,
         nexus::{CreateNexus, DestroyNexus, Nexus, ShareNexus, UnshareNexus},
-        AddNexusChild, RemoveNexusChild,
+        AddNexusChild, RemoveNexusChild, ShutdownNexus,
     },
 };
 
@@ -243,5 +245,41 @@ impl ResourceOffspring for Option<&mut OperationGuardArc<NexusSpec>> {
         } else {
             node.remove_child(request).await
         }
+    }
+}
+
+#[async_trait::async_trait]
+impl ResourceShutdownOperations for OperationGuardArc<NexusSpec> {
+    type RemoveShutdownTargets = ();
+    type Shutdown = ShutdownNexus;
+
+    async fn shutdown(
+        &mut self,
+        registry: &Registry,
+        request: &Self::Shutdown,
+    ) -> Result<(), SvcError> {
+        let node_id = self.as_ref().node.clone();
+        let nexus_id = self.uuid();
+        let node = registry.get_node_wrapper(&node_id).await?;
+        let nexus_state = match node.nexus(nexus_id).await {
+            None => self.as_ref().into(),
+            Some(state) => state,
+        };
+        let spec_clone = self
+            .start_update(registry, &nexus_state, NexusOperation::Shutdown)
+            .await?;
+
+        let result = node.shutdown_nexus(request).await;
+        self.complete_update(registry, result, spec_clone).await?;
+        Ok(())
+    }
+
+    async fn remove_shutdown_targets(
+        &mut self,
+        _registry: &Registry,
+        _request: &Self::RemoveShutdownTargets,
+    ) -> Result<(), SvcError> {
+        // Not applicable for nexus
+        unimplemented!()
     }
 }
--- a/control-plane/agents/src/bin/core/volume/operations.rs
+++ b/control-plane/agents/src/bin/core/volume/operations.rs
@@ -5,7 +5,7 @@ use crate::{
         resources::{
             operations::{
                 ResourceLifecycle, ResourceOwnerUpdate, ResourcePublishing, ResourceReplicas,
-                ResourceSharing,
+                ResourceSharing, ResourceShutdownOperations,
             },
             operations_helper::{
                 GuardedOperationsHelper, OperationSequenceGuard, ResourceSpecsLocked,
@@ -25,13 +25,15 @@ use common_lib::{
             volume::{VolumeOperation, VolumeSpec},
         },
         transport::{
-            CreateVolume, DestroyNexus, DestroyReplica, DestroyVolume, NexusId, Protocol,
-            PublishVolume, Replica, ReplicaOwners, SetVolumeReplica, ShareNexus, ShareVolume,
-            UnpublishVolume, UnshareNexus, UnshareVolume, Volume,
+            CreateVolume, DestroyNexus, DestroyReplica, DestroyShutdownTargets, DestroyVolume,
+            NexusId, Protocol, PublishVolume, Replica, ReplicaOwners, RepublishVolume,
+            SetVolumeReplica, ShareNexus, ShareVolume, ShutdownNexus, UnpublishVolume,
+            UnshareNexus, UnshareVolume, Volume,
         },
     },
 };
 use std::ops::Deref;
+use tracing::debug;
 
 #[async_trait::async_trait]
 impl ResourceLifecycle for OperationGuardArc<VolumeSpec> {
@@ -263,6 +265,7 @@ impl ResourcePublishing for OperationGuardArc<VolumeSpec> {
     type Publish = PublishVolume;
     type PublishOutput = Volume;
     type Unpublish = UnpublishVolume;
+    type Republish = RepublishVolume;
 
     async fn publish(
         &mut self,
@@ -271,7 +274,7 @@ impl ResourcePublishing for OperationGuardArc<VolumeSpec> {
     ) -> Result<Self::PublishOutput, SvcError> {
         let specs = registry.specs();
         let state = registry.get_volume_state(&request.uuid).await?;
-        let nexus_node = get_volume_target_node(registry, &state, request).await?;
+        let nexus_node = get_volume_target_node(registry, &state, request, false).await?;
         let nexus_id = NexusId::new();
 
         let operation =
@@ -376,6 +379,95 @@ impl ResourcePublishing for OperationGuardArc<VolumeSpec> {
 
         self.complete_update(registry, result, spec_clone).await
     }
+
+    async fn republish(
+        &mut self,
+        registry: &Registry,
+        request: &Self::Republish,
+    ) -> Result<Self::PublishOutput, SvcError> {
+        let specs = registry.specs();
+        let spec = self.as_ref().clone();
+        let state = registry.get_volume_state(&request.uuid).await?;
+        // If the volume is not published then it should issue publish call rather than republish.
+        let volume_target = match spec.target {
+            None => {
+                return Err(SvcError::VolumeNotPublished {
+                    vol_id: request.uuid.to_string(),
+                })
+            }
+            Some(target) => target,
+        };
+        let nexus_node = get_volume_target_node(registry, &state, request, true).await?;
+        let newer_nexus_id = NexusId::new();
+        let operation =
+            VolumeOperation::Republish((nexus_node.clone(), newer_nexus_id.clone(), request.share));
+        let spec_clone = self.start_update(registry, &state, operation).await?;
+        let result = specs.nexus(volume_target.nexus()).await;
+        let mut older_nexus = self
+            .validate_update_step(registry, result, &spec_clone)
+            .await?;
+
+        let older_nexus_id = older_nexus.uuid().clone();
+        // shutdown the older nexus before newer nexus creation.
+        // todo: Handle scenario when node is offline and shutdown fails.
+        let result = older_nexus
+            .shutdown(registry, &ShutdownNexus::new(older_nexus_id))
+            .await;
+        self.validate_update_step(registry, result, &spec_clone)
+            .await?;
+
+        // Create a Nexus on the requested or auto-selected node
+        let result = specs
+            .volume_create_nexus(registry, &nexus_node, &newer_nexus_id, &spec_clone)
+            .await;
+        let (mut nexus, nexus_state) = self
+            .validate_update_step(registry, result, &spec_clone)
+            .await?;
+
+        let (volume_id, last_nexus_id) = {
+            let volume_spec = self.lock();
+            (volume_spec.uuid.clone(), volume_spec.last_nexus_id.clone())
+        };
+
+        // Share the Nexus
+        let result = match nexus
+            .share(
+                registry,
+                &ShareNexus::from((&nexus_state, None, request.share)),
+            )
+            .await
+        {
+            Ok(_) => Ok(()),
+            Err(error) => {
+                // Since we failed to share, we'll revert back to the previous state.
+                // If we fail to do this inline, the reconcilers will pick up the slack.
+                nexus
+                    .destroy(registry, &DestroyNexus::from(nexus_state).with_disown_all())
+                    .await
+                    .ok();
+                Err(error)
+            }
+        };
+
+        self.complete_update(registry, result, spec_clone).await?;
+
+        // If there was a previous nexus we should delete the persisted NexusInfo structure.
+        // todo: This needs to be revisited, as there might be inconsistency on child state after
+        // creation of nexus.
+        if let Some(nexus_id) = last_nexus_id {
+            ResourceSpecsLocked::delete_nexus_info(
+                &NexusInfoKey::new(&Some(volume_id), &nexus_id),
+                registry,
+            )
+            .await;
+        }
+
+        let volume = registry.get_volume(&request.uuid).await?;
+        registry
+            .notify_if_degraded(&volume, PollTriggerEvent::VolumeDegraded)
+            .await;
+        Ok(volume)
+    }
 }
 
 #[async_trait::async_trait]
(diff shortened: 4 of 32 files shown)


control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -274,19 +274,20 @@ impl ResourcePublishing for OperationGuardArc<VolumeSpec> {
...
-        let operation =
-            VolumeOperation::Publish((nexus_node.clone(), nexus_id.clone(), request.share));
+        let operation = VolumeOperation::Publish((
+            nexus_node.clone(),
+            nexus_id.clone(),
+            request.share,
+            request.republish(),
+        ));
         let spec_clone = self.start_update(registry, &state, operation).await?;
-
```
tiagolobocastro: keep the newlines to separate the `start_update` and `validate_update_step` from the rest

control-plane/agents/src/bin/core/volume/specs.rs
```diff
@@ -1111,15 +1111,6 @@ pub(crate) async fn get_volume_target_node(
     status: &VolumeState,
     request: &PublishVolume,
 ) -> Result<NodeId, SvcError> {
-    // We can't configure a new target_node if the volume is currently published
```
tiagolobocastro, line 1114: we could keep this and check if request is not a republish?

control-plane/agents/src/bin/core/volume/specs.rs
```diff
@@ -1219,16 +1210,20 @@ impl SpecOperationsHelper for VolumeSpec {
...
-                            Err(SvcError::VolumeAlreadyPublished {
-                                vol_id: self.uuid_str(),
-                                node: target.node().to_string(),
-                                protocol: format!("{:?}", target.protocol()),
-                            })
+                            if !republish {
+                                Err(SvcError::VolumeAlreadyPublished {
+                                    vol_id: self.uuid_str(),
+                                    node: target.node().to_string(),
+                                    protocol: format!("{:?}", target.protocol()),
```
tiagolobocastro: does protocol have a to_string?

control-plane/csi-driver/controller/src/client.rs
```diff
@@ -283,11 +283,12 @@ impl IoEngineApiClient {
         volume_id: &uuid::Uuid,
         node: Option<&str>,
         protocol: VolumeShareProtocol,
+        republish: bool,
```
tiagolobocastro: I'm not sure that we should be exposing this via rest to prevent misuse..

control-plane/grpc/proto/v1/volume/volume.proto
```diff
@@ -169,6 +169,8 @@ message PublishVolumeRequest {
   optional string target_node = 2;
   // share protocol
   optional VolumeShareProtocol share = 3;
+  // republish the volume on the node, without removing existing nexus if any
```
```suggestion
  // allows republishing the volume on the node by shutting down the existing target first
  // once the new target is ready you must destroy the previous target
```

tiagolobocastro: Actually I'm starting to wonder if we should have a republish message, which fails if the volume is not currently published and maybe returns the old target uuid so that we may provide the cluster agent a uuid to destroy old targets, wdyt?

tests/bdd/features/ha/nexus-switchover/nexus_switchover.feature
```diff
@@ -0,0 +1,10 @@
+Feature: Nexus Switchover IO test
```
tiagolobocastro: Let's replace all nexuses here with target please

control-plane/agents/src/bin/core/volume/operations.rs
Abhinandan-Purkait: Oops, will add it back.

control-plane/agents/src/bin/core/volume/specs.rs
Abhinandan-Purkait: Yeah. Right

control-plane/agents/src/bin/core/volume/specs.rs
Abhinandan-Purkait: This was exsisting code, and seems like we need to change the SvcError's parameter. Will take this up as refactor PR.

control-plane/csi-driver/controller/src/client.rs
Abhinandan-Purkait: Yeah, I also had this thought but I added it so that we can have the bdd-tests, as the bdd tests use the rest api.

control-plane/grpc/proto/v1/volume/volume.proto
Abhinandan-Purkait: That's a very good suggestion. This will take some undoing of the things in this PR.

tests/bdd/features/ha/nexus-switchover/nexus_switchover.feature
Abhinandan-Purkait: Done.

control-plane/grpc/proto/v1/volume/volume.proto
Abhinandan-Purkait: > by shutting down the existing target first

Should we add this? This step will be taken care of by the cluster agent, so would not be a part of publish call

tiagolobocastro: Isn't shutdown done by the core agent during republish?

Abhinandan-Purkait: I think the cluster agent does that. It shut downs the nexus and then make a republish call.

tiagolobocastro: I'm not sure that sounds right, the cluster agent does not own the nexus, it doesn't even know its uuid, only it's name.

Abhinandan-Purkait: Can it not get it from the volume spec? I was told the cluster agent drives all the steps.

Abhinandan-Purkait:
bors try

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -376,6 +376,93 @@ impl ResourcePublishing for OperationGuardArc<VolumeSpec> {
...
+
+        let nexus_node = get_volume_target_node(registry, &state, request, true).await?;
+        let newer_nexus_id = NexusId::new();
+
+        let operation =
+            VolumeOperation::Republish((nexus_node.clone(), newer_nexus_id.clone(), request.share));
+        let spec_clone = self.start_update(registry, &state, operation).await?;
+
+        let volume_target = spec_clone.target.as_ref().expect("already validated");
+        let mut older_nexus = specs.nexus(volume_target.nexus()).await?;
```
tiagolobocastro: cannot use `?` here, must use validate_update_step ?

common/src/types/v0/transport/nexus.rs
```diff
@@ -374,6 +378,14 @@ impl DestroyNexus {
     }
 }
 
+/// Shutdown Nexus Request.
+#[derive(Serialize, Deserialize, Default, Debug, Clone, Eq, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct ShutdownNexus {
+    /// The uuid of the nexus.
+    pub uuid: NexusId,
```
tiagolobocastro: we've got loads of pub which around these files for "historical reasons", going forwards let's user getter instead, eg:
`fn uuid(&self) -> &NexusId { ... }`

control-plane/agents/src/bin/core/controller/reconciler/volume/hot_spare.rs
```diff
@@ -64,6 +64,7 @@ async fn hot_spare_reconcile(
             hot_spare_nexus_reconcile(&mut volume, &volume_state, context).await
         }
         VolumeStatus::Faulted => PollResult::Ok(PollerState::Idle),
+        VolumeStatus::Shutdown => PollResult::Ok(PollerState::Idle),
```
tiagolobocastro, line 67: I think it's time for the VolumeStatus to break out of the NexusStatus?

control-plane/agents/src/bin/core/nexus/operations.rs
```diff
@@ -245,3 +247,39 @@ impl ResourceOffspring for Option<&mut OperationGuardArc<NexusSpec>> {
...
+    type RemoveShutdownTargets = ();
+    type Shutdown = ShutdownNexus;
+
+    async fn shutdown(
+        &mut self,
+        registry: &Registry,
+        request: &Self::Shutdown,
+    ) -> Result<(), SvcError> {
+        let node_id = &self.lock().clone().node;
+        let nexus_id = &self.lock().clone().uuid;
```
tiagolobocastro:
```suggestion
        let node_id = self.as_ref().node.clone();
        let nexus_id = self.uuid();
```

control-plane/agents/src/bin/core/nexus/operations.rs
```diff
@@ -245,3 +247,39 @@ impl ResourceOffspring for Option<&mut OperationGuardArc<NexusSpec>> {
...
+        let node = registry.get_node_wrapper(node_id).await?;
+        let nexus_state = match node.nexus(nexus_id).await {
+            None => (&(self.lock().clone())).into(),
```
tiagolobocastro:
```suggestion
            None => self.as_ref().into(),
```

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -376,6 +376,93 @@ impl ResourcePublishing for OperationGuardArc<VolumeSpec> {
...
+        let older_nexus_id = older_nexus.lock().clone().uuid;
```
```suggestion
        let older_nexus_id = older_nexus.as_ref().uuid;
```

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -376,6 +376,93 @@ impl ResourcePublishing for OperationGuardArc<VolumeSpec> {
...
+        // shutdown the older nexus before newer nexus creation.
+        older_nexus
+            .shutdown(
+                registry,
+                &ShutdownNexus {
+                    uuid: older_nexus_id,
+                },
+            )
+            .await?;
```
tiagolobocastro, line 417: can't return here

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -376,6 +376,93 @@ impl ResourcePublishing for OperationGuardArc<VolumeSpec> {
...
+        let (mut nexus, nexus_state) = self
+            .validate_update_step(registry, result, &spec_clone)
+            .await?;
+
+        let (volume_id, last_nexus_id) = {
+            let volume_spec = self.lock();
+            (volume_spec.uuid.clone(), volume_spec.last_nexus_id.clone())
+        };
+
+        // Share the Nexus if it was requested
```
tiagolobocastro: I think republish should always imply re-sharing, so make it a requirement?

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -376,6 +376,93 @@ impl ResourcePublishing for OperationGuardArc<VolumeSpec> {
...
+        self.complete_update(registry, result, spec_clone).await?;
+
+        // If there was a previous nexus we should delete the persisted NexusInfo structure.
+        if let Some(nexus_id) = last_nexus_id {
+            ResourceSpecsLocked::delete_nexus_info(
+                &NexusInfoKey::new(&Some(volume_id), &nexus_id),
+                registry,
+            )
+            .await;
+        }
```
tiagolobocastro: I wonder if this needs to be different for the republish, now.
Should we check if the previous nexus updated its nexus info, and if it did we should re-republish or update the nexus with the new healthy children list?

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -407,3 +494,45 @@ impl ResourceReplicas for OperationGuardArc<VolumeSpec> {
...
+        registry: &Registry,
+        request: &Self::RemoveShutdownTargets,
+    ) -> Result<(), SvcError> {
+        let shutdown_nexuses = registry
+            .specs()
+            .get_volume_shutdown_nexuses_guards(&request.uuid)
+            .await?;
+        let mut all_targets_destroyed = true;
+        for mut nexus in shutdown_nexuses {
+            let nexus_spec = nexus.lock().clone();
```
```suggestion
            let nexus_spec = nexus.as_ref();
```

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -407,3 +494,45 @@ impl ResourceReplicas for OperationGuardArc<VolumeSpec> {
...
+            if nexus.destroy(registry, &destroy_req).await.is_err() {
+                // flag that all targets weren't destroyed
+                all_targets_destroyed = false;
+            }
+        }
+        if all_targets_destroyed {
+            Ok(())
+        } else {
+            Err(SvcError::Deleting {})
+        }
```
tiagolobocastro: I think here we might simply return success as long as the specs are marked as deleting? We'd need to ensure we updated etcd correctly.

control-plane/agents/src/bin/core/volume/service.rs
```diff
@@ -198,6 +227,19 @@ impl Service {
...
+    #[tracing::instrument(level = "info", skip(self), err, fields(volume.uuid = %request.uuid))]
+    pub(super) async fn destroy_shutdown_target(
+        &self,
+        request: &DestroyShutdownTargets,
+    ) -> Result<(), SvcError> {
+        let mut volume = self.specs().volume(&request.uuid).await?;
+        volume
+            .remove_shutdown_targets(&self.registry, request)
+            .await?;
+        Ok(())
```
tiagolobocastro: keeping with the current style of the file:
```suggestion
        volume
            .remove_shutdown_targets(&self.registry, request)
            .await
```

control-plane/agents/src/bin/core/volume/specs.rs
```diff
@@ -394,6 +394,28 @@ impl ResourceSpecsLocked {
             .cloned()
             .collect()
     }
+
+    /// Get a list of protected NexusSpecs's which are associated with the given volume `id`
```
tiagolobocastro, line 399: nexus guards?

control-plane/agents/src/bin/core/volume/specs.rs
```diff
@@ -394,6 +394,28 @@ impl ResourceSpecsLocked {
...
+    /// and are currently in shutdown state.
+    pub(crate) async fn get_volume_shutdown_nexuses_guards(
+        &self,
+        id: &VolumeId,
+    ) -> Result<Vec<OperationGuardArc<NexusSpec>>, SvcError> {
+        let mut shutdown_nexuses = Vec::new();
+        for nexus in self.read().nexuses.values() {
+            let nexus_spec = nexus.lock().clone();
```
tiagolobocastro: why not use iter like the example above?

control-plane/agents/src/bin/core/volume/specs.rs
```diff
@@ -394,6 +394,28 @@ impl ResourceSpecsLocked {
...
+            if nexus_spec.name == id.to_string() && nexus_spec.is_shutdown() {
+                shutdown_nexuses.push(nexus_spec);
+            }
+        }
+        let mut shutdown_nexuses_guards = Vec::with_capacity(shutdown_nexuses.len());
+        for nexus in shutdown_nexuses {
+            let nexus_guard = self.nexus(&nexus.uuid).await?;
```
tiagolobocastro: this will error out if we can't grab one of the nexuses, is this the right thing to do?
What about returning only the Resource<Nexus>, and then try to guard them one at a time, and if we fail to guard one of them skip and fail the shutdown call, but at least we've cleaned up some of them?

control-plane/agents/src/bin/core/volume/specs.rs
```diff
@@ -1240,6 +1265,17 @@ impl SpecOperationsHelper for VolumeSpec {
                     }),
                 },
             },
+            VolumeOperation::Republish((_, _, protocol)) => match protocol {
```
tiagolobocastro, line 1267: I think protocol should not be optional here.

control-plane/rest/openapi-specs/v0_api_spec.yaml
```diff
@@ -1437,6 +1437,15 @@ paths:
...
             $ref: '#/components/schemas/VolumeShareProtocol'
+        - in: query
+          name: republish
+          x-api-scope: internal
+          description: |-
+            Allows republishing the volume on the node by shutting down the existing target first.
+          required: false
+          schema:
+            type: boolean
+            default: false
```
tiagolobocastro: does a default here make sense if republish is already optional?

tiagolobocastro (commented):
Since this is quite large, might want to consider splitting this into different logical commits

control-plane/agents/src/bin/core/volume/operations.rs
Abhinandan-Purkait: should it be rather - > `let older_nexus_id = older_nexus.uuid().clone();` Otherwise its giving moved error.

control-plane/agents/src/bin/core/controller/reconciler/volume/hot_spare.rs
Abhinandan-Purkait: Can we break out? We use the nexus status to set volume status, what will we set if Nexus status is shutdown?

control-plane/agents/src/bin/core/volume/operations.rs
Abhinandan-Purkait: The specs would be marked as deleting as that is the actual destroy call that takes care of it. I have added this error in case cluster agent needs to do something with the error. Otherwise we can simply return success.

control-plane/agents/src/bin/core/volume/operations.rs
tiagolobocastro: Right yes needs to be cloned because it's moved.

control-plane/agents/src/bin/core/volume/operations.rs
tiagolobocastro: I think cluster agent is happy to have us retry on our reconcilers IIRC @mtzaurus ?
But to be able to do that I think we need to check what kind of errors we got, example if we could not reach etcd then we can't ignore the error because if we restart, we'll "forget" that we tried to do this and won't reconcile.

control-plane/agents/src/bin/core/volume/operations.rs
Abhinandan-Purkait: Made it a requirement

control-plane/agents/src/bin/core/volume/operations.rs
Abhinandan-Purkait: Done.

control-plane/agents/src/bin/core/volume/operations.rs
Abhinandan-Purkait: Have added comments for now, but needs revisiting after final discussion.

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -376,6 +379,96 @@ impl ResourcePublishing for OperationGuardArc<VolumeSpec> {
...
+    async fn republish(
+        &mut self,
+        registry: &Registry,
+        request: &Self::Republish,
+    ) -> Result<Self::PublishOutput, SvcError> {
+        let specs = registry.specs();
+        let state = registry.get_volume_state(&request.uuid).await?;
+
+        // If the volume is not published then it should issue publish call rather than republish.
+        if state.target.is_none() {
```
tiagolobocastro: should be from the spec, not state? and if we get spec we can skip the expect below, example:
```rust
let target = match spec.target { Some(target) => target, None => return NotPublish... }
```

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -407,3 +500,67 @@ impl ResourceReplicas for OperationGuardArc<VolumeSpec> {
...
+        registry: &Registry,
+        request: &Self::RemoveShutdownTargets,
+    ) -> Result<(), SvcError> {
+        let shutdown_nexuses = registry
+            .specs()
+            .get_volume_shutdown_nexuses(&request.uuid)
+            .await;
+
+        for mut nexus_res in shutdown_nexuses {
+            let nexus_spec = nexus_res.lock().deref().clone();
```
tiagolobocastro: Don't think we need to do this ahead of time.
Below we can simply do guard.as_ref().clone() ?

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -407,3 +500,67 @@ impl ResourceReplicas for OperationGuardArc<VolumeSpec> {
...
+                        },
+                    }
+                }
+                Err(err) => {
+                    debug!(
+                        err = %err,
+                        nexus.id = nexus_res.uuid().to_string(),
+                        "Encountered error while destroying shutdown nexus, exiting.."
+                    );
+                    return Err(err);
```
tiagolobocastro: Why exit here, just try the next ones.
```suggestion
```

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -407,3 +500,67 @@ impl ResourceReplicas for OperationGuardArc<VolumeSpec> {
...
+                            }
+                        },
+                    }
+                }
+                Err(err) => {
+                    debug!(
+                        err = %err,
+                        nexus.id = nexus_res.uuid().to_string(),
+                        "Encountered error while destroying shutdown nexus, exiting.."
+                    );
```
```suggestion
                    debug!(
                        error = %err,
                        nexus.uuid = %nexus_res.uuid(),
                        "Failed to guard one of the destroying shutdown nexus"
                    );
```

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -407,3 +500,67 @@ impl ResourceReplicas for OperationGuardArc<VolumeSpec> {
...
+                            .await;
+                        }
+                        Err(err) => match err {
+                            SvcError::Store { .. } => return Err(err),
+                            SvcError::StoreSave { .. } => return Err(err),
+                            _ => {
+                                debug!(
+                                    err = %err,
+                                    nexus.id = destroy_req.uuid.to_string(),
+                                    "Encountered error while destroying shutdown nexus"
```
```suggestion
                                debug!(
                                    error = %err,
                                    nexus.uuid = %destroy_req.uuid,
                                    "Encountered error while destroying shutdown nexus"
```

control-plane/agents/src/bin/core/volume/specs.rs
```diff
@@ -394,6 +395,24 @@ impl ResourceSpecsLocked {
...
+    /// and are currently in shutdown state.
+    pub(crate) async fn get_volume_shutdown_nexuses(
+        &self,
+        id: &VolumeId,
+    ) -> Vec<ResourceMutex<NexusSpec>> {
+        self.read()
+            .nexuses
+            .values()
+            .filter(|nexus| {
+                let nexus_spec = nexus.lock().clone();
```
tiagolobocastro: don't need to clone?

control-plane/agents/src/bin/core/volume/specs.rs
```diff
@@ -394,6 +395,24 @@ impl ResourceSpecsLocked {
...
+                nexus_spec.name == id.to_string() && nexus_spec.is_shutdown()
```
tiagolobocastro: `nexus_spec.name == id.as_str()`

control-plane/grpc/src/operations/volume/traits.rs
```diff
@@ -993,6 +1020,16 @@ impl From<&dyn PublishVolumeInfo> for PublishVolume {
...
 
+impl From<&dyn PublishVolumeInfo> for RepublishVolume {
+    fn from(data: &dyn PublishVolumeInfo) -> Self {
+        Self {
+            uuid: data.uuid(),
+            target_node: data.target_node(),
+            share: data.share().unwrap(),
+        }
+    }
+}
```
tiagolobocastro: Is this needed?
and also `impl PublishVolumeInfo for RepublishVolume {` ?

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -376,6 +379,96 @@ impl ResourcePublishing for OperationGuardArc<VolumeSpec> {
...
+            .validate_update_step(registry, result, &spec_clone)
+            .await?;
+
+        let older_nexus_id = older_nexus.uuid().clone();
+        // shutdown the older nexus before newer nexus creation.
+        let result = older_nexus
+            .shutdown(registry, &ShutdownNexus::new(older_nexus_id))
+            .await;
+        self.validate_update_step(registry, result, &spec_clone)
+            .await?;
```
tiagolobocastro, line 417: What happens if the node is offline and we can't shutdown? How should we handle this?
We can leave it for another PR? wdyt?

tiagolobocastro (commented):
What happens if the node is offline and we can't shutdown? How should we handle this?
We can leave it for another PR, wdyt?

control-plane/agents/src/bin/core/volume/operations.rs
Abhinandan-Purkait: Yes, right, this will error out if node goes offline, we need some way to handle this. Maybe we should allow grpc server not implemented errors and use the reservation keys to handle that scenario, and for any other error we error out?

control-plane/grpc/src/operations/volume/traits.rs
Abhinandan-Purkait: Yes, this is needed since we have seperated the structures for PublishVolume and RepublishVolume.

control-plane/agents/src/bin/core/volume/operations.rs
Abhinandan-Purkait: Done.

control-plane/agents/src/bin/core/volume/operations.rs
Abhinandan-Purkait: For now have added a comment. Needs revisiting again.

tiagolobocastro (approved):

Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors merge

PR #333, opened 2022-09-09
feat(csi): add a node-selector label to nodes running the csi-node plugin

tiagolobocastro:
feat(csi): add a node-selector label to nodes running the csi-node plugin

The csi-node returns a new label selector on the `GetNodeInfo` call which is then placed into the node labels by the side-car.
This label is then reported by the csi-controller as a volume topology which is then added to the PV as a required node affinity label, thus pinning the volumes to csi-nodes. The default label selector is `openebs.io/csi-node=mayastor`, but it can be overridden via cli which can be set via the helm chart.


Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

PR #334, opened 2022-09-13
test: enable io-engine grpc v1 api

tiagolobocastro:
Default the io-engine grpc api version to v1, and add the necessary changes to accommodate this. Adds new traits to the rpc crate which the tests used to send grpc directly to the io-engine.


rpc/src/lib.rs
```diff
@@ -12,17 +12,20 @@ extern crate tonic;
 #[allow(clippy::upper_case_acronyms)]
 #[allow(clippy::derive_partial_eq_without_eq)]
 pub mod io_engine {
-    use bdev_rpc_client::BdevRpcClient;
-    /// AutoGenerated Io Engine Client
-    pub use mayastor_client::MayastorClient as IoEngineClient;
+    /// AutoGenerated Io Engine V1 Nexus Client.
+    use super::v1::nexus::nexus_rpc_client::NexusRpcClient as IoEngineClientV1;
```
Ab-hishek: Why only nexus client is treated as `IoEngineClientV1`? Would it not be better if we add all the clients for v1 in a wrapper and them that as `IoEngineClientV1`?

deployer/src/lib.rs
```diff
@@ -452,6 +453,21 @@ impl StartOptions {
...
     }
+
+    /// Get the io-engine api versions.
+    pub fn io_api_versions(&self) -> &Vec<IoEngineApiVersion> {
+        &self.io_engine_api_versions
+    }
+    /// Get the latest io-engine api version.
+    pub fn latest_io_api_version(&self) -> IoEngineApiVersion {
+        let mut sorted = self.io_engine_api_versions.clone();
+        sorted.sort();
```
Ab-hishek, line 464: Wouldn't this return the oldest api version?
Eg: [v0, v2, v1] -> [v0, v1, v2] and storing the first element would give `v0` whereas the latest should have been `v2`?

deployer/src/lib.rs
```diff
@@ -452,6 +453,21 @@ impl StartOptions {
...
+        if !sorted.is_empty() {
```
Abhinandan-Purkait: Do we need this check? We will always have a value right because of the default value ?
Could we return `sorted.last().unwrap_or(IoEngineApiVersion::V1)` instead?

Abhinandan-Purkait (approved):

rpc/src/lib.rs
tiagolobocastro: This is just used by tests and only the nexus was needed, yeah I can do that.

deployer/src/lib.rs
tiagolobocastro: Yes we need latest one instead.

deployer/src/lib.rs
tiagolobocastro: We don't need it, that was me trying to return by value instead of reference, but it doesn't matter here I guess.

tiagolobocastro: Done

rpc/src/lib.rs
tiagolobocastro: Done.

deployer/src/lib.rs
tiagolobocastro: Done.

tiagolobocastro:
bors merge

PR #342, opened 2022-09-22
fix: add checks for fs before returning metrics

vharsh:
(Q) Why is this change required?
(A) Right now, capacity metrics are being shown for a Raw Block Device because according to Kubernetes CSI Block devices' `SupportsMetrics() bool` behaviour returns `true` & the if the corresponding CSI driver supports the `GetVolumeStats` capability the metrics will be exposed for all the volumes provisioned by the said CSI driver.

(Q) Is there another way to stop the metrics from being shown for Raw Block Volumes, as capacity metrics don't really make much sense at a block level?
(A) Nope, another good way would be to hit the kube-apiserver and figure out if the volumeMode is `Filesystem` or `Block` & it'd involve tying the CSI-driver to a certain Container Orchestrator(Kubernetes), so it's not a very good idea, even though this isn't the most elegant solution.

Signed-off-by: Harsh Vardhan <harsh59v@gmail.com>

### DIFF
--- a/control-plane/csi-driver/src/bin/node/node.rs
+++ b/control-plane/csi-driver/src/bin/node/node.rs
@@ -1,4 +1,4 @@
-use nix::{errno::Errno, sys};
+use nix::{errno::Errno, sys, sys::statfs::FsType};
 use rpc::csi;
 use std::{boxed::Box, collections::HashMap, path::Path, time::Duration, vec::Vec};
 use tonic::{Code, Request, Response, Status};
@@ -8,6 +8,7 @@ macro_rules! failure {
     (Code::$code:ident, $msg:literal) => {{ error!($msg); Status::new(Code::$code, $msg) }};
     (Code::$code:ident, $fmt:literal $(,$args:expr)+) => {{ let message = format!($fmt $(,$args)+); error!("{}", message); Status::new(Code::$code, message) }};
 }
+const XFS_MAGIC: i64 = 0x58465342;
 
 use uuid::Uuid;
 
@@ -368,23 +369,29 @@ impl node_server::Node for Node {
             ));
         }
         match sys::statfs::statfs(&*msg.volume_path) {
-            Ok(info) => Ok(Response::new(NodeGetVolumeStatsResponse {
-                usage: vec![
-                    csi::VolumeUsage {
-                        total: info.blocks() as i64 * info.block_size(),
-                        unit: csi::volume_usage::Unit::Bytes as i32,
-                        available: info.blocks_available() as i64 * info.block_size(),
-                        used: (info.blocks() - info.blocks_free()) as i64 * info.block_size(),
-                    },
-                    csi::VolumeUsage {
-                        total: info.files() as i64,
-                        unit: csi::volume_usage::Unit::Inodes as i32,
-                        available: info.files_free() as i64,
-                        used: (info.files() - info.files_free()) as i64,
-                    },
-                ],
-                volume_condition: None,
-            })),
+            Ok(info) => match info.filesystem_type() {
+                sys::statfs::EXT4_SUPER_MAGIC | FsType(XFS_MAGIC) => {
+                    Ok(Response::new(NodeGetVolumeStatsResponse {
+                        usage: vec![
+                            csi::VolumeUsage {
+                                total: info.blocks() as i64 * info.block_size(),
+                                unit: csi::volume_usage::Unit::Bytes as i32,
+                                available: info.blocks_available() as i64 * info.block_size(),
+                                used: (info.blocks() - info.blocks_free()) as i64
+                                    * info.block_size(),
+                            },
+                            csi::VolumeUsage {
+                                total: info.files() as i64,
+                                unit: csi::volume_usage::Unit::Inodes as i32,
+                                available: info.files_free() as i64,
+                                used: (info.files() - info.files_free()) as i64,
+                            },
+                        ],
+                        volume_condition: None,
+                    }))
+                }
+                _ => Err(Status::new(Code::InvalidArgument, "unsupported filesystem")),
+            },
             Err(err) => match err {
                 Errno::ENOENT => Err(Status::new(Code::NotFound, err.to_string())),
                 Errno::EIO => Err(Status::new(Code::Internal, err.to_string())),


control-plane/csi-driver/src/bin/node/node.rs
```diff
@@ -1,4 +1,4 @@
-use nix::{errno::Errno, sys};
+use nix::{errno::Errno, sys, sys::statfs::FsType};
```
vharsh, line 1: Should I use this from FsType from the `use` construct or refer the fully qualified FsType module path in code directly? what'd be more idiomatic/natural Rust?

control-plane/csi-driver/src/bin/node/node.rs
```diff
@@ -8,6 +8,7 @@ macro_rules! failure {
     (Code::$code:ident, $msg:literal) => {{ error!($msg); Status::new(Code::$code, $msg) }};
     (Code::$code:ident, $fmt:literal $(,$args:expr)+) => {{ let message = format!($fmt $(,$args)+); error!("{}", message); Status::new(Code::$code, message) }};
 }
+const XFS_MAGIC: i64 = 0x58465342;
```
vharsh, line 11: I don't have a better way to get this done at the moment, but this magical constant might be better fetched from the `libc` crate of Rust as the current (and maybe even the latest) `nix` crate lacks this constant & it'd be safer to easier to future readers of this codebase.

control-plane/csi-driver/src/bin/node/node.rs
```diff
@@ -367,23 +368,29 @@ impl node_server::Node for Node {
...
-                        available: info.files_free() as i64,
-                        used: (info.files() - info.files_free()) as i64,
-                    },
-                ],
-                volume_condition: None,
-            })),
+            Ok(info) => match info.filesystem_type() {
+                sys::statfs::EXT4_SUPER_MAGIC | FsType(XFS_MAGIC) => {
+                    Ok(Response::new(NodeGetVolumeStatsResponse {
+                        usage: vec![
```
vharsh, line 375: @tiagolobocastro @Abhinandan-Purkait I really had trouble writing it by hand and I'm sure reading this is confusing as well, so I'm suggesting two ways to probably write it in a better-rust-way.

1. Use the `map` way as suggested by [this StackOverflow answer](https://stackoverflow.com/a/38523776) or just drop-braces with which I'm not super comfortable atm?

2. Use `ok_or` for `Result<T, err>` types as suggested [here](https://users.rust-lang.org/t/nested-match-hell-in-rust/57628/10) and use the match for matching against `filesystem_type()`.

tiagolobocastro:
I don't think we should do this, the spec does not highlight this as an error and otherwise we wouldn't be able to report volume condition in the future.
What's the issue with showing stats for block volumes? AFAIK other csi-drivers don't have this check.

blaisedias:
Block and files system volumes are "mounted" differently. It may be possible to work this out but examining the mount point.
I note that volume_path and staging_target_path are supplied as part of NodeGetVolumeStatsRequest
block volumes are bind mounted.

vharsh:
> the spec does not highlight this as an error

Correct, it suggests that we can remove used & available fields from response. 

>What's the issue with showing stats for block volumes

The discussion around this suggested that capacity metrics don't make much sense for block-volumes and currently those tests are failing.

> we wouldn't be able to report volume condition in the future

Right, this would need a refactor for BlockVolumes when we'd like to expose relevant volume info as metrics & maybe also as events.

vharsh:
> I note that volume_path and staging_target_path are supplied as part 

Taking a note of this, will evaluate this for this purpose

tiagolobocastro:
> The discussion around this suggested that capacity metrics don't make much sense for block-volumes and currently those tests are failing.

So the tests expect the call to fail for block volumes? Not something to be said lightly, but if I'm so bold this time I'd say the tests are what needs changing then :)

blaisedias:
> > The discussion around this suggested that capacity metrics don't make much sense for block-volumes and currently those tests are failing.
> 
> So the tests expect the call to fail for block volumes? Not something to be said lightly, but if I'm so bold this time I'd say the tests are what needs changing then :)

Or the BDD changed. TBH returning things like inodes on raw-block volumes is definitely wrong.

vharsh:
> So the tests expect the call to fail for block volumes? 

No, when the tests collect metrics from kubelet, they won't find the metrics from RBD, this matches the discussion during Low Level Design discussion. At that time I had wrongly stated that metrics don't show up for RBD(during testing, I had a wrong image and I had made this conclusion & destroyed the cluster). Now when the BDD tests failed for more than 80% of the time for the past few weeks, I went inside K8s codebase and looked in to see why & how the kubelet was even considering to figure out metrics for a CSI Raw block Device and there I saw that it's set to true and this doesn't figure it out via some capability matrix of Volumes or something else.

// from k8s' source ->>>>
```go
// SupportsMetrics returns true for csiBlockMapper as it initializes the
// MetricsProvider.
func (m *csiBlockMapper) SupportsMetrics() bool {
	return true
}
```

> Or the BDD changed. TBH returning things like inodes on raw-block volumes is definitely wrong.

Surprisingly the `statfs` system call actually returns something & I hope it's a garbage value.
```prometheus
kubelet_volume_stats_inodes_used{namespace="default",persistentvolumeclaim="csi-volume-metrics-test-repl-3"} 639
```

On the flip-side, it's good that RBD's capacity metrics for used size shows up as `0`(I hope this isn't because no fio job has been run on it).

```prometheus
kubelet_volume_stats_available_bytes{namespace="default",persistentvolumeclaim="csi-volume-metrics-test-repl-3"} 3.3419616256e+10
# HELP kubelet_volume_stats_capacity_bytes [ALPHA] Capacity in bytes of the volume
# TYPE kubelet_volume_stats_capacity_bytes gauge
kubelet_volume_stats_capacity_bytes{namespace="default",persistentvolumeclaim="csi-volume-metrics-test-repl-3"} 3.3419616256e+10
# HELP kubelet_volume_stats_inodes [ALPHA] Maximum number of inodes in the volume
# TYPE kubelet_volume_stats_inodes gauge
kubelet_volume_stats_inodes{namespace="default",persistentvolumeclaim="csi-volume-metrics-test-repl-3"} 8.159086e+06
# HELP kubelet_volume_stats_inodes_free [ALPHA] Number of free inodes in the volume
# TYPE kubelet_volume_stats_inodes_free gauge
kubelet_volume_stats_inodes_free{namespace="default",persistentvolumeclaim="csi-volume-metrics-test-repl-3"} 8.158447e+06
# HELP kubelet_volume_stats_inodes_used [ALPHA] Number of used inodes in the volume
# TYPE kubelet_volume_stats_inodes_used gauge
kubelet_volume_stats_inodes_used{namespace="default",persistentvolumeclaim="csi-volume-metrics-test-repl-3"} 639
# HELP kubelet_volume_stats_used_bytes [ALPHA] Number of used bytes in the volume
# TYPE kubelet_volume_stats_used_bytes gauge
kubelet_volume_stats_used_bytes{namespace="default",persistentvolumeclaim="csi-volume-metrics-test-repl-3"} 0
```

tiagolobocastro:
Hopefully `statfs` returns all zeroes for blocks, if not then we should be able to check if it's a block device and if so return empty usage?

vharsh:
I'll do a fio on an RBD & check once. I hope it's all zeros, I'm particularly confused about the non-zero metrics for inodes.

control-plane/csi-driver/src/bin/node/node.rs
```diff
@@ -368,23 +369,29 @@ impl node_server::Node for Node {
...
-                        total: info.files() as i64,
-                        unit: csi::volume_usage::Unit::Inodes as i32,
-                        available: info.files_free() as i64,
-                        used: (info.files() - info.files_free()) as i64,
-                    },
-                ],
-                volume_condition: None,
-            })),
+            Ok(info) => match info.filesystem_type() {
+                sys::statfs::EXT4_SUPER_MAGIC | FsType(XFS_MAGIC) => {
```
avishnu, line 373: Is the match in line. 72 not enough?

vharsh: It is, but this way is sort of a hack. Let's say if tomorrow MayaStor is supporting 10 other file systems, this part of code would need to be changed each time a new fs support is added. Hence I've plans to figure out what should be done here such that zeros are returned for volume capacity used.

vharsh: Also right now it lazily throws an error, but actually it might be better if it returns zero values for used capacity and the total capacity is retuned as is, I'll take a look at it sometime soon and target to ACK all issues in a week

tiagolobocastro:
Replaced by: #414

PR #345, opened 2022-09-28
feat(ha-target-switchover): add component level bdd for switchover

Abhinandan-Purkait:
- Exposes `del_volume_shutdown_target` in rest.
- Adds changes in common nvme module to handle nvme connection if ana is not enabled.
- Adds the component level bdd tests for target switchover.

### DIFF
--- a/control-plane/agents/src/bin/core/controller/wrapper.rs
+++ b/control-plane/agents/src/bin/core/controller/wrapper.rs
@@ -1023,7 +1023,7 @@ impl ClientOps for Arc<tokio::sync::RwLock<NodeWrapper>> {
                     .await
                     .nexuses()
                     .iter()
-                    .find(|nexus| nexus.uuid == request.uuid || nexus.name == nexus_name)
+                    .find(|nexus| nexus.uuid == request.uuid && nexus.name == nexus_name)
                 {
                     Some(nexus) => {
                         // return Ok if nexus with the requested uuid and name already exists
@@ -1151,7 +1151,8 @@ impl ClientOps for Arc<tokio::sync::RwLock<NodeWrapper>> {
 
     async fn shutdown_nexus(&self, request: &ShutdownNexus) -> Result<(), SvcError> {
         let dataplane = self.grpc_client_locked(request.id()).await?;
-        if let Some(nexus) = self.read().await.nexus(&request.uuid()) {
+        let nexus = self.read().await.nexus(&request.uuid());
+        if let Some(nexus) = nexus {
             if nexus.status == NexusStatus::Shutdown {
                 Ok(())
             } else {
--- a/scripts/python/test.sh
+++ b/scripts/python/test.sh
@@ -28,4 +28,8 @@ else
   pytest "$@"
 fi
 
-sh "$ROOT_DIR"/scripts/python/test_residue_cleanup.sh
+cleanup_handler() {
+  "$ROOT_DIR"/scripts/python/test_residue_cleanup.sh
+}
+
+trap cleanup_handler ERR INT QUIT TERM HUP
--- a/scripts/python/test.sh
+++ b/scripts/python/test.sh
@@ -28,3 +28,4 @@ else
   pytest "$@"
 fi
 
+sh "$ROOT_DIR"/scripts/python/test_residue_cleanup.sh
--- a/scripts/python/test_residue_cleanup.sh
+++ b/scripts/python/test_residue_cleanup.sh
@@ -2,7 +2,7 @@
 
 # Cleans up the iptables rules added by bdd tests
 
-set -e
+set -euo pipefail
 
 echo "Cleaning all IPTABLE rules added by tests..."
 
--- a/scripts/python/test_residue_cleanup.sh
+++ b/scripts/python/test_residue_cleanup.sh
@@ -1,0 +1,16 @@
+#!/usr/bin/env bash
+
+# Cleans up the iptables rules added by bdd tests
+
+set -e
+
+echo "Cleaning all IPTABLE rules added by tests..."
+
+# First backup the iptables rules
+sudo bash -c "iptables-save > iptables.backup"
+# Remove the rules filtering by comment "added by bdd tests"
+sed -i '/.*--comment.* "added by bdd tests"/d' iptables.backup
+# Restore the rules
+sudo iptables-restore < iptables.backup
+# Remove the temporary backup file
+sudo rm iptables.backup
--- a/tests/bdd/features/ha/core-agent/test_target_switchover.py
+++ b/tests/bdd/features/ha/core-agent/test_target_switchover.py
@@ -1,0 +1,306 @@
+"""Target Switchover test feature tests."""
+import os
+import subprocess
+from time import sleep
+from urllib.parse import urlparse
+
+import grpc
+import pytest
+from pytest_bdd import (
+    given,
+    scenario,
+    then,
+    when,
+    parsers,
+)
+
+from common.apiclient import ApiClient
+from common.deployer import Deployer
+from common.fio import Fio
+
+from openapi.model.create_pool_body import CreatePoolBody
+from openapi.model.create_volume_body import CreateVolumeBody
+from openapi.model.protocol import Protocol
+from openapi.model.volume_policy import VolumePolicy
+
+POOL_UUID_1 = "4cc6ee64-7232-497d-a26f-38284a444980"
+POOL_UUID_2 = "22e1d15f-4dfd-4bf5-a98f-74e4aebf9e62"
+VOLUME_UUID = "5cd5378e-3f05-47f1-a830-a0f5873a1449"
+NODE_NAME_1 = "io-engine-1"
+NODE_NAME_2 = "io-engine-2"
+VOLUME_CTX_KEY = "volume"
+VOLUME_SIZE = 10485761
+DEPLOYER_NETWORK_INTERFACE = "mayabridge0"
+NVME_SVC_PORT = 8420
+IO_ENGINE_GRPC_PORT = 10124
+IO_ENGINE_1_IP = "10.1.0.7"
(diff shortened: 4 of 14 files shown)


tests/bdd/features/ha/target_switchover/target_switchover.feature
```diff
@@ -0,0 +1,35 @@
+Feature: Target Switchover test
+
+  Background:
+    Given a control plane, two Io-Engine instances, two pools
+
+  Scenario: on target switchover older target should not write to replicas
```
tiagolobocastro:
```suggestion
  Scenario: on target switchover, the older target should not have R/W access to the replicas
```

tests/bdd/features/ha/target_switchover/target_switchover.feature
```diff
@@ -0,0 +1,35 @@
...
+    Given a volume with two replicas, the volume published on a node
```
tiagolobocastro:
```suggestion
    Given a published volume with two replicas
```
Same for the cases below

tests/bdd/features/ha/target_switchover/target_switchover.feature
```diff
@@ -0,0 +1,35 @@
...
+    When the volume is republished on the other node
```
tiagolobocastro:
```suggestion
    When the volume is republished on another node
```

tests/bdd/features/ha/target_switchover/target_switchover.feature
```diff
@@ -0,0 +1,35 @@
...
+    Then the newer target should serve the io
```
tiagolobocastro:
```suggestion
    Then the newer target should have R/W access to the replicas
```
Same for the cases below

tests/bdd/features/ha/target_switchover/target_switchover.feature
```diff
@@ -0,0 +1,35 @@
...
+    And the older target should not serve the io
```
tiagolobocastro:
```suggestion
    And the older target should not have R/W access the replicas
```

tests/bdd/features/ha/target_switchover/target_switchover.feature
```diff
@@ -0,0 +1,35 @@
...
+
+  Scenario: if the node hosting the nexus is offline, switchover should not fail
+    Given a volume with two replicas, the volume published on a node
+    When the node hosting the nexus is killed
+    Then the volume republish should not fail
```
tiagolobocastro:
```suggestion
    Then volume republish should not fail
```

tests/bdd/features/ha/target_switchover/target_switchover.feature
```diff
@@ -0,0 +1,35 @@
...
+    Then the newer target should serve the io
+    And the older target should not serve the io
+
+  Scenario: after target switchover if older target is accessible, it should not write to replicas
```
tiagolobocastro:
```suggestion
  Scenario: after target switchover if the older target is accessible, it should not have R/W access to the replicas
```

tests/bdd/features/ha/target_switchover/target_switchover.feature
```diff
@@ -0,0 +1,35 @@
...
+    Given a volume with two replicas, the volume published on a node
+    When the node hosting the nexus is killed
+    Then the volume republish should not fail
+    When the node hosting the older nexus is brought back
+    Then the newer target should serve the io
+    And the older target should not serve the io
```
tiagolobocastro: If it's killed the older target will no longer exist.
I think here we want to test for temporary network failure, where we can't reach the node to send the shutdown signal, but then network is back and the older nexus is still there?

tests/bdd/features/ha/target_switchover/target_switchover.feature
```diff
@@ -0,0 +1,35 @@
...
+
+  Scenario: continous switchover and older nexus destruction should be seamless
```
tiagolobocastro: :+1:

tiagolobocastro: s/nexus/target/

tests/bdd/features/ha/target_switchover/target_switchover.feature
```diff
@@ -0,0 +1,35 @@
...
+    Given a volume with two replicas, the volume published on a node
+    When the node hosting the nexus is killed
+    Then the volume republish for `n` times and it should not fail
```
tiagolobocastro:
```suggestion
    Then the volume should be republished `n` times and it should not fail
```

tests/bdd/features/ha/target_switchover/target_switchover.feature
```diff
@@ -0,0 +1,35 @@
...
+    And the destroy shutdown nexus call succeeds
```
tiagolobocastro:
```suggestion
    And the destroy shutdown target calls should succeed
```

tests/bdd/features/ha/target_switchover/target_switchover.feature
Abhinandan-Purkait: That a great idea, will simulate netwrork failure here. Thanks

Abhinandan-Purkait:
bors try

control-plane/agents/src/bin/core/controller/wrapper.rs
```diff
@@ -1151,7 +1151,8 @@ impl ClientOps for Arc<tokio::sync::RwLock<NodeWrapper>> {
 
     async fn shutdown_nexus(&self, request: &ShutdownNexus) -> Result<(), SvcError> {
         let dataplane = self.grpc_client_locked(request.id()).await?;
-        if let Some(nexus) = self.read().await.nexus(&request.uuid()) {
+        let nexus = self.read().await.nexus(&request.uuid());
```
tiagolobocastro, line 1154: @abhilashshetty04 you might have done it inadvertently but this test https://github.com/openebs/mayastor-control-plane/blob/b9cbbf86d45f0c867dc4f81773ca4d99c644c08f/tests/bdd/features/ha/node-agent/test_path_replacement.py#L96 it is not actually testing the path replacement, it's just listing the volume?
In fact the current code here on shutdown_nexus is a deadlock, so not sure how you've tested it?

tiagolobocastro: Oh I kind get it now, the node agent test is relying on the cluster agent to run things rather than doing it itself?

tests/bdd/features/ha/core-agent/test_target_switchover.py
```diff
@@ -0,0 +1,309 @@
...
+
+def remove_network_failure(io_engine_ip, port):
+    command = RULE_REMOVE.format(DEPLOYER_NETWORK_INTERFACE, io_engine_ip, port)
+    try:
+        subprocess.run(command, shell=True, check=True)
+    except subprocess.CalledProcessError:
+        assert False, "Error while adding IP table rule"
+
+
+def cleanup_iptable_rules(io_engine_ip):
```
tiagolobocastro, line 269: wonder if we should have a script for this as well, in case process is terminated half-way through? So you can just run it manually.

tiagolobocastro (approved):
:+1:

control-plane/agents/src/bin/core/controller/wrapper.rs
tiagolobocastro: Also I get why the test is passing, in case the node is not online and so this doesn't get executed at all, we error out on 1153.

abhilashshetty04: Yes watchdog will report node status as Unknown. I though lock gets released as we return the result (thought scope ends there). Test used to do manual republish. Since stages are implemented now it relies on cluster agent.

abhilashshetty04:
Looks good to me

abhilashshetty04 (approved):

control-plane/agents/src/bin/core/controller/wrapper.rs
tiagolobocastro: Lock guards are release on drop and the lock is still in scope so it doesn't get released.

tests/bdd/features/ha/core-agent/test_target_switchover.py
Abhinandan-Purkait: Although fixtures will cleanup after each test, but if the fixture itself terminated halfway we will have these entries. I will add a script as well. Thanks

Abhinandan-Purkait: Done.

Abhinandan-Purkait:
bors try

scripts/python/test_residue_cleanup.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+
+# Cleans up the iptables rules added by bdd tests
+
+set -e
```
tiagolobocastro: Should this be `set -euo pipefail` ?

scripts/python/test.sh
```diff
@@ -28,3 +28,4 @@ else
   pytest "$@"
 fi
 
+sh "$ROOT_DIR"/scripts/python/test_residue_cleanup.sh
```
tiagolobocastro: Run it inside a trap instead so it's called in case of failure?
```bash
cleanup_handler() {
  "$ROOT_DIR"/scripts/python/test_residue_cleanup.sh
}

trap cleanup_handler ERR INT QUIT TERM HUP
```

scripts/python/test_residue_cleanup.sh
Abhinandan-Purkait: Done

scripts/python/test.sh
Abhinandan-Purkait: Done

Abhinandan-Purkait:
bors merge

PR #348, opened 2022-09-29
fix: upgrade dockerhub url v2 to check for tags

tiagolobocastro:
This allows us to avoid re-publishing release tags "proper". Move the check till after the build is done, so we can still push the branch/alias tags.

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -5,22 +5,22 @@
 # The script assumes that a user is logged on to dockerhub for public images,
 # or has insecure registry access setup for CI.
 
-CI=${CI:false}
+CI=${CI-}
 
 set -euo pipefail
 
 # Test if the image already exists in dockerhub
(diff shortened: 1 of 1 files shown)


Ab-hishek (approved):

chriswldenyer:
Pardon my ignorance, but what does
CI=${CI:false}
do? 
I can only find that parameter substitution used for substring numerical offsets, e.g. ${string:7}

chriswldenyer (approved):

Abhinandan-Purkait (approved):

tiagolobocastro:
> Pardon my ignorance, but what does CI=${CI:false} do? I can only find that parameter substitution used for substring numerical offsets, e.g. ${string:7}

It's probably an undocumented feature then :) 
I'll do it properly, should be `CI=${CI-}`

tiagolobocastro:
bors merge

PR #356, opened 2022-10-06
fix: cherry-pick pool timeouts and silence duplicate logs

tiagolobocastro:
Adds a higher than base timeout for pool lifecycle operations. Add a higher base grpc timeout for the rest service. This is because I've found out that the default tonic impl of timeout prefers the shorter timeout. So we can't easily set a timeout per request.
TODO: use our own timeout layer impl
Silence logs for busy resources when they've been reported once.


Abhinandan-Purkait (approved):

abhilashshetty04 (approved):

tiagolobocastro:
bors merge

PR #364, opened 2022-10-14
feat(ha_nexus): added additional stage in Switchover

abhilashshetty04:
This PR adds additional stage which marks request as Success. Will delete the switchover request from Etcd.

### DIFF
--- a/control-plane/agents/src/bin/ha/cluster/switchover.rs
+++ b/control-plane/agents/src/bin/ha/cluster/switchover.rs
@@ -35,6 +35,8 @@ pub enum Stage {
     PublishPath,
     /// Delete original/old volume target.
     DeleteTarget,
+    /// Marks switchover process as Complete.
+    Successful,
     /// Represent failed switchover request.
     Errored,
 }
@@ -89,8 +91,9 @@ impl SwitchOverRequest {
             Stage::Init => Stage::RepublishVolume,
             Stage::RepublishVolume => Stage::PublishPath,
             Stage::PublishPath => Stage::DeleteTarget,
-            // DeleteTarget and Errored stage mark request as complete, so no need to update
-            Stage::DeleteTarget => Stage::DeleteTarget,
+            Stage::DeleteTarget => Stage::Successful,
+            // Successful and Errored stage mark request as complete, so no need to update
+            Stage::Successful => Stage::Successful,
             Stage::Errored => Stage::Errored,
         };
     }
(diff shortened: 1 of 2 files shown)


control-plane/agents/src/bin/ha/cluster/switchover.rs
```diff
@@ -173,6 +176,19 @@ impl SwitchOverRequest {
         Ok(())
     }
 
+    /// Deletes Swtichover request from Etcd.
```
tiagolobocastro: Switchover

control-plane/agents/src/bin/ha/cluster/switchover.rs
```diff
@@ -173,6 +176,19 @@ impl SwitchOverRequest {
...
+    async fn delete_switchover(&mut self, etcd: &EtcdStore) -> Result<(), anyhow::Error> {
+        self.start_op(Stage::Successful, etcd).await?;
+        info!(volume.uuid=%self.volume_id, "Deleting Switchover request from Etcd as request completed successfully");
```
tiagolobocastro: You can Etcd with different cases (eg: Etcd vs etcd). Please use the same throughout the messages and code comments.

Abhinandan-Purkait (approved):

abhilashshetty04:
bors try

tiagolobocastro (approved):
Can complete requests be in `ETCD`? If so it seems `fetch_incomplete_requests` will ignore them and leave them around, we should delete those too?

control-plane/agents/src/bin/ha/cluster/switchover.rs
abhilashshetty04: Will use lowercase in all comments/docs

control-plane/agents/src/bin/ha/cluster/switchover.rs
abhilashshetty04: Once request completes delete target . We update the stage as Successful. In that stage request is deleted from etcd. In case cluster agent restarts while processing some request. `fetch_incomplete_request` will pick them up and process them according to the last stage.

abhilashshetty04:
bors merge

control-plane/agents/src/bin/ha/cluster/switchover.rs
abhilashshetty04: I misunderstood your question. Do you want to keep requests that succeded in etcd?

control-plane/agents/src/bin/ha/cluster/switchover.rs
tiagolobocastro: No, I want to delete them. And the existing code on startup just ignores completed entries.

abhilashshetty04: We delete the swicthover request when stage is updates as Successful. ideally there should be no requests in Successful stage at startup. It can happen if stage is updated for a request but before deleting from etcd cluster agent crashed. 

In this case also. At startup if we encounter any Successful request. We would delete that request. Sortof like a cleanup workflow. Is that okay?

tiagolobocastro: Yeah sounds good to me

PR #365, opened 2022-10-15
chore: use rustup with path

tiagolobocastro:
On non-nixos when using rustup and the system's libc is not the same as the nix-shell's libc seems
 we get some odd errors:
https://discourse.nixos.org/t/nix-shell-with-rustup/22452 This is likely because the binaries from rustup haven't been patched to point to the nix env. So, instead of using the binaries from rustup just point rustup to a copy of the nix binaries. This allows us to both use the right binaries and carry on using rustup so we can configure our favourite IDE to the same rustup path.

### DIFF
--- a/scripts/rust/env-setup.sh
+++ b/scripts/rust/env-setup.sh
@@ -1,0 +1,57 @@
+#!/usr/bin/env bash
+
+# Meant to be called only from the nix-shell shellHook
+
+# these are provided by the shellHook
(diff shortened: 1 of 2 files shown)


mtzaurus (approved):

Abhinandan-Purkait (approved):

abhilashshetty04 (approved):

tiagolobocastro:
bors merge

PR #366, opened 2022-10-18
feat(ha_nexus): adding support to handle switchover request asynchronously

abhilashshetty04:
This PR adds support to create worker tasks to process switchover requests parallely.

### DIFF
--- a/control-plane/agents/src/bin/ha/cluster/switchover.rs
+++ b/control-plane/agents/src/bin/ha/cluster/switchover.rs
@@ -17,7 +17,8 @@ use utils::NVME_TARGET_NQN_PREFIX;
 
 use anyhow::anyhow;
 use grpc::operations::ha_node::traits::NodeAgentOperations;
-use std::{convert::TryFrom, net::SocketAddr};
+use std::{convert::TryFrom, net::SocketAddr, sync::Arc};
+use tokio::sync::Mutex;
 use tonic::transport::Uri;
 
 fn client() -> impl VolumeOperations {
@@ -232,6 +233,9 @@ pub struct SwitchOverEngine {
     channel: UnboundedSender<SwitchOverRequest>,
 }
 
+/// Number of Switchover worker tasks to be spawned.
+const WORKER_NUM: u8 = 4;
+
 impl SwitchOverEngine {
     /// Creates a new switchover engine to process Nvme path failures.
     pub fn new(etcd: EtcdStore, nodes: NodeList) -> Self {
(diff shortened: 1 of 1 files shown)


control-plane/agents/src/bin/ha/cluster/switchover.rs
```diff
@@ -243,48 +247,55 @@ impl SwitchOverEngine {
...
-        });
+        }
+    }
+
+    /// Mover instantiates worker task to asynchronously process Switchover request.
+    pub fn mover(&self, recv: Arc<Mutex<UnboundedReceiver<SwitchOverRequest>>>) {
+        for i in 0 .. WORKER_NUM {
+            info!("Starting worker-{} of Switchover Engine", i.to_string());
+            let cloned_self = self.clone();
+            let cloned_channel = Arc::clone(&recv);
```
tiagolobocastro: Clone takes ref self so you can just call clone:
```suggestion
            let cloned_channel = recv.clone();
```

control-plane/agents/src/bin/ha/cluster/switchover.rs
```diff
@@ -243,48 +247,55 @@ impl SwitchOverEngine {
...
-                        match result {
-                            Ok(_) => {
-                                // reset retry delay back to the start.
-                                retry_delay = retry_delay_min;
-                            }
-                            Err(_) => {
-                                retry_delay = (retry_delay * 3).min(retry_delay_max);
-                                tokio::time::sleep(retry_delay).await;
-                            }
+        loop {
```
tiagolobocastro, line 266: You don't need to clone etcd and nodes now?

control-plane/agents/src/bin/ha/cluster/switchover.rs
```diff
@@ -243,48 +247,55 @@ impl SwitchOverEngine {
...
+        for i in 0 .. WORKER_NUM {
+            info!("Starting worker-{} of Switchover Engine", i.to_string());
+            let cloned_self = self.clone();
+            let cloned_channel = Arc::clone(&recv);
+            tokio::spawn(async move { cloned_self.process_request(cloned_channel).await });
+        }
+    }
+
+    /// Switchover request to be handled synchronously in each worker task.
+    async fn process_request(&self, recv: Arc<Mutex<UnboundedReceiver<SwitchOverRequest>>>) {
```
tiagolobocastro: think we can just consume self here:
```suggestion
    async fn process_request(self, recv: Arc<Mutex<UnboundedReceiver<SwitchOverRequest>>>) {
```

control-plane/agents/src/bin/ha/cluster/switchover.rs
```diff
@@ -232,6 +233,9 @@ pub struct SwitchOverEngine {
     channel: UnboundedSender<SwitchOverRequest>,
 }
 
+/// Num of Switchover instances to be spawned.
```
```suggestion
/// Number of Switchover instances to be spawned.
```

tiagolobocastro (approved):
A possible problem with this approach is that the work per worker is still blocking, so if one volume gets stuck for whatever reason, we block that whole Switchover instance.
Let's merge this as it's already better but I think we still need a solution for that..

control-plane/agents/src/bin/ha/cluster/switchover.rs
abhilashshetty04: Incorporated

control-plane/agents/src/bin/ha/cluster/switchover.rs
abhilashshetty04: Incorporated

control-plane/agents/src/bin/ha/cluster/switchover.rs
abhilashshetty04: Incorporated

control-plane/agents/src/bin/ha/cluster/switchover.rs
abhilashshetty04: Incorporated

control-plane/agents/src/bin/ha/cluster/switchover.rs
```diff
@@ -243,48 +247,53 @@ impl SwitchOverEngine {
...
+            let cloned_self = self.clone();
+            let cloned_channel = recv.clone();
+            tokio::spawn(async move { cloned_self.process_request(cloned_channel).await });
+        }
+    }
+
+    /// Switchover request to be handled synchronously in each worker task.
+    async fn process_request(self, recv: Arc<Mutex<UnboundedReceiver<SwitchOverRequest>>>) {
+        loop {
+            if let Some(mut q) = recv.lock().await.recv().await {
```
abhilashshetty04, line 267: @tiagolobocastro , Will this create lock contention?

abhilashshetty04: I guess not.

tiagolobocastro: It's just taking switchover requests out of the queue, it shouldn't be a problem.

abhilashshetty04: Thanks

Abhinandan-Purkait (approved):

abhilashshetty04:
bors try

abhilashshetty04:
bors try

abhilashshetty04:
bors merge

PR #371, opened 2022-10-28
feat(csi-node): allow controlling of the ctrl_loss_tmo

tiagolobocastro:
This sets up the maximum number of reconnect attempts before the kernel initiator gives up.
Nvme-tcp is not able to reconnect a device after the max_reconnections has been hit and the device
is lost, so we should set a known limit which is sufficiently large to allow for failover.
Conversely a value that is too large will make rebooting tricky as the kernel will keep trying..

Refactored the way storage class parameters are setup from the controller to the node service.
The new parameter name is: nvmeCtrlLossTmo.
TODO: node-specific nvmf options should also be refactored in the same way.

A default global value for the ctrl_loss_tmo can be added via cmdline arg nvme-ctrl-loss-tmo.
This value is overridden by the storageClass parameter.


Abhinandan-Purkait (approved):

Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

PR #372, opened 2022-10-28
fix(ha_nexus): enhance error handling in ha cluster agent

abhilashshetty04:
This PR handles path lookup failure by the node agent. Switchover request will be marked as Errored and will not be retried post that.

 Fix in Serializing/Deserializing of Switchover Spec to Etcd.

When a Switchover request fails it is retried by the same Worker. This can block the worker unnecessarily. This PR also includes changes in this logic. We will send the request back to the channel. 

Changes had to be made in the Exponential backoff as well to accomodate above modification.  Now we have faster requeue 
of failed request upto certain configurable number. Then requests are requeued with longer delay. Sleep is out of the Worker as well. 

### DIFF
--- a/common/src/transport_api/mod.rs
+++ b/common/src/transport_api/mod.rs
@@ -163,6 +163,19 @@ impl From<ReplyError> for tonic::Status {
     fn from(error: ReplyError) -> Self {
         match error.kind {
             ReplyErrorKind::InvalidArgument => tonic::Status::invalid_argument(error.full_string()),
+            ReplyErrorKind::DeadlineExceeded => {
+                tonic::Status::deadline_exceeded(error.full_string())
+            }
+            ReplyErrorKind::FailedPrecondition => {
+                tonic::Status::failed_precondition(error.full_string())
+            }
+            ReplyErrorKind::AlreadyExists => tonic::Status::already_exists(error.full_string()),
+            ReplyErrorKind::Aborted => tonic::Status::aborted(error.full_string()),
+            ReplyErrorKind::NotFound => tonic::Status::not_found(error.full_string()),
+            ReplyErrorKind::ResourceExhausted => {
+                tonic::Status::resource_exhausted(error.full_string())
+            }
+            ReplyErrorKind::Unimplemented => tonic::Status::unimplemented(error.full_string()),
             _ => tonic::Status::internal(error.full_string()),
         }
     }
@@ -189,14 +202,14 @@ impl From<JoinError> for ReplyError {
 
 impl StdError for ReplyError {}
 impl ReplyError {
-    /// extend error with source.
-    /// useful when another error wraps around a `ReplyError` and we want to
+    /// Extend error with source.
+    /// Useful when another error wraps around a `ReplyError` and we want to
     /// convert back to `ReplyError` so we can send it over the wire.
     pub fn extend(&mut self, source: &str, extra: &str) {
         self.source = format!("{}::{}", source, self.source);
         self.extra = format!("{}::{}", extra, self.extra);
     }
-    /// useful when the grpc server is dropped due to panic.
+    /// Useful when the grpc server is dropped due to panic.
     pub fn aborted_error(error: JoinError) -> Self {
         Self {
             kind: ReplyErrorKind::Aborted,
@@ -205,7 +218,7 @@ impl ReplyError {
             extra: "Failed to wait for thread".to_string(),
         }
     }
-    /// useful when the grpc server is dropped due to panic.
+    /// Useful when the grpc server is dropped due to panic.
     pub fn tonic_reply_error(kind: ReplyErrorKind, source: String, extra: String) -> Self {
         Self {
             kind,
@@ -214,7 +227,7 @@ impl ReplyError {
             extra,
         }
     }
-    /// used only for testing, not used in code.
+    /// Used only for testing, not used in code.
     pub fn invalid_reply_error(msg: String) -> Self {
         Self {
             kind: ReplyErrorKind::Aborted,
@@ -223,7 +236,7 @@ impl ReplyError {
             extra: msg,
         }
     }
-    /// used when we get an empty response from the grpc server.
+    /// Used when we get an empty response from the grpc server.
     pub fn invalid_response(resource: ResourceKind) -> Self {
         Self {
             kind: ReplyErrorKind::Aborted,
@@ -232,7 +245,7 @@ impl ReplyError {
             extra: "".to_string(),
         }
     }
-    /// used when we get an invalid argument.
+    /// Used when we get an invalid argument.
     pub fn invalid_argument(resource: ResourceKind, arg_name: &str, error: String) -> Self {
         Self {
             kind: ReplyErrorKind::InvalidArgument,
@@ -241,7 +254,7 @@ impl ReplyError {
             extra: format!("Invalid {} was provided", arg_name),
         }
     }
-    /// used when we encounter a missing argument.
+    /// Used when we encounter a missing argument.
     pub fn missing_argument(resource: ResourceKind, arg_name: &str) -> Self {
         Self {
             kind: ReplyErrorKind::InvalidArgument,
@@ -250,7 +263,7 @@ impl ReplyError {
             extra: format!("Argument {} was not provided", arg_name),
         }
     }
-    /// for errors that can occur when serializing or deserializing JSON data.
+    /// For errors that can occur when serializing or deserializing JSON data.
     pub fn serde_error(
         resource: ResourceKind,
         error_kind: ReplyErrorKind,
@@ -263,7 +276,7 @@ impl ReplyError {
             extra: "".to_string(),
         }
     }
-    /// for errors that represent unimplemented functionality.
+    /// For errors that represent unimplemented functionality.
     pub fn unimplemented(msg: String) -> Self {
         Self {
             kind: ReplyErrorKind::Unimplemented,
@@ -272,10 +285,60 @@ impl ReplyError {
             extra: msg,
         }
     }
-    /// for internal errors.
+    /// For internal errors.
     pub fn internal_error(resource: ResourceKind, source: String, extra: String) -> Self {
         Self {
             kind: ReplyErrorKind::Internal,
+            resource,
+            source,
+            extra,
+        }
+    }
+
+    /// For deadline exceeded errors.
+    pub fn deadline_exceeded(resource: ResourceKind, source: String, extra: String) -> Self {
+        Self {
+            kind: ReplyErrorKind::DeadlineExceeded,
+            resource,
+            source,
+            extra,
+        }
+    }
+
+    /// For failed precondition errors.
+    pub fn failed_precondition(resource: ResourceKind, source: String, extra: String) -> Self {
+        Self {
+            kind: ReplyErrorKind::FailedPrecondition,
+            resource,
+            source,
+            extra,
+        }
+    }
+
+    /// For already exist errors.
+    pub fn already_exist(resource: ResourceKind, source: String, extra: String) -> Self {
+        Self {
+            kind: ReplyErrorKind::AlreadyExists,
+            resource,
+            source,
+            extra,
+        }
+    }
+
+    /// For not found errors.
+    pub fn not_found(resource: ResourceKind, source: String, extra: String) -> Self {
+        Self {
+            kind: ReplyErrorKind::NotFound,
+            resource,
+            source,
+            extra,
+        }
+    }
+
+    /// For resource exhausted errors.
+    pub fn resource_exhausted(resource: ResourceKind, source: String, extra: String) -> Self {
+        Self {
+            kind: ReplyErrorKind::ResourceExhausted,
             resource,
             source,
             extra,
--- a/control-plane/agents/src/bin/ha/cluster/switchover.rs
+++ b/control-plane/agents/src/bin/ha/cluster/switchover.rs
@@ -12,12 +12,13 @@ use common_lib::types::v0::{
 use grpc::operations::{ha_node::client::NodeAgentClient, volume::traits::VolumeOperations};
 use serde::{Deserialize, Serialize};
 use tokio::sync::mpsc::{unbounded_channel, UnboundedReceiver, UnboundedSender};
-use tracing::info;
+use tracing::{error, info};
 use utils::NVME_TARGET_NQN_PREFIX;
 
 use anyhow::anyhow;
+use common_lib::transport_api::ReplyErrorKind;
 use grpc::operations::ha_node::traits::NodeAgentOperations;
-use std::{convert::TryFrom, net::SocketAddr, sync::Arc};
+use std::{convert::TryFrom, net::SocketAddr, sync::Arc, time::Duration};
 use tokio::sync::Mutex;
 use tonic::transport::Uri;
 
@@ -54,6 +55,8 @@ pub struct SwitchOverRequest {
     existing_nqn: String,
     // New nexus path of the volume.
     new_path: Option<String>,
+    // Number of failed attempts in the current Stage.
+    retry_count: u64,
 }
 
 impl SwitchOverRequest {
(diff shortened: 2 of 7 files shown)


common/src/transport_api/mod.rs
```diff
@@ -281,6 +287,26 @@ impl ReplyError {
             extra,
         }
     }
+
+    /// For deadline exceeded errors.
+    pub fn deadline_exceeded(resource: ResourceKind, source: String, extra: String) -> Self {
```
Abhinandan-Purkait, line 299: Where are we using this?

common/src/transport_api/mod.rs
```diff
@@ -163,6 +163,12 @@ impl From<ReplyError> for tonic::Status {
     fn from(error: ReplyError) -> Self {
         match error.kind {
             ReplyErrorKind::InvalidArgument => tonic::Status::invalid_argument(error.full_string()),
+            ReplyErrorKind::DeadlineExceeded => {
+                tonic::Status::deadline_exceeded(error.full_string())
```
Abhinandan-Purkait, line 167: Do we need this? Don't seem to be using it.

common/src/transport_api/mod.rs
abhilashshetty04: I had used it initially. Kept it in the PR because it can be used when there is appropriate scenario. We have DeadlineExceeded in ReplyErrorKind already.

common/src/transport_api/mod.rs
abhilashshetty04: Not using in this PR. But can be used later so thought of keeping it.

tiagolobocastro: We should probably add others as well

control-plane/agents/src/bin/ha/cluster/switchover.rs
```diff
@@ -209,7 +227,16 @@ impl SwitchOverRequest {
             if let Some(new_path) = self.new_path.clone() {
                 let replace_request = ReplacePath::new(self.existing_nqn.clone(), new_path.clone());
                 let client = NodeAgentClient::new(uri, None).await;
-                client.replace_path(&replace_request, None).await?;
+                if let Err(e) = client.replace_path(&replace_request, None).await {
+                    return if e.kind == ReplyErrorKind::FailedPrecondition {
+                        error!(nexus.path=%self.existing_nqn, "HA Node agent could not find failed Nvme path. Must have been removed after timeout.");
```
tiagolobocastro: We're already in the HA Node agent, so can just say "Could not find...."

control-plane/agents/src/bin/ha/cluster/switchover.rs
```diff
@@ -156,6 +157,11 @@ impl SwitchOverRequest {
             Some(target) => Some(target.device_uri),
             _ => None,
         };
+        if self.new_path.is_none() {
+            self.stage = Stage::Errored;
+            error!(volume.uuid=%self.volume_id,"Could not find device uri for the volume. Marking request as errored");
```
tiagolobocastro: nit: missing space between `,` and `"`.

control-plane/agents/src/bin/ha/cluster/switchover.rs
```diff
@@ -220,7 +247,9 @@ impl SwitchOverRequest {
                 Err(anyhow!("Could not to get new nexus target for the volume"))
             }
         } else {
-            Err(anyhow!("Could not get grpc address for the node"))
+            self.stage = Stage::Errored;
+            error!(volume.uuid=%self.volume_id,"Could not get grpc address of the node agent. Moving Switchover req to Errored");
```
tiagolobocastro: nit: missing space between `,` and `"`.

tiagolobocastro (approved):
I might be misreading it, but things like "Must have been removed after timeout" might be a bit misleading, eg: what about idempotency? Let's say we've already removed it but crashed? If we retry later, will we flag an error? Or what about if the application is removed?

common/src/transport_api/mod.rs
abhilashshetty04: Will Do

control-plane/agents/src/bin/ha/cluster/switchover.rs
abhilashshetty04: No this is in Cluster agent. Here we are matching expected error type for the scenario when Device is removed from Subsystem.

control-plane/agents/src/bin/ha/cluster/switchover.rs
abhilashshetty04: Will add

control-plane/agents/src/bin/ha/cluster/switchover.rs
abhilashshetty04: Incorporated

control-plane/agents/src/bin/ha/cluster/switchover.rs
abhilashshetty04: Incorporated

abhilashshetty04: Incorporated

common/src/transport_api/mod.rs
abhilashshetty04: Added few more conversions

common/src/transport_api/mod.rs
```diff
@@ -163,6 +163,19 @@ impl From<ReplyError> for tonic::Status {
...
+            ReplyErrorKind::FailedPrecondition => {
+                tonic::Status::failed_precondition(error.full_string())
+            }
+            ReplyErrorKind::AlreadyExists => tonic::Status::already_exists(error.full_string()),
+            ReplyErrorKind::Aborted => tonic::Status::aborted(error.full_string()),
+            ReplyErrorKind::NotFound => tonic::Status::not_found(error.full_string()),
+            ReplyErrorKind::ResourceExhausted => {
+                tonic::Status::resource_exhausted(error.full_string())
+            }
+            ReplyErrorKind::Unimplemented => tonic::Status::unimplemented(error.full_string()),
```
tiagolobocastro, line 178: Cool, thanks for adding all these :+1:

control-plane/agents/src/bin/ha/cluster/switchover.rs
```diff
@@ -235,6 +266,12 @@ pub struct SwitchOverEngine {
 
 /// Number of Switchover worker tasks to be spawned.
 const WORKER_NUM: u8 = 4;
+/// Sleep duration during Fast requeue phase.
+const FAST_REQUEUE: u64 = 10;
+/// Sleep duration during Slow requeue phase.
+const SLOW_REQUEUE: u64 = 30;
+/// Number of Fast request requeue per stage.
+const NUM_OF_FAST_REQUEUE: u64 = 18;
```
tiagolobocastro: You can use these as an enum, eg:
```rust
enum ReQueue {
    /// Sleep duration during Fast requeue phase.
    Fast = 10,
    /// Sleep duration during Slow requeue phase.
    Slow = 30,
    /// Number of Fast request requeue per stage.
    NumFast = 18
}
// and then just use them like
ReQueue::Fast as u64/whatever
```

tiagolobocastro: Have you thought about using a proper backoff algorithm? There's a rust crate for that.
Also is 30 "slow" enough for slow?

control-plane/agents/src/bin/ha/cluster/switchover.rs
```diff
@@ -257,59 +294,77 @@ impl SwitchOverEngine {
...
-        worker_num: u8,
-    ) {
+    async fn worker(self, recv: Arc<Mutex<UnboundedReceiver<SwitchOverRequest>>>, worker_num: u8) {
         loop {
             let mut lo = recv.lock().await;
             if let Some(mut q) = lo.recv().await {
+                drop(lo);
                 info!(
-                    "switchover for vol: {} picked by worker-{}",
+                    "switchover for volume: {} picked by worker-{}",
```
```suggestion
    info!(
        volume.uuid = q.volume_id,
        worker = worker_num,
        "Volume switchover picked by worker",
    );
```

control-plane/agents/src/bin/ha/cluster/switchover.rs
```diff
@@ -257,59 +294,77 @@ impl SwitchOverEngine {
...
         }
     }
 
-    pub fn initiate(&self, req: SwitchOverRequest) -> anyhow::Result<()> {
-        self.channel
-            .send(req)
-            .map_err(|error| anyhow::anyhow!("Failed to send switchover request: {error}"))
+    /// Sends Switchover request to the Channel after sleeping for a specified duration if
+    /// specified.
+    pub fn initiate(
```
tiagolobocastro: nit: can we improve the name here, example: requeue?

control-plane/agents/src/bin/ha/cluster/switchover.rs
```diff
@@ -257,59 +294,77 @@ impl SwitchOverEngine {
...
-                            retry_delay = retry_delay_min;
+                            // reset retry count back to the start after successfully completing a
+                            // stage.
+                            q.retry_count = 0;
                         }
                         Err(_) => {
-                            retry_delay = (retry_delay * 3).min(retry_delay_max);
-                            tokio::time::sleep(retry_delay).await;
+                            q.retry_count += 1;
+                            let retry_delay = if q.retry_count < NUM_OF_FAST_REQUEUE {
```
tiagolobocastro: can we calculate the retry_delay in the requeue function itself?

control-plane/agents/src/bin/ha/cluster/switchover.rs
```diff
@@ -257,59 +294,77 @@ impl SwitchOverEngine {
...
+                                FAST_REQUEUE
+                            } else {
+                                SLOW_REQUEUE
+                            };
+                            let _ =
+                                self.initiate(q.clone(), Some(Duration::from_secs(retry_delay)));
```
tiagolobocastro: Do we need to clone q?

control-plane/agents/src/bin/ha/cluster/switchover.rs
```diff
@@ -257,59 +294,77 @@ impl SwitchOverEngine {
...
-    pub fn initiate(&self, req: SwitchOverRequest) -> anyhow::Result<()> {
-        self.channel
-            .send(req)
-            .map_err(|error| anyhow::anyhow!("Failed to send switchover request: {error}"))
+    /// Sends Switchover request to the Channel after sleeping for a specified duration if
+    /// specified.
+    pub fn initiate(
+        &self,
+        req: SwitchOverRequest,
+        duration: Option<Duration>,
```
tiagolobocastro: If `None` we unwrap to 0 and sleep 0 which is a bit odd.
Why not remove the duration completely, and calculate here, and add a check where if error count is 0 then don't even bother sleeping at all.

control-plane/agents/src/bin/ha/cluster/switchover.rs
```diff
@@ -257,59 +294,77 @@ impl SwitchOverEngine {
...
+    ) -> anyhow::Result<()> {
+        let clone_self = self.clone();
+        let requeue_after = duration.unwrap_or(Duration::from_secs(0));
+        tokio::spawn(async move {
+            tokio::time::sleep(requeue_after).await;
+            clone_self
+                .channel
+                .send(req)
+                .map_err(|error| anyhow::anyhow!("Failed to send switchover request: {error}"))
```
tiagolobocastro: There's no point mapping since we're not doing anything with it?

control-plane/agents/src/bin/ha/cluster/switchover.rs
```diff
@@ -257,59 +294,77 @@ impl SwitchOverEngine {
...
-            .send(req)
-            .map_err(|error| anyhow::anyhow!("Failed to send switchover request: {error}"))
+    /// Sends Switchover request to the Channel after sleeping for a specified duration if
+    /// specified.
+    pub fn initiate(
+        &self,
+        req: SwitchOverRequest,
+        duration: Option<Duration>,
+    ) -> anyhow::Result<()> {
+        let clone_self = self.clone();
```
tiagolobocastro: Could we clone just the channel since we don't seem to use self after?

control-plane/agents/src/bin/ha/cluster/switchover.rs
abhilashshetty04: This function is also used when we receive new Failed path request also. How about enqueue ?

control-plane/agents/src/bin/ha/cluster/switchover.rs
abhilashshetty04: WIll check

control-plane/agents/src/bin/ha/cluster/switchover.rs
tiagolobocastro: Sounds good!

common/src/transport_api/mod.rs
abhilashshetty04: Thank you.

control-plane/agents/src/bin/ha/cluster/switchover.rs
abhilashshetty04: Post 3 minutes we requeue after 30 secs. Do you have any suggestion?

control-plane/agents/src/bin/ha/cluster/switchover.rs
abhilashshetty04: Incorporated

control-plane/agents/src/bin/ha/cluster/switchover.rs
abhilashshetty04: Incorporated

control-plane/agents/src/bin/ha/cluster/switchover.rs
abhilashshetty04: No we dont. I had used logger with q after that function. Moved that before fn hence not required now.

control-plane/agents/src/bin/ha/cluster/switchover.rs
abhilashshetty04: Incorporated

control-plane/agents/src/bin/ha/cluster/switchover.rs
abhilashshetty04: Incorporated

control-plane/agents/src/bin/ha/cluster/switchover.rs
abhilashshetty04: Incorporated

abhilashshetty04:
bors try

Abhinandan-Purkait (approved):

abhilashshetty04:
bors try

control-plane/agents/src/bin/ha/cluster/switchover.rs
tiagolobocastro: I don't know, 1 minute? Not sure.. What do you think @Abhinandan-Purkait ?

control-plane/agents/src/bin/ha/cluster/switchover.rs
```diff
@@ -257,59 +297,72 @@ impl SwitchOverEngine {
...
         }
     }
 
-    pub fn initiate(&self, req: SwitchOverRequest) -> anyhow::Result<()> {
-        self.channel
-            .send(req)
-            .map_err(|error| anyhow::anyhow!("Failed to send switchover request: {error}"))
+    /// Sends Switchover request to the Channel after sleeping for a specified duration if
+    /// specified.
+    pub fn enqueue(&self, req: SwitchOverRequest) -> anyhow::Result<()> {
```
tiagolobocastro: this doesn't return any errors? So doesn't need to return anything?

tiagolobocastro (approved):

control-plane/agents/src/bin/ha/cluster/switchover.rs
abhilashshetty04: Removed returns

control-plane/agents/src/bin/ha/cluster/switchover.rs
abhilashshetty04: Ill keep it as is for now and merge PR. We can always change it when we decide.

abhilashshetty04:
bors merge

PR #373, opened 2022-10-31
fix: idempotent cordon and drain

chriswldenyer:
Ensure that applying a cordon or drain using an existing label will be treated idempotently by the kubectl plugin. i.e. will be handled as if the label had been applied for the first time.

### DIFF
--- a/control-plane/plugin/src/resources/node.rs
+++ b/control-plane/plugin/src/resources/node.rs
@@ -118,11 +118,37 @@ fn drain_labels_from_state(ds: &CordonDrainState) -> Vec<String> {
 impl Cordoning for Node {
     type ID = NodeId;
     async fn cordon(id: &Self::ID, label: &str, output: &OutputFormat) {
-        match RestClient::client()
-            .nodes_api()
-            .put_node_cordon(id, label)
-            .await
-        {
+        // is node already cordoned with the label?
+        let already_has_cordon_label: bool =
+            match RestClient::client().nodes_api().get_node(id).await {
+                Ok(node) => {
+                    let node_body = &node.into_body();
+                    match &node_body.spec {
+                        Some(spec) => match &spec.cordondrainstate {
+                            Some(ds) => cordon_labels_from_state(ds).contains(&label.to_string()),
+                            None => false,
+                        },
+                        None => {
+                            println!("Node {} is not registered", id);
+                            return;
+                        }
+                    }
+                }
+                Err(e) => {
+                    println!("Failed to get node {}. Error {}", id, e);
+                    return;
+                }
+            };
+        let result = match already_has_cordon_label {
+            false => {
+                RestClient::client()
+                    .nodes_api()
+                    .put_node_cordon(id, label)
+                    .await
+            }
+            true => RestClient::client().nodes_api().get_node(id).await,
+        };
+        match result {
             Ok(node) => match output {
                 OutputFormat::Yaml | OutputFormat::Json => {
                     // Print json or yaml based on output format.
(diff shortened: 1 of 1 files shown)


control-plane/plugin/src/resources/node.rs
```diff
@@ -118,23 +118,58 @@ fn drain_labels_from_state(ds: &CordonDrainState) -> Vec<String> {
...
-                OutputFormat::Yaml | OutputFormat::Json => {
-                    // Print json or yaml based on output format.
-                    utils::print_table(output, node.into_body());
+        // is node already cordoned with the label?
+        let already_has_cordon_label: bool =
+            match RestClient::client().nodes_api().get_node(id).await {
+                Ok(node) => {
+                    let node_body = &node.into_body();
+
+                    match &node_body.spec.as_ref().unwrap().cordondrainstate {
```
tiagolobocastro: handle `unwrap`, if no spec then error with node is not registered.

control-plane/plugin/src/resources/node.rs
```diff
@@ -118,23 +118,58 @@ fn drain_labels_from_state(ds: &CordonDrainState) -> Vec<String> {
...
                 }
-                OutputFormat::None => {
-                    // In case the output format is not specified, show a success message.
-                    println!("Node {} cordoned successfully", id)
+                Err(e) => {
+                    println!("Failed to get node {}. Error {}", id, e);
+                    return;
+                }
+            };
+        if !already_has_cordon_label {
```
tiagolobocastro: the code inside these two matches seem to be exactly the same? If so then I'd suggest reusing it to avoid duplication, eg:
```rust
let result = match already_has_cordon_label {
  true => RestClient::client().nodes_api().put_node_cordon(id, label).await,
  false => RestClient::client().nodes_api().get_node(id).await,
};
match result {
  Ok(node) => match output {
      OutputFormat::Yaml | OutputFormat::Json => {
          // Print json or yaml based on output format.
          utils::print_table(output, node.into_body());
      }
      OutputFormat::None => {
          println!("Node {} cordoned successfully", id)
      }
  },
  Err(e) => {
      println!("Failed to cordon node {}. Error {}", id, e)
  }
}
```

control-plane/plugin/src/resources/node.rs
```diff
@@ -378,77 +413,93 @@ impl Drain for Node {
...
+                .nodes_api()
+                .put_node_drain(id, &label)
+                .await
+            {
+                Ok(_node) => {}
+                Err(e) => {
+                    println!("Failed to put node drain {}. Error {}", id, e);
+                    return;
+                }
+            }
```
```suggestion
            if let Err(error) = RestClient::client()
                .nodes_api()
                .put_node_drain(id, &label)
                .await
            {
                    println!("Failed to put node drain {}. Error {}", id, error);
                    return;
            }
```

control-plane/plugin/src/resources/node.rs
```diff
@@ -378,77 +413,93 @@ impl Drain for Node {
...
+        }
+        // loop this call until no longer draining
+        loop {
+            match RestClient::client().nodes_api().get_node(id).await {
+                Ok(node) => {
+                    let node_body = &node.clone().into_body();
+
+                    match &node_body.spec.as_ref().unwrap().cordondrainstate {
```
tiagolobocastro: missing handle unwrap here.

control-plane/plugin/src/resources/node.rs
```diff
@@ -378,77 +413,93 @@ impl Drain for Node {
...
-                    tokio::time::sleep(sleep).await;
+                }
+                Err(e) => {
+                    println!("Failed to get node {}. Error {}", id, e);
+                    break;
                 }
             }
-            Err(e) => {
-                println!("Failed to get node {}. Error {}", id, e)
+            if timeout_instant.is_some() && time::Instant::now() > timeout_instant.unwrap() {
```
tiagolobocastro, line 497: nit: btw you could also "bundle" the code above into a future, and then call it with tokio timeout, which will automatically cancel the future and return when the timeout is reached.

abhilashshetty04 (approved):

control-plane/plugin/src/resources/node.rs
chriswldenyer: fixed

control-plane/plugin/src/resources/node.rs
chriswldenyer: fixed

control-plane/plugin/src/resources/node.rs
chriswldenyer: fixed

control-plane/plugin/src/resources/node.rs
chriswldenyer: fixed

control-plane/plugin/src/resources/node.rs
```diff
@@ -432,23 +483,22 @@ impl Drain for Node {
...
+                println!("Node {} drain command timed out", id);
+                break;
             }
+            let sleep = Duration::from_secs(1);
```
tiagolobocastro: nit: watch default period is 2 seconds, might want to consider doing the same here?

tiagolobocastro (approved):

control-plane/plugin/src/resources/node.rs
chriswldenyer: Sure, now fixed.

chriswldenyer:
bors merge

PR #374, opened 2022-11-01
feat(csi-node): cherry-pick allow controlling of the ctrl_loss_tmo

tiagolobocastro:
feat(csi-node): allow controlling of the ctrl_loss_tmo

This sets up the maximum number of reconnect attempts before the kernel initiator gives up.
Nvme-tcp is not able to reconnect a device after the max_reconnections has been hit and the device
is lost, so we should set a known limit which is sufficiently large to allow for failover.
Conversely a value that is too large will make rebooting tricky as the kernel will keep trying..

Refactored the way storage class parameters are setup from the controller to the node service.
The new parameter name is: nvmeCtrlLossTmo.
TODO: node-specific nvmf options should also be refactored in the same way.

A default global value for the ctrl_loss_tmo can be added via cmdline arg nvme-ctrl-loss-tmo.
This value is overridden by the storageClass parameter.

---

chore(cherry-pick): use rustup with path

On non-nixos when using rustup and the system's libc is not the same as the nix-shell's libc seems
 we get some odd errors:
https://discourse.nixos.org/t/nix-shell-with-rustup/22452
This is likely because the binaries from rustup haven't been patched to point to the nix env.

When this happens set rust-toolchain.toml to our cloned rust toolchain aso the IDE can work
correctly but use the RUSTUP_TOOLCHAIN under the nix-shell so we can use rustup properly.


Abhinandan-Purkait (approved):

control-plane/csi-driver/src/context.rs
```diff
@@ -0,0 +1,162 @@
...
+        let share_protocol = parse_protocol(args.get(Parameters::ShareProtocol.as_ref()))?;
+
+        let io_timeout = Parameters::io_timeout(args.get(Parameters::IoTimeout.as_ref()))
+            .map_err(|_| tonic::Status::invalid_argument("Invalid I/O timeout"))?;
+        let ctrl_loss_tmo =
+            Parameters::ctrl_loss_tmo(args.get(Parameters::NvmeCtrlLossTmo.as_ref()))
+                .map_err(|_| tonic::Status::invalid_argument("Invalid ctrl_loss_tmo"))?;
+
+        let replica_count = match args.get(Parameters::ReplicaCount.as_ref()) {
+            Some(c) => match c.parse::<u8>() {
```
chriswldenyer, line 140: Does this mean it will fail with >= 256 replicas?

control-plane/csi-driver/src/bin/node/config.rs
```diff
@@ -46,35 +64,46 @@ pub(crate) fn config<'a>() -> MutexGuard<'a, Config> {
...
-        // nothing parsed for now
-        Ok(Default::default())
+    fn try_from(value: NvmeParseParams) -> Result<Self, Self::Error> {
+        fn add_param(from: &NvmeParseParams, to: &mut NvmeArgValues, name: &str) {
+            if let Some(value) = from.get(name) {
+                to.0.insert(name.to_string(), value.to_string());
+            }
+        }
+        let mut us = Self::default();
+        add_param(&value, &mut us, Parameters::NvmeCtrlLossTmo.as_ref());
```
chriswldenyer, line 105: Why do we need the embedded function add_param()? Could we not just inline its code here?

chriswldenyer (approved):
LGTM

control-plane/csi-driver/src/bin/node/config.rs
tiagolobocastro: Yes we could indeed, this was just to simplify adding more parameters, like the NrIoQueues, but I ended up removing the NrIoQueues.

control-plane/csi-driver/src/context.rs
tiagolobocastro: Yes correct, I think the core also uses u8 for replica count.

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

abhilashshetty04 (approved):

tiagolobocastro:
bors ping

tiagolobocastro:
bors cancel

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

PR #375, opened 2022-11-03
fix(switchover): lazy shutdown and destroy

tiagolobocastro:
fix(shutdown): don't fail when lazily shutting down the nexus

When the core agent restarts at the same time as the older nexus node, then we don't have
the node state and as such we fail the shutdown call.
For a republish call we want to shutdown in a lazy manner, so ignore this error.
The reconcilers will pick up the slack, if need be.

---

fix(switchover): destroy shutdown targets lazily

The cluster agent has some issues handling this itself, eg: it doesn't know when nodes are
online. It would have to ask the core-agent. Since it doesn't have this functionality yet
it's better to have the core agent handle these destroys lazily.

Also added a switchover test for this effect.


common/src/types/v0/store/mod.rs
```diff
@@ -45,18 +45,26 @@ impl<T> From<SpecStatus<T>> for models::SpecStatus {
...
     }
+    /// Check if resource is being deleted.
     pub fn deleting(&self) -> bool {
         self == &Self::Deleting
     }
+    /// Check if resource is deleted.
     pub fn deleted(&self) -> bool {
         self == &Self::Deleted
     }
+    /// Check if resource is being delete or is deleted.
```
Abhinandan-Purkait: `is being delete` -> `is being deleted`

common/src/types/v0/store/mod.rs
```diff
@@ -45,18 +45,26 @@ impl<T> From<SpecStatus<T>> for models::SpecStatus {
...
+    pub fn deleting_or_deleted(&self) -> bool {
+        self == &Self::Deleting
```
Abhinandan-Purkait: Should we also add `||  self == &Self::Deleted` ?

common/src/types/v0/transport/nexus.rs
```diff
@@ -409,8 +412,18 @@ impl DestroyNexus {
             node,
             uuid,
             disowners: NexusOwners::None,
+            lazy: false,
```
Abhinandan-Purkait, line 415: Should we take the lazy from parameters here as well like in ShutdownNexus below?

Abhinandan-Purkait (approved):

common/src/types/v0/store/mod.rs
tiagolobocastro: oops, well spotted!

common/src/types/v0/transport/nexus.rs
tiagolobocastro: Most calls would set it to false anyway, which is why I didn't add it, it can be add with the builder type method.

tiagolobocastro:
bors merge

PR #377, opened 2022-11-07
fix(csi-node): add several fixes

tiagolobocastro:
fix(csi-node): don't unstage if bind mount exists

If an application creates a bind mount for our volume unstaging can be very troublesome as it
leaves the mount in a really bad state.
Even if we reconnect the device, it's very unlikely things won't recover.

This commit adds a check on node_unstage for unknown mounts (as in, not a csi publish mount).
Should there be any we output this as a trace and simply fail the unstage.
It's up to the user to clean up non-csi mounts, otherwise we can never unstage.

---

fix(csi-node): remount staging path as ro before final umount

Sometimes after/during disconnect we get some page write errors, triggered by a journal update by
the JBD2 worker task.
We don't really know how we can "flush" these tasks so at the moment the best solution we have is
to remount the staging path as RO before we umount it for good, which seems to help.

---

fix(csi-node): add noddy sleep between umount and disconnect

Sometimes when disconnecting we see page read errors due to ENXIO.
There seems to be some race in the kernel when removing a device with queued IOs.
While this is not strictly an issue, it may confuse or hide other problems.
Although an ugly WA, sleeping between umount and disconnect seems to alleviate this.

### DIFF
--- a/control-plane/csi-driver/src/bin/node/filesystem_vol.rs
+++ b/control-plane/csi-driver/src/bin/node/filesystem_vol.rs
@@ -59,15 +59,21 @@ pub(crate) async fn stage_fs_volume(
         }
     };
 
-    if mount::find_mount(Some(device_path), Some(fs_staging_path)).is_some() {
+    if let Some(existing) = mount::find_mount(Some(device_path), Some(fs_staging_path)) {
         debug!(
             "Device {} is already mounted onto {}",
             device_path, fs_staging_path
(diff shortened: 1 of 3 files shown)


chriswldenyer (approved):
Looks reasonable

Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

PR #380, opened 2022-11-14
chore: cluster agent to reuse existing target for switchover

Abhinandan-Purkait:
Cluster agent would now try to reuse the existing target first and if that fails, it would do a mandatory move of the target for switchover.

Signed-off-by: Abhinandan Purkait <purkaitabhinandan@gmail.com>

### DIFF
--- a/control-plane/agents/src/bin/core/controller/scheduling/mod.rs
+++ b/control-plane/agents/src/bin/core/controller/scheduling/mod.rs
@@ -83,6 +83,15 @@ impl NodeFilters {
             .into_iter()
             .any(|node_spec| node_spec.id() == item.node_wrapper().id())
     }
+
+    /// Should only attempt to use node where current target is not present.
+    pub(crate) fn current_target(request: &GetSuitableNodesContext, item: &NodeItem) -> bool {
+        if let Some(target) = &request.target {
+            target.node() != item.node_wrapper().id()
+        } else {
+            true
+        }
+    }
 }
 
 /// Filter pools used for replica creation.
--- a/control-plane/agents/src/bin/ha/cluster/switchover.rs
+++ b/control-plane/agents/src/bin/ha/cluster/switchover.rs
@@ -1,26 +1,31 @@
 use crate::{core_grpc, etcd::EtcdStore, nodes::NodeList};
+use anyhow::anyhow;
 use chrono::Utc;
-use common_lib::types::v0::{
-    store::{
-        switchover::{Operation, OperationState, SwitchOverSpec, SwitchOverTime},
-        SpecTransaction,
-    },
-    transport::{
-        DestroyShutdownTargets, ReplacePath, RepublishVolume, VolumeId, VolumeShareProtocol,
+use common_lib::{
+    transport_api::ReplyErrorKind,
+    types::v0::{
+        store::{
+            switchover::{Operation, OperationState, SwitchOverSpec, SwitchOverTime},
+            SpecTransaction,
+        },
+        transport::{
+            DestroyShutdownTargets, ReplacePath, RepublishVolume, VolumeId, VolumeShareProtocol,
+        },
     },
 };
-use grpc::operations::{ha_node::client::NodeAgentClient, volume::traits::VolumeOperations};
+use grpc::operations::{
+    ha_node::{client::NodeAgentClient, traits::NodeAgentOperations},
+    volume::traits::VolumeOperations,
+};
 use serde::{Deserialize, Serialize};
-use tokio::sync::mpsc::{unbounded_channel, UnboundedReceiver, UnboundedSender};
+use std::{convert::TryFrom, net::SocketAddr, sync::Arc, time::Duration};
+use tokio::sync::{
+    mpsc::{unbounded_channel, UnboundedReceiver, UnboundedSender},
+    Mutex,
+};
+use tonic::transport::Uri;
 use tracing::{error, info};
 use utils::NVME_TARGET_NQN_PREFIX;
-
-use anyhow::anyhow;
-use common_lib::transport_api::ReplyErrorKind;
-use grpc::operations::ha_node::traits::NodeAgentOperations;
-use std::{convert::TryFrom, net::SocketAddr, sync::Arc, time::Duration};
-use tokio::sync::Mutex;
-use tonic::transport::Uri;
 
 fn client() -> impl VolumeOperations {
     core_grpc().volume()
@@ -57,6 +62,8 @@ pub struct SwitchOverRequest {
     new_path: Option<String>,
     // Number of failed attempts in the current Stage.
     retry_count: u64,
+    // Reuse existing target.
+    reuse_existing: bool,
 }
 
 impl SwitchOverRequest {
@@ -74,11 +81,16 @@ impl SwitchOverRequest {
             existing_nqn: existing_path,
             new_path: None,
             retry_count: 0,
+            reuse_existing: true,
         }
     }
 
     pub fn with_stage(&mut self, stage: Stage) {
         self.stage = stage;
+    }
+
+    pub fn with_reuse_existing(&mut self, reuse_existing: bool) {
+        self.reuse_existing = reuse_existing;
     }
 
     pub fn stage(&self) -> Stage {
@@ -154,7 +166,7 @@ impl SwitchOverRequest {
             uuid: self.volume_id.clone(),
             target_node: None,
             share: VolumeShareProtocol::Nvmf,
-            reuse_existing: false,
+            reuse_existing: self.reuse_existing,
         };
         let vol = client().republish(&republish_req, None).await?;
         self.new_path = match vol.state().target {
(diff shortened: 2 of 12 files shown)


Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors try

control-plane/agents/src/bin/core/controller/scheduling/mod.rs
```diff
@@ -83,6 +83,15 @@ impl NodeFilters {
             .into_iter()
             .any(|node_spec| node_spec.id() == item.node_wrapper().id())
     }
+
+    /// Should only attempt to use node where current target is not present.
+    pub(crate) fn current_target(request: &GetSuitableNodesContext, item: &NodeItem) -> bool {
+        if let Some(target) = request.target.clone() {
```
tiagolobocastro: don't need to clone?

control-plane/agents/src/bin/ha/cluster/switchover.rs
```diff
@@ -230,12 +242,21 @@ impl SwitchOverRequest {
...
                 if let Err(e) = client.replace_path(&replace_request, None).await {
                     return if e.kind == ReplyErrorKind::FailedPrecondition {
                         error!(nexus.path=%self.existing_nqn, "HA Node agent could not find failed Nvme path");
                         info!(volume.uuid=%self.volume_id, "Moving Switchover request to Errored state");
                         self.stage = Stage::Errored;
                         Err(anyhow!("HA Node agent could not lookup old Nvme path"))
+                    } else if e.kind == ReplyErrorKind::Aborted
+                        || e.kind == ReplyErrorKind::Timeout
+                        || e.kind == ReplyErrorKind::NotFound
+                    {
```
```suggestion
                    } else if matches!(e.kind, ReplyErrorKind::Aborted | ReplyErrorKind::Timeout | ReplyErrorKind::NotFound)
                    {
```

control-plane/agents/src/bin/ha/cluster/switchover.rs
```diff
@@ -230,12 +242,21 @@ impl SwitchOverRequest {
...
+                        info!("Retrying Republish without older nexus reuse");
```
```suggestion
                        info!("Retrying Republish without older target reuse");
```

control-plane/agents/src/bin/ha/node/server.rs
```diff
@@ -65,24 +65,37 @@ impl NodeAgentSvc {
...
+    let port = parsed_path.1;
+
     match get_nvme_path_buf(&ctrlr.path) {
         Some(pbuf) => {
             let subsystem = Subsystem::new(pbuf.as_path()).map_err(|_| SvcError::Internal {
                 details: "Failed to get NVMe subsystem for controller".to_string(),
             })?;
 
-            tracing::info!(
+            if subsystem.address == format!("traddr={},trsvcid={}", host, port) {
```
tiagolobocastro: Can we add a new type SubsystemAddr, and here use SubsystemAddr::new(host, port) ?

control-plane/agents/src/bin/ha/node/server.rs
```diff
@@ -100,11 +113,11 @@ fn disconnect_controller(ctrlr: &NvmeController) -> Result<(), SvcError> {
 impl NodeAgentSvc {
     /// Translate new path URI into connection arguments for Subsystem connect API.
     fn get_nvmf_connection_args(&self, new_path: &str) -> Option<ConnectArgs> {
-        let uri = new_path.parse::<Uri>().ok()?;
+        let parsed_path = parse_uri(new_path)?;
```
tiagolobocastro: Any reason for changing these? The previous one seemed cleaner.

control-plane/agents/src/bin/ha/node/server.rs
```diff
@@ -132,26 +145,31 @@ impl NodeAgentSvc {
...
-                );
-                return Err(SvcError::Internal { details: nvme_err });
-            }
+        // which immediately resumes all stalled I/O
+
+        let parsed_uri = parse_uri(new_path.as_str()).ok_or(SvcError::InvalidArguments {})?;
+        let mut subsystem = match get_subsystem(parsed_uri) {
+            Ok(subsystem) => subsystem,
+            Err(_) => match connect_args.connect() {
+                Ok(c) => {
```
tiagolobocastro: s/c/subsystem

control-plane/agents/src/bin/ha/node/server.rs
```diff
@@ -239,13 +257,38 @@ impl NodeAgentOperations for NodeAgentSvc {
...
+    let uri = new_path.parse::<Uri>().ok()?;
+
+    let host = uri.host()?;
+    let port = uri.port()?.to_string();
+    let nqn = &uri.path()[1 ..];
+    Some((host.to_string(), port, nqn.to_string()))
+}
+
+// Returns the particular subsystem based on the nqn and address.
+fn get_subsystem(parsed_uri: (String, String, String)) -> Result<Subsystem, SvcError> {
```
tiagolobocastro: having an Uri instead of an unnamed tuple is cleaner here IMHO.
another alternative is to create our own parsed uri type with pub getters.

control-plane/agents/src/bin/ha/node/server.rs
```diff
@@ -239,13 +257,38 @@ impl NodeAgentOperations for NodeAgentSvc {
...
+    let nvme_subsystems = NvmeSubsystems::new().map_err(|_| SvcError::SubsystemNotFound {
+        nqn: parsed_uri.2.clone(),
+    })?;
+    for subsys in nvme_subsystems.flatten() {
+        if subsys.nqn == parsed_uri.2.clone()
+            && subsys.address == format!("traddr={},trsvcid={}", parsed_uri.0, parsed_uri.1)
```
tiagolobocastro: Again here, I think we should have a proper type for subsys addr rather than manipulating strings everywhere.

control-plane/agents/src/bin/ha/node/server.rs
```diff
@@ -239,13 +257,38 @@ impl NodeAgentOperations for NodeAgentSvc {
...
+    Some((host.to_string(), port, nqn.to_string()))
+}
+
+// Returns the particular subsystem based on the nqn and address.
+fn get_subsystem(parsed_uri: (String, String, String)) -> Result<Subsystem, SvcError> {
+    let nvme_subsystems = NvmeSubsystems::new().map_err(|_| SvcError::SubsystemNotFound {
+        nqn: parsed_uri.2.clone(),
+    })?;
+    for subsys in nvme_subsystems.flatten() {
+        if subsys.nqn == parsed_uri.2.clone()
```
tiagolobocastro: don't need to clone to compare?

control-plane/agents/src/common/errors.rs
```diff
@@ -216,6 +216,8 @@ pub enum SvcError {
     MaxRebuilds { max_rebuilds: u32 },
     #[snafu(display("The api version: {:?} is not valid", api_version))]
     InvalidApiVersion { api_version: Option<ApiVersion> },
+    #[snafu(display("The subsystem with nqn: {:?} is not found", nqn))]
```
tiagolobocastro: String doesn't need Debug formatter?

tests/bdd/features/ha/node-agent/test_path_replacement.py
```diff
@@ -125,7 +125,7 @@ def background():
         VOLUME_UUID, Protocol("nvmf"), node=TARGET_NODE_1
     )
     yield volume
-    Deployer.stop()
+    # Deployer.stop()
```
tiagolobocastro: guess this is from debug, we should add stop back.

control-plane/agents/src/bin/ha/node/server.rs
Abhinandan-Purkait: Yeah, that would be good. Thanks

control-plane/agents/src/bin/ha/node/server.rs
Abhinandan-Purkait: The reason for changing it is, since now its being used at multiple places, I wanted to map and return error from one place rather than from each getter, but yeah using a tuple is indeed not cleaner. I will use a `ParsedUri` type for the same.

control-plane/agents/src/bin/ha/node/server.rs
Abhinandan-Purkait: Going with `ParsedUri` type as that suits this use case.

control-plane/agents/src/common/errors.rs
Abhinandan-Purkait: Ahh right. Thanks

tests/bdd/features/ha/node-agent/test_path_replacement.py
Abhinandan-Purkait: Oops, reverting it back

control-plane/agents/src/bin/ha/node/server.rs
Abhinandan-Purkait: Removed.

Abhinandan-Purkait:
bors try

abhilashshetty04 (approved):

control-plane/agents/src/bin/ha/node/server.rs
```diff
@@ -65,24 +65,36 @@ impl NodeAgentSvc {
...
+
     match get_nvme_path_buf(&ctrlr.path) {
         Some(pbuf) => {
             let subsystem = Subsystem::new(pbuf.as_path()).map_err(|_| SvcError::Internal {
                 details: "Failed to get NVMe subsystem for controller".to_string(),
             })?;
 
-            tracing::info!(
+            if subsystem.address
+                == SubsystemAddr::new(parsed_path.host(), parsed_path.port()).string_value()
```
tiagolobocastro: You can impl Eq for SubsystemAddr so you don't need to call string_value explicitly

tiagolobocastro (approved):

control-plane/agents/src/bin/ha/node/server.rs
Abhinandan-Purkait: Right. Would do that. Thank you

Abhinandan-Purkait:
bors merge

PR #381, opened 2022-11-15
chore(ha_nexus): allow errored and successful requests to be enqueued at startup

abhilashshetty04:
When cluster agent starts. It gets Switchover request from Etcd and enqueues them in channel so they can get process. We were sending only In progress requests. Now with this change all requests present in Etcd irrespective of the stage will be enqueued and precossed to completion.

Added debug in Etcd store failure to get actual store error.

### DIFF
--- a/control-plane/agents/src/bin/ha/cluster/etcd.rs
+++ b/control-plane/agents/src/bin/ha/cluster/etcd.rs
@@ -11,7 +11,7 @@ use common_lib::{
 use futures::lock::Mutex;
 use http::Uri;
 use std::{convert::TryFrom, sync::Arc, time::Duration};
-use tracing::debug;
+use tracing::{debug, error};
 
 /// Represent object to access Etcd.
 #[derive(Debug, Clone)]
@@ -32,10 +32,13 @@ impl EtcdStore {
                     timeout,
                 })
             }
-            Err(_) => Err(StoreError::Timeout {
-                operation: "connect".to_string(),
-                timeout,
-            }),
+            Err(error) => {
+                error!(%error, "Failed to create persistent store client");
+                Err(StoreError::Timeout {
+                    operation: "connect".to_string(),
+                    timeout,
+                })
+            }
         }
     }
 
@@ -44,11 +47,14 @@ impl EtcdStore {
         let mut store = self.store.lock().await;
         match tokio::time::timeout(self.timeout, async move { store.put_obj(object).await }).await {
             Ok(result) => result.map_err(Into::into),
-            Err(_) => Err(StoreError::Timeout {
-                operation: "Put".to_string(),
-                timeout: self.timeout,
+            Err(error) => {
+                error!(%error, "Failed to write to persistent store");
+                Err(StoreError::Timeout {
+                    operation: "Put".to_string(),
+                    timeout: self.timeout,
+                }
+                .into())
             }
-            .into()),
         }
     }
 
@@ -61,11 +67,14 @@ impl EtcdStore {
         .await
         {
             Ok(_) => Ok(()),
-            Err(_) => Err(StoreError::Timeout {
-                operation: "Delete".to_string(),
-                timeout: self.timeout,
+            Err(error) => {
+                error!(%error, "Failed to delete from persistent store");
+                Err(StoreError::Timeout {
+                    operation: "Delete".to_string(),
+                    timeout: self.timeout,
+                }
+                .into())
             }
-            .into()),
         }
     }
 
@@ -86,10 +95,8 @@ impl EtcdStore {
 
         let mut partial_entries = vec![];
         for spec in entries {
-            if !spec.is_completed() {
-                let req = SwitchOverRequest::try_from(&spec)?;
-                partial_entries.push(req)
-            }
+            let req = SwitchOverRequest::try_from(&spec)?;
+            partial_entries.push(req)
         }
 
         Ok(partial_entries)
--- a/control-plane/agents/src/bin/ha/cluster/volume.rs
+++ b/control-plane/agents/src/bin/ha/cluster/volume.rs
@@ -48,8 +48,7 @@ impl VolumeMover {
         &self,
         mut req: Vec<SwitchOverRequest>,
     ) -> Result<(), anyhow::Error> {
-        req.sort_by_key(|r| r.timestamp());
-
+        req.sort();
         for entry in req {
             entry.start_op(entry.stage(), &self.etcd).await?;
             self.engine.enqueue(entry);
--- a/common/src/types/v0/store/switchover.rs
+++ b/common/src/types/v0/store/switchover.rs
@@ -78,7 +78,7 @@ impl SwitchOverSpec {
         if let Some(op) = &self.operation {
             match op.operation {
                 Operation::Errored(_) => true,
-                Operation::ReplacePath => matches!(op.result.unwrap_or(false), true),
+                Operation::Successful => matches!(op.result.unwrap_or(false), true),
                 _ => false,
             }
         } else {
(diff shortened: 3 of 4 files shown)


Abhinandan-Purkait:
> When cluster agent starts. It gets Switchover request from Etcd and enqueues them in channel so they can get process. We were sending only In progress requests. Now with this change all requests present in Etcd irrespective of the stage will be enqueued and precossed to completion.
> 
> Added debug in Etcd store failure to get actual store error.
> 
> Signed-off-by: Abhilash Shetty [abhilash.shetty@datacore.com](mailto:abhilash.shetty@datacore.com)

Why do we want to enqueue errored and successful requests? How that would be useful?

abhilashshetty04:
@Abhinandan-Purkait , By current logic Errored and Switchover request should be deleted. If cluster agent crashes/restarts before that happens by sending them again to Switchover engine when cluster agent starts we are just cleaning up those entries.

Abhinandan-Purkait:
> @Abhinandan-Purkait , By current logic Errored and Switchover request should be deleted. If cluster agent crashes/restarts before that happens by sending them again to Switchover engine when cluster agent starts we are just cleaning up those entries.

I understand that part, but just wondering would that not have a toll on the in progress requests, as the in progress requests are more priority IMO. One thing that can be done I think is, we can sort the requests keeping errored/successful at the end of the queue, so that in progress ones are picked up first always.

control-plane/agents/src/bin/ha/cluster/etcd.rs
```diff
@@ -44,11 +44,14 @@ impl EtcdStore {
         let mut store = self.store.lock().await;
         match tokio::time::timeout(self.timeout, async move { store.put_obj(object).await }).await {
             Ok(result) => result.map_err(Into::into),
-            Err(_) => Err(StoreError::Timeout {
-                operation: "Put".to_string(),
-                timeout: self.timeout,
+            Err(err) => {
+                error!("Failed to write to etcd. error: {}", err);
```
tiagolobocastro:
```suggestion
                error!(error=%error, "Failed to write to persistent store");
```

tiagolobocastro:
> One thing that can be done I think is, we can sort the requests keeping errored/successful at the end of the queue, so that in progress ones are picked up first always.

That's a good idea, we should order requests in a way that would allow us to process those ones last. We can impl `Ord` IIRC on `SwitchoverRequest` and sort the entries.

abhilashshetty04:
> > @Abhinandan-Purkait , By current logic Errored and Switchover request should be deleted. If cluster agent crashes/restarts before that happens by sending them again to Switchover engine when cluster agent starts we are just cleaning up those entries.
> 
> I understand that part, but just wondering would that not have a toll on the in progress requests, as the in progress requests are more priority IMO. One thing that can be done I think is, we can sort the requests keeping errored/successful at the end of the queue, so that in progress ones are picked up first always.

There is already sorting happening in terms of creation timestamp for the request.  See this..

https://github.com/openebs/mayastor-control-plane/blob/develop/control-plane/agents/src/bin/ha/cluster/volume.rs#L51

I believe its more fair to sort it via timestamp. Even if Errored/Successful enqueues first it will straightaway get deleted right?

Abhinandan-Purkait:
> > > @Abhinandan-Purkait , By current logic Errored and Switchover request should be deleted. If cluster agent crashes/restarts before that happens by sending them again to Switchover engine when cluster agent starts we are just cleaning up those entries.
> > 
> > 
> > I understand that part, but just wondering would that not have a toll on the in progress requests, as the in progress requests are more priority IMO. One thing that can be done I think is, we can sort the requests keeping errored/successful at the end of the queue, so that in progress ones are picked up first always.
> 
> There is already sorting happening in terms of creation timestamp for the request. See this..
> 
> https://github.com/openebs/mayastor-control-plane/blob/develop/control-plane/agents/src/bin/ha/cluster/volume.rs#L51
> 
> I believe its more fair to sort it via timestamp. Even if Errored/Successful enqueues first it will straightaway get deleted right?

Cleanup is less important than serving in progress requests, and since you said its sorted by timestamp, if we have a lot of errored and successful requests prior to the in progress ones, I believe there would be unnecessary delay in serving in-progress ones.

abhilashshetty04:
> 
Sure Will change the enqueue ordering.

control-plane/agents/src/bin/ha/cluster/etcd.rs
abhilashshetty04: Incorporated

abhilashshetty04:
> > 
> 
> That's a good idea, we should order requests in a way that would allow us to process those ones last. We can impl Ord IIRC on SwitchoverRequest and sort the entries.

I have changed how we enqueue requests from etcd. I thought of maintaining another list for Errored/Successful. Send them once others are done.  Let me know if its fine.

abhilashshetty04:
> > 
> 
> Sure Will change the enqueue ordering.

Incorporated

Abhinandan-Purkait (approved):

control-plane/agents/src/bin/ha/cluster/etcd.rs
```diff
@@ -32,10 +32,13 @@ impl EtcdStore {
                     timeout,
                 })
             }
-            Err(_) => Err(StoreError::Timeout {
-                operation: "connect".to_string(),
-                timeout,
-            }),
+            Err(err) => {
```
tiagolobocastro: nit: why not simply call err error, then can simply use error!(%error, "Failed...

control-plane/agents/src/bin/ha/cluster/etcd.rs
```diff
@@ -32,10 +32,13 @@ impl EtcdStore {
...
+                error!(error=%err, "Failed to write to persistent store");
```
tiagolobocastro: Should this be Failed to connect rather than write?

control-plane/agents/src/bin/ha/cluster/etcd.rs
```diff
@@ -61,11 +67,14 @@ impl EtcdStore {
         .await
         {
             Ok(_) => Ok(()),
-            Err(_) => Err(StoreError::Timeout {
-                operation: "Delete".to_string(),
-                timeout: self.timeout,
+            Err(err) => {
+                error!(error=%err, "Failed to write to persistent store");
```
tiagolobocastro: Failed to delete from persistent store?

control-plane/agents/src/bin/ha/cluster/volume.rs
```diff
@@ -49,10 +49,19 @@ impl VolumeMover {
         mut req: Vec<SwitchOverRequest>,
     ) -> Result<(), anyhow::Error> {
         req.sort_by_key(|r| r.timestamp());
-
+        let mut delayed_enqueue = vec![];
```
tiagolobocastro: this seems quite convoluted, why not simply sort? Just impl `Ord` trait for a `SwitchOverRequest` and then simply req.sort() ?

control-plane/agents/src/bin/ha/cluster/etcd.rs
abhilashshetty04: Incorporated

control-plane/agents/src/bin/ha/cluster/etcd.rs
abhilashshetty04: Incorporated

control-plane/agents/src/bin/ha/cluster/volume.rs
abhilashshetty04: Incorporated

tiagolobocastro (approved):

abhilashshetty04:
bors try

abhilashshetty04:
bors try

abhilashshetty04:
bors try

abhilashshetty04:
bors merge

PR #390, opened 2022-11-24
fix(csi-node): don't detach if mounts are present

tiagolobocastro:
When staging fails, we must not stage if another mount point exists already, which can happen when trying to stage with a different staging path. When unstaging, don't detach if any mount points exist for the device.

### DIFF
--- a/control-plane/csi-driver/src/bin/node/node.rs
+++ b/control-plane/csi-driver/src/bin/node/node.rs
@@ -143,6 +143,17 @@ async fn detach(uuid: &Uuid, errheader: String) -> Result<(), Status> {
         let device_path = device.devname();
         debug!("Detaching device {}", device_path);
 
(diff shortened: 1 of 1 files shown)


control-plane/csi-driver/src/bin/node/node.rs
```diff
@@ -550,14 +561,18 @@ impl node_server::Node for Node {
...
                 {
-                    detach(
-                        &uuid,
-                        format!(
-                            "Failed to stage volume {}: {};",
-                            &msg.volume_id, fsmount_error
-                        ),
-                    )
-                    .await?;
+                    let mounts = crate::mount::find_src_mounts(&device_path, None);
```
blaisedias, line 564: Multiple mounts on the same device? How would that occur?

blaisedias (approved):
Ah ignore my previous comment

tiagolobocastro:
bors merge

PR #394, opened 2022-11-30
feat: implement node drain reconciler

chriswldenyer:
Implement the reconciler for draining nodes.
This uses republish_volume to initiate the move for all volumes on a node in the draining state,
and then waits for the nexus to be removed.
On success, the node's state is set to drained.
Nodes are processed serially within the loop.

### DIFF
--- a/common/src/types/v0/store/node.rs
+++ b/common/src/types/v0/store/node.rs
@@ -112,6 +112,14 @@ impl CordonDrainState {
             CordonDrainState::Drained(_) => false,
         }
     }
+    /// Returns whether the state corresponds to drained.
+    pub fn is_drained(&self) -> bool {
+        match self {
+            CordonDrainState::Cordoned(_) => false,
+            CordonDrainState::Draining(_) => false,
+            CordonDrainState::Drained(_) => true,
+        }
+    }
 }
 
 /// Node information.
@@ -194,10 +202,39 @@ impl NodeSpec {
         self.endpoint = endpoint;
     }
 
+    /// Ensure the state is consistent with the labels.
+    pub fn resolve(&mut self) {
+        match &mut self.cordon_drain_state {
+            Some(ds) => match ds {
+                CordonDrainState::Cordoned(state) => {
+                    if state.cordonlabels.is_empty() {
+                        self.cordon_drain_state = None;
+                    }
+                }
+                CordonDrainState::Draining(state) | CordonDrainState::Drained(state) => {
+                    // the drain labels should not be empty, if not - fix the state
+                    if state.drainlabels.is_empty() {
+                        if state.cordonlabels.is_empty() {
+                            self.cordon_drain_state = None;
+                        } else {
+                            self.cordon_drain_state = Some(CordonDrainState::Cordoned(
+                                CordonedState::new(state.cordonlabels.clone()),
+                            ));
+                        }
+                    }
+                }
+            },
+            None => {}
+        }
+    }
+
     /// Cordon node by applying the label.
     pub fn cordon(&mut self, label: String) {
         match &mut self.cordon_drain_state {
-            Some(ds) => ds.add_cordon_label(&label),
+            Some(ds) => {
+                ds.add_cordon_label(&label);
+                self.resolve();
+            }
             None => {
                 //add the label and set the state to cordoned
                 self.cordon_drain_state = Some(CordonDrainState::cordon(&label));
@@ -235,36 +272,34 @@ impl NodeSpec {
             }
         }
     }
+
+    /// Move state from Draining to Drained, no change to the labels.
+    pub fn set_drained(&mut self) {
+        if let Some(CordonDrainState::Draining(state)) = &mut self.cordon_drain_state {
+            self.cordon_drain_state = Some(CordonDrainState::Drained(DrainState::new(
+                state.cordonlabels.clone(),
+                state.drainlabels.clone(),
+            )));
+            self.resolve();
+        }
+    }
+
     /// Uncordon node by removing the corresponding label.
     pub fn uncordon(&mut self, label: String) {
         match &mut self.cordon_drain_state {
             Some(ds) => match ds {
                 CordonDrainState::Cordoned(state) => {
-                    // remove the cordon label
-                    // if none left set ds to none
                     state.remove_label(&label);
-                    if state.cordonlabels.is_empty() {
-                        self.cordon_drain_state = None;
-                    }
                 }
                 CordonDrainState::Draining(state) | CordonDrainState::Drained(state) => {
-                    // if no drain labels left, set to cordoned or none
                     state.remove_label(&label);
-                    if state.drainlabels.is_empty() {
-                        if state.cordonlabels.is_empty() {
-                            self.cordon_drain_state = None;
-                        } else {
-                            self.cordon_drain_state = Some(CordonDrainState::Cordoned(
-                                CordonedState::new(state.cordonlabels.clone()),
-                            ));
-                        }
-                    }
                 }
             },
             None => {
                 // should not be possible
             }
         }
+        self.resolve();
     }
 
     /// Returns whether or not the node is cordoned.
@@ -297,6 +332,13 @@ impl NodeSpec {
     pub fn is_draining(&self) -> bool {
         match &self.cordon_drain_state {
             Some(ds) => ds.is_draining(),
+            None => false,
+        }
+    }
+    /// Returns true if the node is in the drained state.
+    pub fn is_drained(&self) -> bool {
+        match &self.cordon_drain_state {
+            Some(ds) => ds.is_drained(),
             None => false,
         }
     }
--- a/control-plane/agents/src/bin/core/controller/reconciler/node/mod.rs
+++ b/control-plane/agents/src/bin/core/controller/reconciler/node/mod.rs
@@ -1,0 +1,41 @@
+mod nexus;
+
+use crate::controller::{
+    reconciler::node::nexus::NodeNexusReconciler,
+    task_poller::{PollContext, PollPeriods, PollResult, PollTimer, TaskPoller},
+};
+
+/// Node reconciler loop which moves nexuses from draining nodes.
+#[derive(Debug)]
+pub(crate) struct NodeReconciler {
+    counter: PollTimer,
+    poll_targets: Vec<Box<dyn TaskPoller>>,
+}
+impl NodeReconciler {
+    /// Return new `Self` with the provided period.
+    pub(crate) fn from(period: PollPeriods) -> Self {
+        NodeReconciler {
+            counter: PollTimer::from(period),
+            poll_targets: vec![Box::new(NodeNexusReconciler::new())],
+        }
+    }
+    /// Return new `Self` with the default period.
+    pub(crate) fn new() -> Self {
+        Self::from(1)
+    }
+}
+
+#[async_trait::async_trait]
+impl TaskPoller for NodeReconciler {
+    async fn poll(&mut self, context: &PollContext) -> PollResult {
+        let mut results = Vec::with_capacity(self.poll_targets.len());
+        for target in &mut self.poll_targets {
+            results.push(target.try_poll(context).await);
+        }
+        Self::squash_results(results)
+    }
+
+    async fn poll_timer(&mut self, _context: &PollContext) -> bool {
+        self.counter.poll()
+    }
+}
--- a/control-plane/agents/src/bin/core/controller/reconciler/node/nexus.rs
+++ b/control-plane/agents/src/bin/core/controller/reconciler/node/nexus.rs
@@ -1,0 +1,164 @@
+use crate::controller::{
+    reconciler::{PollContext, TaskPoller},
+    resources::{
+        operations::ResourcePublishing, operations_helper::OperationSequenceGuard, OperationGuard,
+        ResourceMutex,
+    },
+    task_poller::{PollEvent, PollResult, PollTimer, PollerState},
+};
+use agents::errors::SvcError;
+use common_lib::types::v0::{
+    store::{node::NodeSpec, volume::VolumeSpec},
+    transport::{NodeId, RepublishVolume, VolumeId, VolumeShareProtocol},
+};
+/// Node drain reconciler.
+#[derive(Debug)]
+pub(super) struct NodeNexusReconciler {
+    counter: PollTimer,
+}
+impl NodeNexusReconciler {
+    /// Return a new `Self`.
+    pub(super) fn new() -> Self {
+        Self {
+            counter: PollTimer::from(1), // sets the reconciler polling rate
+        }
+    }
+}
+
+#[async_trait::async_trait]
+impl TaskPoller for NodeNexusReconciler {
+    async fn poll(&mut self, context: &PollContext) -> PollResult {
+        let nodes = context.specs().get_nodes();
+        let mut results = Vec::with_capacity(nodes.len());
+
+        for node in nodes {
+            results.push(check_and_drain_node(context, &node).await);
+        }
+        Self::squash_results(results)
+    }
+
+    async fn poll_timer(&mut self, _context: &PollContext) -> bool {
+        self.counter.poll()
+    }
+
+    async fn poll_event(&mut self, context: &PollContext) -> bool {
+        matches!(context.event(), PollEvent::TimedRun)
+    }
+}
+
+/// Republish the specified volume on any node other than its current node.
+async fn republish_volume(
+    volume: &mut OperationGuard<ResourceMutex<VolumeSpec>, VolumeSpec>,
+    context: &PollContext,
+    vol_uuid: &VolumeId,
+    frontend_node: &NodeId,
+) -> Result<(), SvcError> {
+    let request = RepublishVolume::new(
+        vol_uuid.clone(),
+        None,
+        frontend_node.clone(),
+        VolumeShareProtocol::Nvmf,
+        false,
+    );
+    tracing::info!(
+        volume.uuid = vol_uuid.as_str(),
+        "Attempting to republish volume"
+    );
+
+    volume.republish(context.registry(), &request).await?;
+
+    Ok(())
+}
+
+/// Drain the specified node if in draining state
+async fn check_and_drain_node(context: &PollContext, node_spec: &NodeSpec) -> PollResult {
+    if !node_spec.is_draining() {
+        return PollResult::Ok(PollerState::Idle);
+    }
+
+    let node_id = node_spec.id();
+    tracing::trace!(node.id = node_spec.id().as_str(), "Draining node");
+    let vol_specs = context.specs().get_locked_volumes();
+
+    let mut move_failures = false;
+
+    // Iterate through all the volumes, find those with a nexus hosted on the
+    // node and move each one away.
+    for vol_spec in vol_specs {
+        match vol_spec.operation_guard() {
+            Ok(mut guarded_vol_spec) => {
+                if let Some(target) = guarded_vol_spec.as_ref().target() {
+                    if target.node() != node_id {
+                        continue; // some other node's volume, ignore
+                    }
+                    let nexus_id = target.nexus().clone();
+                    let vol_id = guarded_vol_spec.as_ref().uuid.clone();
+
+                    let config = match guarded_vol_spec.as_ref().config() {
+                        None => {
+                            tracing::error!("Failed to get config");
+                            move_failures = true;
+                            continue;
+                        }
+                        Some(config) => config,
+                    };
+                    let frontend_node = config
+                        .frontend()
+                        .node_names()
+                        .first()
+                        .unwrap_or(&String::default())
+                        .clone();
+                    let frontend_node_id: NodeId = frontend_node.into();
+                    // frontend_node could be "", republish will still be allowed.
+                    tracing::info!(
+                        volume.uuid = vol_id.as_str(),
+                        nexus.uuid = nexus_id.as_str(),
+                        node.id = node_spec.id().as_str(),
+                        "Moving volume"
+                    );
+                    if let Err(e) =
+                        republish_volume(&mut guarded_vol_spec, context, &vol_id, &frontend_node_id)
+                            .await
+                    {
+                        tracing::error!(
+                            error=%e,
+                            volume.uuid = vol_id.as_str(),
+                            nexus.uuid = nexus_id.as_str(),
+                            node.id = node_spec.id().as_str(),
+                            "Failed to republish volume"
+                        );
+                        move_failures = true;
+                        continue;
+                    }
+                    tracing::info!(
+                        volume.uuid = vol_id.as_str(),
+                        nexus.uuid = nexus_id.as_str(),
+                        node.id = node_spec.id().as_str(),
+                        "Moved volume"
+                    );
+                }
+            }
+            Err(_) => {
+                // we can't get to the volume so we don't know if it belongs to this node
+                move_failures = true;
+            }
+        };
+    }
+    // Change the node state to "drained"
+    if !move_failures {
+        if let Err(e) = context
+            .specs()
+            .set_node_drained(context.registry(), node_spec.id())
+            .await
+        {
+            tracing::error!(
+                error=%e,
+                node.id = node_id.as_str(),
+                "Failed to set node to state drained"
+            );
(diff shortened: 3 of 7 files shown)


common/src/types/v0/store/node.rs
```diff
@@ -112,6 +112,13 @@ impl CordonDrainState {
             CordonDrainState::Drained(_) => false,
         }
     }
+    pub fn is_drained(&self) -> bool {
```
tiagolobocastro, line 116: missing doc comment.

common/src/types/v0/store/node.rs
```diff
@@ -235,6 +242,17 @@ impl NodeSpec {
             }
         }
     }
+
+    /// Move state from Draining to Drained, no change to the labels
```
tiagolobocastro: missing `.`

control-plane/agents/src/bin/core/controller/reconciler/node/mod.rs
```diff
@@ -0,0 +1,41 @@
...
+    /// Return new `Self` with the default period.
+    pub(crate) fn new() -> Self {
+        Self::from(1)
+    }
+}
+
+#[async_trait::async_trait]
+impl TaskPoller for NodeReconciler {
+    async fn poll(&mut self, context: &PollContext) -> PollResult {
+        let mut results = vec![];
```
tiagolobocastro:
```suggestion
        let mut results = Vec::with_capacity(self.poll_targets.len());
```

control-plane/agents/src/bin/core/node/specs.rs
```diff
@@ -169,4 +170,19 @@ impl ResourceSpecsLocked {
         registry.store_obj(&drained_node_spec).await?;
         Ok(drained_node_spec)
     }
+
+    pub(crate) async fn set_node_drained(
+        &self,
+        registry: &Registry,
+        node_id: &NodeId,
+    ) -> Result<NodeSpec, SvcError> {
+        let node = self.get_locked_node(node_id)?;
```
tiagolobocastro, line 180: IIRC we previously discussed adding a ResourceMutex and OperationGuard for nodes, guess that can come after?

control-plane/agents/src/bin/core/volume/registry.rs
```diff
@@ -103,7 +103,7 @@ impl Registry {
     }
 
     /// Get all volumes
-    pub(super) async fn get_volumes(&self) -> Vec<Volume> {
+    pub(crate) async fn get_volumes(&self) -> Vec<Volume> {
```
tiagolobocastro: Since you're here, could you add missing `.` above? thanks

control-plane/agents/src/bin/core/node/specs.rs
```diff
@@ -169,4 +170,19 @@ impl ResourceSpecsLocked {
         registry.store_obj(&drained_node_spec).await?;
         Ok(drained_node_spec)
     }
+
+    pub(crate) async fn set_node_drained(
```
tiagolobocastro, line 175: missing doc comment

control-plane/agents/src/bin/core/controller/reconciler/node/nexus.rs
```diff
@@ -0,0 +1,138 @@
...
+/// Drain the specified node if in draining state
+async fn check_and_drain_node(context: &PollContext, node_spec: &NodeSpec) -> PollResult {
+    if node_spec.is_draining() {
+        let node_id = node_spec.id().clone();
+        tracing::info!("Draining node {}", node_id);
+        // iterate through all the volumes, find those with a nexus hosted on the node
+        let volumes = context.registry().get_volumes().await;
+
+        for volume in volumes.iter() {
+            if volume.state().target.is_some() && volume.state().target.unwrap().node == node_id {
```
tiagolobocastro: Shouldn't we check for the specs rather than the states?

tiagolobocastro: I wonder if drain could be an operation implemented by the OpeerationGuard<VolumeSpec> ?

control-plane/agents/src/bin/core/controller/reconciler/node/nexus.rs
```diff
@@ -0,0 +1,138 @@
...
+
+    async fn poll_event(&mut self, context: &PollContext) -> bool {
+        matches!(context.event(), PollEvent::TimedRun)
+    }
+}
+
+/// Republish the specified volume on any node other than its current node
+async fn republish_volume(context: &PollContext, vol_uuid: &VolumeId) -> Result<(), SvcError> {
+    let request = RepublishVolume::new(vol_uuid.clone(), None, VolumeShareProtocol::Nvmf, false);
+    tracing::info!("Attempting to republish volume, id {}", vol_uuid);
```
tiagolobocastro:
```suggestion
    tracing::info!(volume.uuid=vol_uuid.as_str(), "Attempting to republish volume");
```

control-plane/agents/src/bin/core/controller/reconciler/node/nexus.rs
```diff
@@ -0,0 +1,138 @@
...
+        }
+    };
+    Ok(nodespec.is_draining())
+}
+
+/// Drain the specified node if in draining state
+async fn check_and_drain_node(context: &PollContext, node_spec: &NodeSpec) -> PollResult {
+    if node_spec.is_draining() {
+        let node_id = node_spec.id().clone();
+        tracing::info!("Draining node {}", node_id);
```
tiagolobocastro:
```suggestion
        tracing::trace!(node.id=node_spec.id().as_str(), "Draining node");
```

tiagolobocastro (commented):
Please change the trace logs to use tracing fields. I've made a few example suggestions.

control-plane/agents/src/bin/core/node/specs.rs
chriswldenyer: fixed

control-plane/agents/src/bin/core/volume/registry.rs
chriswldenyer: fixed (plus some others)

control-plane/agents/src/bin/core/controller/reconciler/node/mod.rs
chriswldenyer: fixed

common/src/types/v0/store/node.rs
chriswldenyer: fixed

common/src/types/v0/store/node.rs
chriswldenyer: fixed

control-plane/agents/src/bin/core/controller/reconciler/node/nexus.rs
chriswldenyer: fixed

control-plane/agents/src/bin/core/controller/reconciler/node/nexus.rs
chriswldenyer: fixed (plus the other trace calls in this file)

Abhinandan-Purkait (approved):

control-plane/agents/src/bin/core/controller/reconciler/node/nexus.rs
chriswldenyer: I've changed it to using the volume specs.
Looking at OperationGuard now.

control-plane/agents/src/bin/core/controller/reconciler/node/nexus.rs
```diff
@@ -0,0 +1,146 @@
...
+            counter: PollTimer::from(1), // sets the reconciler polling rate
+        }
+    }
+}
+
+#[async_trait::async_trait]
+impl TaskPoller for NodeNexusReconciler {
+    async fn poll(&mut self, context: &PollContext) -> PollResult {
+        let mut results = vec![];
+        let nodes = context.specs().get_nodes();
```
tiagolobocastro:
```suggestion
        let nodes = context.specs().get_nodes();
        let mut results = Vec::with_capacity(nodes.len());
```

control-plane/agents/src/bin/core/controller/reconciler/node/nexus.rs
```diff
@@ -0,0 +1,146 @@
...
+    async fn poll_timer(&mut self, _context: &PollContext) -> bool {
+        self.counter.poll()
+    }
+
+    async fn poll_event(&mut self, context: &PollContext) -> bool {
+        matches!(context.event(), PollEvent::TimedRun)
+    }
+}
+
+/// Republish the specified volume on any node other than its current node
```
tiagolobocastro:
```suggestion
/// Republish the specified volume on any node other than its current node.
```

control-plane/agents/src/bin/core/controller/reconciler/node/nexus.rs
```diff
@@ -0,0 +1,146 @@
...
+}
+
+/// Returns true if the specified node is in the draining state
+fn is_node_draining(context: &PollContext, nodeid: &NodeId) -> Result<bool, SvcError> {
+    let nodespec = match context.specs().get_node(nodeid) {
+        Ok(ns) => ns,
+        Err(e) => {
+            return Err(e);
+        }
+    };
```
tiagolobocastro:
```suggestion
    let nodespec = match context.specs().get_node(nodeid)?;
```

control-plane/agents/src/bin/core/controller/reconciler/node/nexus.rs
```diff
@@ -0,0 +1,146 @@
...
+    Ok(nodespec.is_draining())
+}
+
+/// Drain the specified node if in draining state
+async fn check_and_drain_node(context: &PollContext, node_spec: &NodeSpec) -> PollResult {
+    if node_spec.is_draining() {
```
tiagolobocastro: nit: reduce nesting suggestion:
```rust
if !node_spec.is_draining() { return PollResult::Ok(PollerState::Idle); }
```

control-plane/agents/src/bin/core/controller/reconciler/node/nexus.rs
```diff
@@ -0,0 +1,146 @@
...
+        let node_id = node_spec.id().clone();
```
tiagolobocastro: does it need to be cloned ?

control-plane/agents/src/bin/core/controller/reconciler/node/nexus.rs
```diff
@@ -0,0 +1,146 @@
...
+        tracing::trace!(node.id = node_spec.id().as_str(), "Draining node");
+        // iterate through all the volumes, find those with a nexus hosted on the node
+        let volumes = context.registry().get_volumes().await;
+
+        for volume in volumes.iter() {
+            if volume.spec().target.is_some()
+                && volume.spec().target.unwrap().node().clone() == node_id
```
tiagolobocastro:
```suggestion
            let target = match volume.spec().target() { Some(t) => t, None => continue };
            if target.node() == node_id
```

control-plane/agents/src/bin/core/controller/reconciler/node/nexus.rs
```diff
@@ -0,0 +1,146 @@
...
+/// Drain the specified node if in draining state
+async fn check_and_drain_node(context: &PollContext, node_spec: &NodeSpec) -> PollResult {
+    if node_spec.is_draining() {
+        let node_id = node_spec.id().clone();
+        tracing::trace!(node.id = node_spec.id().as_str(), "Draining node");
+        // iterate through all the volumes, find those with a nexus hosted on the node
+        let volumes = context.registry().get_volumes().await;
+
+        for volume in volumes.iter() {
+            if volume.spec().target.is_some()
```
tiagolobocastro: I think probably here you need to take the volume guard first to ensure nothing else is trying to modify the volume

control-plane/agents/src/bin/core/controller/reconciler/node/nexus.rs
```diff
@@ -0,0 +1,146 @@
...
+            {
+                let nexus_id = volume.spec().target.unwrap().nexus().clone();
+                let vol_id = volume.spec().uuid;
+                tracing::info!(
+                    volume.uuid = vol_id.as_str(),
+                    nexus.uuid = nexus_id.as_str(),
+                    node.id = node_spec.id().as_str(),
+                    "Moving volume"
+                );
+                match republish_volume(context, &vol_id).await {
```
tiagolobocastro: if let Err(error) = republish_volume(context, &vol_id).await { .. }

control-plane/agents/src/bin/core/controller/reconciler/node/nexus.rs
```diff
@@ -0,0 +1,146 @@
...
+                        tracing::error!(
+                            "Failed to republish volume {} on node {}, error {}",
+                            vol_id,
+                            node_id,
+                            e,
+                        );
+                        return PollResult::Err(e);
+                    }
+                };
+                // wait for the nexus to be removed, or for the drain to be cancelled
```
tiagolobocastro: the reconcilers don't quite work like this atm, they don't wait inline, rather they try again next time they run...
But I don't think we need it at all here because the republish already succeeds when it completes, so it's already moved.

control-plane/agents/src/bin/core/controller/reconciler/node/nexus.rs
```diff
@@ -0,0 +1,146 @@
...
+                match republish_volume(context, &vol_id).await {
+                    Ok(()) => {}
+                    Err(e) => {
+                        tracing::error!(
+                            "Failed to republish volume {} on node {}, error {}",
+                            vol_id,
+                            node_id,
+                            e,
+                        );
+                        return PollResult::Err(e);
```
tiagolobocastro: probably want to record the error and `continue` instead otherwise if 1 volume is "bad" we won't move any other volume?

control-plane/agents/src/bin/core/node/specs.rs
tiagolobocastro: As discussed let's drop this for now.

control-plane/agents/src/bin/core/controller/reconciler/node/nexus.rs
chriswldenyer: From the logs, when testing this, republish_volume() seems to return before the the nexus is moved. The loop (line 105) runs 3 or 4 times before the nexus is found to be gone, so it seems to be asynchronous AFAICT.
If I do remove the loop, I suppose there is a risk that republish_volume will be called multiple times for the same volume (i.e. when the reconciler gets triggered again) if it has not yet been moved from the node by that time. Is that safe?

tiagolobocastro: Hmm that doesn't seem right, the nexus should be ready and shared when republish completes.

control-plane/agents/src/bin/core/controller/reconciler/node/nexus.rs
chriswldenyer: fixed

control-plane/agents/src/bin/core/controller/reconciler/node/nexus.rs
chriswldenyer: fixed

control-plane/agents/src/bin/core/controller/reconciler/node/nexus.rs
chriswldenyer: fixed

control-plane/agents/src/bin/core/controller/reconciler/node/nexus.rs
chriswldenyer: fixed

control-plane/agents/src/bin/core/controller/reconciler/node/nexus.rs
chriswldenyer: It seems not, fixed.

control-plane/agents/src/bin/core/controller/reconciler/node/nexus.rs
chriswldenyer: fixed

control-plane/agents/src/bin/core/controller/reconciler/node/nexus.rs
chriswldenyer: fixed

control-plane/agents/src/bin/core/controller/reconciler/node/nexus.rs
chriswldenyer: It now checks the volume specs, leaving the operation guard for drain for another PR.

control-plane/agents/src/bin/core/controller/reconciler/node/nexus.rs
chriswldenyer: That part of the code was rewritten in the form: if let Some(target) = ....

control-plane/agents/src/bin/core/controller/reconciler/node/nexus.rs
chriswldenyer: Removed wait loop as superfluous.

control-plane/agents/src/bin/core/controller/reconciler/node/nexus.rs
chriswldenyer: Now uses a volume operation guard.

control-plane/agents/src/bin/core/controller/reconciler/node/nexus.rs
```diff
@@ -0,0 +1,173 @@
...
+/// Drain the specified node if in draining state
+async fn check_and_drain_node(context: &PollContext, node_spec: &NodeSpec) -> PollResult {
+    if !node_spec.is_draining() {
+        return PollResult::Ok(PollerState::Idle);
+    }
+
+    let node_id = node_spec.id();
+    tracing::trace!(node.id = node_spec.id().as_str(), "Draining node");
+    // iterate through all the volumes, find those with a nexus hosted on the node
+    let volumes = context.registry().get_volumes().await;
```
tiagolobocastro: Looks like we're not using state, so we could instead use `context.specs().get_locked_volumes`?

control-plane/agents/src/bin/core/controller/reconciler/node/nexus.rs
```diff
@@ -0,0 +1,173 @@
...
+
+    let mut volumes_moved = 0;
+    let mut volumes_on_node = 0;
+
+    for volume in volumes.iter() {
```
tiagolobocastro: can probably skip the iter or use into_iter to consume the volumes?

control-plane/agents/src/bin/core/controller/reconciler/node/nexus.rs
```diff
@@ -0,0 +1,173 @@
...
+        let vol_spec = volume.spec();
```
tiagolobocastro: shouldn't the guard be taken at the start of the loop? Otherwise we still have a race?

control-plane/agents/src/bin/core/controller/reconciler/node/nexus.rs
```diff
@@ -0,0 +1,173 @@
...
+                            volumes_moved += 1;
+                            tracing::info!(
+                                volume.uuid = vol_id.as_str(),
+                                nexus.uuid = nexus_id.as_str(),
+                                node.id = node_spec.id().as_str(),
+                                "Moved volume"
+                            );
+                        }
+                        Err(e) => {
+                            tracing::error!(error=%e, "Failed to get operation guard");
```
tiagolobocastro: IIRC that's already traced automatically so we can simply continue

control-plane/agents/src/bin/core/controller/reconciler/node/nexus.rs
```diff
@@ -0,0 +1,173 @@
...
+    if volumes_moved == volumes_on_node {
+        if let Err(e) = context
+            .specs()
+            .set_node_drained(context.registry(), node_spec.id())
+            .await
+        {
+            tracing::error!(
+                error=%e,
+                node.id = node_id.as_str(),
+                "Failed to set node to state drained");
```
tiagolobocastro: formatting seems off here?

tiagolobocastro (approved):

control-plane/agents/src/bin/core/controller/reconciler/node/nexus.rs
chriswldenyer: Fixed. Not sure why the commit-checker didn't catch that.

control-plane/agents/src/bin/core/controller/reconciler/node/nexus.rs
chriswldenyer: Removed. Also the continue is redundant, so removed that as well.

control-plane/agents/src/bin/core/controller/reconciler/node/nexus.rs
chriswldenyer: Now takes the guard at the top of the loop.

control-plane/agents/src/bin/core/controller/reconciler/node/nexus.rs
chriswldenyer: Fixed.

control-plane/agents/src/bin/core/controller/reconciler/node/nexus.rs
chriswldenyer: Done.

control-plane/agents/src/bin/core/controller/reconciler/node/nexus.rs
tiagolobocastro: Yeah sadly rustfmt doesn't always work properly when macros are involved :(

control-plane/agents/src/bin/core/controller/reconciler/node/nexus.rs
```diff
@@ -0,0 +1,161 @@
...
+                        tracing::error!("Failed to get config");
+                        move_failures = true;
+                        continue;
+                    }
+                    let frontend_nodes = config.unwrap().frontend().node_names();
+                    if frontend_nodes.is_empty() {
+                        tracing::error!("Cannot republish, volume has no front end node");
+                        move_failures = true;
+                        continue;
+                    }
```
tiagolobocastro: nit: We can avoid the unwraps here, example:
```rust
let config = match guarded_vol_spec.as_ref().config() {
  None => {
    tracing::error!("Failed to get config");
    move_failures = true;
    continue;
  }
  Some(config) => config,
};

let frontend_node = config.frontend().node_names().first().unwrap_or_default();
```

control-plane/agents/src/bin/core/controller/reconciler/node/nexus.rs
```diff
@@ -0,0 +1,161 @@
...
+                    let vol_id = guarded_vol_spec.as_ref().uuid.clone();
+
+                    let config = guarded_vol_spec.as_ref().config().clone();
+                    if config.is_none() {
+                        tracing::error!("Failed to get config");
+                        move_failures = true;
+                        continue;
+                    }
+                    let frontend_nodes = config.unwrap().frontend().node_names();
+                    if frontend_nodes.is_empty() {
```
tiagolobocastro: Maybe just pass "" in this case?
`let frontend_node = config.frontend().node_names().first().unwrap_or_default();`
Otherwise the node will never be drained?

chriswldenyer: Because "" is an acceptable value for the frontend node?
Got it.

tiagolobocastro: IIRC "" was being used on some tests (I then changed it to a proper name during code refactoring), so I guess it's acceptable? Not sure then, might be worth checking

chriswldenyer: As discussed, the republish uses "" via String::default() if the frontend node isn't defined.

Abhinandan-Purkait (approved):

chriswldenyer:
bors try

chriswldenyer:
bors merge

PR #405, opened 2022-12-12
fix(republish): filter out nodes with existing volume targets

tiagolobocastro:
Ensures we don't attempt to create nexus on a node which already has a nexus for a given volume.

### DIFF
--- a/control-plane/agents/src/bin/core/controller/scheduling/mod.rs
+++ b/control-plane/agents/src/bin/core/controller/scheduling/mod.rs
@@ -92,6 +92,13 @@ impl NodeFilters {
             true
         }
     }
+    /// Should only attempt to use node where there are no targets for the current volume.
+    pub(crate) fn no_targets(request: &GetSuitableNodesContext, item: &NodeItem) -> bool {
+        let volume_targets = request.registry().specs().get_volume_nexuses(&request.uuid);
+        !volume_targets
+            .into_iter()
+            .any(|n| &n.lock().node == item.node_wrapper().id())
+    }
 }
 
 /// Filter pools used for replica creation.
--- a/control-plane/grpc/proto/v1/volume/volume.proto
+++ b/control-plane/grpc/proto/v1/volume/volume.proto
@@ -150,6 +150,8 @@ message GetVolumesRequest {
   }
   // pagination to allow for multiple requests to get all volumes
   common.Pagination pagination = 2;
+  // ignore 404 not found errors
+  bool ignore_notfound = 3;
 }
 
 // volume creation request
--- a/common/src/transport_api/mod.rs
+++ b/common/src/transport_api/mod.rs
@@ -108,7 +108,7 @@ pub trait Message {
 }
 
 /// All the different variants of Resources
-#[derive(Serialize, Deserialize, Debug, Clone, AsRefStr, ToString)]
+#[derive(Serialize, Deserialize, Debug, Clone, AsRefStr, ToString, Eq, PartialEq)]
 pub enum ResourceKind {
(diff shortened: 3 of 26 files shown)


control-plane/grpc/proto/v1/volume/volume.proto
```diff
@@ -150,6 +150,8 @@ message GetVolumesRequest {
   }
   // pagination to allow for multiple requests to get all volumes
   common.Pagination pagination = 2;
+  // ignore 404 not found errors
```
Abhinandan-Purkait, line 153: Question: Why is this needed?

control-plane/agents/src/bin/core/controller/scheduling/mod.rs
```diff
@@ -92,6 +92,13 @@ impl NodeFilters {
             true
         }
     }
+    /// Should only attempt to use node where there are no targets for the current volume.
+    pub(crate) fn no_targets(request: &GetSuitableNodesContext, item: &NodeItem) -> bool {
+        let volume_targets = request.registry().specs().get_volume_nexuses(&request.uuid);
+        !volume_targets
+            .into_iter()
+            .any(|n| &n.lock().node == item.node_wrapper().id())
```
avishnu, line 100: 1. Why is locking needed here? If it's needed, should it be on the 'volume_targets' instead of per item?
2. Is there a possibility that there is only one healthy node at the moment and a previous nexus for the same volume on the same node is in a shutdown/deletion state? Should we not select the node in that case? OR will it happen in the future reconciliation?

tiagolobocastro: 1. Because that's how things are atm (irrespective of this PR), each  target lives on its own and has its own lock
2. Yes it's possible, in such a case we need something to clean up shutdown targets, which will be done as part of @abhilashshetty04 's PR, which will detect when we run out of nodes and clean up unused shutdown targets.

control-plane/grpc/proto/v1/volume/volume.proto
tiagolobocastro: While debugging a cluster I noticed we end up with a ton of error messages which are actually not an error because on initial creation we do Get Volume xyz to find out if it exists or not, and first time it won't exist.
This prevents the error..
Usually we wouldn't need this flag but it seems we use the same grpc call to get 1 or many, so that's why it's needed as we cannot differentiate..

control-plane/agents/src/bin/core/controller/scheduling/mod.rs
avishnu: > 1. Because that's how things are atm (irrespective of this PR), each  target lives on its own and has its own lock

Why do we need to n.lock() here? Just trying to understanding.

tiagolobocastro: Because the object in case is protected by a Mutex, eg: `Mutex<Obj>`, so to access its inner value (Obj) you have to take the lock which gives you a guard over it,eg: `Guard<Obj>`. When the guard is dropped, the mutex is unlocked.

avishnu: I see, thanks for the explanation !!

control-plane/grpc/proto/v1/volume/volume.proto
Abhinandan-Purkait: RIght. Thank you.

Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

PR #406, opened 2022-12-12
feat(csi-node): parse and use keep-alive-tmo when connecting to a target

tiagolobocastro:
Adds nvme-keep-alive-tmo via the cmdline which is used as a default for all targets. Alternatively, it can also be overridden via the storage class with nvmeKeepAliveTmo.

### DIFF
--- a/control-plane/agents/src/bin/core/volume/service.rs
+++ b/control-plane/agents/src/bin/core/volume/service.rs
@@ -197,7 +197,6 @@ impl Service {
             },
             Filter::Volume(volume_id) => {
                 tracing::Span::current().record("volume.uuid", &volume_id.as_str());
-                tracing::error!("I: {}", ignore_notfound);
                 match self.registry.get_volume(&volume_id).await {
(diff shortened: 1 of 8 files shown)


Abhinandan-Purkait:
Since this can be set via the storage class, should we add this as a part of the publish context as well so that we can use it in HA?

Abhinandan-Purkait (approved):

avishnu:
nitpick: Should the string have Timeout instead of abbr Tmo?

tiagolobocastro:
> nitpick: Should the string have Timeout instead of abbr Tmo?

It's inkeeping with the actual nvme parameter name, and it's the same as the controller loss timeout.

abhilashshetty04 (approved):

tiagolobocastro:
> Since this can be set via the storage class, should we add this as a part of the publish context as well so that we can use it in HA?

Ah good point, done!

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

PR #411, opened 2022-12-15
fix(csi-driver): ensure only 1 in-flight request per volume 

tiagolobocastro:
fix(csi-driver): ensure only 1 in-flight request per volume

As per the CSI spec there should be no more than one call “in-flight” per volume at a given time.
This is now ensured on both the node and controller services.

---

refactor(csi-node): concurrency limiter for volume operations

Move serializer into the csi-driver library module so that it may be used by the controller.
Decode the volume uuid from the nqn so that we may use it to concurrency limit rpc calls.
Rpc calls for a given volume should be aborted when there is one already in-flight,
as per the CSI spec.

---

fix(concurrency): enable openapi client concurrency

Upgrade openapi-generator which includes a new concurrency_limit option.
todo: limit concurrency per volume in the csi stack


chriswldenyer (approved):
Looks reasonable

blaisedias (approved):
Looks plausible

tiagolobocastro:
bors merge

PR #413, opened 2022-12-19
fix: handle device or resource busy error for unpublish volume

nearora-msft:
Issue: We are running some scale tests with mayastor and mayastor-control-plane on AKS. When the number of pods per node is large(~80), there are some issues while scaling in and the pods remain in Terminating state.

This happens because, the find_mount check [here ](https://github.com/openebs/mayastor-control-plane/blob/develop/control-plane/csi-driver/src/bin/node/filesystem_vol.rs#L322) returns false, even though the target mount is there in the /proc/mounts file.
As a result, we run into : "Device or resource busy (os error 16)" errors on the csi-node.

Fix: Whenever, this issue happens, instead of returning a Success, return a failure so that the kubelet can retry unmounting. Tested this with a scale run and all the pods got terminated successfully.

### DIFF
--- a/control-plane/csi-driver/src/bin/node/filesystem_vol.rs
+++ b/control-plane/csi-driver/src/bin/node/filesystem_vol.rs
@@ -322,11 +322,16 @@ pub(crate) fn unpublish_fs_volume(msg: &NodeUnpublishVolumeRequest) -> Result<()
     if mount::find_mount(None, Some(target_path)).is_none() {
         // No mount found for target_path.
         // The idempotency requirement means this is not an error.
-        // Just clean up as best we can and claim success.
-
+        // Just clean up as best we can
         if let Err(error) = fs::remove_dir(PathBuf::from(target_path)) {
             if error.kind() != ErrorKind::NotFound {
-                error!("Failed to remove directory {}: {}", target_path, error);
+                // Return error so that kubelet can retry
+                return Err(failure!(
+                    Code::Internal,
+                    "Failed to remove directory {}: {}",
+                    target_path,
+                    error
+                ));
             }
         }
 
@@ -354,7 +359,12 @@ pub(crate) fn unpublish_fs_volume(msg: &NodeUnpublishVolumeRequest) -> Result<()
 
     if let Err(error) = fs::remove_dir(PathBuf::from(target_path)) {
         if error.kind() != ErrorKind::NotFound {
-            error!("Failed to remove directory {}: {}", target_path, error);
+            return Err(failure!(
+                Code::Internal,
+                "Failed to remove directory {}: {}",
+                target_path,
+                error
+            ));
         }
     }
 


control-plane/csi-driver/src/bin/node/filesystem_vol.rs
```diff
@@ -324,6 +326,14 @@ pub(crate) fn unpublish_fs_volume(msg: &NodeUnpublishVolumeRequest) -> Result<()
         // The idempotency requirement means this is not an error.
         // Just clean up as best we can and claim success.
 
+        // If the Device or Resource busy error is encountered, return error so that 
+        // unmounting and removing is retried 
+        if error.to_string().contains(RESOURCE_BUSY_ERROR_STRING){
```
nearora-msft: Checking error by parsing error string here, because ErrorKind::ResourceBusy is an unstable rust feature. 
https://doc.rust-lang.org/stable/std/io/enum.ErrorKind.html#variant.ResourceBusy

tiagolobocastro: I may be looking at this wrongly in the gh website, but is this check if the right place? I can't see the variable error in the current scope?
That aside, I wonder if we should switch the logic, and NotFound becomes the only "accepted" error, and everything else should lead into a failure, what do you think?

nearora-msft: Yep, that's what I was thinking too, but I wasn't sure why we return success in other cases currently. Sure, I would make the change to return error in all other cases.

tiagolobocastro:
I guess we also must make an error on L360 if != not found?
But I'm more curious why the `find_mount` is returning `None` if the mount is still there, because it means we'd never amount it? Or is this part of a "retry"?
Btw could you please sign your commits?

nearora-msft:
> I guess we also must make an error on L360 if != not found? But I'm more curious why the `find_mount` is returning `None` if the mount is still there, because it means we'd never amount it? Or is this part of a "retry"? Btw could you please sign your commits?

Right! I checked that the mount was present in the /proc/mounts file and yet it find_mount was returning None. I checked that find_mount uses MountIter(https://docs.rs/proc-mounts/0.1.1/proc_mounts/struct.MountIter.html)  which essentially reads the /proc/mounts iteratively. 

We run into this issue when we are running unmounts on a node at once while scaling in our workload pods. We'll probably have to check if MountIter is behaving incorrectly under stress. I wonder if you have run into anything like this before.

tiagolobocastro:
I haven't seen it myself, tagging @blaisedias who might know.

@nearora-msft would it be possible to dump the the entries we get from `MountIter` to stdout to see if we're somehow not parsing it properly or something like that?

nearora-msft:
> I haven't seen it myself, tagging @blaisedias who might know.
> 
> @nearora-msft would it be possible to dump the the entries we get from `MountIter` to stdout to see if we're somehow not parsing it properly or something like that?

Sure, we could do that. But I guess these error handling changes will still be valid?

tiagolobocastro:
> Sure, we could do that. But I guess these error handling changes will still be valid?

Yes I think so. May I suggest handling the error the same way on L360 as well and leave the rest to another PR?

nearora-msft:
> > Sure, we could do that. But I guess these error handling changes will still be valid?
> 
> Yes I think so. May I suggest handling the error the same way on L360 as well and leave the rest to another PR?

Done

tiagolobocastro:
Can you fix the commitment failure please? thanks

tiagolobocastro (approved):

abhilashshetty04 (approved):

tiagolobocastro:
Great thanks @nearora-msft !
Let's merge this now but if you can please check for the "origin" of the find mount not finding the mount, otherwise I think the system might be left in a bad state.
bors merge

nearora-msft:
> Great thanks @nearora-msft ! Let's merge this now but if you can please check for the "origin" of the find mount not finding the mount, otherwise I think the system might be left in a bad state. bors merge

Thank you for merging @tiagolobocastro. Shall I create an issue for it, so that I can pick it up whenever I have cycles?

tiagolobocastro:
Sounds good, please raise it on the [main mayastor repo](https://github.com/openebs/mayastor/issues).

nearora-msft:
> Sounds good, please raise it on the [main mayastor repo](https://github.com/openebs/mayastor/issues).

Done : https://github.com/openebs/mayastor/issues/1274

PR #414, opened 2022-12-20
fix: ignore metrics for block volume

Abhinandan-Purkait:
Signed-off-by: Abhinandan Purkait <purkaitabhinandan@gmail.com>

### DIFF
--- a/control-plane/csi-driver/src/bin/node/node.rs
+++ b/control-plane/csi-driver/src/bin/node/node.rs
@@ -383,31 +383,47 @@ impl node_server::Node for Node {
             ));
         }
         let _guard = csi_driver::limiter::VolumeOpGuard::new_str(&msg.volume_id)?;
-        match sys::statfs::statfs(&*msg.volume_path) {
-            Ok(info) => Ok(Response::new(NodeGetVolumeStatsResponse {
-                usage: vec![
-                    csi::VolumeUsage {
-                        total: info.blocks() as i64 * info.block_size(),
-                        unit: csi::volume_usage::Unit::Bytes as i32,
-                        available: info.blocks_available() as i64 * info.block_size(),
-                        used: (info.blocks() - info.blocks_free()) as i64 * info.block_size(),
+
+        let volume_path = Path::new(&msg.volume_path);
+        if volume_path.exists() {
+            // Check if its a filesystem.
+            if volume_path.is_dir() {
+                trace!("Getting statfs metrics for : {:?}", volume_path);
+                match sys::statfs::statfs(&*msg.volume_path) {
+                    Ok(info) => Ok(Response::new(NodeGetVolumeStatsResponse {
+                        usage: vec![
+                            csi::VolumeUsage {
+                                total: info.blocks() as i64 * info.block_size(),
+                                unit: csi::volume_usage::Unit::Bytes as i32,
+                                available: info.blocks_available() as i64 * info.block_size(),
+                                used: (info.blocks() - info.blocks_free()) as i64
+                                    * info.block_size(),
+                            },
+                            csi::VolumeUsage {
+                                total: info.files() as i64,
+                                unit: csi::volume_usage::Unit::Inodes as i32,
+                                available: info.files_free() as i64,
+                                used: (info.files() - info.files_free()) as i64,
+                            },
+                        ],
+                        volume_condition: None,
+                    })),
+                    Err(err) => match err {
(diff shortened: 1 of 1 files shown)


control-plane/csi-driver/src/bin/node/node.rs
```diff
@@ -383,31 +383,49 @@ impl node_server::Node for Node {
...
-                    csi::VolumeUsage {
-                        total: info.files() as i64,
-                        unit: csi::volume_usage::Unit::Inodes as i32,
-                        available: info.files_free() as i64,
-                        used: (info.files() - info.files_free()) as i64,
+
+        let volume_path = Path::new(&msg.volume_path);
+        if volume_path.exists() {
+            // Check if its a filesystem.
+            if volume_path.is_dir() {
```
Abhinandan-Purkait, line 390: For block volume the volume path would be a file. Thus the above condition will only be true for the filesystem. This  is in accordance to the logic we use in other methods. We have tested this.
ref:- https://github.com/openebs/mayastor-control-plane/blob/d6f8664b47414a47b9c3c17e6597467b64e12e4d/control-plane/csi-driver/src/bin/node/node.rs#L338

control-plane/csi-driver/src/bin/node/node.rs
```diff
@@ -383,31 +383,49 @@ impl node_server::Node for Node {
...
+                            },
+                        ],
+                        volume_condition: None,
+                    })),
+                    Err(err) => match err {
+                        Errno::ENOENT => Err(Status::new(Code::NotFound, err.to_string())),
+                        Errno::EIO => Err(Status::new(Code::Internal, err.to_string())),
+                        Errno::ENOSYS => Err(Status::new(Code::Unavailable, err.to_string())),
+                        Errno::ENOTDIR => {
+                            Err(Status::new(Code::FailedPrecondition, err.to_string()))
```
tiagolobocastro: Should be internal now because we already entered if dir?

control-plane/csi-driver/src/bin/node/node.rs
```diff
@@ -383,31 +383,49 @@ impl node_server::Node for Node {
...
-            Err(err) => match err {
-                Errno::ENOENT => Err(Status::new(Code::NotFound, err.to_string())),
-                Errno::EIO => Err(Status::new(Code::Internal, err.to_string())),
-                Errno::ENOSYS => Err(Status::new(Code::Unavailable, err.to_string())),
-                Errno::ENOTDIR => Err(Status::new(Code::FailedPrecondition, err.to_string())),
-                _ => Err(Status::new(Code::InvalidArgument, err.to_string())),
-            },
+                }
+            } else {
+                Err(Status::new(Code::InvalidArgument, "not filesystem"))
```
tiagolobocastro: this is not an error, we should just report stats with empty vec.

control-plane/csi-driver/src/bin/node/node.rs
```diff
@@ -383,31 +383,49 @@ impl node_server::Node for Node {
...
+            }
+        } else {
+            Err(Status::new(
+                Code::InvalidArgument,
```
tiagolobocastro: As per spec guess we should return NotFound?

Abhinandan-Purkait: Right, it should be Notfound.

control-plane/csi-driver/src/bin/node/node.rs
Abhinandan-Purkait: Yeah, need to check if that would fail the test or not.

control-plane/csi-driver/src/bin/node/node.rs
Abhinandan-Purkait: Right

control-plane/csi-driver/src/bin/node/node.rs
Abhinandan-Purkait: Done. The tests pass, with empty vector.

tiagolobocastro (approved):

control-plane/csi-driver/src/bin/node/node.rs
avishnu: Good description, please add this as a code comment itself, so that it will be self-explanatory in the future.

avishnu (approved):
lgtm

Abhinandan-Purkait:
bors merge

PR #417, opened 2023-01-08
fix(agent-node/subsystem): a subsystem can have more than 1 controller

tiagolobocastro:
chore(ci): set target test branch automatically
    
    Add script that determines the appropriate branch of dataplane to test against and writes
    this to a rust file.
    This is to avoid having to modify this file when we create a release branch.
    
---

    fix(agent-node/subsystem): handle multiple controller paths
    
    The existing code seems to treat a subsystem as having a single path.
    The nvmeadm library subsystem type seems to be quite confusing using the term controller
    and subsystem interchangeably.
    This poses a problem when swithching over as temporarily the subsystem will have two controllers,
    the new one (good) and the old one (bad).
    
    Here we can see the old controller nvme0 and the new one nvme1 by probing sysfs at the right time:
    tiago@ksnode-2:~$ ls /sys/class/nvme-subsystem/nvme-subsys0/
    model  ng0n1  nvme0n1 (block device is still 0)  nvme0 (old) nvme1 (new)  subsysnqn ..
    
    What was then happening is that when fixing a path to live we were setting the suspected_path nvme0
    as Good, rather than leave nvme0 as Bad and set nvme1 as Good.
    This posed a problem on a subsequent failure as we would even try to disconnect nvme0 instead of
    nvme1 - this could cause us to boot the wrong controller and causing issues on another disk!!!
    
    This fix is to keep a list of paths per NQN on the detector. Additionally when we fetch a controller
    per NQN we have to fetch a list instead (again, multiple paths for the same NQN...)
    
    Ideally we should fix things properly (nvmeadm) but nearing the release this is probably less risky.
    
---

    fix(agent-node): verify nqn when disconnecting
    
    When attempting to disconnect a path, make sure that the nqn is the same.
    Why? We found a bug where we would keep a stale path, and it's possible that a new device
    may be reuse that path (new nqn ofc) and so we should ensure the nqn is the expected
    when disconnecting to make sure we don't remove the wrong device!
    
---

    fix(nix-build): speedup incremental builds
    
    When building a single binary, don't bother building the rest of the binaries from a project.
    
---

    fix(release-ci): override registry properly
    
    Previously the existing registry prefix was being kept.
    To keep existing behaviour (in case it's being used :() we only strip out the existing
    registry if the new one has a slash.
    Example:
    ./scripts/release.sh --registry mayadata would push image mayadata/openebs/$image
    ./scripts/release.sh --registry docker.io/mayadata would push image docker.io/mayadata/$image
    

avishnu (approved):
LGTM
On a relative front, does any of this apply to the NVMe connection handling from the CSI node agent as well?
I meant the regular connections (not HA).

mtzaurus (approved):

abhilashshetty04 (approved):

tiagolobocastro:
> LGTM On a relative front, does any of this apply to the NVMe connection handling from the CSI node agent as well? I meant the regular connections (not HA).

It's generally applicable but it's only a problem when we switchover and have more than 1, so for all intents and purposes only with HA

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

PR #422, opened 2023-01-18
fix(pstor/etcd): include queuing as part of the timeout

tiagolobocastro:
As things stands, store operations are serialized. This is simply a side-effect of the etcd client library taking operation methods in a mutable fashion. todo: whilst we probably don't want 1000x requests to hit the store at once, we probably can open the tap for more than 1 at a time :)

For now, simply include the queuing for our "slot" as part of the timeout.

### DIFF
--- a/control-plane/agents/src/bin/core/controller/registry.rs
+++ b/control-plane/agents/src/bin/core/controller/registry.rs
@@ -34,6 +34,7 @@ use common_lib::{
 };
 use std::{
     collections::HashMap,
+    future::Future,
     ops::{Deref, DerefMut},
     sync::Arc,
 };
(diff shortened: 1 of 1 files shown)


control-plane/agents/src/bin/core/controller/registry.rs
```diff
@@ -234,17 +240,33 @@ impl Registry {
         }
     }
 
+    async fn op_with_threshold<F, O>(future: F) -> O
+    where
+        F: Future<Output = O>,
+    {
+        let start = std::time::Instant::now();
+        let result = future.await;
+        let warn_threshold = std::time::Duration::from_secs(1);
```
Abhinandan-Purkait, line 249: Should the threshold be more, is 1s too fast ? How did we calculate this threshold?

Abhinandan-Purkait (approved):

dsavitskiy (approved):

tiagolobocastro:
bors merge

PR #427, opened 2023-01-24
refactor(dsp-operator): refactor disk pool operator

abhilashshetty04:
### mark dsp as terminating before deleting it

Currently when dsp cr is deleted, cmd gets hung if for some reason pool deletion is failing. Pool status stays as Online evnthough operator is trying to delete pool as part of reconcilation.

Signed-off-by: Abhilash Shetty [abhilash.shetty@datacore.com](mailto:abhilash.shetty@datacore.com)

### add respective control plane pool status in dsp cr
adding control plane pool status in CR. This is as is representation of respective control plane Pool resource state.

Signed-off-by: Abhilash Shetty [abhilash.shetty@datacore.com](mailto:abhilash.shetty@datacore.com)

### added new cr states to align with control plane states
added new cr state field. Reconciling on new states.

### DIFF
--- a/k8s/operators/src/diskpool/crd.rs
+++ b/k8s/operators/src/diskpool/crd.rs
@@ -56,6 +56,8 @@ pub enum PoolState {
     /// The resource is present, and the pool has been created. The schema MUST
     /// have a status and spec field.
     Online,
+    /// This state is set when we receive delete event on the dsp cr.
+    Terminating,
     /// The resource is present but the control plane did not return the pool state.
     Unknown,
     /// Trying to converge to the next state has exceeded the maximum retry
@@ -118,6 +120,21 @@ impl DiskPoolStatus {
             available: 0,
         }
     }
+
+    pub fn terminating(p: Pool) -> Self {
+        let state = p.state.unwrap_or_default();
+        let free = if state.capacity > state.used {
+            state.capacity - state.used
+        } else {
+            0
+        };
+        Self {
+            state: PoolState::Terminating,
+            capacity: state.capacity,
+            used: state.used,
+            available: free,
+        }
+    }
 }
 
 impl From<Pool> for DiskPoolStatus {
@@ -147,6 +164,7 @@ impl ToString for PoolState {
             PoolState::Online => "Online",
             PoolState::Unknown => "Unknown",
             PoolState::Error => "Error",
+            PoolState::Terminating => "Terminating",
         }
         .to_string()
     }
--- a/k8s/operators/src/diskpool/crd.rs
+++ b/k8s/operators/src/diskpool/crd.rs
@@ -18,8 +18,8 @@ derive = "PartialEq",
 derive = "Default",
 shortname = "dsp",
 printcolumn = r#"{ "name":"node", "type":"string", "description":"node the pool is on", "jsonPath":".spec.node"}"#,
-printcolumn = r#"{ "name":"status", "type":"string", "description":"pool status", "jsonPath":".status.state"}"#,
-printcolumn = r#"{ "name":"poolstatus", "type":"string", "description":"control plane pool status", "jsonPath":".status.status"}"#,
+printcolumn = r#"{ "name":"state", "type":"string", "description":"dsp cr state", "jsonPath":".status.state"}"#,
+printcolumn = r#"{ "name":"pool_status", "type":"string", "description":"Control plane pool status", "jsonPath":".status.pool_status"}"#,
 printcolumn = r#"{ "name":"capacity", "type":"integer", "format": "int64", "minimum" : "0", "description":"total bytes", "jsonPath":".status.capacity"}"#,
 printcolumn = r#"{ "name":"used", "type":"integer", "format": "int64", "minimum" : "0", "description":"used bytes", "jsonPath":".status.used"}"#,
 printcolumn = r#"{ "name":"available", "type":"integer", "format": "int64", "minimum" : "0", "description":"available bytes", "jsonPath":".status.available"}"#
@@ -46,7 +46,6 @@ impl DiskPoolSpec {
 
 #[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq, JsonSchema)]
 #[non_exhaustive]
-/// PoolState represents operator specific states for DSP CR.
 pub enum PoolState {
     /// The pool is a new OR missing resource, and it has not been created or
     /// imported yet by the operator. The pool spec MAY be but DOES
@@ -58,8 +57,6 @@ pub enum PoolState {
     /// The resource is present, and the pool has been created. The schema MUST
     /// have a status and spec field.
     Online,
-    /// This state is set when we receive delete event on the dsp cr.
-    Terminating,
     /// The resource is present but the control plane did not return the pool state.
     Unknown,
     /// Trying to converge to the next state has exceeded the maximum retry
@@ -68,6 +65,22 @@ pub enum PoolState {
     /// reconciliation stops. Only external events (a new resource version)
     /// will trigger a new attempt.
     Error,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq, JsonSchema, Default)]
+#[non_exhaustive]
+/// PoolState represents operator specific states for DSP CR.
+pub enum CrPoolState {
+    /// The pool is a new OR missing resource, and it has not been created or
+    /// imported yet by the operator. The pool spec MAY be but DOES
+    /// NOT have a status field.
+    #[default]
+    Creating,
+    /// The resource spec has been created, and the pool is getting created by
+    /// the control plane.
+    Created,
+    /// This state is set when we receive delete event on the dsp cr.
+    Terminating,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq, JsonSchema)]
@@ -87,10 +100,12 @@ pub enum PoolStatus {
 #[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq, JsonSchema)]
 /// Status of the pool which is driven and changed by the controller loop.
 pub struct DiskPoolStatus {
-    /// The state of the pool.
     pub state: PoolState,
+    /// The state of the pool.
+    #[serde(default)]
+    pub cr_state: CrPoolState,
     /// Pool status from respective control plane object.
-    pub status: Option<PoolStatus>,
+    pub pool_status: Option<PoolStatus>,
     /// Capacity as number of bytes.
     capacity: u64,
     /// Used number of bytes.
@@ -103,7 +118,8 @@ impl Default for DiskPoolStatus {
     fn default() -> Self {
         Self {
             state: PoolState::Creating,
-            status: None,
+            cr_state: CrPoolState::Creating,
+            pool_status: None,
             capacity: 0,
             used: 0,
             available: 0,
@@ -112,28 +128,15 @@ impl Default for DiskPoolStatus {
 }
 
 impl DiskPoolStatus {
-    /// Set when Pool creation fails for some reason.
-    pub fn error() -> Self {
-        Self {
-            state: PoolState::Error,
-            ..Default::default()
-        }
-    }
-    /// Set when create pool api call is successful.
-    pub fn created() -> Self {
+    /// Set when Pool is not found for some reason.
+    pub fn not_found() -> Self {
         Self {
             state: PoolState::Created,
+            pool_status: None,
             ..Default::default()
         }
     }
-    /// Set when we cant get the Pool from control plane.
-    pub fn unknown() -> Self {
-        Self {
-            state: PoolState::Unknown,
-            status: Some(PoolStatus::Unknown),
-            ..Default::default()
-        }
-    }
+
     /// Set when operator is attempting delete on pool.
     pub fn terminating(p: Pool) -> Self {
         let state = p.state.unwrap_or_default();
@@ -143,8 +146,9 @@ impl DiskPoolStatus {
             0
         };
         Self {
-            state: PoolState::Terminating,
-            status: Some(state.status.into()),
+            state: PoolState::Online,
+            cr_state: CrPoolState::Terminating,
+            pool_status: Some(state.status.into()),
             capacity: state.capacity,
             used: state.used,
             available: free,
@@ -154,8 +158,18 @@ impl DiskPoolStatus {
     /// Set when deleting a Pool which is not accessible.
     pub fn terminating_when_unknown() -> Self {
         Self {
-            state: PoolState::Terminating,
-            status: Some(PoolStatus::Unknown),
+            state: PoolState::Unknown,
+            cr_state: CrPoolState::Terminating,
+            pool_status: Some(PoolStatus::Unknown),
+            ..Default::default()
+        }
+    }
+
+    pub fn mark_unknown() -> Self {
+        Self {
+            state: PoolState::Unknown,
+            cr_state: CrPoolState::Created,
+            pool_status: Some(PoolStatus::Unknown),
             ..Default::default()
         }
     }
@@ -175,40 +189,26 @@ impl From<RestPoolStatus> for PoolStatus {
 /// Returns DiskPoolStatus from Control plane pool object.
 impl From<Pool> for DiskPoolStatus {
     fn from(p: Pool) -> Self {
-        let state = p.state.expect("pool does not have state");
-        // todo: Should we set the pool to some sort of error state?
-        let free = if state.capacity > state.used {
-            state.capacity - state.used
+        if let Some(state) = p.state {
+            let free = if state.capacity > state.used {
+                state.capacity - state.used
+            } else {
+                0
+            };
+            Self {
+                state: PoolState::Online,
+                cr_state: CrPoolState::Created,
+                pool_status: Some(state.status.into()),
+                capacity: state.capacity,
+                used: state.used,
+                available: free,
+            }
         } else {
-            0
-        };
-        Self {
-            state: PoolState::Online,
-            status: Some(state.status.into()),
-            capacity: state.capacity,
-            used: state.used,
-            available: free,
+            Self {
+                state: PoolState::Online,
+                cr_state: CrPoolState::Created,
+                ..Default::default()
+            }
         }
-    }
-}
-
-/// converts the pool state into a string
-impl ToString for PoolState {
-    fn to_string(&self) -> String {
-        match self {
-            PoolState::Creating => "Creating",
-            PoolState::Created => "Created",
-            PoolState::Online => "Online",
-            PoolState::Unknown => "Unknown",
-            PoolState::Error => "Error",
-            PoolState::Terminating => "Terminating",
-        }
-        .to_string()
-    }
-}
-/// Pool state into a string
-impl From<PoolState> for String {
-    fn from(p: PoolState) -> Self {
-        p.to_string()
     }
 }
--- a/k8s/operators/src/diskpool/crd.rs
+++ b/k8s/operators/src/diskpool/crd.rs
@@ -1,5 +1,5 @@
 use kube::CustomResource;
-use openapi::models::Pool;
+use openapi::models::{pool_status::PoolStatus as RestPoolStatus, Pool};
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 
@@ -19,6 +19,7 @@ derive = "Default",
 shortname = "dsp",
 printcolumn = r#"{ "name":"node", "type":"string", "description":"node the pool is on", "jsonPath":".spec.node"}"#,
 printcolumn = r#"{ "name":"status", "type":"string", "description":"pool status", "jsonPath":".status.state"}"#,
+printcolumn = r#"{ "name":"poolstatus", "type":"string", "description":"control plane pool status", "jsonPath":".status.status"}"#,
 printcolumn = r#"{ "name":"capacity", "type":"integer", "format": "int64", "minimum" : "0", "description":"total bytes", "jsonPath":".status.capacity"}"#,
 printcolumn = r#"{ "name":"used", "type":"integer", "format": "int64", "minimum" : "0", "description":"used bytes", "jsonPath":".status.used"}"#,
 printcolumn = r#"{ "name":"available", "type":"integer", "format": "int64", "minimum" : "0", "description":"available bytes", "jsonPath":".status.available"}"#
@@ -45,6 +46,7 @@ impl DiskPoolSpec {
 
 #[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq, JsonSchema)]
 #[non_exhaustive]
+/// PoolState represents operator specific states for DSP CR.
 pub enum PoolState {
     /// The pool is a new OR missing resource, and it has not been created or
     /// imported yet by the operator. The pool spec MAY be but DOES
@@ -69,15 +71,31 @@ pub enum PoolState {
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq, JsonSchema)]
-/// Status of the pool which is driven and changed by the controller loop
+#[non_exhaustive]
+/// PoolStatus is Control plane status of a given DSP CR.
+pub enum PoolStatus {
+    /// State is Unknown.
+    Unknown,
+    /// The pool is in normal working order.
+    Online,
+    /// The pool has experienced a failure but can still function.
+    Degraded,
+    /// The pool is completely inaccessible.
+    Faulted,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq, JsonSchema)]
+/// Status of the pool which is driven and changed by the controller loop.
 pub struct DiskPoolStatus {
-    /// The state of the pool
+    /// The state of the pool.
     pub state: PoolState,
-    /// Capacity as number of bytes
+    /// Pool status from respective control plane object.
+    pub status: Option<PoolStatus>,
+    /// Capacity as number of bytes.
     capacity: u64,
-    /// Used number of bytes
+    /// Used number of bytes.
     used: u64,
-    /// Available number of bytes
+    /// Available number of bytes.
     available: u64,
 }
 
@@ -85,6 +103,7 @@ impl Default for DiskPoolStatus {
     fn default() -> Self {
         Self {
             state: PoolState::Creating,
+            status: None,
             capacity: 0,
             used: 0,
             available: 0,
@@ -93,34 +112,29 @@ impl Default for DiskPoolStatus {
 }
 
 impl DiskPoolStatus {
-    /// error pool status
+    /// Set when Pool creation fails for some reason.
     pub fn error() -> Self {
         Self {
             state: PoolState::Error,
-            capacity: 0,
-            used: 0,
-            available: 0,
+            ..Default::default()
         }
     }
-    /// created pool status
+    /// Set when create pool api call is successful.
     pub fn created() -> Self {
         Self {
             state: PoolState::Created,
-            capacity: 0,
-            used: 0,
-            available: 0,
+            ..Default::default()
         }
     }
-    /// unknown pool status
+    /// Set when we cant get the Pool from control plane.
     pub fn unknown() -> Self {
         Self {
             state: PoolState::Unknown,
-            capacity: 0,
-            used: 0,
-            available: 0,
+            status: Some(PoolStatus::Unknown),
+            ..Default::default()
         }
     }
-
+    /// Set when operator is attempting delete on pool.
     pub fn terminating(p: Pool) -> Self {
         let state = p.state.unwrap_or_default();
         let free = if state.capacity > state.used {
@@ -130,13 +144,35 @@ impl DiskPoolStatus {
         };
         Self {
             state: PoolState::Terminating,
+            status: Some(state.status.into()),
             capacity: state.capacity,
             used: state.used,
             available: free,
         }
     }
+
+    /// Set when deleting a Pool which is not accessible.
+    pub fn terminating_when_unknown() -> Self {
+        Self {
+            state: PoolState::Terminating,
+            status: Some(PoolStatus::Unknown),
+            ..Default::default()
+        }
+    }
 }
 
+impl From<RestPoolStatus> for PoolStatus {
+    fn from(p: RestPoolStatus) -> Self {
+        match p {
+            RestPoolStatus::Unknown => Self::Unknown,
+            RestPoolStatus::Online => Self::Online,
+            RestPoolStatus::Degraded => Self::Degraded,
+            RestPoolStatus::Faulted => Self::Faulted,
+        }
+    }
+}
+
+/// Returns DiskPoolStatus from Control plane pool object.
 impl From<Pool> for DiskPoolStatus {
     fn from(p: Pool) -> Self {
         let state = p.state.expect("pool does not have state");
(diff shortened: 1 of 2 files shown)


k8s/operators/src/diskpool/crd.rs
```diff
@@ -118,6 +120,21 @@ impl DiskPoolStatus {
             available: 0,
         }
     }
+
```
Abhinandan-Purkait, line 139: doc comment? Can it be pub(crate) ?

Abhinandan-Purkait (approved):

k8s/operators/src/diskpool/crd.rs
```diff
@@ -56,6 +56,8 @@ pub enum PoolState {
     /// The resource is present, and the pool has been created. The schema MUST
     /// have a status and spec field.
     Online,
+    /// This state is set when we receive delete event on the dsp cr.
+    Terminating,
```
tiagolobocastro: I thought we discussed yesterday not doing this and instead separating into 2 different states? Am I misremembering @blaisedias ?

abhilashshetty04: @tiagolobocastro, Now i have added one more field i.e status. This will represent control plane status for the pool.

k8s/operators/src/diskpool/crd.rs
abhilashshetty04: Added doc. WIll check on (crate)

k8s/operators/src/diskpool/crd.rs
```diff
@@ -1,5 +1,5 @@
 use kube::CustomResource;
-use openapi::models::Pool;
+use openapi::models::{pool_status::PoolStatus as RestPoolState, Pool};
```
tiagolobocastro: This is a bit confusing, is it status or state?

k8s/operators/src/diskpool/crd.rs
```diff
@@ -66,11 +69,26 @@ pub enum PoolState {
     Error,
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq, JsonSchema)]
+#[non_exhaustive]
+pub enum PoolStatus {
```
tiagolobocastro, line 89: missing doc comment for PoolStatus type.

k8s/operators/src/diskpool/crd.rs
```diff
@@ -66,11 +69,26 @@ pub enum PoolState {
...
+    /// State is Unknown
```
tiagolobocastro: doc comments should be `C`apitalized and end with a `.`

k8s/operators/src/diskpool/crd.rs
```diff
@@ -66,11 +69,26 @@ pub enum PoolState {
...
+    /// the pool is completely inaccessible
+    Faulted,
+}
+
 #[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq, JsonSchema)]
 /// Status of the pool which is driven and changed by the controller loop
 pub struct DiskPoolStatus {
     /// The state of the pool
     pub state: PoolState,
+    /// Pool state from respective control plane object.
```
tiagolobocastro:
```suggestion
    /// Pool status from respective control plane object.
```

k8s/operators/src/diskpool/main.rs
```diff
@@ -647,7 +672,26 @@ async fn ensure_crd(k8s: Client) {
...
+        info!(
+            "Creating CRD: {}",
+            serde_json::to_string_pretty(&crd).unwrap()
+        );
+
+        let pp = PostParams::default();
+        match dsp.create(&pp, &crd).await {
+            Ok(o) => {
+                info!(crd = ?o.name_any(), "created");
+                // let the CRD settle this purely to avoid errors messages in the console
```
tiagolobocastro: this piece of code is duplicated, please reuse.

k8s/operators/src/diskpool/main.rs
```diff
@@ -672,14 +716,14 @@ fn error_policy(error: &Error, _ctx: Arc<OperatorContext>) -> Action {
         when.to_rfc2822(),
         duration.as_secs()
     );
+    info!("requeue after :{:?}", duration);
```
tiagolobocastro: we already have a message above, why is this needed?

k8s/operators/src/diskpool/main.rs
```diff
@@ -619,20 +645,19 @@ impl ResourceContext {
...
 
-    // the CRD has not been installed yet, to avoid overwriting (and create upgrade issues) only
-    // install it when there is no crd with the given name
-    if crds.iter().count() == 0 {
-        let crd = DiskPool::crd();
+    // Check if there is an existing dsp crd. If yes, Replace it with new generated one.
+    // Create new if it doesnt exist.
+    let mut crd = DiskPool::crd();
+    if let Ok(exiting) = dsp.get("diskpools.openebs.io").await {
+        crd.metadata.resource_version = exiting.resource_version();
```
tiagolobocastro: Should we only patch if version differs?

k8s/operators/src/diskpool/main.rs
```diff
@@ -619,20 +645,19 @@ impl ResourceContext {
...
         info!(
             "Creating CRD: {}",
             serde_json::to_string_pretty(&crd).unwrap()
         );
 
         let pp = PostParams::default();
-        match dsp.create(&pp, &crd).await {
+        match dsp.replace("diskpools.openebs.io", &pp, &crd).await {
```
tiagolobocastro: below says created, but instead should it be `replaced`?

k8s/operators/src/diskpool/main.rs
```diff
@@ -619,20 +645,19 @@ impl ResourceContext {
...
-    let crds = dsp.list(&lp).await.expect("failed to list CRDS");
 
-    // the CRD has not been installed yet, to avoid overwriting (and create upgrade issues) only
-    // install it when there is no crd with the given name
-    if crds.iter().count() == 0 {
-        let crd = DiskPool::crd();
+    // Check if there is an existing dsp crd. If yes, Replace it with new generated one.
+    // Create new if it doesnt exist.
+    let mut crd = DiskPool::crd();
+    if let Ok(exiting) = dsp.get("diskpools.openebs.io").await {
```
tiagolobocastro: I know it was already hardcoded like so, but can we retrieve this value from somewhere instead, eg: `CRD` type?

k8s/operators/src/diskpool/main.rs
abhilashshetty04: Will remove it . It was logger i added for testing

k8s/operators/src/diskpool/main.rs
abhilashshetty04: resource_version is not present in auto generated crd. it is added by k8s once object is updated. How would we know if version is different?

tiagolobocastro: Not this one, we should bump the CRD api version from v1alpha1 to something higher. Then we can check this, and if != then update.

abhilashshetty04: Could be tedious to update crd resource version manually for any changes in crd. If we use this validation we have to bump it everytime right?  With current code it just replaces with crd from the operator. If someone doesnt update crd version new changes will not be patched and new field will not reflect in CR. So thought of just replacing it regardless. Should not impact if there is no change in crd.

k8s/operators/src/diskpool/crd.rs
abhilashshetty04: Added

k8s/operators/src/diskpool/crd.rs
abhilashshetty04: Done

k8s/operators/src/diskpool/crd.rs
abhilashshetty04: Done

k8s/operators/src/diskpool/main.rs
abhilashshetty04: now getting crd name from the crd instance that gets created on line 651

k8s/operators/src/diskpool/main.rs
tiagolobocastro: If it's tedious then we automate it, but we should update the version, otherwise what's the point of having a version in the CRD?
I'm fine with updating everytime then, but let's update the api version anyway.

k8s/operators/src/diskpool/main.rs
abhilashshetty04: We create if crd is not there in the cluster. If its there we replace with crd from the dsp operator

k8s/operators/src/diskpool/main.rs
abhilashshetty04: Okay will add it.

k8s/operators/src/diskpool/crd.rs
```diff
@@ -19,6 +19,7 @@ derive = "Default",
 shortname = "dsp",
 printcolumn = r#"{ "name":"node", "type":"string", "description":"node the pool is on", "jsonPath":".spec.node"}"#,
 printcolumn = r#"{ "name":"status", "type":"string", "description":"pool status", "jsonPath":".status.state"}"#,
+printcolumn = r#"{ "name":"poolstatus", "type":"string", "description":"control plane pool status", "jsonPath":".status.status"}"#,
```
tiagolobocastro: what should the control-plane status be named as? @blaisedias @avishnu @Abhinandan-Purkait

abhilashshetty04:
Any suggesting on what should be control plane Pool status on dsp cr be called?

k8s/operators/src/diskpool/crd.rs
```diff
@@ -67,22 +71,39 @@ pub enum PoolState {
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq, JsonSchema)]
-/// Status of the pool which is driven and changed by the controller loop
+#[non_exhaustive]
+/// PoolStatus is Control plane status of a given DSP CR.
+pub enum PoolStatus {
```
chriswldenyer, line 89: PoolStatus sounds a lot like PoolState. IMHO it needs a more descriptive name to distinguish it from PoolState, so that it is obvious to developers what it means.

k8s/operators/src/diskpool/crd.rs
```diff
@@ -67,22 +71,39 @@ pub enum PoolState {
...
+
+#[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq, JsonSchema)]
+/// Status of the pool which is driven and changed by the controller loop.
 pub struct DiskPoolStatus {
-    /// The state of the pool
+    /// The state of the pool.
     pub state: PoolState,
-    /// Capacity as number of bytes
+    /// Pool status from respective control plane object.
+    pub status: Option<PoolStatus>,
```
chriswldenyer: See my earlier comment, state and status mean the same thing, so a more descriptive name for this variable will make it easier for developers.

k8s/operators/src/diskpool/crd.rs
Abhinandan-Purkait: This can a bit confusing for user, why should a user be interested in the CR status? Can we not display only the control-plane status by default i.e for `kubectl get dsp -n mayastor` and show the CR status if the user does a `kubectl get dsp -owide -n mayastor`. For the naming convention, we can call the control-plane status as the `status` and the existing `status` as `CR Status` which will be only shown on `-owide`.

k8s/operators/src/diskpool/main.rs
```diff
@@ -660,7 +692,7 @@ fn error_policy(error: &Error, _ctx: Arc<OperatorContext>) -> Action {
         Error::ReconcileError { .. } => {
             return Action::await_change();
         }
-        _ => 5,
+        _ => 20,
```
chriswldenyer: It is better practice to use a constant here IMHO.

k8s/operators/src/diskpool/crd.rs
Abhinandan-Purkait: Yeah, the variable names can be `control_plane_status` and `cr_status` respectively for clarity. I agree.

k8s/operators/src/diskpool/main.rs
```diff
@@ -366,6 +351,15 @@ impl ResourceContext {
                 // 'Created' to reflect this.
             }
             Err(e) => {
+                if self.num_retries == 10 {
```
chriswldenyer: It looks like the limit is now hardcoded to 10 instead of being configurable. Why was this done?

k8s/operators/src/diskpool/main.rs
```diff
@@ -619,20 +632,20 @@ impl ResourceContext {
...
 
-    // the CRD has not been installed yet, to avoid overwriting (and create upgrade issues) only
-    // install it when there is no crd with the given name
-    if crds.iter().count() == 0 {
-        let crd = DiskPool::crd();
+    // Check if there is an existing dsp crd. If yes, Replace it with new generated one.
+    // Create new if it doesnt exist.
+    let mut crd = DiskPool::crd();
+    let crd_name = crd.metadata.name.as_ref().unwrap().as_str();
+    if let Ok(exiting) = dsp.get(crd_name).await {
```
chriswldenyer: typo "existing" ?

k8s/operators/src/diskpool/main.rs
```diff
@@ -619,20 +632,20 @@ impl ResourceContext {
...
+        crd.metadata.resource_version = exiting.resource_version();
         info!(
             "Creating CRD: {}",
```
chriswldenyer: It looks like this message should be changed to "Replacing CRD" or something like that?

k8s/operators/src/diskpool/crd.rs
tiagolobocastro: What about `cr_status` and `pool_status`?

Abhinandan-Purkait: Sounds good.

k8s/operators/src/diskpool/main.rs
abhilashshetty04: When the 10th retry results in failure we add failure event on the CR. Hence hardcoded it. This doesnt need to be a user configurable paramater IMO. Do you think this should be constant?

k8s/operators/src/diskpool/crd.rs
abhilashshetty04: I agree. We need to make names more clear.

k8s/operators/src/diskpool/crd.rs
tiagolobocastro: This is mirroring the control-plane where State is composed of a Status and other fields, eg: `VolumeState { status: Online, replicas: 3, .... }`
But it becomes messy here with the PoolState above which I think we might be able to rename to `CrStatus` instead?

k8s/operators/src/diskpool/main.rs
chriswldenyer: Ok, sounds reasonable for it to be hardcoded.
Having a constant to define it would be prefereable, I think.

k8s/operators/src/diskpool/crd.rs
abhilashshetty04: Im okay with CrStatus.

k8s/operators/src/diskpool/main.rs
abhilashshetty04: Will correct

k8s/operators/src/diskpool/main.rs
abhilashshetty04: Removed this hardcoding

k8s/operators/src/diskpool/crd.rs
```diff
@@ -18,7 +18,8 @@ derive = "PartialEq",
 derive = "Default",
 shortname = "dsp",
 printcolumn = r#"{ "name":"node", "type":"string", "description":"node the pool is on", "jsonPath":".spec.node"}"#,
-printcolumn = r#"{ "name":"status", "type":"string", "description":"pool status", "jsonPath":".status.state"}"#,
+printcolumn = r#"{ "name":"state", "type":"string", "description":"pool status", "jsonPath":".status.state"}"#,
```
tiagolobocastro: description here should be pool state?

k8s/operators/src/diskpool/main.rs
```diff
@@ -239,55 +245,41 @@ impl ResourceContext {
...
-    /// error state will not be deleted.
-    async fn mark_error(&self) -> Result<Action, Error> {
-        let _ = self.patch_status(DiskPoolStatus::error()).await?;
+    /// Mark Pool state as None as couldnt find already provisioned pool in control plane.
+    async fn mark_pool_not_found(&self) -> Result<Action, Error> {
+        self.patch_status(DiskPoolStatus::not_found()).await?;
 
-        error!(name = ?self.name_any(),"status set to error");
-        Ok(Action::await_change())
+        error!(name = ?self.name_any(),"status pool state to blank");
```
tiagolobocastro:
```suggestion
        error!(name = ?self.name_any(), "Pool not found, clearing status");
```

k8s/operators/src/diskpool/main.rs
```diff
@@ -497,7 +497,7 @@ impl ResourceContext {
             }
             Err(clients::tower::Error::Request(_)) => {
                 // Probably grpc server is not yet up
-                return self.mark_unknown().await;
+                return Ok(Action::requeue(Duration::from_secs(3)))
```
tiagolobocastro: Shouldn't we be marking as unknown here?

k8s/operators/src/diskpool/main.rs
```diff
@@ -479,11 +479,11 @@ impl ResourceContext {
 
                         // We expected the control plane to have a spec for this pool. It didn't so
                         // set the CRD to the error state and don't try to recreate it.
-                        self.mark_error().await
+                        self.mark_pool_not_found().await
                     }
                 } else if response.status() == clients::tower::StatusCode::SERVICE_UNAVAILABLE || response.status() == clients::tower::StatusCode::REQUEST_TIMEOUT {
                     // Probably grpc server is not yet up
-                    self.mark_unknown().await
+                    Ok(Action::requeue(Duration::from_secs(5)))
```
tiagolobocastro: Shouldn't we be marking as unknown here?

k8s/operators/src/diskpool/main.rs
```diff
@@ -600,13 +601,25 @@ impl ResourceContext {
             |event| async move {
                 match event {
                     finalizer::Event::Apply(dsp) => Self::put_finalizer(dsp).await,
-                    finalizer::Event::Cleanup(_dsp) => Self::delete_finalizer(self.clone()).await,
+                    finalizer::Event::Cleanup(dsp) => {
+                        if dsp.status.as_ref().unwrap().cr_state != CrPoolState::Terminating {
+                            info!("set state as terminating");
```
tiagolobocastro: do we need this info here and below?

k8s/operators/src/diskpool/main.rs
```diff
@@ -619,13 +632,34 @@ impl ResourceContext {
...
+                // let the CRD settle this purely to avoid errors messages in the console
+                // that are harmless but can cause some confusion maybe.
+                tokio::time::sleep(Duration::from_secs(5)).await;
+            }
+
+            Err(e) => {
+                error!("failed to create CRD error {}", e);
+                tokio::time::sleep(Duration::from_secs(1)).await;
+                std::process::exit(1);
+            }
```
tiagolobocastro: seems this code is duplicated below, can we reuse?

tiagolobocastro (commented):
You will have to rebase.
But before that, do we want to raise it here first or in the release/2.0?
@Abhinandan-Purkait  @blaisedias @avishnu

k8s/operators/src/diskpool/crd.rs
blaisedias: I can see that the new CR status is named as `cr_state` and what I assume is control-plane pool state is `state` I think to disambiguate it should not be named `state`
Maybe `pool_state`  or  pedantically `control_plane_pool_state`.

k8s/operators/src/diskpool/main.rs
abhilashshetty04: I think so . Should we do it at first ServiceUnavailable itself? Thinking of marking it as None. Control plane status is Option anyway.

k8s/operators/src/diskpool/main.rs
abhilashshetty04: No needed will remove

k8s/operators/src/diskpool/crd.rs
abhilashshetty04: Yes

k8s/operators/src/diskpool/main.rs
tiagolobocastro: Yes I think so, because we cannot determine what the status is, should be None (or unknown ?).

abhilashshetty04: we dont know so maybe lets not post anything. These erros are connectivity/gRPC issues. So dont know what is the status .

k8s/operators/src/diskpool/crd.rs
chriswldenyer: I thought the suggestion was pool_status?
Using "state " and "status" to distinguish between different things arbitrarily isn't good IMHO.

chriswldenyer (commented):
Please use a different name for the pool status.

k8s/operators/src/diskpool/crd.rs
abhilashshetty04: state is old cr states. status was control plane status. Made it pool_status now for more clarity.

k8s/operators/src/diskpool/crd.rs
abhilashshetty04: Yes, made it pool_status. This is control plane status for the dsp.

k8s/operators/src/diskpool/main.rs
abhilashshetty04: Done

k8s/operators/src/diskpool/main.rs
tiagolobocastro: The problem is that if we don't change state, and the rest/grpc is not back up for whatever reason, we'll keep saying a pool is online for example, when it might not be.
I don't know what the common practice is in k8s for this, anyone knows? @Abhinandan-Purkait  @niladrih

abhilashshetty04: I was thinking of not showing it at all. marking as None. We dont know will not show

k8s/operators/src/diskpool/main.rs
abhilashshetty04: Removed

k8s/operators/src/diskpool/crd.rs
abhilashshetty04: made it control plane pool state

k8s/operators/src/diskpool/main.rs
abhilashshetty04: Done

k8s/operators/src/diskpool/main.rs
abhilashshetty04: Setting as None

k8s/operators/src/diskpool/main.rs
abhilashshetty04: Setting as None

k8s/operators/src/diskpool/main.rs
```diff
@@ -838,12 +814,6 @@ async fn main() -> anyhow::Result<()> {
                 .env("JAEGER_ENDPOINT")
                 .help("enable open telemetry and forward to jaeger"),
         )
-        .arg(
```
tiagolobocastro: Don't remove this flag from CLI as it would be breaking change, instead simply not use it.

k8s/operators/src/diskpool/main.rs
abhilashshetty04: Incorporated

k8s/operators/src/diskpool/main.rs
abhilashshetty04: Incorporated

k8s/operators/src/diskpool/main.rs
abhilashshetty04: Alright

abhilashshetty04: Done

k8s/operators/src/diskpool/main.rs
```diff
@@ -362,13 +300,75 @@ impl ResourceContext {
...
+                            Ok(Action::requeue(Duration::from_secs(self.ctx.interval)))
+                        }
+                    }
+                    Err(clients::tower::Error::Response(response))
+                        if response.status() == StatusCode::NOT_FOUND =>
+                    {
+                        self.k8s_notify(
+                            "Create or Import Failure",
+                            "Failure",
+                            format!("Unable to find Io engine node {}", &self.spec.node()).as_str(),
```
tiagolobocastro: s/Io engine/io-engine

k8s/operators/src/diskpool/main.rs
```diff
@@ -362,13 +300,75 @@ impl ResourceContext {
...
+                            "Critical",
+                        )
+                        .await;
+                        error!("Unable to find Io engine node {}", &self.spec.node());
```
tiagolobocastro: s/Io engine/io-engine

k8s/operators/src/diskpool/main.rs
```diff
@@ -362,13 +300,75 @@ impl ResourceContext {
...
+                        } else {
+                            self.k8s_notify(
+                                "Create or Import Failure",
+                                "Failure",
+                                format!("Unable to create or import pool {}", error).as_str(),
+                                "Critical",
+                            )
+                            .await;
+                            error!("Unable to create or import pool {}", error);
+                            Ok(Action::requeue(Duration::from_secs(self.ctx.interval)))
```
tiagolobocastro: return error

k8s/operators/src/diskpool/main.rs
```diff
@@ -362,13 +300,75 @@ impl ResourceContext {
...
+                    {
+                        self.k8s_notify(
+                            "Create or Import Failure",
+                            "Failure",
+                            format!("Unable to find Io engine node {}", &self.spec.node()).as_str(),
+                            "Critical",
+                        )
+                        .await;
+                        error!("Unable to find Io engine node {}", &self.spec.node());
+                        Ok(Action::requeue(Duration::from_secs(self.ctx.interval)))
```
tiagolobocastro: return error

k8s/operators/src/diskpool/main.rs
```diff
@@ -362,13 +300,75 @@ impl ResourceContext {
...
+                    _ => {
+                        self.k8s_notify(
+                            "Create or Import Failure",
+                            "Failure",
+                            format!("Unable to create or import pool {}", error).as_str(),
+                            "Critical",
+                        )
+                        .await;
+                        error!("Unable to create or import pool {}", error);
+                        Ok(Action::requeue(Duration::from_secs(self.ctx.interval)))
```
tiagolobocastro: return error

k8s/operators/src/diskpool/main.rs
```diff
@@ -470,20 +451,28 @@ impl ResourceContext {
...
-                    self.mark_unknown().await
+                    self.k8s_notify(
+                        "Unreachable",
+                        "Check",
+                        "Could not reach Rest API service. Please check control plane health",
+                        "Warning",
+                    )
+                        .await;
+                    self.mark_pool_not_found().await?;
+                    Ok(Action::requeue(Duration::from_secs(10)))
```
tiagolobocastro:
```suggestion
                    self.mark_pool_not_found().await
```

k8s/operators/src/diskpool/main.rs
```diff
@@ -559,36 +541,44 @@ impl ResourceContext {
...
+                            std::env::var("MY_POD_NAME")
+                                .ok()
+                                .unwrap_or_else(|| WHO_AM_I_SHORT.into()),
+                        ),
+                        message: Some(message.into()),
+                        ..Default::default()
+                    },
+                )
+                .await
+                .map_err(|e| error!(?e));
```
tiagolobocastro:
```suggestion
                .map_err(|error| error!(?error));
```

k8s/operators/src/diskpool/main.rs
```diff
@@ -600,13 +590,32 @@ impl ResourceContext {
...
+                        match self
+                            .pools_api()
+                            .get_node_pool(&self.spec.node(), &self.name_any())
+                            .await
+                        {
+                            Ok(pool) => {
+                                if dsp.status.as_ref().unwrap().cr_state == CrPoolState::Created {
+                                    let new_status = DiskPoolStatus::terminating(pool.into_body());
+                                    let _ = self.patch_status(new_status).await?;
+                                }
```
tiagolobocastro: Why do we check for Create? Otherwise if Creating then we don't transition to Terminating?

tiagolobocastro (commented):
Still needs rebasing @abhilashshetty04

k8s/operators/src/diskpool/main.rs
abhilashshetty04: We want to show Pool usage when marking Terminating. Since we couldnt get pool from the control plane we just delete finalizer for that cr and remove from inventory. We dont need to transition to Terminating for this IMO. There should be nothing holding off on directly deleting cr? Still, we do check if grpc status is NotFound before deletingcr. Or else requeue

tiagolobocastro: Here we could get pool from control-plane, it's just the `CR` that is `Creating`?
What if the pool has replicas? Or if the node is offline? Then we cannot delete the pool, so meanwhile we should show the `cr_state` as `Terminating`, even if the if the previous state was `Creating` and not `Created`? Otherwise it will stay as `Creating` even after user issued delete, which doesn't seem right to me.

@blaisedias is this covered by the BDD, and if so what is the expected behaviour?

k8s/operators/src/diskpool/main.rs
abhilashshetty04: Done

k8s/operators/src/diskpool/main.rs
abhilashshetty04: done

k8s/operators/src/diskpool/main.rs
abhilashshetty04: done

k8s/operators/src/diskpool/main.rs
abhilashshetty04: done

k8s/operators/src/diskpool/main.rs
abhilashshetty04: done

k8s/operators/src/diskpool/main.rs
abhilashshetty04: done

k8s/operators/src/diskpool/main.rs
abhilashshetty04: done

k8s/operators/src/diskpool/main.rs
abhilashshetty04: @tiagolobocastro We will tarnsition to Terminating now on all states. Made changes

tiagolobocastro (approved):

abhilashshetty04:
bors try

abhilashshetty04:
bors merge

k8s/operators/src/diskpool/main.rs
blaisedias: It is indeed covered by the BDD- create and a delete on a k8s node not provisioned with the IO engine.

PR #429, opened 2023-01-30
fix(release): use latest annotated tag only

tiagolobocastro:
When building the images, there may be the case where 2 tags point at HEAD. In this case, use the latest annotated tag only.


pchandra19 (approved):

tiagolobocastro:
bors merge

utils/utils-lib/src/test_constants.rs
```diff
@@ -1,3 +1,3 @@
 /// Automatically generated by ROOT/scripts/rust/branch_ancestor.sh
```
blaisedias, line 1: `automatically generated by ....` so this change should be made elsewhere?

tiagolobocastro: No, it doesn't matter, this file is always re-generated when compiling, but it's checked in so I thought might as well keep it up to date for when browsing on github.

PR #431, opened 2023-01-31
Merge develop fixes to release/2.0

tiagolobocastro:
fix(nexus/create): destroy unused nexuses on creation
    
    When creation fails due to an unused nexus (no equivalent spec), or deleting nexus,
    simply destroy said nexus and retry the creation attempt once.
    
---

    fix(volume/publish): clean up nexus/spec on failure
    
    For some weird reasons, after each nexus creation we check if the nexus is listed by the io=engine.
    This is due to the new v1 idempotency where if already exists an error is returned.
    
    When creating the first time we don't need this listing because we already have the nexus returned.
    The listing was actually creating a weird bug where it was timing out after a successful creation.
    To fix this properly, a new enum `OnCreateFail` was introduced which allows us to control the
    behaviour when create fails. By default it will be set to `SetDeleting` which means we'll set the
    spec to Deleting on a failed creation allowing the garbage collector to clean up..
    
### DIFF
--- a/common/src/types/v0/store/mod.rs
+++ b/common/src/types/v0/store/mod.rs
@@ -79,6 +79,10 @@ pub trait SpecTransaction<Operation> {
     fn start_op(&mut self, operation: Operation);
     /// Sets the result of the operation
     fn set_op_result(&mut self, result: bool);
+    /// Allow this operation while deleting a spec.
+    fn allow_op_deleting(&mut self, _operation: &Operation) -> bool {
+        false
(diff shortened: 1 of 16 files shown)


chriswldenyer (approved):
Looks reasonable

Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

PR #432, opened 2023-02-02
refactor grpc versioning interface

tiagolobocastro:
refactor(io-engine/grpc): move api differences to the interface impl
    
    This simplifies the node wrapper which is the "user" of the api.
    Separated the nexus traits into 3, allowing for a saner customization of the returns
    that are needed to bridge v0 and v1 api's.
    
    TODO: noticed fault_child is not returning the nexus as the other v1 nexus api's.
    
---

    refactor(io-engine/rpc): tidy up msg translation between core-io
    
---

    fix(agent/core): avoid making unnecessary grpc to io-engine
    
    When the results are retrived via gRPC simply update the registry.
    Note that in some situations that is no possible, example:
    Creating a replica affects the pool free space, so good to update pool?
    Share/Unshare doesn't return the object itself :(
    AddChild/Remove doesn't return the nexus object either..
    
    Some of these are fixed in v1 api, so todo:
    Move idempotency to each version impl allowing v1 to skip some of these..
    
---

    refactor(io-engine/rpc): split the version handling in separate modules
    
    Adds new traits for the pool/replica/nexus/host api's which are implemented by each
    version handler rpc.
    This should make it much easier to add newer versions in the future.
    Also split the node/pool wrapper in separate modules.
    

Abhinandan-Purkait (approved):
Thanks for the refactor. 👍🏻

tiagolobocastro:
bors merge

PR #435, opened 2023-02-10
chore: don't rebuild when git HEAD and index changes

tiagolobocastro:
Container images are built with the version provided by nix so we don't necesserily need to keep local builds with the right version.
This makes development a lot faster as otherwise for every git stage change or commit we have to rebuild everything.

### DIFF
--- a/utils/utils-lib/src/version.rs
+++ b/utils/utils-lib/src/version.rs
@@ -1,0 +1,88 @@
+pub mod macros {
+    /// Makes a version info instance.
+    #[macro_export]
+    macro_rules! version_info {
+        () => {{
+            $crate::version_info_inner!($crate::long_raw_version_str())
+        }};
+    }
+
+    /// Returns a version info instance.
+    #[macro_export]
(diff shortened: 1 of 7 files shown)


utils/utils-lib/src/version.rs
```diff
@@ -0,0 +1,85 @@
...
+    /// This includes the package name and version, and commit info.
+    #[macro_export]
+    macro_rules! print_package_info {
+        () => {
+            println!("{}", $crate::fmt_package_info!());
+        };
+    }
+}
+
+pub fn raw_version_string() -> String {
```
Abhinandan-Purkait, line 56: Doc comments needed? Below as well?

Abhinandan-Purkait (approved):
Thanks.  🚀

utils/utils-lib/src/version.rs
tiagolobocastro: Done

tiagolobocastro:
bors merge

PR #437, opened 2023-02-14
refactor: split out the common into separate crates

tiagolobocastro:
Carry on previous work of splitting crates into utils. At the moment the control-plane types/aka transport cannot be split from the store types.
TODO:
More work needs to be done there to decouple those types, will be taken on another PR.
Also we should stop using ETCd directly in some cases and use the PSTor interface.


Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

PR #446, opened 2023-02-22
Capacity - remove largest enospc replica and online enospc child

tiagolobocastro:
feat(capacity): remove largest enospc replicas

When a pool is exhausted multiple thin-provisioned replicas may fail with ENOSPC.
In this situation we'll currently adopt the simplistic strategy of faulting the largest replica
on the pool, or more specifically the replica with largest actual allocation.
A pre-condition for the replica faulting is that the volume to which the replica belongs to
should retain "enough" remaining healthy replicas! For example, we can't fault the last replica
of a volume as we can't rebuild from "thin" air.

todo: online nexus children if their respective pools have sufficient capacity.

---

feat(capacity): online enospc children if there is enough capacity in the pool

It's not trivial to find out if we should online a child as the pool allocation can be forever
changing.
For capacity v1 start with a kiss approach of simply onlining children if we have *some* space.
We take into account the current replica and volume allocation.
Example, if the failed child's replica has been allocated up to 90% of the volume full capacity,
then we don't need free space for the whole volume capacity.
At the same time, if the volume current allocation is 90% then we need to make sure we have free
space to reach 90% of the volume allocation.
A dummy slack is added on top, which is mostly a placeholder for a future replacement that would
take into account pool allocation growth.

### DIFF
--- a/control-plane/agents/src/bin/core/controller/io_engine/v1/nexus.rs
+++ b/control-plane/agents/src/bin/core/controller/io_engine/v1/nexus.rs
@@ -188,7 +188,13 @@ impl crate::controller::io_engine::NexusShareApi<Nexus, Nexus> for super::RpcCli
 
 #[async_trait::async_trait]
 impl crate::controller::io_engine::NexusChildApi<Nexus, Nexus, ()> for super::RpcClient {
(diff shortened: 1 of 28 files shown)


dsavitskiy (approved):

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

PR #447, opened 2023-02-28
feat(volume group): add openapi spec and stor_port type for VolumeGroup

Abhinandan-Purkait:
Adds REST API Changes.

### DIFF
--- a/control-plane/rest/openapi-specs/v0_api_spec.yaml
+++ b/control-plane/rest/openapi-specs/v0_api_spec.yaml
@@ -2051,6 +2051,7 @@ components:
         size: 10485761
         thin: false
         topology: null
+        volume_group: null
       description: Create Volume Body
       type: object
       properties:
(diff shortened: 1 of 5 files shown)


abhilashshetty04 (approved):

control-plane/rest/openapi-specs/v0_api_spec.yaml
```diff
@@ -2077,11 +2078,25 @@ components:
           type: object
           additionalProperties:
             type: string
+        volume_group:
+          description: Volume Group related information.
```
tiagolobocastro, line 2082:
```suggestion
```
This doesn't do anything, simply remove it since the desc is the same as the defined in the type itself or use an allOf.

control-plane/stor-port/src/types/v0/transport/volume.rs
```diff
@@ -352,6 +352,26 @@ pub struct CreateVolume {
     pub thin: bool,
 }
 
+/// Volume Group related information.
+#[derive(Serialize, Deserialize, Default, Debug, Clone, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct VolumeGroup {
+    /// The name of the volume group.
+    pub name: String,
```
tiagolobocastro: nit: remove pub and add `fn name(&self) -> &str` ?

tiagolobocastro (approved):

control-plane/rest/openapi-specs/v0_api_spec.yaml
Abhinandan-Purkait: Ahh right, I missed this. Thanks. Added the `allOf`.

control-plane/stor-port/src/types/v0/transport/volume.rs
Abhinandan-Purkait: Right. Added the `fn new()` and `fn name()`

Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors merge

PR #448, opened 2023-03-01
feat(rebuild/policy): add policies for rebuild orchestration

dsharma-dc:
Adding policies to orchestrate the behaviour of rebuilds. Based on these policies the control-plane decides whether to online the existing faulted child, in effect doing partial rebuild, OR immediately remove faulted child and add new one, in effect doing full rebuild.

### DIFF
--- a/control-plane/agents/src/bin/core/controller/policies/rebuild_policies.rs
+++ b/control-plane/agents/src/bin/core/controller/policies/rebuild_policies.rs
@@ -1,0 +1,107 @@
+#![allow(unused)]
+use stor_port::types::v0::transport::Nexus;
+
+use serde::{Deserialize, Serialize};
+use std::time::Duration;
+
+/*
+    This file defines the policies that control plane uses to make
+    decisions about rebuild workflow behaviour in case of a child
+    becoming faulted. We can logically provide multiple different policies.
+    These policies are represented as an enum, since only one of the policy
+    can be applied to a given volume/nexus at a time. Once applied, the
+    behaviour of the policy can vary depending on runtime state of the nexus.
+    Any new policy introduced in future will have to implement Policy trait
+    for defining the policy behaviour.
+*/
+
+// time constants, in seconds
+const TWAIT_SPERF: std::time::Duration = Duration::new(600, 0);
+const TWAIT_SAVAIL: std::time::Duration = Duration::new(300, 0);
+const TWAIT_ZERO: std::time::Duration = Duration::new(0, 0);
+// 100GiB = 100 * 1024 * 1024 * 1024 bytes
+const VOL_SIZE_100GIB: u64 = 107374182400;
+
+#[derive(Serialize, Deserialize, Debug, Clone, Copy, Eq, PartialEq)]
+/// SystemPerf policy, optimized for rebuild performance.
+pub struct SystemPerf {}
+#[derive(Serialize, Deserialize, Debug, Clone, Copy, Eq, PartialEq)]
+/// SystemAvail policy, optimized to maintain replica redundancy.
+pub struct SystemAvail {}
+
+/// Set of policies that internally define their rules regarding
+/// partial rebuild feasibility.
+#[derive(Serialize, Deserialize, Debug, PartialEq, Eq, Clone, Copy)]
+pub enum RuleSet {
+    SystemPerf(SystemPerf),
+    SystemAvail(SystemAvail),
+}
+
+impl RuleSet {
+    /// Returns a duration value. The caller can use this duration
+    /// to timeout between two time Instants, particularly to wait
+    /// upon a faulted child to possibly be healthy again.
+    pub fn faulted_child_wait(nexus: &Nexus) -> Duration {
+        let _pol = Self::assign(nexus.size);
+        Self::default_faulted_child_timewait()
+
+        // TODO: Enable when policies are dynamic
+        /* match pol {
+            RuleSet::SystemAvail(pol) => pol.faulted_child_wait_duration(nexus),
+            RuleSet::SystemPerf(pol) => pol.faulted_child_wait_duration(nexus),
+        } */
+    }
+
+    /// Assign a rebuild policy to the nexus.
+    fn assign(size: u64) -> Self {
+        match size > VOL_SIZE_100GIB {
+            true => RuleSet::SystemPerf(SystemPerf {}),
+            false => RuleSet::default(),
+        }
+    }
+
+    fn default_faulted_child_timewait() -> Duration {
+        tracing::info!("default timed-wait TWAIT_SPERF(10 mins)");
+        TWAIT_SPERF
+    }
+}
+
+impl Default for RuleSet {
+    fn default() -> Self {
+        RuleSet::SystemAvail(SystemAvail {})
+    }
+}
+
+trait Policy {
+    fn faulted_child_wait_duration(&self, nexus: &Nexus) -> Duration;
+    fn faulted_child_state_allows_wait(&self, nexus: &Nexus) -> bool;
+}
+
+impl Policy for SystemPerf {
+    fn faulted_child_wait_duration(&self, _nexus: &Nexus) -> Duration {
+        TWAIT_SPERF
+    }
+
+    fn faulted_child_state_allows_wait(&self, _nexus: &Nexus) -> bool {
+        todo!()
+    }
+}
+
+impl Policy for SystemAvail {
+    fn faulted_child_wait_duration(&self, nexus: &Nexus) -> Duration {
+        let mut online_child_cnt = 0;
+        for _ in nexus.children.iter().filter(|c| c.state.online()) {
+            online_child_cnt += 1;
+        }
+        if online_child_cnt <= 1 {
+            // immediate full rebuild
+            return TWAIT_ZERO;
+        }
+
+        TWAIT_SAVAIL
+    }
+
+    fn faulted_child_state_allows_wait(&self, _nexus: &Nexus) -> bool {
+        todo!()
+    }
+}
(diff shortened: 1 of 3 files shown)


dsharma-dc:
bors try

control-plane/stor-port/src/types/v0/store/nexus.rs
```diff
@@ -107,6 +108,8 @@ pub struct NexusSpec {
     #[serde(default)]
     /// Hosts allowed to access the nexus.
     pub allowed_hosts: Vec<HostNqn>,
+    /// rebuild policy for this nexus
+    pub rpolicy: RuleSet,
```
tiagolobocastro: Let's not add these in the objects until we're set on how the policy should look like, otherwise these may be breaking changes.
We can take a default policy for now.

control-plane/stor-port/src/types/v0/transport/policies.rs
```diff
@@ -0,0 +1,86 @@
...
+// SystemAvail policy
+pub struct SystemAvail {}
+
+#[derive(Serialize, Deserialize, Debug, PartialEq, Eq, Clone, Copy)]
+pub enum RuleSet {
+    SystemPerf(SystemPerf),
+    SystemAvail(SystemAvail),
+}
+
+impl RuleSet {
```
tiagolobocastro: The logic should probably live in the core agent itself?

control-plane/stor-port/src/types/v0/transport/policies.rs
```diff
@@ -0,0 +1,86 @@
...
+    // returning the absolute Instant to compare with.
+    pub fn faulted_child_wait(&self, nexus: &Nexus) -> Duration {
+        match self {
+            RuleSet::SystemAvail(pol) => pol.faulted_child_wait_duration(nexus),
+            RuleSet::SystemPerf(pol) => pol.faulted_child_wait_duration(nexus),
+        }
+    }
+
+    pub fn assign(size: u64) -> Self {
+        match size > VOL_SIZE_100GIB {
```
tiagolobocastro: I thought we decided on the hl meeting not to do policies based on anything yet (just add some basic scaffolding), has this been revised?

dsharma-dc: Not revised, I just sanity tested it for both cases this way. For now I'll modify assign() call to take only default() path.

control-plane/stor-port/src/types/v0/store/nexus.rs
dsharma-dc: ok. Commented the field with a TODO tag. The get call for duration will return default 10mins for now.

control-plane/stor-port/src/types/v0/transport/policies.rs
dsharma-dc: Are you suggesting that policies.rs be moved under agents/core hierarchy somewhere? Perhaps we can, I'm not too clear on control-plane code organisation yet.

dsharma-dc:
bors try

control-plane/stor-port/src/types/v0/transport/policies.rs
tiagolobocastro: Yes, indeed. Perhaps we need a module in core for policies. On one of my PR's (perhaps in draft or merged can't remember), I added a policy for the volume scheduling, but that was specific for scheduling new replicas and thus in the scheduling module..
Maybe controller/policies/{mod.rs, nexus/mod.rs}, something like this?@Abhinandan-Purkait any suggestions?

Abhinandan-Purkait: Yeah, that sounds good. Since we have started to have policy driven approaches and that will only increase in future, better to have seperate modules.

control-plane/grpc/src/operations/nexus/traits.rs
```diff
@@ -370,6 +370,7 @@ impl TryFrom<nexus::NexusSpec> for NexusSpec {
                 .into_iter()
                 .map(TryInto::try_into)
                 .collect::<Result<_, _>>()?,
+            //rpolicy: RuleSet::assign(value.size),
```
Abhinandan-Purkait: Let's just remove the commented code and add the actual code whenever we need to.

control-plane/stor-port/src/types/v0/transport/policies.rs
```diff
@@ -0,0 +1,97 @@
...
+// time constants, in seconds
+const TWAIT_SPERF: std::time::Duration = Duration::new(600, 0);
+const TWAIT_SAVAIL: std::time::Duration = Duration::new(300, 0);
+const TWAIT_ZERO: std::time::Duration = Duration::new(0, 0);
+// 100GiB = 100 * 1024 * 1024 * 1024 bytes
+const VOL_SIZE_100GIB: u64 = 107374182400;
+
+#[derive(Serialize, Deserialize, Debug, Clone, Copy, Eq, PartialEq)]
+// SystemPerf policy
+pub struct SystemPerf {}
```
Abhinandan-Purkait: Doc comments start with `///`. The format for doc comments are. `Should start with capital end with .`

control-plane/stor-port/src/types/v0/transport/policies.rs
```diff
@@ -0,0 +1,97 @@
...
+#[derive(Serialize, Deserialize, Debug, Clone, Copy, Eq, PartialEq)]
+// SystemAvail policy
+pub struct SystemAvail {}
```
Abhinandan-Purkait: Doc comments, needed as above.

control-plane/stor-port/src/types/v0/transport/policies.rs
```diff
@@ -0,0 +1,97 @@
...
+
+#[derive(Serialize, Deserialize, Debug, PartialEq, Eq, Clone, Copy)]
+pub enum RuleSet {
```
Abhinandan-Purkait: Doc comments needed.

control-plane/stor-port/src/types/v0/transport/policies.rs
dsharma-dc: So while rearranging this, realised that it'll be a problem to move everything into core agent because when we add a new RuleSet type field to Nexus, it'll not be identified in store-port.
So either:
1. Need to split the definitions and implementation. Keeping definitions in stor-port/transport, and implementation in core-agent.
2. OR, not binding the policy to nexus statically, rather call assign() every time a child goes faulted and we need to get the timed-wait value. It's not a costly function call so should be ok. 

Exploring option 2.

dsharma-dc: Updated the PR with 2.) as mentioned above.
The assign() call exists in the faulted_child_wait() path but we don't use it for returning value now. We'll simply return default 10mins. Once we decide to handle based on size and other factors, we can just remove the default path.

dsharma-dc:
bors try

abhilashshetty04 (approved):

dsharma-dc:
Testing done:

1. Provision a volume smaller than VOL_SIZE_100GIB, and initiate fault and rebuild path. Check logs that SystemAvail policy is chosen (but default value returned for now) - 

> 2023-03-03T18:23:27.689488Z  INFO core::controller::policies::rebuild_policies: Policy SystemAvail TWAIT_AVAIL(5 mins)
2023-03-03T18:23:27.689410Z  INFO core::controller::policies::rebuild_policies: default timed-wait TWAIT_SPERF(10 mins)

2. Provision a volume larger than VOL_SIZE_100GIB, and initiate fault and rebuild path. Check logs that SystemPerf policy is chosen (but default value returned for now) -

> 2023-03-03T18:26:27.691801Z  INFO core::controller::policies::rebuild_policies: Policy SystemPerf TWAIT_SPERF(10 mins)
2023-03-03T18:26:27.691862Z  INFO core::controller::policies::rebuild_policies: default timed-wait TWAIT_SPERF(10 mins)

Abhinandan-Purkait (approved):

control-plane/agents/src/bin/core/controller/policies/rebuild_policies.rs
```diff
@@ -0,0 +1,107 @@
...
+    fn faulted_child_state_allows_wait(&self, nexus: &Nexus) -> bool;
+}
+
+impl Policy for SystemPerf {
+    fn faulted_child_wait_duration(&self, _nexus: &Nexus) -> Duration {
+        TWAIT_SPERF
+    }
+
+    fn faulted_child_state_allows_wait(&self, _nexus: &Nexus) -> bool {
+        todo!()
```
Abhinandan-Purkait, line 86: nit: Instead of `todo!()` you can use `unimplemented!()`

dsharma-dc: I'd think we should prefer todo here because we would be implementing this in future. Moreover, unimplemented panics as per documentation:

> Indicates unimplemented code by panicking with a message of “not implemented”.
[...]
> The difference between unimplemented! and [todo!](https://doc.rust-lang.org/std/macro.todo.html) is that while todo! conveys an intent of implementing the functionality later and the message is “not yet implemented”, unimplemented! makes no such claims

dsharma-dc:
bors merge

control-plane/agents/src/bin/core/controller/policies/rebuild_policies.rs
Abhinandan-Purkait: That's not true, even `todo!()` will panic. `unimplemented` is always better as in code which we have pushed to organisation level and todo is good for own code. But again its subjective to person to person.

PR #451, opened 2023-03-09
fix(ha): nvme path replacement timeout

mtzaurus:
Currently HA node agent waits indefinitely till new path is connected while replacing a broken NVMe path, which can lead to gRPC timeouts on HA Cluster agent and may leave dangling NVMe controllers in the system. The fix introduces a tuneable timeout with the default value of 4 seconds.

### DIFF
--- a/control-plane/agents/src/bin/ha/node/server.rs
+++ b/control-plane/agents/src/bin/ha/node/server.rs
@@ -2,6 +2,7 @@ use crate::{
     csi_node_nvme_client,
     detector::{NvmeController, NvmePathCache},
     path_provider::get_nvme_path_entry,
+    Cli,
 };
 use agents::errors::{SvcError, SvcError::SubsystemNotFound};
 use common_lib::transport_api::{ErrorChain, ReplyError, ResourceKind};
@@ -24,7 +25,7 @@ use nvmeadm::nvmf_subsystem::SubsystemAddr;
 
 use http::Uri;
 use nvmeadm::nvmf_subsystem::Subsystem;
-use std::{collections::HashMap, net::SocketAddr, sync::Arc};
+use std::{collections::HashMap, net::SocketAddr, sync::Arc, time::Instant};
 use tokio::time::{sleep, Duration};
 use utils::NVME_TARGET_NQN_PREFIX;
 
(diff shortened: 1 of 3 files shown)


mtzaurus:
bors try

control-plane/agents/src/bin/ha/node/server.rs
```diff
@@ -229,6 +245,31 @@ impl NodeAgentSvc {
...
+
+            // Check if path reconnection timeout has elapsed.
+            if now.elapsed() >= timeout {
+                tracing::error!(
+                    new_path,
+                    nqn,
+                    "Timeout while connecting NVMe path, disconnecting NVMe subsystem"
+                );
+                // Forcibly disconnect subsystem incomplete subsystem in case of timeout.
+                if let Err(error) = subsystem.disconnect() {
```
Abhinandan-Purkait: There could be an issue here, if we are reusing the same nexus. Then the subsystem will have the older path and that would be the only path, and if we disconnect that we will not be able to switchover later.

control-plane/agents/src/bin/ha/node/server.rs
```diff
@@ -229,6 +245,31 @@ impl NodeAgentSvc {
...
+                    tracing::error!(
+                        ?subsystem,
+                        ?error,
+                        "NVMe subsystem successfully disconnected"
```
chriswldenyer: Should this tracing message be "not successfully disconnected" ?

control-plane/agents/src/bin/ha/node/server.rs
mtzaurus: Yes, indeed.

Abhinandan-Purkait: Yeah, we should probably skip the disconnect here if its the only path in the subsystem.

Abhinandan-Purkait (approved):
LGTM. Thanks

mtzaurus:
bors try

mtzaurus:
bors try

chriswldenyer (approved):
Looks reasonable.

mtzaurus:
bors merge

PR #452, opened 2023-03-09
chore(test/rebuild_policy): add component test for policy api

dsharma-dc:
Simple cargo test for policy API. Will be enhanced as the policies are enriched.

### DIFF
--- a/control-plane/agents/src/bin/core/controller/policies/rebuild_policies.rs
+++ b/control-plane/agents/src/bin/core/controller/policies/rebuild_policies.rs
@@ -1,8 +1,8 @@
-#![allow(unused)]
-use stor_port::types::v0::transport::Nexus;
-
+#[allow(unused)]
 use serde::{Deserialize, Serialize};
+
 use std::time::Duration;
+use stor_port::types::v0::transport::Nexus;
 
 /*
     This file defines the policies that control plane uses to make
@@ -16,11 +16,12 @@ use std::time::Duration;
 */
 
 // time constants, in seconds
-const TWAIT_SPERF: std::time::Duration = Duration::new(600, 0);
-const TWAIT_SAVAIL: std::time::Duration = Duration::new(300, 0);
-const TWAIT_ZERO: std::time::Duration = Duration::new(0, 0);
+pub const TWAIT_SPERF: std::time::Duration = Duration::new(600, 0);
+pub const TWAIT_SAVAIL: std::time::Duration = Duration::new(300, 0);
+pub const TWAIT_ZERO: std::time::Duration = Duration::new(0, 0);
 // 100GiB = 100 * 1024 * 1024 * 1024 bytes
-const VOL_SIZE_100GIB: u64 = 107374182400;
+#[allow(unused)]
+pub const VOL_SIZE_100GIB: u64 = 107374182400;
 
 #[derive(Serialize, Deserialize, Debug, Clone, Copy, Eq, PartialEq)]
 /// SystemPerf policy, optimized for rebuild performance.
@@ -41,6 +42,7 @@ impl RuleSet {
     /// Returns a duration value. The caller can use this duration
     /// to timeout between two time Instants, particularly to wait
     /// upon a faulted child to possibly be healthy again.
+    #[allow(dead_code)]
     pub fn faulted_child_wait(nexus: &Nexus) -> Duration {
         let _pol = Self::assign(nexus.size);
         Self::default_faulted_child_timewait()
@@ -53,6 +55,7 @@ impl RuleSet {
     }
 
     /// Assign a rebuild policy to the nexus.
+    #[allow(dead_code)]
     fn assign(size: u64) -> Self {
         match size > VOL_SIZE_100GIB {
             true => RuleSet::SystemPerf(SystemPerf {}),
@@ -60,6 +63,7 @@ impl RuleSet {
         }
     }
 
+    #[allow(dead_code)]
     fn default_faulted_child_timewait() -> Duration {
         tracing::info!("default timed-wait TWAIT_SPERF(10 mins)");
         TWAIT_SPERF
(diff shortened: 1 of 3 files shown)


abhilashshetty04 (approved):

control-plane/agents/src/bin/core/controller/policies/rebuild_policies.rs
```diff
@@ -1,8 +1,8 @@
-#![allow(unused)]
-use stor_port::types::v0::transport::Nexus;
-
+#[allow(unused)]
```
tiagolobocastro, line 1: why is there so much dead code around?

control-plane/agents/src/bin/core/controller/policies/rebuild_policies.rs
```diff
@@ -16,11 +16,12 @@ use std::time::Duration;
 */
 
 // time constants, in seconds
-const TWAIT_SPERF: std::time::Duration = Duration::new(600, 0);
-const TWAIT_SAVAIL: std::time::Duration = Duration::new(300, 0);
-const TWAIT_ZERO: std::time::Duration = Duration::new(0, 0);
+pub const TWAIT_SPERF: std::time::Duration = Duration::new(600, 0);
```
tiagolobocastro, line 19: please fix doc comments above

control-plane/agents/src/bin/core/controller/policies/rebuild_policies.rs
```diff
@@ -16,11 +16,12 @@ use std::time::Duration;
...
+pub const TWAIT_SAVAIL: std::time::Duration = Duration::new(300, 0);
+pub const TWAIT_ZERO: std::time::Duration = Duration::new(0, 0);
```
tiagolobocastro, line 21: what do these mean, the name is not very descriptive, at the very least need some comments.

control-plane/agents/src/bin/core/controller/policies/rebuild_policies.rs
```diff
@@ -53,13 +55,15 @@ impl RuleSet {
...
     /// Assign a rebuild policy to the nexus.
+    #[allow(dead_code)]
     fn assign(size: u64) -> Self {
         match size > VOL_SIZE_100GIB {
             true => RuleSet::SystemPerf(SystemPerf {}),
             false => RuleSet::default(),
         }
     }
 
+    #[allow(dead_code)]
```
tiagolobocastro, line 66: can't we just add there when we need them rather than keep it as dead code?

control-plane/agents/src/bin/core/tests/policies/mod.rs
```diff
@@ -0,0 +1,61 @@
...
+This file contains tests for any policies that are present
+in the control plane.
+*/
+use deployer_cluster::{Cluster, ClusterBuilder};
+use grpc::operations::{nexus::traits::NexusOperations, node::traits::NodeOperations};
+
+use stor_port::types::v0::transport::{CreateNexus, Filter, Nexus, NexusId};
+
+mod pmodule {
+    include!("../../controller/policies/rebuild_policies.rs");
```
tiagolobocastro, line 13: why is this being included like this?

control-plane/agents/src/bin/core/tests/policies/mod.rs
```diff
@@ -0,0 +1,61 @@
+#![allow(unused)]
```
tiagolobocastro, line 1: why are we allowing unused on a test file?

control-plane/agents/src/bin/core/tests/policies/mod.rs
```diff
@@ -0,0 +1,61 @@
...
+
+/**
```
tiagolobocastro, line 3: please use regular doc comments

tiagolobocastro (commented):
Not related to this PR but:
```rust
trait Policy {
    fn faulted_child_wait_duration(&self, nexus: &Nexus) -> Duration;
    fn faulted_child_state_allows_wait(&self, nexus: &Nexus) -> bool;
}
```
Can we doc comment the trait and the methods? It's not immediately obvious to me how that's meant to be used.

control-plane/agents/src/bin/core/tests/policies/mod.rs
dsharma-dc: Apparently, spending couple of hours, it didn't seem feasible to use the types from rebuild_policies.rs in the tests/ by any other means :( . Every other module outside  `tests` is able to see rebuild_policies module, but not `tests`. I'm all ears if there is a possible better way. This is possibly the reason I ended up having to mitigate lot of unused/dead code warnings. The product(non-test) code that's going to use rebuild_policies.rs isn't merged yet.

dsharma-dc: The reason appears to be that the file being included has unused attribute at module level(inner attribute), and it's being imported into a different module so compiler complains about attribute validity. I'll likely wait till the attribute can be removed from included file, i.e. when Abhilash's changes are ready.

control-plane/agents/src/bin/core/controller/policies/rebuild_policies.rs
abhilashshetty04: This is probably because policy functions are going to be used by reconciler changes that i am working on.

control-plane/agents/src/bin/core/controller/policies/rebuild_policies.rs
tiagolobocastro: nit: this file can simply be called rebuild.rs since we're already within policies.

control-plane/agents/src/bin/core/controller/policies/rebuild_policies.rs
tiagolobocastro: Ah right

control-plane/agents/src/bin/core/controller/policies/rebuild_policies.rs
```diff
@@ -53,13 +55,15 @@ impl RuleSet {
     }
 
     /// Assign a rebuild policy to the nexus.
+    #[allow(dead_code)]
```
tiagolobocastro, line 58: s/assign/new

PR #453, opened 2023-03-10
feat(volume group): load volume group specs in memory in startup

Abhinandan-Purkait:


control-plane/stor-port/src/types/v0/store/volume.rs
```diff
@@ -591,3 +591,42 @@ impl From<VolumeSpec> for models::VolumeSpec {
...
+    pub fn name(&self) -> &String {
+        &self.volume_group_name
+    }
+    /// Add a new volume to the list of volumes.
+    pub fn append_volume(&mut self, id: VolumeId) {
+        self.volumes.push(id)
+    }
+    /// Add a new volume to the list of volumes.
+    pub fn remove_volume(&mut self, id: VolumeId) {
+        self.volumes.retain(|vol| vol.clone() != id)
```
tiagolobocastro: why clone?

control-plane/stor-port/src/types/v0/store/volume.rs
```diff
@@ -591,3 +591,42 @@ impl From<VolumeSpec> for models::VolumeSpec {
         )
     }
 }
+
+/// In memory representation of volume group.
+#[derive(Debug, Clone, PartialEq, Default)]
+pub struct VolumeGroupSpec {
+    /// Name of the volume group.
+    volume_group_name: String,
```
```suggestion
    name: String,
```

tiagolobocastro: Should we have a type rather than just String?

control-plane/agents/src/bin/core/controller/resources/operations_helper.rs
```diff
@@ -937,3 +942,23 @@ impl ResourceSpecsLocked {
...
+fn get_volume_group_specs(volume_specs: &[VolumeSpec]) -> Vec<VolumeGroupSpec> {
+    let mut volume_group_specs: Vec<VolumeGroupSpec> = Vec::new();
+    for volume_spec in volume_specs.iter() {
+        if let Some(volume_group) = &volume_spec.volume_group {
+            let volume_group_spec =
+                VolumeGroupSpec::new(volume_group.name(), vec![volume_spec.uuid.clone()]);
+            let existing_index = volume_group_specs
+                .iter()
+                .position(|x| x.name() == volume_group_spec.name());
+            if let Some(index) = existing_index {
```
tiagolobocastro: why not simply use find_mut ?

control-plane/stor-port/src/types/v0/store/volume.rs
Abhinandan-Purkait: `self.volumes.retain(|vol| vol != &id)` How about this?

control-plane/stor-port/src/types/v0/store/volume.rs
Abhinandan-Purkait: We can but would there be any advantages? It would be always a string I guess?

control-plane/agents/src/bin/core/controller/resources/operations_helper.rs
Abhinandan-Purkait: Good point.

control-plane/stor-port/src/types/v0/store/volume.rs
tiagolobocastro: You can use eg type VolumeGroupId = String;
Then if we ever need to add validation for example, we simply replace that with the actual definition we want.

control-plane/stor-port/src/types/v0/store/volume.rs
tiagolobocastro: Yep that's it!

control-plane/stor-port/src/types/v0/store/volume.rs
Abhinandan-Purkait: Sounds good. Thanks.

Abhinandan-Purkait: Do we need a openapi-spec entry for the type so that the rest layer passes of that type or is it fine in the control-plane level?

tiagolobocastro: Hmm, I guess so?

tiagolobocastro: I might be inclined to leave that for another PR and maybe we need to decide if we really need it.

control-plane/agents/src/bin/core/controller/resources/operations_helper.rs
Abhinandan-Purkait: Done. But time complexity is same though.

control-plane/stor-port/src/types/v0/store/volume.rs
Abhinandan-Purkait: Right.

control-plane/agents/src/bin/core/controller/resources/operations_helper.rs
tiagolobocastro: Not for performance, just to avoid avoid to index since we can get the object straight away.

Abhinandan-Purkait: Ohh, Right.

control-plane/stor-port/src/types/v0/store/volume.rs
```diff
@@ -591,3 +591,45 @@ impl From<VolumeSpec> for models::VolumeSpec {
...
+    /// Name of the volume group.
+    pub fn name(&self) -> &VolumeGroupId {
+        &self.name
+    }
+    /// Add a new volume to the list of volumes.
+    pub fn append_volume(&mut self, id: VolumeId) {
+        self.volumes.push(id)
+    }
+    /// Add a new volume to the list of volumes.
+    pub fn remove_volume(&mut self, id: VolumeId) {
```
tiagolobocastro: suggest passing &VolumeId instead, maybe on append too?

control-plane/stor-port/src/types/v0/transport/volume.rs
```diff
@@ -359,16 +362,16 @@ pub struct CreateVolume {
...
 
 impl VolumeGroup {
     /// Create a new VolumeGroup from the params.
-    pub fn new(name: String) -> Self {
+    pub fn new(name: VolumeGroupId) -> Self {
         Self { name }
     }
     /// The name of the volume group.
-    pub fn name(&self) -> String {
+    pub fn name(&self) -> VolumeGroupId {
```
tiagolobocastro: return reference instead

control-plane/stor-port/src/types/v0/store/volume.rs
Abhinandan-Purkait: Okay.

control-plane/stor-port/src/types/v0/store/volume.rs
```diff
@@ -591,3 +591,45 @@ impl From<VolumeSpec> for models::VolumeSpec {
...
 }
+
+/// VolumeGroupId type.
+pub type VolumeGroupId = String;
+
+/// In memory representation of volume group.
+#[derive(Debug, Clone, PartialEq, Default)]
+pub struct VolumeGroupSpec {
+    /// Name of the volume group.
+    name: VolumeGroupId,
```
tiagolobocastro: should this be id and not name? If so also fixup getter method,.

Abhinandan-Purkait: I initially thought of id. But its not really an id right? by id I mean uuid format. What do you think, is id a better term here? Also is there a future possibility of vg uuid?

Abhinandan-Purkait: The same way we have pool name and pool uuid. In those terms

control-plane/stor-port/src/types/v0/store/volume.rs
Abhinandan-Purkait: Done.

control-plane/stor-port/src/types/v0/transport/volume.rs
Abhinandan-Purkait: Done.

control-plane/stor-port/src/types/v0/store/volume.rs
tiagolobocastro: Id is just an Id, UUID is different

Abhinandan-Purkait: I see, alright. Will change the names to id.

tiagolobocastro: Hmm we might want a uuid indeed, in fact I'm wondering if we should have a uuid though it would make it harder for csi layer since the vg id is created based on funky pvc and ss logic.

Abhinandan-Purkait: Yeah. Right.

control-plane/stor-port/src/types/v0/store/volume.rs
```diff
@@ -591,3 +591,44 @@ impl From<VolumeSpec> for models::VolumeSpec {
...
+            volumes,
+            sequencer: OperationSequence::new(id),
+        }
+    }
+    /// Name of the volume group.
+    pub fn id(&self) -> &VolumeGroupId {
+        &self.id
+    }
+    /// Add a new volume to the list of volumes.
+    pub fn append_volume(&mut self, id: VolumeId) {
```
```suggestion
    pub fn append(&mut self, id: VolumeId) {
```

control-plane/stor-port/src/types/v0/store/volume.rs
```diff
@@ -591,3 +591,44 @@ impl From<VolumeSpec> for models::VolumeSpec {
...
+        self.volumes.push(id)
+    }
+    /// Add a new volume to the list of volumes.
+    pub fn remove_volume(&mut self, id: &VolumeId) {
```
```suggestion
    pub fn remove(&mut self, id: &VolumeId) {
```

control-plane/stor-port/src/types/v0/store/volume.rs
```diff
@@ -591,3 +591,44 @@ impl From<VolumeSpec> for models::VolumeSpec {
...
+        self.volumes.retain(|vol| vol != id)
+    }
+    /// Check if the volume group has any more volumes.
+    pub fn volume_list_empty(&self) -> bool {
```
```suggestion
    pub fn is_empty(&self) -> bool {
```

tiagolobocastro (approved):

abhilashshetty04 (approved):

Abhinandan-Purkait:
bors merge

PR #454, opened 2023-03-12
feat(core_reconciler): add partial rebuild support in control plane

abhilashshetty04:
This PR adds Partial/Log-based rebuild support in control plane. Currently control pane removes Faulted child and adds new child to nexus children list. Post which nexus uses one of the healthy child to Full rebuild the empty child.

With this change depending on number of healthy replicas in nexus we wait for the time defined by the rebuild policy. If replica comes back within timed wait duration then we mark child as Online after which nexus performs log based/ partial rebuild of the child. 

If after TimedWait period child doesnt come back we do Full rebuild of the child.

### DIFF
--- a/control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
+++ b/control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
@@ -147,9 +147,7 @@ pub(super) async fn handle_faulted_children(
     let child_count = nexus_state.children.len();
     if nexus_state.status == NexusStatus::Degraded && child_count > 1 {
         for child in nexus_state.children.iter().filter(|c| c.state.faulted()) {
-            if child.state_reason != ChildStateReason::NoSpace {
-                let _ = handle_child_rebuild(nexus, child, &nexus_state, context).await;
-            }
+            let _ = handle_child_rebuild(nexus, child, &nexus_state, context).await;
         }
     }
     Ok(PollerState::Idle)
--- a/control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
+++ b/control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
@@ -193,12 +193,12 @@ pub(super) async fn faulted_children_remover(
     let nexus_uuid = nexus.uuid();
     let nexus_state = context.registry().nexus(nexus_uuid).await?;
 
-    info!(%child, "Attempting to remove faulted child '{}'", child);
+    info!(%child, "Attempting to remove faulted child");
     nexus
         .remove_child_by_uri(context.registry(), &nexus_state, child, true)
         .await?;
 
-    info!(%child ,"Successfully removed faulted child",);
+    info!(%child ,"Successfully removed faulted child");
     Ok(())
 }
 
@@ -258,14 +258,14 @@ pub(super) async fn initialize_partial_rebuild(
     context: &PollContext,
 ) -> Result<(), SvcError> {
     let nexus_state = context.registry().nexus(nexus).await?;
-    info!("Making {:?} online of {:?} nexus", child, nexus);
+    info!(%child, "Making online of {:?} nexus", nexus);
     let node = context.registry().node_wrapper(&nexus_state.node).await?;
     let online_context = NexusChildActionContext::new(&nexus_state.node, &nexus.clone(), child);
     let ctx = NexusChildActionContextNode::new(online_context, context.registry());
     // TODO: Check if we can handle things differently for different SvcError.
     node.online_child(ctx).await?;
     info!(
-        %child, %nexus,"Successfully made child online"
+        %child, "Successfully made child online in nexus :{:?}", nexus
     );
     Ok(())
 }
--- a/control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
+++ b/control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
@@ -19,6 +19,7 @@ use crate::{
         wrapper::NodeWrapper,
     },
     nexus::scheduling::healthy_nexus_children,
+    node::wrapper::NexusChildActionContextNode,
 };
 use agents::errors::SvcError;
 use capacity::enospc_children_onliner;
@@ -43,7 +44,6 @@ use stor_port::{
 };
 use tokio::sync::RwLock;
 use tracing::{info, Instrument};
-use crate::node::wrapper::NexusChildActionContextNode;
 /// Nexus Reconciler loop
 #[derive(Debug)]
 pub(crate) struct NexusReconciler {
--- a/control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
+++ b/control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
@@ -3,7 +3,8 @@ mod garbage_collector;
 
 use crate::{
     controller::{
-        io_engine::NexusApi,
+        io_engine::{NexusApi, NexusChildActionApi},
+        policies::rebuild_policies::RuleSet,
         reconciler::{ReCreate, Reconciler},
         resources::{
             operations::ResourceSharing,
@@ -22,23 +23,27 @@ use crate::{
 use agents::errors::SvcError;
 use capacity::enospc_children_onliner;
 use garbage_collector::GarbageCollector;
+use std::{
+    convert::TryFrom,
+    sync::Arc,
+    time::{Duration, SystemTime},
+};
 use stor_port::{
-    transport_api::ErrorChain,
+    transport_api::{ErrorChain, ResourceKind},
     types::v0::{
         store::{
             nexus::{NexusSpec, ReplicaUri},
             nexus_child::NexusChild,
         },
         transport::{
-            CreateNexus, NexusShareProtocol, NexusStatus, NodeStatus, ShareNexus, UnshareNexus,
+            Child, ChildStateReason, ChildUri, CreateNexus, NexusChildActionContext, NexusId,
+            NexusShareProtocol, NexusStatus, NodeStatus, Replica, ShareNexus, UnshareNexus,
         },
     },
 };
-
-use std::{convert::TryFrom, sync::Arc};
 use tokio::sync::RwLock;
-use tracing::Instrument;
-
+use tracing::{info, Instrument};
+use crate::node::wrapper::NexusChildActionContextNode;
 /// Nexus Reconciler loop
 #[derive(Debug)]
 pub(crate) struct NexusReconciler {
@@ -116,7 +121,7 @@ async fn nexus_reconciler(
 
     if reconcile {
         match squash_results(vec![
-            faulted_children_remover(nexus, context).await,
+            handle_faulted_child(nexus, context).await,
             unknown_children_remover(nexus, context).await,
             missing_children_remover(nexus, context).await,
             fixup_nexus_protocol(nexus, context).await,
@@ -130,61 +135,139 @@ async fn nexus_reconciler(
     }
 }
 
-/// Find and removes faulted children from the given nexus
-/// If the child is a replica it also disowns and destroys it
+/// Checks if nexus is Degraded and any child is Faulted. If yes, Depending on rebuild policy for
+/// child it performs rebuild operation. We exclude NoSpace Degrade.
 #[tracing::instrument(skip(nexus, context), level = "trace", fields(nexus.uuid = %nexus.uuid(), request.reconcile = true))]
-pub(super) async fn faulted_children_remover(
+pub(super) async fn handle_faulted_child(
     nexus: &mut OperationGuardArc<NexusSpec>,
     context: &PollContext,
 ) -> PollResult {
     let nexus_uuid = nexus.uuid();
     let nexus_state = context.registry().nexus(nexus_uuid).await?;
     let child_count = nexus_state.children.len();
-
-    // Remove faulted children only from a degraded nexus with other healthy children left
     if nexus_state.status == NexusStatus::Degraded && child_count > 1 {
-        let span = tracing::info_span!("faulted_children_remover", nexus.uuid = %nexus_uuid, request.reconcile = true);
-        async {
-            let nexus_spec_clone = nexus.lock().clone();
-            for child in nexus_state.children.iter().filter(|c| c.state.faulted()) {
-                nexus_spec_clone.warn_span(|| {
-                    tracing::warn!(%child.state, %child.state_reason, "Attempting to remove faulted child '{}'", child.uri)
-                });
-                if let Err(error) = nexus
-                    .remove_child_by_uri(
-                        context.registry(),
-                        &nexus_state,
-                        &child.uri,
-                        true,
-                    )
-                    .await
+        let nexus_spec_clone = nexus.lock().clone();
+        for child in nexus_state.children.iter().filter(|c| c.state.faulted()) {
+            if child.state_reason != ChildStateReason::NoSpace {
+                let wait_time = RuleSet::faulted_child_wait(&nexus_state, context.registry());
+                if wait_time.is_zero()
+                    || child.state_reason.clone() == ChildStateReason::RebuildFailed
                 {
-                    nexus_spec_clone.error_span(|| {
-                        tracing::error!(
-                            error = error.full_string().as_str(),
-                            %child.state,
-                            %child.state_reason,
-                            child.uri = %child.uri.as_str(),
-                            "Failed to remove faulted child"
-                        )
-                    });
+                    info!(%child.uri, "Start full rebuild for child");
+                    faulted_children_remover(nexus, &child.uri, context).await?
+                } else if is_time_elapsed(child.faulted_at, Some(wait_time), &child.uri).await {
+                    info!(
+                        %child.uri, "Start full rebuild for child,  Wait time has elapsed"
+
+                    );
+                    faulted_children_remover(nexus, &child.uri, context).await?;
+                } else if get_child_replica(nexus_spec_clone.clone(), &child.clone(), context)
+                    .await
+                    .is_err()
+                {
+                    info!(%child.uri,"Replica is still not online for child");
                 } else {
-                    nexus_spec_clone.info_span(|| {
-                        tracing::info!(
-                            %child.uri,
-                            %child.state,
-                            %child.state_reason,
-                            "Successfully removed faulted child",
-                        )
-                    });
+                    info!(
+                        %child.uri, "Replica for child is Online, starting partial rebuild"
+                    );
+                    if let Err(error) =
+                        initialize_partial_rebuild(nexus.uuid(), &child.uri, context).await
+                    {
+                        info!(%child.uri, "Initializing Full rebuild as partial rebuild failed for child, failed with : {:?}", error);
+                        faulted_children_remover(nexus, &child.uri, context).await?
+                    }
                 }
             }
         }
-        .instrument(span)
-        .await
     }
+    Ok(PollerState::Idle)
+}
 
-    PollResult::Ok(PollerState::Idle)
+/// Removes child from nexus children list to initiate Full rebuild of child.
+#[tracing::instrument(skip(nexus, context), level = "trace", fields(nexus.uuid = %nexus.uuid(), request.reconcile = true))]
+pub(super) async fn faulted_children_remover(
+    nexus: &mut OperationGuardArc<NexusSpec>,
+    child: &ChildUri,
+    context: &PollContext,
+) -> Result<(), SvcError> {
+    let nexus_uuid = nexus.uuid();
+    let nexus_state = context.registry().nexus(nexus_uuid).await?;
+
+    info!(%child, "Attempting to remove faulted child '{}'", child);
+    nexus
+        .remove_child_by_uri(context.registry(), &nexus_state, child, true)
+        .await?;
+
+    info!(%child ,"Successfully removed faulted child",);
+    Ok(())
+}
+
+/// Returns true if time elapsed from child fault time is greater or equal to wait time duration.
+async fn is_time_elapsed(
+    fault_time: Option<SystemTime>,
+    wait_time: Option<Duration>,
+    uri: &ChildUri,
+) -> bool {
+    if let Some(fault_time) = fault_time {
+        if let Some(wait_time) = wait_time {
+            if let Ok(elapsed) = fault_time.elapsed() {
+                info!(
+                    child.uri = %uri,
+                    "waited {:?} for child to comeback. Wait time is {:?}",
+                    elapsed, wait_time
+                );
+                return elapsed >= wait_time;
+            };
+        }
+    }
+    false
+}
+
+/// If we can get replica for a child from registry. It means that child node is online, Pool is
+/// imported containing the child/replica.
+async fn get_child_replica(
+    nexus: NexusSpec,
+    child: &Child,
+    context: &PollContext,
+) -> Result<Replica, SvcError> {
+    for nexus_child in nexus.children.iter() {
+        if let Some(replica) = nexus_child.as_replica() {
+            if replica.uri() == &child.uri {
+                return if let Ok(replica) = context.registry().replica(replica.uuid()).await {
+                    Ok(replica)
+                } else {
+                    Err(SvcError::ReplicaNotFound {
+                        replica_id: replica.uuid().clone(),
+                    })
+                };
+            } else {
+                continue;
+            }
+        }
+    }
+    Err(SvcError::NotFound {
+        kind: ResourceKind::Node,
+        id: "Child replica not found".to_string(),
+    })
+}
+
+/// Tries to mark child as Online. Returns error on failing to do so.
+pub(super) async fn initialize_partial_rebuild(
+    nexus: &NexusId,
+    child: &ChildUri,
+    context: &PollContext,
+) -> Result<(), SvcError> {
+    let nexus_state = context.registry().nexus(nexus).await?;
+    info!("Making {:?} online of {:?} nexus", child, nexus);
+    let node = context.registry().node_wrapper(&nexus_state.node).await?;
+    let online_context = NexusChildActionContext::new(&nexus_state.node, &nexus.clone(), child);
+    let ctx = NexusChildActionContextNode::new(online_context, context.registry());
+    // TODO: Check if we can handle things differently for different SvcError.
+    node.online_child(ctx).await?;
+    info!(
+        %child, %nexus,"Successfully made child online"
+    );
+    Ok(())
 }
 
 /// Find and removes unknown children from the given nexus
--- a/control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
+++ b/control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
@@ -149,13 +149,13 @@ pub(super) async fn handle_faulted_child(
         let nexus_spec_clone = nexus.lock().clone();
         for child in nexus_state.children.iter().filter(|c| c.state.faulted()) {
             if child.state_reason != ChildStateReason::NoSpace {
-                let wait_time = RuleSet::faulted_child_wait(&nexus_state, context.registry());
-                if wait_time.is_zero()
+                let wait_duration = RuleSet::faulted_child_wait(&nexus_state, context.registry());
+                if wait_duration.is_zero()
                     || child.state_reason.clone() == ChildStateReason::RebuildFailed
                 {
                     info!(%child.uri, "Start full rebuild for child");
                     faulted_children_remover(nexus, &child.uri, context).await?
-                } else if is_time_elapsed(child.faulted_at, Some(wait_time), &child.uri).await {
+                } else if is_time_elapsed(child.faulted_at, wait_duration, &child.uri).await {
                     info!(
                         %child.uri, "Start full rebuild for child,  Wait time has elapsed"
 
@@ -205,20 +205,18 @@ pub(super) async fn faulted_children_remover(
 /// Returns true if time elapsed from child fault time is greater or equal to wait time duration.
 async fn is_time_elapsed(
     fault_time: Option<SystemTime>,
-    wait_time: Option<Duration>,
+    wait_time: Duration,
     uri: &ChildUri,
 ) -> bool {
     if let Some(fault_time) = fault_time {
-        if let Some(wait_time) = wait_time {
-            if let Ok(elapsed) = fault_time.elapsed() {
-                info!(
-                    child.uri = %uri,
-                    "waited {:?} for child to comeback. Wait time is {:?}",
-                    elapsed, wait_time
-                );
-                return elapsed >= wait_time;
-            };
-        }
+        if let Ok(elapsed) = fault_time.elapsed() {
+            info!(
+                child.uri = %uri,
+                "waited {:?} for child to comeback. Wait time is {:?}",
+                elapsed, wait_time
+            );
+            return elapsed >= wait_time;
+        };
     }
     false
 }
--- a/control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
+++ b/control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
@@ -37,8 +37,9 @@ use stor_port::{
             nexus_child::NexusChild,
         },
         transport::{
-            Child, ChildStateReason, ChildUri, CreateNexus, NexusChildActionContext, NexusId,
-            NexusShareProtocol, NexusStatus, NodeStatus, Replica, ShareNexus, UnshareNexus,
+            Child, ChildStateReason, ChildUri, CreateNexus, Nexus, NexusChildActionContext,
+            NexusId, NexusShareProtocol, NexusStatus, NodeStatus, Replica, ShareNexus,
+            UnshareNexus,
         },
     },
 };
@@ -121,7 +122,7 @@ async fn nexus_reconciler(
 
     if reconcile {
         match squash_results(vec![
-            handle_faulted_child(nexus, context).await,
+            handle_faulted_children(nexus, context).await,
             unknown_children_remover(nexus, context).await,
             missing_children_remover(nexus, context).await,
             fixup_nexus_protocol(nexus, context).await,
@@ -138,7 +139,7 @@ async fn nexus_reconciler(
 /// Checks if nexus is Degraded and any child is Faulted. If yes, Depending on rebuild policy for
 /// child it performs rebuild operation. We exclude NoSpace Degrade.
 #[tracing::instrument(skip(nexus, context), level = "trace", fields(nexus.uuid = %nexus.uuid(), request.reconcile = true))]
-pub(super) async fn handle_faulted_child(
+pub(super) async fn handle_faulted_children(
     nexus: &mut OperationGuardArc<NexusSpec>,
     context: &PollContext,
 ) -> PollResult {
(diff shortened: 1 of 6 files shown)


abhilashshetty04:
@tiagolobocastro  Will handle all unwraps. WIP as of now.

control-plane/grpc/proto/v1/nexus/nexus.proto
```diff
@@ -108,6 +109,19 @@ message NexusSpec {
...
   repeated string allowed_hosts = 12;
+  map<string, RebuildInfo> rebuild_state = 13;
+}
+
+message RebuildInfo {
+  optional google.protobuf.Duration wait_period = 1;
+  RebuildStage stage = 2;
+}
+
+enum RebuildStage {
```
dsharma-dc: Need to put comments describing these stages.

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
```diff
@@ -126,6 +132,201 @@ async fn nexus_reconciler(
...
+    nexus: &mut OperationGuardArc<NexusSpec>,
+    context: &PollContext,
+) -> PollResult {
+    info!("entered initialize partial rebuild function");
+    let nexus_uuid = nexus.uuid().clone();
+    let nexus_state = context.registry().nexus(&nexus_uuid).await?;
+    let nexus_spec = context.specs().nexus(&nexus_uuid).await?;
+    let node = context.registry().node_wrapper(&nexus_state.node).await?;
+    let rebuild_state = nexus_spec.lock().rebuild_state.clone();
+    let nexus_name = nexus_spec.lock().clone().name;
```
dsharma-dc: Since we are using nexus _name just for tracing, can remove this and use nexus_uuid itself to avoid a lock call.

control-plane/stor-port/src/types/v0/store/nexus.rs
```diff
@@ -232,6 +255,13 @@ impl SpecTransaction<NexusOperation> for NexusSpec {
                 }
                 NexusOperation::AddChild(uri) => self.children.push(uri),
                 NexusOperation::RemoveChild(uri) => self.children.retain(|c| c != &uri),
+                NexusOperation::RebuildState(child_uri, stage) => {
+                    if let Some(stage) = stage {
```
dsharma-dc: Do you intend to check `stage.is_some()` here?

dsharma-dc (commented):
Cursory look. There is some cleanup required as well it seems.

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
abhilashshetty04: Done.

control-plane/stor-port/src/types/v0/store/nexus.rs
abhilashshetty04: Mostly yes. If i used is_some() we get bool and will have to do unwrap while inserting to hashmap. with if let Some(stage) this does unwrap if its Some or else we just delete key from the hashmap.

control-plane/grpc/proto/v1/nexus/nexus.proto
abhilashshetty04: Will do

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
```diff
@@ -130,49 +134,247 @@ async fn nexus_reconciler(
     }
 }
 
-/// Find and removes faulted children from the given nexus
-/// If the child is a replica it also disowns and destroys it
+/// Checks if nexus is Degraded and any child is Faulted. If yes, Depending on rebuild policy for
+/// child sets the rebuild stage in hashmap to drive the rebuild. We exclude NoSpace Degrade.
 #[tracing::instrument(skip(nexus, context), level = "trace", fields(nexus.uuid = %nexus.uuid(), request.reconcile = true))]
-pub(super) async fn faulted_children_remover(
+pub(super) async fn handle_faulted_child(
```
tiagolobocastro:
```suggestion
pub(super) async fn handle_faulted_children(
```

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
```diff
@@ -130,49 +134,247 @@ async fn nexus_reconciler(
...
     nexus: &mut OperationGuardArc<NexusSpec>,
     context: &PollContext,
 ) -> PollResult {
     let nexus_uuid = nexus.uuid();
     let nexus_state = context.registry().nexus(nexus_uuid).await?;
     let child_count = nexus_state.children.len();
+    if nexus_state.status == NexusStatus::Degraded && child_count > 1 {
+        let nexus_spec_clone = nexus.lock().clone();
```
tiagolobocastro: don't need to clone it

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
```diff
@@ -130,49 +134,247 @@ async fn nexus_reconciler(
...
+        for child in nexus_state.children.iter().filter(|c| c.state.faulted()) {
+            if !nexus_spec_clone.rebuild_state.contains_key(&child.uri)
+                && child.state_reason.clone() != ChildStateReason::NoSpace
+            {
+                let wait_time = RuleSet::faulted_child_wait(&nexus_state);
+                if wait_time.is_zero() {
+                    info!("Start full rebuild");
```
tiagolobocastro: What if the replica is not online?

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
```diff
@@ -130,49 +134,247 @@ async fn nexus_reconciler(
...
+    })
+}
+
+/// Iterates child rebuild stage and makes child online if RebuildStage is PartialRebuildInit. Marks
+/// stage as PartialRebuild.
+pub(super) async fn initialize_partial_rebuild(
+    nexus: &mut OperationGuardArc<NexusSpec>,
+    context: &PollContext,
+) -> PollResult {
+    let nexus_uuid = nexus.uuid().clone();
```
tiagolobocastro: we don't need all these clones..

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
```diff
@@ -130,49 +134,247 @@ async fn nexus_reconciler(
...
+/// Iterates over child rebuild stage. If stage is TimedWait and wait duration isn't elapsed,
+/// checks if child node came back. If yes, Marks RebuildStage as PartialRebuildInit. If wait_time
+/// is elapsed then marks stage as FullRebuildInit.
+pub(super) async fn wait_for_child(
+    nexus: &mut OperationGuardArc<NexusSpec>,
+    context: &PollContext,
+) -> PollResult {
+    let nexus_spec = nexus.lock().clone();
+    let rebuild_state = nexus_spec.rebuild_state.clone();
+    for (uri, state) in rebuild_state.iter() {
```
tiagolobocastro: there's way too much nesting in this function, please try to reduce it.

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
```diff
@@ -130,49 +134,247 @@ async fn nexus_reconciler(
...
+                                                        .await?;
+                                                }
+                                            } else {
+                                                info!("Child pool is not imported yet");
+                                            }
+                                        }
+                                    };
+                                }
+                            }
+                        }
```
tiagolobocastro: we can simplify this by simply checking if we have the replica in the registry.

control-plane/stor-port/src/types/v0/store/nexus.rs
```diff
@@ -109,7 +109,40 @@ pub struct NexusSpec {
...
+    pub rebuild_state: HashMap<ChildUri, RebuildInfo>,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, Eq, PartialEq)]
+/// Rebuild specific information for a child.
+pub struct RebuildInfo {
+    /// Reconciler needs to wait for specified duration before starting Full Rebuild.
+    pub wait_time: Option<Duration>,
+    /// Current Rebuild stage of the child.
+    pub stage: RebuildStage,
```
tiagolobocastro: Why do we need this at all?
Can't we just check what to do based on the time and avoid having to keep and track this state altogether? Would also avoid making so many etcd calls which we should avoid as much as possible!

abhilashshetty04: 1. Yes, we can operate without having state but with that approach we might have to timewait, start partial rebuild / Full rebuild within same function. there is more logic coming in next phase as well. Thought of in memory state how we do in dsp but that doesnt seem feasible here as we go out of scope after every reconcile run. 
2. Rebuildstage is like an intent that control plane reconciler has for a child hence we write that to nexus Spec and carry out each stage on separate reconciler fn.
3. There will be one more reconciler coming which will monitor PartialRebuild child and will delete state if its complete.
4. In this phase we delete state for a child after removing child from children list. This will change in phase 2 with some modifications in volume count reconciler. We will have to add one more reconciler for FullRebuild state and check rebuild stats incase it comes back if want to switch to partial rebuild.
5. Hashmap is empty when there is no rebuild. Any child rebuild will have max 4 spec transactions to etcd. transactions are spread across 3 etcd cluster and rebuild phase, depending on when node comes back and wait time.

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
abhilashshetty04: This will remove the child from children list to trigger Full rebuild. mostly in scenario when 2nd child gets Faulted in 3 replica volume. First will be in TimedWait. 2nd will be removed in FullRebuildInit stage

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
abhilashshetty04: Will do

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
abhilashshetty04: I think i was getting replica from the registry even when node is Offline. Thats how i get child node. You want me to check status/state of the Replica?

control-plane/stor-port/src/types/v0/store/nexus.rs
tiagolobocastro: >  there is more logic coming in next phase as well

Let's add the logic when we need to not ahead of time, baby steps, and be very conscious that when we add things to etcd we can't then remove them easily :)

abhilashshetty04: Sure Tiago,Not adding them now. Its for phase 2.  I will be testing removing those keys properly. FullRebuild key gets removed fine. That i checked in etcd. For PartialRebuild need to add logic to track the progress :)

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
abhilashshetty04: We have mutable borrow later in this function to update stage. Before that we have immutable borrow. Without clone complier will complain.

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
abhilashshetty04: Incorporated

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
```diff
@@ -130,57 +139,282 @@ async fn nexus_reconciler(
...
+    context: &PollContext,
+) -> PollResult {
+    let nexus_uuid = nexus.uuid();
+    let nexus_state = context.registry().nexus(nexus_uuid).await?;
+    let child_count = nexus_state.children.len();
+    if nexus_state.status == NexusStatus::Degraded && child_count > 1 {
+        let span = tracing::info_span!("faulted_children_remover", nexus.uuid = %nexus_uuid, request.reconcile = true);
+        async {
+            let nexus_spec_clone = nexus.lock().clone();
+            for (child, state) in nexus_spec_clone.rebuild_state.iter() {
```
dsharma-dc: Instead of iterating the rebuild_state hashmap, let it iterate over nexus_state.children as it does today. Then for each children, get  RebuildInfo from hashmap by child key.

abhilashshetty04: That will increase number of operations right? By iterating on stages we check if we need to proceed further on the reconciler by checking stage. In this case we move further if stage == FullRebuildInit meaning remove child from nexus. This way we wont do any operations that is not required for a child in stage != FullRebuildInit

If this reconciler runs on healthy nexus which will have empty hashmap. We will exit early.

dsharma-dc: > That will increase number of operations right

We're only dealing with 3-4 child so doesn't matter. Moreover, we'd be comparing against state as it is today instead of spec, lesser code churn.

control-plane/stor-port/src/types/v0/store/nexus.rs
abhilashshetty04: We have removed stages from this PR. Will revisit its usage during the next phase if there is a need

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
abhilashshetty04: Removed rebuild stages from implementation.

abhilashshetty04:
bors try

abhilashshetty04:
bors try

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
```diff
@@ -130,61 +135,139 @@ async fn nexus_reconciler(
...
-                    .remove_child_by_uri(
-                        context.registry(),
-                        &nexus_state,
-                        &child.uri,
-                        true,
-                    )
+        let nexus_spec_clone = nexus.lock().clone();
+        for child in nexus_state.children.iter().filter(|c| c.state.faulted()) {
+            if child.state_reason != ChildStateReason::NoSpace {
+                let wait_time = RuleSet::faulted_child_wait(&nexus_state, context.registry());
```
dsharma-dc: nitpick: let's call this wait_period or wait_duration

dsharma-dc: Please combine first two branches - 

 ```
             let elapsed = is_time_elapsed(child.faulted_at, Some(wait_time), &child.uri).await;
              if wait_time.is_zero()
                    || child.state_reason.clone() == ChildStateReason::RebuildFailed
                    || elapsed
                {
                    info!(%child.uri, "Start full rebuild for child, wait elapsed {elapsed}");
                    faulted_children_remover(nexus, &child.uri, context).await?
                } 
```

**Comment by abhilashshetty04** [comment]
bors try

File: control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs [hunk 13]
```diff
@@ -130,61 +135,139 @@ async fn nexus_reconciler(
...
+                {
+                    info!(%child.uri, "Start full rebuild for child");
+                    faulted_children_remover(nexus, &child.uri, context).await?
+                } else if is_time_elapsed(child.faulted_at, Some(wait_time), &child.uri).await {
+                    info!(
+                        %child.uri, "Start full rebuild for child,  Wait time has elapsed"
+
+                    );
+                    faulted_children_remover(nexus, &child.uri, context).await?;
+                } else if get_child_replica(nexus_spec_clone.clone(), &child.clone(), context)
```
dsharma-dc: I'm sure there is some compact and fancy way in rust to make these two branches look cleaner. Please try. @tiagolobocastro , any suggestion?

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
```diff
@@ -130,61 +135,139 @@ async fn nexus_reconciler(
...
+    child: &ChildUri,
+    context: &PollContext,
+) -> Result<(), SvcError> {
+    let nexus_state = context.registry().nexus(nexus).await?;
+    info!(%child, "Making online of {:?} nexus", nexus);
+    let node = context.registry().node_wrapper(&nexus_state.node).await?;
+    let online_context = NexusChildActionContext::new(&nexus_state.node, &nexus.clone(), child);
+    let ctx = NexusChildActionContextNode::new(online_context, context.registry());
+    // TODO: Check if we can handle things differently for different SvcError.
+    node.online_child(ctx).await?;
```
dsharma-dc, line 253: need to handle return failures here as you were doing in earlier version of change.

abhilashshetty04: Its handled, In the caller end if this returns err, We start Full rebuild of the child.

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
abhilashshetty04: It can be done but we will lose the logger in that case when is_time_elapsed() is true. Without a separate condition. We do have a log in is_time_elapsed() that lets us know how much we waited and wait_duration. But this gives us clear log before we start Full rebuild

abhilashshetty04: Incorporated

dsharma-dc: We won't lose logger, I've taken care of that in suggestion.
`info!(%child.uri, "Start full rebuild for child, wait elapsed {elapsed}");`

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
abhilashshetty04: both else if check different conditions. You might feel there is lots of branching as we have removed stages. We are checking all possiblities in single reconciler now.

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
```diff
@@ -130,61 +136,139 @@ async fn nexus_reconciler(
...
+        faulted_children_remover(nexus_spec, &child.uri, context).await?;
+    } else if get_child_replica(nexus_spec.as_ref(), child, context)
         .await
+        .is_err()
+    {
+        info!(%child.uri,"Replica is still not online for child");
+    } else {
+        info!(%child.uri, "Replica for child is Online, starting partial rebuild");
+        if let Err(error) = initialize_partial_rebuild(&nexus.uuid, &child.uri, context).await {
+            info!(%child.uri, "Initializing Full rebuild as partial rebuild failed for child, failed with : {:?}", error);
```
dsharma-dc: Let's not say `partial rebuild failed`, rather say `online_child attempt` failed

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
```diff
@@ -130,61 +136,139 @@ async fn nexus_reconciler(
...
-                        tracing::info!(
-                            %child.uri,
-                            %child.state,
-                            %child.state_reason,
-                            "Successfully removed faulted child",
-                        )
-                    });
-                }
+        for child in nexus_state.children.iter().filter(|c| c.state.faulted()) {
+            if child.state_reason != ChildStateReason::NoSpace {
```
tiagolobocastro: ah, I think enospc won't actually be here, since that's degraded rather than faulted?

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
abhilashshetty04: Ive made a separate fn for wait now

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
```diff
@@ -130,61 +136,139 @@ async fn nexus_reconciler(
...
+#[tracing::instrument(skip(nexus, context), level = "trace", fields(nexus.uuid = %nexus.uuid(), request.reconcile = true))]
+pub(super) async fn faulted_children_remover(
+    nexus: &mut OperationGuardArc<NexusSpec>,
+    child: &ChildUri,
+    context: &PollContext,
+) -> Result<(), SvcError> {
+    let nexus_uuid = nexus.uuid();
+    let nexus_state = context.registry().nexus(nexus_uuid).await?;
+
+    info!(%child, "Attempting to remove faulted child");
```
tiagolobocastro:
```suggestion
    nexus_spec_clone.warn_span(|| {
        tracing::warn!(%child.uri, %child.state, %child.state_reason, "Attempting to remove faulted child")
    });
```

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
```diff
@@ -130,61 +136,139 @@ async fn nexus_reconciler(
...
+    nexus
+        .remove_child_by_uri(context.registry(), &nexus_state, child, true)
+        .await?;
+
+    info!(%child ,"Successfully removed faulted child");
```
tiagolobocastro:
```suggestion
        nexus_spec_clone.warn_span(|| {
        tracing::warn!(%child.uri, %child.state, %child.state_reason, "Successfully removed faulted child")
    });
```

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
```diff
@@ -130,61 +136,139 @@ async fn nexus_reconciler(
...
+    Ok(())
+}
+
+/// Returns true if time elapsed from child fault time is greater or equal to wait time duration.
+async fn is_time_elapsed(
```
tiagolobocastro: this does not need to be async

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
```diff
@@ -130,61 +136,139 @@ async fn nexus_reconciler(
...
+    wait_duration: Duration,
+    uri: &ChildUri,
+) -> bool {
+    if let Some(fault_time) = fault_time {
+        if let Ok(elapsed) = fault_time.elapsed() {
+            info!(
+                child.uri = %uri,
+                "waited {:?} for child to comeback. Wait time is {:?}",
+                elapsed, wait_duration
+            );
```
tiagolobocastro, line 220: don't need this since we have logs above

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
```diff
@@ -130,61 +136,139 @@ async fn nexus_reconciler(
...
+            } else {
+                continue;
+            }
+        }
+    }
+    Err(SvcError::NotFound {
+        kind: ResourceKind::Node,
+        id: "Child replica not found".to_string(),
+    })
+}
```
tiagolobocastro:
```suggestion
let replica_uri = nexus
        .replica_uri(&child.uri)
        .ok_or(SvcError::ChildNotFound {
            child: child.uri.to_string(),
            nexus: nexus.uuid.to_string(),
        })?;
    registry.replica(replica_uri.uuid()).await
}
```

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
```diff
@@ -130,61 +136,139 @@ async fn nexus_reconciler(
...
+    info!(%child, "Making online of {:?} nexus", nexus);
+    let node = context.registry().node_wrapper(&nexus_state.node).await?;
+    let online_context = NexusChildActionContext::new(&nexus_state.node, &nexus.clone(), child);
+    let ctx = NexusChildActionContextNode::new(online_context, context.registry());
+    // TODO: Check if we can handle things differently for different SvcError.
+    node.online_child(ctx).await?;
+    info!(
+        %child, "Successfully made child online in nexus :{:?}", nexus
+    );
+    Ok(())
```
tiagolobocastro:
```suggestion
pub(super) async fn online_nexus_child(
    nexus: &mut OperationGuardArc<NexusSpec>,
    child: &ChildUri,
    replica: &Replica,
    context: &PollContext,
) -> Result<(), SvcError> {
    let nexus_node = &nexus.as_ref().node;
    let node = registry.node_wrapper(nexus_node).await?;
    node.online_child(NexusChildActionContextNode::new(
        NexusChildActionContext::new(nexus_node, nexus.uuid(), &child.uri),
        registry,
    ))
    .await?;
     tracing::info!(
            child.uri = %child.uri.as_str(),
            replica.uuid = %replica.uid(),
            "Successfully onlined child",
        );
     Ok(())
```

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
```diff
@@ -130,61 +136,139 @@ async fn nexus_reconciler(
...
+) -> bool {
+    if let Some(fault_time) = fault_time {
+        if let Ok(elapsed) = fault_time.elapsed() {
+            info!(
+                child.uri = %uri,
+                "waited {:?} for child to comeback. Wait time is {:?}",
+                elapsed, wait_duration
+            );
+            return elapsed >= wait_duration;
+        };
```
tiagolobocastro:
```suggestion
        }
```

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
```diff
@@ -130,61 +136,139 @@ async fn nexus_reconciler(
...
+        info!(%child.uri,"Replica is still not online for child");
+    } else {
+        info!(%child.uri, "Replica for child is Online, starting partial rebuild");
+        if let Err(error) = initialize_partial_rebuild(&nexus.uuid, &child.uri, context).await {
+            info!(%child.uri, "Initializing Full rebuild as partial rebuild failed for child, failed with : {:?}", error);
+            faulted_children_remover(nexus_spec, &child.uri, context).await?
+        }
     }
+    Ok(())
+}
```
tiagolobocastro: I think we can simplify this a bit:

```suggestion
async fn handle_faulted_child(
    nexus_spec: &mut OperationGuardArc<NexusSpec>,
    child: &Child,
    nexus: &Nexus,
    context: &PollContext,
) -> Result<(), SvcError> {
    // Child is permanently faulted, we have remove it and rebuild anew.
    if child.state_reason == ChildStateReason::RebuildFailed {
        return faulted_children_remover(nexus_spec, &child.uri, context).await;
    }

    if let Ok(replica) = get_child_replica(nexus_spec.as_ref(), child, context).await {
        info!(%child.uri, "Child's Replica is Online within the partial rebuild window");
        if online_nexus_child(nexus_spec, child, replica, context).await.is_err() {
            // we can also not do this and simply wait till the timeout..
            faulted_children_remover(nexus_spec, &child.uri, context).await?
        }
    }

    // Partial rebuild time window expired, we have to rebuild anew.
    if partial_rebuild_elapsed(child.faulted_at, &child.uri) {
        info!(%child.uri, "Child's Replica failed to become online within the partial rebuild window");
        return faulted_children_remover(nexus_spec, &child.uri, context).await;
    }
    
    Ok(())
}


```

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
abhilashshetty04: This log doesnt hold true for RebuildFailed child right? We do get duration from policy for that child but dont really care about time_elapsed for that child. elapsed can very well be false in this scenario.

abhilashshetty04: info!(%child.uri, "Start full rebuild for child, wait elapsed {elapsed}, child state reason: {}", child.state_reason);

I think this would help us figure out why we started Full rebuild

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
abhilashshetty04: If its Degraded then it wont be needed. Are you sure its going to be Degraded.

dsharma-dc (approved):

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
abhilashshetty04: Ill print nexus and child in success log. Replica is not required in this.

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
abhilashshetty04: Yes

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
abhilashshetty04: Here it prints how much we have waited for the child during every attempt.

abhilashshetty04:
bors try

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
tiagolobocastro: I believe so yes.

abhilashshetty04: Ill remove the check

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
abhilashshetty04: I have combined checking time_elapsed() and wait_time.is_zero() together. @dsharma-dc asked if i can combine it.

abhilashshetty04: Shouldnt have different if statements in this. code should enter anyone of all checks. Just tested, even if remove child (1st condition) we attempt get_child_replica. I am checking time elapsed before anything else.

tests/bdd/features/volume/replicas/test_feature.py
```diff
@@ -219,6 +227,7 @@ def the_volume_spec_should_show_1_replica():
 # Wait for the number of runtime volume replicas to reach the expected number of replicas.
 @retry(wait_fixed=1000, stop_max_attempt_number=10)
 def wait_for_volume_replica_count(expected_num_replicas):
+    time.sleep(SLEEP_BEFORE_CHECK)
```
tiagolobocastro: We don't need to sleep here, this function uses the retry decorator

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
```diff
@@ -130,61 +135,126 @@ async fn nexus_reconciler(
...
+/// Removes child from nexus children list to initiate Full rebuild of child.
+#[tracing::instrument(skip(nexus, context), level = "trace", fields(nexus.uuid = %nexus.uuid(), request.reconcile = true))]
+pub(super) async fn faulted_children_remover(
+    nexus: &mut OperationGuardArc<NexusSpec>,
+    child: &Child,
+    context: &PollContext,
+) -> Result<(), SvcError> {
+    let nexus_uuid = nexus.uuid();
+    let nexus_state = context.registry().nexus(nexus_uuid).await?;
+    let nexus_spec_clone = nexus.lock().clone();
```
tiagolobocastro: hmm can you try not cloning and using nexus.as_ref() ?

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
```diff
@@ -130,61 +135,126 @@ async fn nexus_reconciler(
...
+        .await?;
+
+    nexus_spec_clone.warn_span(|| {
+        tracing::warn!(%child.uri, %child.state, %child.state_reason, "Successfully removed faulted child")
+    });
+    Ok(())
+}
+
+/// Returns true if time elapsed from child fault time is greater or equal to wait time duration.
+fn is_time_elapsed(
```
tiagolobocastro: suggest using something like `partial_rebuild_elapsed`

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
tiagolobocastro: I don't think that's necessary.

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
```diff
@@ -130,61 +135,126 @@ async fn nexus_reconciler(
...
+) -> Result<Replica, SvcError> {
+    let replica_uri = nexus.replica_uri(&child.uri).ok_or(SvcError::NotFound {
+        kind: ResourceKind::Replica,
+        id: "Child replica not found".to_string(),
+    })?;
+    context.registry().replica(replica_uri.uuid()).await
+}
+
+/// Tries to mark child as Online. Returns error on failing to do so.
+pub(super) async fn initialize_partial_rebuild(
```
tiagolobocastro: Suggest a more representative name such as `online_nexus_child` since there is no special partial rebuild call.

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
```diff
@@ -130,61 +135,126 @@ async fn nexus_reconciler(
...
+) -> Result<(), SvcError> {
+    let nexus_node = &nexus.as_ref().node;
+    info!(%child, "Making online of {:?} nexus", nexus);
+    let node = context.registry().node_wrapper(nexus_node).await?;
+    let online_context = NexusChildActionContext::new(nexus_node, nexus.uuid(), child);
+    let ctx = NexusChildActionContextNode::new(online_context, context.registry());
+    // TODO: Check if we can handle things differently for different SvcError.
+    node.online_child(ctx).await?;
+    info!(
+        %child, "Successfully made child online in nexus :{:?}", nexus
```
tiagolobocastro: We don't need to trace the entire nexus, you can take a look at existing traces.

tiagolobocastro (approved):

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
abhilashshetty04: Yes, nexus arg used to be NexusId before. Made changes

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
abhilashshetty04: I think partial_rebuild_elapsed doesnt indicate wait duration getting elapsed. How about wait_duration_elapsed() ?

abhilashshetty04:
bors try

abhilashshetty04:
bors merge

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
dsharma-dc: `NoSpace` reason appears to be associated with state `faulted` as of today. This may probably change with GTM-469 later.

```
    /// Faults the device by its name, with the given I/O error.
    /// The faulted device is scheduled to be retired.
    fn fault_device(
 [...]
        let reason = match io_status {
            IoCompletionStatus::LvolError(LvolFailure::NoSpace) => {
                FaultReason::NoSpace
```

tiagolobocastro: That's only within the Nexus, once it's translated to grpc it's degraded I think. But yes this all may change when we refactor the states.

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
tiagolobocastro: `wait_duration_elapsed` also doesn't tell you that's for the partial rebuild.

abhilashshetty04: We have explained why we wait in the PR comment.

PR #458, opened 2023-03-13
Import Pool on Reconciler State recreation

tiagolobocastro:
test(agents/core): pool recreate should import only
    
    On recreation, switch the backend disk and make sure recreation/import fails!
    
---

    fix(agents/core): fetch pools and replicas on import
    
    This helps keep a consistent view of the pool/replicas as the io-engine pool api
    has a separate interface for the pool replicas..
    
---

    fix(agents/core): import pool on recreation
    
    When recreating a pool we should use the v1 import api to avoid creating a new pool if the
    backend disk has changed.
    
### DIFF
--- a/control-plane/agents/src/bin/core/controller/reconciler/pool/mod.rs
+++ b/control-plane/agents/src/bin/core/controller/reconciler/pool/mod.rs
@@ -11,7 +11,7 @@ use crate::controller::{
 use crate::controller::io_engine::PoolApi;
 use stor_port::types::v0::{
     store::pool::PoolSpec,
-    transport::{CreatePool, DestroyPool, NodeStatus},
+    transport::{DestroyPool, ImportPool, NodeStatus},
 };
 use tracing::Instrument;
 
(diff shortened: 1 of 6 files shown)


control-plane/agents/src/bin/core/controller/reconciler/pool/mod.rs
```diff
@@ -141,8 +141,8 @@ async fn missing_pool_state_reconciler(
         async {
             pool_spec.warn_span(|| tracing::warn!("Attempting to recreate missing pool"));
 
-            let request = CreatePool::new(&pool_spec.node, &pool_spec.id, &pool_spec.disks, &pool_spec.labels);
-            match node.create_pool(&request).await {
+            let request = ImportPool::new(&pool_spec.node, &pool_spec.id, &pool_spec.disks);
```
Abhinandan-Purkait, line 144: How do we check the disks have not been swapped? Also what happens if the disk is correct and the pool is missing?

tiagolobocastro: If the disks have not swapped then no problem. If the disk is the same but somehow was zeroed/corrupted then import will fail.

Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

PR #461, opened 2023-03-16
fix(capacity/thin): don't fault largest replica if we can't move it

tiagolobocastro:
fix(agents/core/publish): check volume spec target
    
    When publishing we should check the volume spec and not the state.
    Otherwise if a nexus is down we might think the volume is not published.
    
---

    chore: use feature to avoid .git rebuild dependency
    
---

    refactor(agents/core): continue making use of operation guard
    
    Carry on previous refactor work moving functionality to the operations modules and
    out of the specs.
    
---

    fix(deployer): isolate io-engine cores
    
    A previous change had broken the isolation when used without specifying core count.
    
---

    fix(capacity/thin): don't fault largest replica if we can't move it
    
    Changes the faulting logic of the largest enospc replica. The replica is now moved first and only
    then it is faulted&deleted.
    Otherwise we might end up faulting all enospc replicas as each second largest replica eventually
    becomes the largest replica :)
    
    We might still need a way to delete the largest replica even if we can't move it as it might allow
    us to get enough space for the other replicas.
    

Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

PR #463, opened 2023-03-21
chore(child/fault): make fault timestamp available in control plane

dsharma-dc:


rpc/build.rs
```diff
@@ -26,6 +26,7 @@ fn main() {
 
     tonic_build::configure()
         .build_server(true)
+        .extern_path(".google.protobuf.Timestamp", "::prost_wkt_types::Timestamp")
```
tiagolobocastro: We don't need this, we don't set serde traits here?

control-plane/stor-port/src/types/v0/transport/child.rs
```diff
@@ -17,6 +18,8 @@ pub struct Child {
     pub rebuild_progress: Option<u8>,
     /// Reason for the child state.
     pub state_reason: ChildStateReason,
+    /// Last faulted timestamp of this child
```
tiagolobocastro:
```suggestion
    /// Last faulted timestamp of this child.
```

control-plane/agents/src/bin/core/controller/io_engine/v1/translation.rs
```diff
@@ -266,6 +266,7 @@ impl IoEngineToAgent for v1::nexus::Child {
             state_reason: v1::nexus::ChildStateReason::from_i32(self.state_reason)
                 .map(|f| From::from(ExternalType(f)))
                 .unwrap_or(ChildStateReason::Unknown),
+            faulted_at: self.fault_timestamp.clone().map(|d| d.into()),
```
tiagolobocastro:
```suggestion
            faulted_at: self.fault_timestamp.into_opt(),
``` 
?

dsharma-dc: I don't see any method `into_opt()` in the `Option` type. Any reference ?

tiagolobocastro: It's a trait we've added: `use stor_port::IntoOption`.

rpc/build.rs
dsharma-dc: I'm not sure about serde applicability, but this is added here so that protobuf timestamp maps correctly to `prost_wkt_types::Timestamp`. Without this, the autogen code maps that to `prost_types::Timestamp` which doesn't have `From` trait implementation for conversion.

tiagolobocastro: I was under the impression that we don't need to use `prost_wkt_types::Timestamp` as its main drive was to get the serde traits. We can simply use the default protobuf Timestamp.

control-plane/agents/src/bin/core/controller/io_engine/v1/translation.rs
tiagolobocastro: To avoid the `prost_wkt_types` dependency:

```suggestion
            faulted_at: self
                .fault_timestamp
                .clone()
                .and_then(|t| std::time::SystemTime::try_from(t).ok())
                .map(chrono::DateTime::from),
```

dsharma-dc: This is building ok with this suggestion, but I'm seeing timestamp value zero for a faulted child. Debugging if this is due to this usage.

tiagolobocastro: Then you can try how prost does it, something like this:

```suggestion
            faulted_at: self.fault_timestamp.clone().and_then(|t| {
                t.normalize();
                NaiveDateTime::from_timestamp_opt(t.seconds, t.nanos as u32)            
            ).and_then(|dt| DateTime::from_utc(dt, Utc)),
```

dsharma-dc: @tiagolobocastro  Since the bug was identified in dataplane (https://github.com/openebs/mayastor/pull/1340/), it's working fine as per earlier suggestion(without NaiveDateTime).

`TSDEBUG: Child nvmf://10.1.0.7:8420/nqn.2019-05.io.openebs:c4e7ad8b-56cf-4c51-a26f-d9fce2680d53?uuid=c4e7ad8b-56cf-4c51-a26f-d9fce2680d53, state 3, fault_time 2023-03-28T07:41:38Z`

tiagolobocastro: Great! please revert Cargo.lock.

tiagolobocastro (approved):

control-plane/agents/src/bin/core/controller/io_engine/v1/translation.rs
dsharma-dc: Cargo.lock update looks unrelated to previous prost-wkt-types dependency.

dsharma-dc:
bors try

control-plane/agents/src/bin/core/controller/io_engine/v1/translation.rs
tiagolobocastro: It is related, why would it change otherwise?

abhilashshetty04 (approved):

dsharma-dc:
bors try

Abhinandan-Purkait (approved):

dsharma-dc:
bors merge

PR #465, opened 2023-03-22
fix(openapi): don't rebuild openapi twice

tiagolobocastro:
Cargo keeps an internal timestamp of files to trigger rebuilds. If the version of the library did not change then don't touch the file otherwise we could rebuild twice.


niladrih (approved):

Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

PR #466, opened 2023-03-23
feat(upgrade): etcd key space migration from legacy on core-agent starup

Abhinandan-Purkait:
ETCD Key Space Migration from product v1 to v2


control-plane/agents/src/bin/core/controller/registry.rs
```diff
@@ -138,21 +147,26 @@ impl Registry {
...
-            ),
+    async fn put_config_or_panic<S: Store>(
+        mut store: S,
+        legacy_prefix_present: bool,
+    ) -> CoreRegistryConfig {
+        let mut config = CoreRegistryConfig::new(NodeRegistration::Automatic);
+        // If legacy prefix is present enable the compat mode.
+        if legacy_prefix_present {
+            config.enable_mayastor_compat_v1();
+            info!("Legacy prefix present, enabling compat mode");
```
```suggestion
            warn!("Legacy V1 prefix present, enabling compatibility mode");
```

control-plane/agents/src/bin/core/controller/registry.rs
```diff
@@ -138,21 +147,26 @@ impl Registry {
...
-            Err(StoreError::MissingEntry { .. }) => {
-                store.put_obj(&config).await.expect(
-                    "Must be able to access the persistent store to persist configuration information",
-                );
-                config
-            },
-            Err(error) => panic!(
-                "Must be able to access the persistent store to load configuration information. Got error: '{error:#?}'"
-            ),
+    async fn put_config_or_panic<S: Store>(
```
tiagolobocastro: We're not loading the existing configuration at all, we're just overwriting it?

control-plane/agents/src/bin/core/controller/registry.rs
```diff
@@ -119,12 +128,12 @@ impl Registry {
...
                 reconcile_idle_period,
                 reconciler: ReconcilerControl::new(),
-                config: Self::get_config_or_panic(store).await,
+                config: Self::put_config_or_panic(store, legacy_prefix_present).await,
                 max_rebuilds,
                 host_acl,
             }),
         };
-        registry.init().await;
+        registry.init(legacy_prefix_present).await;
```
tiagolobocastro: can't we read this from the self?

control-plane/stor-port/src/types/v0/store/registry.rs
```diff
@@ -29,6 +29,10 @@ impl CoreRegistryConfig {
     pub fn mayastor_compat_v1(&self) -> bool {
         self.mayastor_compat_v1.unwrap_or(false)
     }
+    /// Set the `mayastor_compat_v1` to true.
+    pub fn enable_mayastor_compat_v1(&mut self) {
```
tiagolobocastro: maybe set instead, so we can disable it when it's not needed anymore?

utils/pstor/src/legacy.rs
```diff
@@ -0,0 +1,49 @@
+use crate::{key_prefix_obj, Error, StorableObjectType, Store};
+use tracing::info;
+
+/// Returns the legacy key prefix that was used for the keys, when running from within the cluster.
+fn legacy_key_prefix() -> String {
```
tiagolobocastro: suggest having legacy module and within legacy have another file called product_v1 ?
Then don't add legacy prefix to all of these since we're already in legacy?

utils/pstor/src/legacy.rs
```diff
@@ -0,0 +1,49 @@
...
+    build_legacy_key_prefix(platform::platform_info())
+}
+
+/// Returns the legacy key prefix that was used for the keys.
+/// The platform info and namespace where the product is running must be specified.
+fn build_legacy_key_prefix(cluster_uid: &dyn platform::PlatformInfo) -> String {
+    format!("/namespace/{}/control-plane", cluster_uid.namespace(),)
```
tiagolobocastro:
```suggestion
    format!("/namespace/{}/control-plane", cluster_uid.namespace())
```

utils/pstor/src/legacy.rs
```diff
@@ -0,0 +1,49 @@
...
+    for (k, v) in &store_entries {
+        // Bail if we are not able extract the resource id. This can leave
+        // incomplete migration entries in etcd, if failed midway.
+        let id = k
+            .split('/')
+            .last()
+            .ok_or_else(|| crate::Error::InvalidKey { key: k.to_string() })?;
+
+        store
+            .put_kv(&format!("{}/{}", key_prefix_obj(spec_type, 0), id), v)
```
tiagolobocastro: Maybe we should have another module for conversions between product_v1 and product_v2 with api v0 ?

control-plane/agents/src/bin/core/controller/registry.rs
Abhinandan-Purkait: Yes, because when its missing we write it. So why not always write it. That saves the additional get call on fallback.

control-plane/agents/src/bin/core/controller/registry.rs
Abhinandan-Purkait: You mean adding a new field to the Registry and use that?

tiagolobocastro: Yeah or just reading the v1 enabled from config?

control-plane/agents/src/bin/core/controller/registry.rs
tiagolobocastro: No because then you're always modifying the contents?

control-plane/stor-port/src/types/v0/store/registry.rs
tiagolobocastro: We should disable it automatically when there are no more nexus info coming from the old location. We can check this in startup?

control-plane/agents/src/bin/core/controller/registry.rs
Abhinandan-Purkait: The contents of the config seems to always remain same. Would revert it back probably

control-plane/stor-port/src/types/v0/store/registry.rs
Abhinandan-Purkait: How do we detect that? Its on root.

control-plane/agents/src/bin/core/controller/registry.rs
tiagolobocastro: Registry can be updated using etdctl, eg: change node registration type. In the future we may also provide ways of modifying them via cli/rest.

utils/pstor/src/legacy.rs
Abhinandan-Purkait: Done.

utils/pstor/src/legacy.rs
Abhinandan-Purkait: Done.

control-plane/agents/src/bin/core/controller/registry.rs
```diff
@@ -109,6 +115,11 @@ impl Registry {
         .await
         .expect("Should connect to the persistent store");
         tracing::info!("Connected to persistent store at {}", store_endpoint);
+
+        let legacy_prefix_present = detect_product_v1_prefix(store.clone())
```
tiagolobocastro: can pass by reference instead?

control-plane/agents/src/bin/core/controller/registry.rs
```diff
@@ -138,21 +150,35 @@ impl Registry {
...
+            Err(error) => panic!("Failed to load configuration information: {error:?}"),
+        };
+
+        if legacy_prefix_present != config.mayastor_compat_v1() {
+            config.set_mayastor_compat_v1(legacy_prefix_present);
+            store
+                .put_obj(&config)
+                .await
+                .expect("Failed to persist configuration information");
+            warn!("Legacy V1 prefix present, enabling compatibility mode");
```
tiagolobocastro, line 183: This doesn't seem right, what if legacy_prefix_present is false but config.mayastor_compat_v1() is true? This will actually clear it which seems wrong?

control-plane/agents/src/bin/core/controller/resources/operations_helper.rs
```diff
@@ -883,7 +887,15 @@ impl ResourceSpecsLocked {
         &self,
         store: &mut S,
         spec_type: StorableObjectType,
+        legacy_prefix_present: bool,
     ) -> Result<(), SpecError> {
+        if legacy_prefix_present {
+            migrate_product_v1_to_v2(store, spec_type)
+                .await
+                .map_err(|e| SpecError::StoreGet {
```
tiagolobocastro: Should this be `StorageMigrate` error?

utils/pstor/src/products/mod.rs
```diff
@@ -0,0 +1,29 @@
...
+
+pub mod v1;
+pub mod v2;
+
+/// Migrate the values from legacy keys to latest keys.
+pub async fn migrate_product_v1_to_v2<S: Store>(
+    store: &mut S,
+    spec_type: StorableObjectType,
+) -> Result<(), crate::Error> {
+    info!("Migrating {spec_type:?} from legacy to current key space");
```
tiagolobocastro:
```suggestion
    info!("Migrating {spec_type:?} from v1 to v2 key space");
```

utils/pstor/src/products/mod.rs
```diff
@@ -0,0 +1,29 @@
...
+    for (k, v) in &store_entries {
+        // Bail if we are not able extract the resource id. This can leave
+        // incomplete migration entries in etcd, if failed midway.
+        let id = k
+            .split('/')
+            .last()
+            .ok_or_else(|| crate::Error::InvalidKey { key: k.to_string() })?;
+
+        store
+            .put_kv(&format!("{}/{}", v2::key_prefix_obj(spec_type, 0), id), v)
```
tiagolobocastro: api version (0) should be passed as arg to the fn ?

tiagolobocastro (commented):
Don't make version default to 0 to avoid future mistakes. Better each type to explicitly set it.

control-plane/agents/src/bin/core/controller/registry.rs
Abhinandan-Purkait: Done.

control-plane/agents/src/bin/core/controller/resources/operations_helper.rs
Abhinandan-Purkait: Good Point. Thanks

control-plane/agents/src/bin/core/controller/registry.rs
Abhinandan-Purkait: Fixed

control-plane/stor-port/src/types/v0/store/registry.rs
Abhinandan-Purkait: We would probably need to go though all the keys and find out which one only uuid4 format. Seems to be a costly operation.

utils/pstor/src/products/mod.rs
Abhinandan-Purkait: Done.

control-plane/stor-port/src/types/v0/store/registry.rs
Abhinandan-Purkait: As discussed we will take this as enhancement

Abhinandan-Purkait:
bors try

tiagolobocastro (approved):

Abhinandan-Purkait:
bors merge

PR #469, opened 2023-03-28
fix(io-engine/replica): destroy replica should be idempotent

tiagolobocastro:
Adding the pool to the request allows us determine whether a pool is loaded or not. Also use a gtm header to detect wether the returned error is expected. In addition, we now handle the v0 separately by making use of the api version.


control-plane/agents/src/bin/core/node/wrapper.rs
```diff
@@ -1082,19 +1082,69 @@ impl ReplicaApi for Arc<tokio::sync::RwLock<NodeWrapper>> {
...
+            Ok(()) => {
+                self.update_replica_state(Either::Remove(request.uuid.clone()))
+                    .await;
+                if let Ok(mut ctx) = dataplane.reconnect(GETS_TIMEOUT).await {
+                    self.update_pool_states(ctx.deref_mut()).await.ok();
+                }
+                Ok(())
             }
+            Err(SvcError::GrpcRequestError { source, .. })
+                if source.metadata().contains_key("gtm-602")
```
avishnu, line 1118: What's up with gtm-602?

control-plane/agents/src/bin/core/node/wrapper.rs
```diff
@@ -1082,19 +1082,69 @@ impl ReplicaApi for Arc<tokio::sync::RwLock<NodeWrapper>> {
...
+                self.update_replica_state(Either::Remove(request.uuid.clone()))
+                    .await;
+                if let Ok(mut ctx) = dataplane.reconnect(GETS_TIMEOUT).await {
+                    self.update_pool_states(ctx.deref_mut()).await.ok();
+                }
+                Err(SvcError::ReplicaNotFound {
+                    replica_id: request.uuid.to_owned(),
+                })
+            }
+            // previous not found error was not not entirely correct as it was being returned
```
avishnu: repeated 'not'

control-plane/agents/src/bin/core/node/wrapper.rs
```diff
@@ -1082,19 +1082,69 @@ impl ReplicaApi for Arc<tokio::sync::RwLock<NodeWrapper>> {
...
-        let mut ctx = dataplane.reconnect(GETS_TIMEOUT).await?;
-        self.update_replica_states(ctx.deref_mut()).await?;
-        // todo: remove when CAS-1107 is resolved
-        if let Some(replica) = self.read().await.replica(&request.uuid) {
-            if replica.pool_id == request.pool_id {
-                return Err(SvcError::Internal {
-                    details: "replica was not destroyed by the io-engine".to_string(),
-                });
+        match dataplane.destroy_replica(request).await {
+            // v0 success was not not entirely correct as it was being returned
```
avishnu: repeated 'not'

rpc/src/lib.rs
```diff
@@ -452,9 +452,9 @@ pub mod v1 {
     /// V1 Replica autogenerated grpc code.
     pub mod replica {
         pub use super::pb::{
-            replica_rpc_client, CreateReplicaRequest, DestroyReplicaRequest, ListReplicaOptions,
-            ListReplicasResponse, Replica, ReplicaSpaceUsage, ShareReplicaRequest,
-            UnshareReplicaRequest,
+            destroy_replica_request, replica_rpc_client, CreateReplicaRequest,
```
avishnu, line 455: Why are there 2 types of destroy replica request?

control-plane/agents/src/bin/core/node/wrapper.rs
tiagolobocastro: good catch, done

rpc/src/lib.rs
tiagolobocastro: the new one is not a request type, it's a module containing the `oneof` type which is declared within the destroy replica message.

control-plane/agents/src/bin/core/node/wrapper.rs
tiagolobocastro: At the moment there's no version being returned from dataplane response header which would allow us to determine if the dataplane which we're talking to has the code fix. Until such time I simply used the gtm key as the indicator. This does need to be implemented properly though as discussed with @dsavitskiy

Abhinandan-Purkait (approved):

abhilashshetty04 (approved):

tiagolobocastro:
bors merge

PR #472, opened 2023-03-30
feat(upgrade): migrate MayastorPool CRs to DiskPool CRs on operator startup

Abhinandan-Purkait:
- #### Restructure the DSP Operator code base
- #### Added the migration from MSP to DSP

### DIFF
--- a/k8s/operators/src/diskpool/mod.rs
+++ b/k8s/operators/src/diskpool/mod.rs
@@ -1,2 +1,0 @@
-/// The DiskPool custom resource definition
-pub mod crd;
--- a/k8s/operators/src/lib.rs
+++ b/k8s/operators/src/lib.rs
@@ -1,2 +1,7 @@
-/// The DiskPool operator
-pub mod diskpool;
+/// This module has the definitions for the diskpool crd.
+pub mod diskpool {
+    /// The DiskPool custom resource definition.
+    pub mod crd {
+        include!("pool/diskpool/crd.rs");
+    }
+}
--- a/k8s/operators/Cargo.toml
+++ b/k8s/operators/Cargo.toml
@@ -7,7 +7,7 @@ authors = ["Jeffry Molanus <jeffry.molanus@gmail.com>"]
 
 [[bin]]
 name = "operator-diskpool"
-path = "src/diskpool/main.rs"
+path = "src/pool/main.rs"
 
 [lib]
 name = "k8s_operators"
--- a/k8s/operators/src/pool/diskpool/client.rs
+++ b/k8s/operators/src/pool/diskpool/client.rs
@@ -1,0 +1,67 @@
+use super::crd::{DiskPool, DiskPoolSpec};
+use crate::error::Error;
+use k8s_openapi::apiextensions_apiserver::pkg::apis::apiextensions::v1::CustomResourceDefinition;
+use kube::{api::PostParams, Api, Client, CustomResourceExt, ResourceExt};
+use openapi::apis::StatusCode;
+use tracing::info;
+
+/// Get the DiskPool api.
+pub(crate) fn api(client: &Client, namespace: &str) -> Api<DiskPool> {
+    Api::namespaced(client.clone(), namespace)
+}
+
+/// Create a disk pool CR, with the given name and spec.
+pub(crate) async fn create(
+    client: &Client,
+    namespace: &str,
+    name: &str,
+    spec: DiskPoolSpec,
+) -> Result<(), Error> {
+    let post_params = PostParams::default();
+    let api = api(client, namespace);
+    let new_disk_pool: DiskPool = DiskPool::new(name, spec);
+    match api.create(&post_params, &new_disk_pool).await {
+        Ok(_) => Ok(()),
+        Err(kube::Error::Api(e)) if e.code == StatusCode::CONFLICT => Ok(()),
+        Err(error) => Err(error.into()),
+    }
+}
+
+/// Ensure the CRD is installed. This creates a chicken and egg problem. When the CRD is removed,
+/// the operator will fail to list the CRD going into a error loop.
+///
+/// To prevent that, we will simply panic, and hope we can make progress after restart. Keep
+/// running is not an option as the operator would be "running" and the only way to know something
+/// is wrong would be to consult the logs.
+pub(crate) async fn ensure_crd(k8s: &Client) -> Result<CustomResourceDefinition, Error> {
+    let crd_api: Api<CustomResourceDefinition> = Api::all(k8s.clone());
+
+    // Check if there is an existing dsp diskpool. If yes, Replace it with new generated one.
+    // Create new if it doesnt exist.
+    let mut crd = DiskPool::crd();
+    let crd_name = crd.metadata.name.as_ref().ok_or(Error::InvalidCRField {
+        field: "diskpool.metadata.name".to_string(),
+    })?;
+
+    // Ensure the diskpool.
+    let result = if let Ok(existing) = crd_api.get(crd_name).await {
+        crd.metadata.resource_version = existing.resource_version();
+        info!(
+            "Replacing CRD: {}",
+            serde_json::to_string_pretty(&crd).unwrap()
+        );
+
+        let pp = PostParams::default();
+        crd_api.replace(crd_name, &pp, &crd).await
+    } else {
+        info!(
+            "Creating CRD: {}",
+            serde_json::to_string_pretty(&crd).unwrap()
+        );
+
+        let pp = PostParams::default();
+        crd_api.create(&pp, &crd).await
+    };
+    let crd = result.map_err(|e| Error::Kube { source: e })?;
+    Ok(crd)
+}
--- a/k8s/operators/src/pool/diskpool/crd.rs
+++ b/k8s/operators/src/pool/diskpool/crd.rs
@@ -34,18 +34,21 @@ pub struct DiskPoolSpec {
 }
 
 impl DiskPoolSpec {
-    /// The node the pool is placed on
+    /// Create a new DiskPoolSpec from the node and the disks.
+    pub fn new(node: String, disks: Vec<String>) -> Self {
+        Self { node, disks }
+    }
+    /// The node the pool is placed on.
     pub fn node(&self) -> String {
         self.node.clone()
     }
-    /// The disk device the pool is located on
+    /// The disk devices that compose the pool.
     pub fn disks(&self) -> Vec<String> {
         self.disks.clone()
     }
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq, JsonSchema)]
-#[non_exhaustive]
 pub enum PoolState {
     /// The pool is a new OR missing resource, and it has not been created or
     /// imported yet by the operator. The pool spec MAY be but DOES
@@ -68,7 +71,6 @@ pub enum PoolState {
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq, JsonSchema, Default)]
-#[non_exhaustive]
 /// PoolState represents operator specific states for DSP CR.
 pub enum CrPoolState {
     /// The pool is a new OR missing resource, and it has not been created or
@@ -84,7 +86,6 @@ pub enum CrPoolState {
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq, JsonSchema)]
-#[non_exhaustive]
 /// PoolStatus is Control plane status of a given DSP CR.
 pub enum PoolStatus {
     /// State is Unknown.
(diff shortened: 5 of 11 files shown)


k8s/operators/src/diskpool/crd.rs
```diff
@@ -34,18 +34,21 @@ pub struct DiskPoolSpec {
...
+    /// Create a new DiskPoolSpec from the node and the disks.
+    pub fn new(node: String, disks: Vec<String>) -> Self {
+        Self { node, disks }
+    }
+    /// The node the pool is placed on.
     pub fn node(&self) -> String {
         self.node.clone()
     }
-    /// The disk device the pool is located on
+    /// The disk device the pool is located on.
```
tiagolobocastro:
```suggestion
    /// The disk devices that compose the pool.
```

k8s/operators/src/diskpool/crd.rs
```diff
@@ -34,18 +34,21 @@ pub struct DiskPoolSpec {
...
     pub fn disks(&self) -> Vec<String> {
         self.disks.clone()
     }
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq, JsonSchema)]
-#[non_exhaustive]
```
tiagolobocastro, line 48: Why did you remove this?

k8s/operators/src/diskpool/main.rs
```diff
@@ -728,17 +635,39 @@ async fn reconcile(dsp: Arc<DiskPool>, ctx: Arc<OperatorContext>) -> Result<Acti
...
+    match diskpool::ensure_crd(&k8s).await {
+        Ok(o) => {
+            info!(crd = ?o.name_any(), "Created");
+            // let the CRD settle this purely to avoid errors messages in the console
+            // that are harmless but can cause some confusion maybe.
+            tokio::time::sleep(Duration::from_secs(5)).await;
+        }
+
+        Err(e) => {
+            error!("Failed to create CRD error {}", e);
```
tiagolobocastro:
```suggestion
            error!(%error, "Failed to create CRD");
```

k8s/operators/src/diskpool/main.rs
```diff
@@ -728,17 +635,39 @@ async fn reconcile(dsp: Arc<DiskPool>, ctx: Arc<OperatorContext>) -> Result<Acti
...
+            tokio::time::sleep(Duration::from_secs(1)).await;
+            std::process::exit(1);
+        }
+    }
+
+    // Migrate the MayastorPool CRs to the DiskPool.
+    let _ = migrate_and_clean_msps(&k8s, namespace)
+        .await
+        .map_err(|error| {
+            error!("{error:?}");
```
tiagolobocastro:
```suggestion
            error!(%error, "Was not able to fully migrate existing v1 CRs");
```

k8s/operators/src/diskpool/mod.rs
```diff
@@ -1,2 +1,11 @@
+//! This module has the definitions and the client specific operations for the
+//! DiskPool CRs. This module is works with the current operator logic.
+
+/// DiskPool client operations.
+mod client;
 /// The DiskPool custom resource definition
-pub mod crd;
+mod crd;
```
tiagolobocastro: not changed, but can you add the `.` to the previous line

k8s/operators/src/lib.rs
```diff
@@ -1,2 +1,59 @@
...
+
+/// The DiskPool operator.
 pub mod diskpool;
+/// The error module for the operator.
+pub mod error;
+/// The MayaStorPool Operator definitions.
+pub mod mayastorpool;
+
+/// Migrate from MayastorPool
+pub async fn migrate_and_clean_msps(k8s: &Client, namespace: &str) -> Result<(), Error> {
```
tiagolobocastro: Shouldn't this be part of the binary rather than the library?

k8s/operators/src/lib.rs
```diff
@@ -1,2 +1,59 @@
...
+    // Check if the MayastorPool CRD is present, and migrate from it if it is.
+    match mayastorpool::check_crd(k8s).await {
+        // Fetch the MayastorPool CRs.
+        Ok(true) => match mayastorpool::list(k8s, namespace, 100).await {
```
tiagolobocastro: What's the 100?

k8s/operators/src/lib.rs
```diff
@@ -1,2 +1,59 @@
...
+                        diskpool::DiskPoolSpec::new(node, disks),
+                    )
+                    .await?;
+                }
+                info!("Migration of CRs from MayastorPool to DiskPool complete");
+                // Patch the finalizers and delete the MayastorPool CRs.
+                match mayastorpool::delete(k8s, namespace, &mut msps).await {
+                    Ok(_) => info!("Cleanup of MayastorPool CRs complete"),
+                    Err(error) => {
+                        return Err(Error::Generic {
```
tiagolobocastro: Shall we have a non-generic error? Here and below as well?
Also shouldn't we try to migrate and delete all other CR's, otherwise if the first fails we don't try to migrate the other ones?

k8s/operators/src/mayastorpool/client.rs
```diff
@@ -0,0 +1,86 @@
...
+    loop {
+        let mut result = pools_api.list(&list_params).await?;
+        pools.append(&mut result.items);
+        // Check for the token, if valid then continue.
+        match result.metadata.continue_ {
+            Some(token) if !token.is_empty() => {
+                list_params = list_params.continue_token(token.as_str())
+            }
+            _ => break,
+        };
```
tiagolobocastro:
```suggestion
        }
```

k8s/operators/src/mayastorpool/client.rs
```diff
@@ -0,0 +1,86 @@
...
+    pools_api
+        .delete_collection(&delete_params, &list_params)
+        .await?;
+
+    Ok(())
+}
+
+/// Check whether the MayastorPool crd exists or not.
+pub async fn check_crd(k8s: &Client) -> Result<bool, Error> {
+    let crd_api: Api<k8s_openapi::apiextensions_apiserver::pkg::apis::apiextensions::v1::CustomResourceDefinition> = Api::all(k8s.clone());
```
tiagolobocastro: wow that's long, should we shorten it with a use? :D

k8s/operators/src/diskpool/crd.rs
Abhinandan-Purkait: Why do we need this? This is causing issues with compilation.

k8s/operators/src/lib.rs
Abhinandan-Purkait: That's the pagination limit.

k8s/operators/src/mayastorpool/client.rs
Abhinandan-Purkait: Yeah, we can :-p

k8s/operators/src/lib.rs
Abhinandan-Purkait: We are deleting after complete migration only. The delete is outside the loop, once migration is done the list of pools we used for migration are passed to delete. If you look into the definition of the delete method. Its a delete_collection call, we aren't deleting one item at a time.

k8s/operators/src/lib.rs
Abhinandan-Purkait: This is a common code between the two modules. That's why kept in the library. By binary do you mean on main.rs?

k8s/operators/src/lib.rs
Abhinandan-Purkait: Maybe that should come from some constant. What do you think?

k8s/operators/src/mayastorpool/client.rs
Abhinandan-Purkait: Done. I did it in the other module but forgot here. 😅

k8s/operators/src/lib.rs
Abhinandan-Purkait: Added a constant. `PAGINATION_LIMIT`

tiagolobocastro: Yeah I saw it was pagination when reviewed the other file :D 
Great, this way it's clearer :+1:

k8s/operators/src/diskpool/crd.rs
tiagolobocastro: IIRC this is to ensure that changes to the fields (adding new enum values) are not breaking by ensuring we always use the catch all _ => {}

k8s/operators/src/lib.rs
tiagolobocastro: So the library is only here for the support tool to inspect the DiskPool CR IIRC. We can move the rest to diskpool which is only used by the diskpool binary IIRC, or perhaps just make sure we don't put the binary modules on the lib?

k8s/operators/src/lib.rs
tiagolobocastro: Ah ok.. even so, shouldn't we try to delete all of them rather than fail on the first error?

Abhinandan-Purkait: As discussed, we keep continuing on errors and the delete would be along with migration.

k8s/operators/src/diskpool/crd.rs
Abhinandan-Purkait: As discussed, removing this doesn't break anything.

k8s/operators/src/lib.rs
Abhinandan-Purkait: Done.

k8s/operators/src/lib.rs
Abhinandan-Purkait: Done. Moved this to binary. And the only binary module exposed via lib is crd.

k8s/operators/src/mayastorpool/client.rs
```diff
@@ -0,0 +1,82 @@
+use crate::{error::Error, mayastorpool::crd::MayastorPool};
+use k8s_openapi::apiextensions_apiserver::pkg::apis::apiextensions::v1::CustomResourceDefinition;
+use kube::{
+    api::{DeleteParams, ListParams, Patch, PatchParams},
+    Api, Client, CustomResourceExt,
+};
+
+const NOT_FOUND_CODE: u16 = 404;
```
tiagolobocastro: Simply use `StatusCode::NOT_FOUND`

k8s/operators/src/diskpool/client.rs
```diff
@@ -0,0 +1,115 @@
...
+use kube::{api::PostParams, Api, Client, CustomResourceExt, ResourceExt};
+
+use k8s_operators::{
+    diskpool::crd::{DiskPool, DiskPoolSpec},
+    error::Error,
+    mayastorpool,
+};
+use tracing::{error, info, warn};
+
+const ALREADY_EXISTS_CODE: u16 = 409;
```
tiagolobocastro: Can use `StatusCode::CONFLICT`

k8s/operators/src/mayastorpool/client.rs
```diff
@@ -0,0 +1,82 @@
...
+                list_params = list_params.continue_token(token.as_str())
+            }
+            _ => break,
+        }
+    }
+
+    Ok(pools)
+}
+
+/// Delete all the mayastor pool CRs forcefully by patching finalizers.
```
tiagolobocastro:
```suggestion
/// Delete the mayastor pool CR forcefully by patching finalizer.
```

k8s/operators/src/mayastorpool/crd.rs
```diff
@@ -0,0 +1,76 @@
...
+shortname = "msp",
+printcolumn = r#"{ "name":"node", "type":"string", "description":"node the pool is on", "jsonPath":".spec.node"}"#,
+printcolumn = r#"{ "name":"status", "type":"string", "description":"pool status", "jsonPath":".status.state"}"#,
+printcolumn = r#"{ "name":"capacity", "type":"integer", "format": "int64", "minimum" : "0", "description":"total bytes", "jsonPath":".status.capacity"}"#,
+printcolumn = r#"{ "name":"used", "type":"integer", "format": "int64", "minimum" : "0", "description":"used bytes", "jsonPath":".status.used"}"#,
+printcolumn = r#"{ "name":"available", "type":"integer", "format": "int64", "minimum" : "0", "description":"available bytes", "jsonPath":".status.available"}"#
+)]
+
+/// The pool spec which contains the parameters we use when creating the pool.
+pub struct MayastorPoolSpec {
```
tiagolobocastro: do these need to be pub or is pub(crate) enough? We don't need to expose these to the library?

k8s/operators/src/lib.rs
```diff
@@ -1,2 +1,8 @@
-/// The DiskPool operator
+/// The DiskPool operator.
 pub mod diskpool;
+
+/// Mayastor pool module.
+pub mod mayastorpool;
```
tiagolobocastro: We don't need to expose these right?

Abhinandan-Purkait: Changed them to pub(crate)

k8s/operators/src/mayastorpool/crd.rs
Abhinandan-Purkait: Changed them to pub(crate)

tiagolobocastro (approved):

Abhinandan-Purkait:
bors merge

PR #475, opened 2023-04-03
fix(ha): nvme path replacement timeout

mtzaurus:
Currently HA node agent waits indefinitely till new path is connected while replacing a broken NVMe path, which can lead to gRPC timeouts on HA Cluster agent and may leave dangling NVMe controllers in the system. The fix introduces a tuneable timeout with the default value of 11 seconds.


mtzaurus:
bors try

tiagolobocastro (approved):

mtzaurus:
bors merge

PR #477, opened 2023-04-04
feat(upgrade): disable v1 compatibility mode when v1 nexus_infos are migrated

Abhinandan-Purkait:


control-plane/agents/src/bin/core/controller/registry.rs
```diff
@@ -186,9 +188,26 @@ impl Registry {
         config
     }
 
+    /// Disable the v1 compat mode, once all nexus info keys are migrated.
+    pub(crate) async fn disable_v1_compat<S: Store>(store: &mut S) -> CoreRegistryConfig {
+        let config = CoreRegistryConfig::new(NodeRegistration::Automatic);
```
tiagolobocastro: we should not do this, we're overwriting the config altogether..

Abhinandan-Purkait:
~~Need to add the cleanup part.~~

Abhinandan-Purkait:
bors try

control-plane/agents/src/bin/core/controller/registry.rs
```diff
@@ -130,19 +131,25 @@ impl Registry {
...
         };
         registry.init().await;
 
-        // Remove all entries of v1 key prefix.
-        store
-            .delete_values_prefix(&product_v1_key_prefix())
-            .await
-            .expect("Product v1 key prefix deletion should not fail");
+        // Disable v1 compat if nexus_info keys are migrated.
+        if !registry.nexus_info_v1_present().await {
```
tiagolobocastro: hmm shouldn't this be negated, if v1 is present?
Also I think this should pre preceded by if compat is enable to avoid doing this everytime?
registry.compat_v1() && registry.nexus_info_v1_present()

control-plane/agents/src/bin/core/controller/registry.rs
```diff
@@ -130,19 +131,25 @@ impl Registry {
...
+            // Delete the v1 nexus_info keys by brute force.
+            delete_all_v1_nexus_info(&mut store)
+                .await
+                .expect("v1 nexus_info deletion should not fail");
+            // Disable the v1 compat mode.
+            registry.set_config(Self::disable_v1_compat(&mut store).await);
```
tiagolobocastro: can we move this entire block to a separate function and have a single catch all for the Result::Err instead of separate panics?
And the panic message should be something related to persistent store I guess?

utils/pstor/src/products/v1.rs
```diff
@@ -26,3 +28,16 @@ pub async fn detect_product_v1_prefix<S: Store>(store: &mut S) -> Result<bool, c
     let prefix = store.get_values_prefix(&key_prefix()).await?;
     Ok(!prefix.is_empty())
 }
+
+/// Deletes all v1 nexus_info by fetching all keys and parsing the key to UUID and deletes on
+/// success.
+pub async fn delete_all_v1_nexus_info<S: Store>(store: &mut S) -> Result<(), crate::Error> {
+    let kvs = store.get_values_prefix("").await.unwrap();
```
tiagolobocastro: return error

tiagolobocastro: Also we should probably use paging here, as I'm sure the clusters will be filled with these...

control-plane/agents/src/bin/core/controller/registry.rs
Abhinandan-Purkait: Which should be fine right we don't have any field expect v1 compat that changes?

control-plane/agents/src/bin/core/controller/registry.rs
Abhinandan-Purkait: The method returns true if the some v1 nexus_info is not migrated so we are cleaning up only when the method returns false, i.e. all keys migrated.

The v1 compat enable check is already a part of the method. We don't need to check it here.

utils/pstor/src/products/v1.rs
Abhinandan-Purkait: Ahh, sorry missed this.

Abhinandan-Purkait: I did try pagination, and it doesn't seem to work as expected. Would give another try.

control-plane/agents/src/bin/core/controller/registry.rs
Abhinandan-Purkait: Yeah probably the method name is wrong here, any suggestions?

tiagolobocastro: hmm the method doesn't check if compat is enabled at the start AFAICT... only if it failed to load the nexus info.. 
if compat is not enabled, then we don't need to do anything right? which is why I suggested we check if it's enabled here, though we can also check at the start of that function

Abhinandan-Purkait: Ahh okay. You mean don't do anything at all. Cool. I will move that check up. Thanks

control-plane/agents/src/bin/core/controller/registry.rs
tiagolobocastro: yes we do... you can modify etcd directly.. also let's make it ready for the future when we do expose the knobs to modify it, otherwise this will very silently corrupt the config.
Also this should take self Registry so we can also modify the in-memory config.

Abhinandan-Purkait: Sure. I will take the config from self and just modify the v1 compat and update accordingly. Thanks

control-plane/agents/src/bin/core/controller/registry.rs
```diff
@@ -186,9 +193,26 @@ impl Registry {
...
+            .await
+            .expect("Failed to persist configuration information");
+        info!("v1 compatibility is disabled");
+        config
+    }
+
     /// Get the `CoreRegistryConfig`
-    pub(crate) fn config(&self) -> &CoreRegistryConfig {
-        &self.config
+    pub(crate) fn config(&self) -> RwLockReadGuard<'_, RawRwLock, CoreRegistryConfig> {
```
```suggestion
    pub(crate) fn config(&self) -> parking_lot::RwLockReadGuard<CoreRegistryConfig> {
```
This should suffice..

utils/pstor/src/products/v1.rs
tiagolobocastro: hmm maybe a bug on the paged method

Abhinandan-Purkait: It worked. Have added the pagination.

control-plane/agents/src/bin/core/nexus/registry.rs
```diff
@@ -101,4 +110,45 @@ impl Registry {
...
+        let volspecs: Vec<VolumeSpec> = self
+            .specs()
+            .volumes()
+            .into_iter()
+            .filter(|vol| vol.target_config.is_some())
+            .collect();
+
+        for volspec in volspecs {
+            let volume_uuid = volspec.uuid;
+            let nexus_uuid = match &volspec.target_config {
```
tiagolobocastro:
```suggestion
            let nexus_uuid = match volspec.target() {
```

control-plane/agents/src/bin/core/nexus/registry.rs
```diff
@@ -101,4 +110,45 @@ impl Registry {
...
+
+    /// Check for the presence of v1 nexus_info associated to the current volumes.
+    /// This method only checks for v1 nexus_info key, if v1_compat is enabled.
+    pub(crate) async fn nexus_info_v1_migrated(&self) -> bool {
+        // Filter the volumes having a target.
+        let volspecs: Vec<VolumeSpec> = self
+            .specs()
+            .volumes()
+            .into_iter()
+            .filter(|vol| vol.target_config.is_some())
```
tiagolobocastro:
```suggestion
            .filter(|vol| vol.target())
```

control-plane/agents/src/bin/core/nexus/registry.rs
```diff
@@ -101,4 +110,45 @@ impl Registry {
...
+                // On entry missing error, load the nexus info with v1 compat.
+                if self
+                    .load_obj::<NexusInfo>(&nexus_info_key.with_mayastor_compat_v1(true))
+                    .await
+                    .is_ok()
+                {
+                    // return and terminate the loop.
+                    return true;
+                }
+            };
```
tiagolobocastro:
```suggestion
            }
```

control-plane/agents/src/bin/core/nexus/registry.rs
```diff
@@ -101,4 +110,45 @@ impl Registry {
...
+
+            // Load the nexus info with compat disabled.
+            if let Err(SvcError::StoreMissingEntry { .. }) =
+                self.load_obj::<NexusInfo>(&nexus_info_key).await
+            {
+                // On entry missing error, load the nexus info with v1 compat.
+                if self
+                    .load_obj::<NexusInfo>(&nexus_info_key.with_mayastor_compat_v1(true))
+                    .await
+                    .is_ok()
```
tiagolobocastro: Do we need to check for the actual error? Let's say the load failed because of a network error?

utils/pstor/src/products/v1.rs
```diff
@@ -26,3 +31,33 @@ pub async fn detect_product_v1_prefix<S: Store>(store: &mut S) -> Result<bool, c
...
+            }
+        }
+
+        if let Some((key, _)) = kvs.last() {
+            prefix = key;
+        } else {
+            break;
+        }
+    }
+    info!("v1 nexus_info cleaned up successfully");
```
tiagolobocastro:
```suggestion
    info!("v1.0.x nexus_info cleaned up successfully");
```

control-plane/agents/src/bin/core/nexus/registry.rs
Abhinandan-Purkait: Done.

control-plane/stor-port/src/types/v0/store/volume.rs
```diff
@@ -310,6 +310,10 @@ impl VolumeSpec {
         }
         self.target_config.as_ref().map(|c| &c.target.nexus)
     }
+    /// Get `uuid`.
+    pub fn uuid(&self) -> &VolumeId {
```
tiagolobocastro: There's already a trait for this: `ResourceUid`.

tiagolobocastro (approved):

control-plane/stor-port/src/types/v0/store/volume.rs
Abhinandan-Purkait: Ahh okay. Will use that.

Abhinandan-Purkait:
bors merge

Abhinandan-Purkait:
bors merge

PR #482, opened 2023-04-09
fix(nexusInfo) : Delete NexusInfo during VolumeDestroy when Nexus was already deleted

hasethuraman:
When a Nexus is destroyed and later when the Volume destroy comes, the existing delete_nexus_info will not be invoked since nexus target is already deleted as part of Nexus destroy. Hence in this change is to call it explicitly using the target information present in VolumeSpec. 

The delete_nexus_info is safe to call again because the underlying delete function returns ok if the KV is not present. Can be avoided if this is going to be a frequently hit condition. Otherwise leaving it for code readability.

Fixing https://github.com/openebs/mayastor/issues/1157

### DIFF
--- a/control-plane/agents/src/bin/core/volume/operations.rs
+++ b/control-plane/agents/src/bin/core/volume/operations.rs
@@ -178,6 +178,19 @@ impl ResourceLifecycle for OperationGuardArc<VolumeSpec> {
             }
         }
 
+        // When nexus is destroyed ahead of the volume destroy, then
+        // delete_nexus_info in previous will not be called since nexus won't be present.
+        // So invoke delete_nexus_info explicitly using the nexus id in target_config if present.
+        if let Some(config) = self.as_ref().config() {
+            let nexus_id = config.target().nexus();
+            // Delete the NexusInfo entry persisted by the IoEngine.
+            ResourceSpecsLocked::delete_nexus_info(
+                &NexusInfoKey::new(&Some(self.uuid().clone()), nexus_id),
+                registry,
+            )
+            .await;
+        }
+
         let replicas = specs.volume_replicas(&request.uuid);
         for replica in replicas {
             let mut replica = match replica.operation_guard_wait().await {


Abhinandan-Purkait:
@hasethuraman Can you please sign your commit? The DCO expects the commits to be signed.
Also your commit message needs to be of this format, ref:- https://github.com/conventional-changelog/commitlint/#what-is-commitlint

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -178,6 +178,19 @@ impl ResourceLifecycle for OperationGuardArc<VolumeSpec> {
             }
         }
 
+        // When nexus is destroyed ahead of the volume destroy, then
+        // delete_nexus_info in previous will not be called since nexus wont be present.
```
```suggestion
        // delete_nexus_info in previous will not be called since nexus won't be present.
```

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -178,6 +178,19 @@ impl ResourceLifecycle for OperationGuardArc<VolumeSpec> {
...
+        // So invoke delete_nexus_info explicitly using the nexus id in target_config if present.
+        if let Some(config) = self.as_ref().config() {
+            let nexus_id = config.target().nexus().clone();
+            // Delete the NexusInfo entry persisted by the IoEngine.
+            ResourceSpecsLocked::delete_nexus_info(
+                &NexusInfoKey::new(&Some(self.uuid().clone()), &nexus_id),
```
tiagolobocastro: looks like we take a reference to the nexus_id, so maybe we don't need to clone it on L185?

tiagolobocastro (approved):
Great, thank you @hasethuraman!
Just a small nit (if it works).

tiagolobocastro:
This solves the issue going forward but we presumably still have a bunch of stale entries leftover.
Is it worth coming up with a way to clean those up?

tiagolobocastro:
@hasethuraman could you please amend your commit message to comply with conventional commits? Just need to change Delete to lowercase delete I think.

Abhinandan-Purkait (approved):

control-plane/agents/src/bin/core/volume/operations.rs
hasethuraman: Thank you. Addressed

hasethuraman:
> @hasethuraman could you please amend your commit message to comply with conventional commits? Just need to change Delete to lowercase delete I think.

Thanks @tiagolobocastro. Pushed them. Also, I am yet to figure out why commitlint is failing to get installed in my setup.

hasethuraman:
> This solves the issue going forward but we presumably still have a bunch of stale entries leftover. Is it worth coming up with a way to clean those up?

Do you remember any other KVs stale scenarios?

Probably we can compare the states to specs and detect the dangling KVs in etcd in a reconciler. And at some point in future, this reconciler can be disabled once we find the codes that is supposed to do it.

tiagolobocastro:
> > @hasethuraman could you please amend your commit message to comply with conventional commits? Just need to change Delete to lowercase delete I think.
> 
> Thanks @tiagolobocastro. Pushed them. Also, I am yet to figure out why commitlint is failing to get installed in my setup.

Ah interesting... should be installed via nix-shell when you enter it IIRC

One final ask @hasethuraman, could you please squash your commits into a single 1? We use bors which does a merge and includes all commits from a PR so we always squash logically related commits into 1 (eg: review comments)
Thanks!

tiagolobocastro:
> > This solves the issue going forward but we presumably still have a bunch of stale entries leftover. Is it worth coming up with a way to clean those up?
> 
> Do you remember any other KVs stale scenarios?
> 
> Probably we can compare the states to specs and detect the dangling KVs in etcd in a reconciler. And at some point in future, this reconciler can be disabled once we find the codes that is supposed to do it.

Yes, we might be able to do something like that, 1-time "jobs" or maybe when the core-agent starts after an update.
Another approach could be exposing a set of "jobs" that the CLI may be able to request via REST, though that would be more changes.

tiagolobocastro:
bors merge

tiagolobocastro:
@hasethuraman could you fix the linter error please:
```
error: this expression creates a reference which is immediately dereferenced by the compiler

   --> control-plane/agents/src/bin/core/volume/operations.rs:188:64

    |

188 |                 &NexusInfoKey::new(&Some(self.uuid().clone()), &nexus_id),

    |                                                                ^^^^^^^^^ help: change this to: `nexus_id`

    |

    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_borrow

    = note: `-D clippy::needless-borrow` implied by `-D warnings`

```

hasethuraman:
> @hasethuraman could you fix the linter error please:
> 
> ```
> error: this expression creates a reference which is immediately dereferenced by the compiler
> 
>    --> control-plane/agents/src/bin/core/volume/operations.rs:188:64
> 
>     |
> 
> 188 |                 &NexusInfoKey::new(&Some(self.uuid().clone()), &nexus_id),
> 
>     |                                                                ^^^^^^^^^ help: change this to: `nexus_id`
> 
>     |
> 
>     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#needless_borrow
> 
>     = note: `-D clippy::needless-borrow` implied by `-D warnings`
> ```

Fixed it and I dont see any linter errors locally. I hope it must go through fine there.

tiagolobocastro:
bors merge

hasethuraman:
Thanks @tiagolobocastro for helping out.

tiagolobocastro:
Thanks for finding and fixing this :+1: 
I'll cherry-pick it to the release/2.1 branch tomorrow.

PR #481, opened 2023-04-06
Thin provisioning - add configuration arguments and pool commitment filter

tiagolobocastro:
refactor(capacity/thin): expose cli/env knobs to control overcommitment

Signed-off-by: Tiago Castro <tiagolobocastro@gmail.com>

---

feat(capacity/thin): add pool commitment filter

Allows us to configure the max pool commitment in percentage.
Example: if the commitment is 250 and the pool is 10GiB we can overcommit the pool up to 25GiB
(create 2 10GiB and 1 5GiB volume) but no further.

Signed-off-by: Tiago Castro <tiagolobocastro@gmail.com>

---

feat(openapi/pool): expose pool committed bytes

This allows us to easily see if a pool is overcommitted.

Signed-off-by: Tiago Castro <tiagolobocastro@gmail.com>

PR #484, opened 2023-04-12
Cherry-pick #481 and #482 into release/2.1

tiagolobocastro:
482: fix(nexusInfo) : Delete NexusInfo during VolumeDestroy when Nexus was already deleted r=tiagolobocastro a=hasethuraman

When a Nexus is destroyed and later when the Volume destroy comes, the existing delete_nexus_info will not be invoked since nexus target is already deleted as part of Nexus destroy. Hence in this change is to call it explicitly using the target information present in VolumeSpec. 

The delete_nexus_info is safe to call again because the underlying delete function returns ok if the KV is not present. Can be avoided if this is going to be a frequently hit condition. Otherwise leaving it for code readability.

Fixing https://github.com/openebs/mayastor/issues/1157

Co-authored-by: hsethuraman <hsethuraman@microsoft.com>

---

481: Thin provisioning - add configuration arguments and pool commitment filter r=tiagolobocastro a=tiagolobocastro

refactor(capacity/thin): expose cli/env knobs to control overcommitment

Co-authored-by: Tiago Castro <tiagolobocastro@gmail.com>


Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

PR #490, opened 2023-04-14
fix(capacity/thin): don't add children for rebuild if there's no space

tiagolobocastro:
Don't attempt to add replicas to a nexus for rebuild if the replicas has no space to grow. This ensures we follow the same logic for add and online. Also, delete these replicas otherwise we're keeping them around even though we can't use them.


Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

PR #494, opened 2023-04-19
chore(rebuild/policies): invariant -  return zero duration if only one online child remains

dsharma-dc:


control-plane/agents/src/bin/core/controller/policies/rebuild_policies.rs
```diff
@@ -95,13 +93,17 @@ impl Policy for SystemPerf {
     }
 }
 
+fn online_child_count_check(nexus: &Nexus) -> i32 {
+    let mut online_child_cnt = 0;
+    for _ in nexus.children.iter().filter(|c| c.state.online()) {
+        online_child_cnt += 1;
+    }
+    online_child_cnt
+}
```
tiagolobocastro: You could consider even adding this to Nexus object itself I guess, or just calling the code directly since it can be a one liner like this:
```suggestion
fn online_child_count_check(nexus: &Nexus) -> usize {
    nexus.children.iter().filter(|c| c.state.online()).count()
}
```

control-plane/agents/src/bin/core/controller/policies/rebuild_policies.rs
```diff
@@ -50,27 +48,23 @@ impl RuleSet {
...
             RuleSet::SystemPerf(pol) => pol.faulted_child_wait_duration(nexus),
-        } */
+        }
     }
 
     /// Assign a rebuild policy to the nexus.
-    fn assign(size: u64) -> Self {
+    fn assign(_size: u64) -> Self {
+        RuleSet::SystemPerf(SystemPerf {})
+        /*
```
tiagolobocastro, line 62: shall we remove this commented out code?

dsharma-dc: This will be required at some point, it can remain harmlessly here I guess.

control-plane/agents/src/bin/core/controller/policies/rebuild_policies.rs
dsharma-dc: Sure, this one liner looks good.

control-plane/agents/src/bin/core/controller/policies/rebuild_policies.rs
tiagolobocastro: It's harmless but it's also pointless, we use git anyway so it's in the history anyway if you want to recover it

control-plane/agents/src/bin/core/controller/policies/rebuild_policies.rs
```diff
@@ -86,7 +80,11 @@ trait Policy {
 }
 
 impl Policy for SystemPerf {
-    fn faulted_child_wait_duration(&self, _nexus: &Nexus) -> Duration {
+    fn faulted_child_wait_duration(&self, nexus: &Nexus) -> Duration {
+        if 1 == online_child_count_check(nexus) {
```
abhilashshetty04, line 84: 1 == online_child_count_check(nexus) isnt that clear. can we make it online_child_count_check(nexus) == 1 ? We would like to check online child count to a number right?

dsharma-dc:
Converted this PR to draft. Keeping status quo for policy behaviour. With this change,  a reconciler issue is uncovered that is documented in GTM-669.

tiagolobocastro:
Reopen if required.

PR #495, opened 2023-04-20
fix: wait for no shutdown nexuses

chriswldenyer:
POC. Wait until there are no shutdown nexuses before marking a node as drained.
TODO: add a timeout

### DIFF
--- a/control-plane/agents/src/bin/core/controller/reconciler/node/nexus.rs
+++ b/control-plane/agents/src/bin/core/controller/reconciler/node/nexus.rs
@@ -145,19 +145,56 @@ async fn check_and_drain_node(context: &PollContext, node_spec: &NodeSpec) -> Po
     }
     // Change the node state to "drained"
     if !move_failures {
-        if let Err(error) = context
-            .specs()
-            .set_node_drained(context.registry(), node_spec.id())
-            .await
-        {
-            tracing::error!(
-                %error,
-                node.id = node_id.as_str(),
-                "Failed to set node to state drained"
-            );
-            return PollResult::Err(error);
+        // get the list of volumes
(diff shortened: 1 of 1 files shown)


control-plane/agents/src/bin/core/controller/reconciler/node/nexus.rs
```diff
@@ -145,19 +145,56 @@ async fn check_and_drain_node(context: &PollContext, node_spec: &NodeSpec) -> Po
...
-            .set_node_drained(context.registry(), node_spec.id())
-            .await
-        {
-            tracing::error!(
-                %error,
-                node.id = node_id.as_str(),
-                "Failed to set node to state drained"
-            );
-            return PollResult::Err(error);
+        // get the list of volumes
```
tiagolobocastro, line 148: So I think here you probably want to keep a timestamp to start counting towards the timeout? Maybe we can have an in-memory timeout in the nexus draining state?

control-plane/agents/src/bin/core/controller/reconciler/node/nexus.rs
```diff
@@ -145,19 +145,56 @@ async fn check_and_drain_node(context: &PollContext, node_spec: &NodeSpec) -> Po
...
+        let vol_specs = context.specs().volumes_rsc();
+        // for each volume call volume_shutdown_nexuses()
+        let mut count_shutdown_nexuses = 0;
+        for vol_spec in vol_specs {
+            tracing::info!(volume.id = node_id.as_str(), "found volume");
+
+            match vol_spec.operation_guard() {
```
tiagolobocastro, line 155: don't think you need to get the guard here since we're not making changes.

chriswldenyer:
Superseded by a later PR

PR #500, opened 2023-04-24
test(partial_rebuild): unit tests for partial rebuild

dsharma-dc:
Unit tests for:

1. Wait and partial rebuild
2. Wait and full rebuild

### DIFF
--- a/control-plane/agents/src/bin/core/tests/rebuild/mod.rs
+++ b/control-plane/agents/src/bin/core/tests/rebuild/mod.rs
@@ -1,0 +1,217 @@
+//! This file contains tests for rebuild scenarios, specifically around
+//! partial(log-based) rebuild, but not limited to that.
+
+use deployer_cluster::{Cluster, ClusterBuilder};
+
+use std::{collections::HashMap, time::Duration};
+use stor_port::types::v0::{
+    openapi::{
+        models,
+        models::{
+            ChildState, CreateVolumeBody, Pool, PoolStatus, PublishVolumeBody, Replica,
+            ReplicaState, VolumePolicy, VolumeStatus,
+        },
+    },
+    transport::VolumeId,
+};
+
+async fn build_cluster(num_ioe: u32, pool_size: u64, _num_pools: u32) -> Cluster {
+    let reconcile_period = Duration::from_secs(1);
+    let child_twait = Duration::from_secs(15);
+    ClusterBuilder::builder()
+        .with_rest(true)
+        .with_io_engines(num_ioe)
+        .with_faulted_child_wait_period(child_twait)
+        .with_reconcile_period(reconcile_period, reconcile_period)
+        .with_cache_period("1s")
+        .with_tmpfs_pool(pool_size)
+        .build()
+        .await
+        .unwrap()
+}
+
+async fn wait_till_pool_online(
+    cluster: &Cluster,
+    replica: &Replica,
+    timeout: Duration,
+) -> Option<Pool> {
+    let client = cluster.rest_v00();
+    let pool_api = client.pools_api();
+    let start = std::time::Instant::now();
+
+    loop {
+        let pool = pool_api.get_pool(replica.pool.as_str()).await.unwrap();
+        tracing::info!("Waiting for pool to be Online - {:?}", pool);
+        if pool.state.is_some() && (pool.state.clone().unwrap().status == PoolStatus::Online) {
+            tracing::info!("Pool is Online now");
+            return Some(pool);
+        }
+
+        if std::time::Instant::now() > (start + timeout)
+            && (pool.state.is_none() || pool.state.as_ref().unwrap().status != PoolStatus::Online)
+        {
+            tracing::info!("Timeout waiting for pool to be Online - {:?}", pool);
+            return None;
+        }
+        tokio::time::sleep(Duration::from_secs(1)).await;
+    }
+}
+
+// This test
+// 1. creates a volume with 3 replicas, faults one of them by restarting container.
+// 2. Expects child's replica to be online again within 15sec period.
+// 3. Attempts to online the child.
+// 4. Child should get successfully onlined and partial rebuild happen.
+// 5. If online call fails, full rebuild of child happens.
+#[tokio::test]
+async fn replica_wait_and_partial_rebuild() {
+    let cluster = build_cluster(4, 52428800, 0).await; // 50MiB pool size
+
+    let vol_target = cluster.node(0).to_string();
+    let api_client = cluster.rest_v00();
+    let volume_api = api_client.volumes_api();
+
+    let body = CreateVolumeBody::new(VolumePolicy::new(true), 3, 10485760u64, false);
+    let volid = VolumeId::new();
+    let volume = volume_api.put_volume(&volid, body).await.unwrap();
+    let volume = volume_api
+        .put_volume_target(
+            &volume.spec.uuid,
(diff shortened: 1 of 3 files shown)


dsharma-dc:
Opened for review.
@tiagolobocastro The rebuild related rpcs are not exposed via grpc client for NexusOperations so couldn't check that for now to validate eventual volume state. NexusOperations has very minimal operations supported currently.

dsharma-dc:
bors try

control-plane/agents/src/bin/core/tests/rebuild/mod.rs
```diff
@@ -0,0 +1,260 @@
...
+    loop {
+        let pool = pool_api.get_pool(poolid.as_str()).await.unwrap();
+        tracing::info!("Waiting for pool to be Online - {:?}", pool);
+        if pool.state.is_some() && (pool.state.clone().unwrap().status == PoolStatus::Online) {
+            tracing::info!("Pool is Online now");
+            return Some(pool);
+        }
+
+        if std::time::Instant::now() > (start + timeout)
+            && (pool.state.is_none() || pool.state.clone().unwrap().status != PoolStatus::Online)
```
tiagolobocastro:
```suggestion
            && (pool.state.is_none() || pool.state.as_ref().unwrap().status != PoolStatus::Online)
```

control-plane/agents/src/bin/core/tests/rebuild/mod.rs
```diff
@@ -0,0 +1,260 @@
...
+    assert_eq!(nexus.children.len(), 3);
+    nexus
+        .children
+        .iter()
+        .for_each(|c| assert!(c.state == ChildState::Online.into()));
+
+    // Check where the replicas are, apart from vol target node.
+    let replicas = api_client.replicas_api().get_replicas().await.unwrap();
+    tracing::info!({?replicas}, "Available replicas - ");
+    let testrep = replicas.iter().find(|r| r.node != vol_target).unwrap();
```
tiagolobocastro, line 104: what is testrep?

abhilashshetty04 (approved):

control-plane/agents/src/bin/core/tests/rebuild/mod.rs
```diff
@@ -0,0 +1,260 @@
...
+    // Can't see any other better way to get node index from name,
+    // so utilizing the container name format io-engine-<num>.
+    let nodeidx: u32 = testrep_node
+        .as_str()
+        .split('-')
+        .last()
+        .expect("Couldn't parse nodeidx")
+        .parse::<u32>()
+        .unwrap()
+        - 1;
```
tiagolobocastro: this seems a bit brittle, why do we need to do this?
Can't we just use node-name rather than node index?

control-plane/agents/src/bin/core/tests/rebuild/mod.rs
```diff
@@ -0,0 +1,260 @@
...
+    );
+
+    // Stop the container and don't restart.
+    cluster
+        .composer()
+        .stop(&testrep_node.to_string())
+        .await
+        .expect("container stop failure");
+    // Give reconciler time to pitch-in.
+    sleep(Duration::from_secs(5));
```
tiagolobocastro: What are we waiting for here? Should probably have some loop checking for a condition.
Also should not use blocking sleeps on async functions.

control-plane/agents/src/bin/core/tests/rebuild/mod.rs
dsharma-dc: testrep indicates the _test replica_ that is picked up for this test.

control-plane/agents/src/bin/core/tests/rebuild/mod.rs
dsharma-dc: to get the poolid of a pool on a node via cluster.pool() , we need the node index. Since the io-engine node being restarted is randomly chosen here, I didn't find any other way to get the index of node except utilising the name to derive that.

control-plane/agents/src/bin/core/tests/rebuild/mod.rs
dsharma-dc: I can remove this and put a looping check on container state.

control-plane/agents/src/bin/core/tests/rebuild/mod.rs
tiagolobocastro: I'm still confused, we have the node and pool in the replica, we can just pass that?

dsharma-dc: ok, I didn't know models Replica had pool String present. This should be ok I guess then. No need of node.

`let pool = api_client.pools_api().get_pool(testrep.pool.as_str()).await.unwrap();`

dsharma-dc:
bors try

control-plane/agents/src/bin/core/tests/rebuild/mod.rs
```diff
@@ -0,0 +1,217 @@
...
+    let replicas_now = api_client.replicas_api().get_replicas().await.unwrap();
+    // The replica picked for test must be online now again.
+    assert!(replicas_now.iter().any(|r| r.uri == testrep.uri));
+    replicas_now.iter().find(|r| {
+        if r.uri == testrep.uri {
+            assert!(r.state == ReplicaStatus::Online.into());
+            return true;
+        }
+        false
+    });
```
tiagolobocastro: You're not using anything from the find, did you mean to use for_each instead?

```suggestion
    replicas_now.iter().for_each(|r| {
        if r.uri == testrep.uri {
            assert!(r.state == ReplicaStatus::Online.into());
        }
    });    
```
or 
```suggestion
    replicas_now.iter().filter(|r| r.uri == testrep.uri).for_each(|r| {
        assert!(r.state == ReplicaStatus::Online.into());
    });
```

control-plane/agents/src/bin/core/tests/rebuild/mod.rs
```diff
@@ -0,0 +1,217 @@
...
+                models::VolumeShareProtocol::Nvmf,
+                None,
+                cluster.csi_node(0),
+            ),
+        )
+        .await
+        .unwrap();
+
+    let volume_state = volume.state;
+    assert!(volume_state.status == VolumeStatus::Online.into());
```
tiagolobocastro: Why all the into() in the asserts, should not be needed if using the right type.

tiagolobocastro (approved):

control-plane/agents/src/bin/core/tests/rebuild/mod.rs
dsharma-dc: Corrected to use `models` types.

control-plane/agents/src/bin/core/tests/rebuild/mod.rs
dsharma-dc: hmm yeah. preferred `find` because it short-circuits, but yes my usage is clumsy currently. How about this:

```
    if let Some(rep) = replicas_now.iter().find(|r| r.uri == testrep.uri) {
        assert!(rep.state == ReplicaState::Online);
    }
```

tiagolobocastro: ok... lgtm

dsharma-dc:
bors try

dsharma-dc:
bors merge

PR #503, opened 2023-04-28
Several Switchover fixes

tiagolobocastro:
test(switchover/bdd): add robustness tests
    
    Add tests to help exercise some corner cases seen during bug study.
    Tbh these are not very precise as it's difficult to re-create some corner cases,
    but nonetheless they are probably a decent starting point and can be used to help
    manually try to recreate these issues in a more automated fashion.
    
---

    fix(switchover): adds several corner case fixes
    
    Replace path is not idempotent in case the connection was established without moving path.
    gRPC Timeouts for connection are not enough which leads into moving the target
    when not necessary.
    Report of failed paths does not differentiate errors at all.
    Report of failed paths returns only 1 error.. we should have an error per path as we report.
     in batches.
    In case of no resources for republish volume the ha cluster retries in a loop with no delay.
    or backoff..
    In error case we don’t clean up shutdown targets.
    On switchover failure, failed paths are not cleared in the cluster agent.
    
---

    chore(python/bdd): pin to version 6
    
    Newer version (not sure which exactly) seems to generate the test cases without a name.
    todo: determine best version to use
    
---

    fix(ha-node): add subsystem sync for platform none
    
    On the deployer the udev code we have for listening to disk events does not seem to work.
    For now, assume it is incompatible and add a subsystem sync just for platform none.
    
### DIFF
--- a/control-plane/agents/src/bin/ha/cluster/switchover.rs
+++ b/control-plane/agents/src/bin/ha/cluster/switchover.rs
@@ -15,8 +15,8 @@ use stor_port::{
             SpecTransaction,
         },
         transport::{
-            DestroyShutdownTargets, GetController, NodeId, ReplacePath, RepublishVolume, VolumeId,
-            VolumeShareProtocol,
+            DestroyShutdownTargets, GetController, NodeId, NvmeSubsystem, ReplacePath,
+            RepublishVolume, VolumeId, VolumeShareProtocol,
         },
     },
 };
@@ -25,8 +25,7 @@ use tokio::sync::{
     Mutex,
 };
 use tonic::transport::Uri;
-use tracing::{error, info};
-use utils::NVME_TARGET_NQN_PREFIX;
+use tracing::{error, info, warn};
 
 fn client() -> impl VolumeOperations {
     core_grpc().volume()
@@ -34,7 +33,7 @@ fn client() -> impl VolumeOperations {
 
 /// Stage represents the steps for switchover request.
 #[derive(Serialize, Deserialize, Debug, PartialEq, Eq, Copy, Clone, Ord, PartialOrd)]
-pub enum Stage {
+pub(crate) enum Stage {
     /// Initialize switchover request.
     Init,
     /// Shutdown original/old volume target. Create new nexus for existing vol obj.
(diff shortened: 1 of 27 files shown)


control-plane/agents/src/bin/ha/cluster/switchover.rs
```diff
@@ -49,12 +48,33 @@ pub enum Stage {
...
+        *self.0.read()
+    }
+}
+
+impl PartialEq for SwitchOverStage {
+    fn eq(&self, other: &Self) -> bool {
+        *self.0.read() == *other.0.read()
+    }
+}
+impl Eq for SwitchOverStage {}
```
Abhinandan-Purkait, line 70: Is this supposed to be empty?

control-plane/agents/src/bin/ha/node/path_provider.rs
```diff
@@ -92,9 +97,41 @@ impl CachedNvmePathProvider {
     pub fn start(&mut self) -> anyhow::Result<impl std::future::Future<Output = ()> + '_> {
         Ok(ready(()))
     }
+    #[cfg(target_os = "linux")]
+    fn spawn_nvmeadm_syncer(&self) {
```
Abhinandan-Purkait, line 101: 👍🏻

control-plane/stor-port/src/transport_api/mod.rs
```diff
@@ -494,10 +516,14 @@ impl RequestMinTimeout {
     pub fn pool(&self) -> Duration {
         self.pool
     }
-    /// minimum timeout for nexus shutdown
+    /// minimum timeout for nexus shutdown.
```
Abhinandan-Purkait: Start with capital? Below as well?

tests/bdd/common/nvme.py
```diff
@@ -69,6 +69,7 @@ def nvme_connect(uri):
     hostnqn = nvme_hostnqn_arg(u)
 
     command = f"sudo nvme connect -t tcp -s {port} -a {host} -n {nqn} {hostnqn}"
+    print(command)
```
Abhinandan-Purkait, line 72: Do we need this print?

tests/bdd/features/ha/robustness.feature
```diff
@@ -0,0 +1,30 @@
+Feature: Switchover Robustness
+
+  Background:
+    Given a deployer cluster
+
+  Scenario: reconnecting the new target times out
+    Given a single replica volume
+    And a connected nvme initiator
+    And a reconnect_delay set to 15s
+    When we restart the volume target
```
Abhinandan-Purkait: `When we restart the volume target node` ?

tests/bdd/features/ha/robustness.feature
```diff
@@ -0,0 +1,30 @@
...
+    And the ha clustering republishes
+    Then the path should be established
+
+  Scenario: path failure with no free nodes
+    Given a 2 replica volume
+    And a connected nvme initiator
+    When we cordon the non-target node
+    And we stop the volume target
```
Abhinandan-Purkait: `And we stop the volume target node` ? Below as well?

tests/bdd/requirements.txt
```diff
@@ -1,4 +1,4 @@
-pytest-bdd
+pytest-bdd==6.0.0
```
Abhinandan-Purkait, line 1: Should we revert this as this `def _():` seems to be the way forward?
As per - https://github.com/pytest-dev/pytest-bdd/pull/536

Abhinandan-Purkait (approved):
LGTM. Few nits.

control-plane/agents/src/bin/ha/cluster/switchover.rs
tiagolobocastro: Yes, it's simply to indicate equivalence.

control-plane/stor-port/src/transport_api/mod.rs
tiagolobocastro: Modified all of them.

tests/bdd/common/nvme.py
tiagolobocastro: It's just a helper in case things fail, nothing gets printed when running pytest with default args.

tests/bdd/requirements.txt
tiagolobocastro: Let's keep it as is for now, and let this soak, just in case it gets changed upstream..
Then to remove this we should change all our testcode to def _()

tests/bdd/features/ha/robustness.feature
tiagolobocastro: Done to all

tiagolobocastro:
bors merge

abhilashshetty04 (approved):

tiagolobocastro:
bors merge

PR #504, opened 2023-05-02
fix(switchover/cluster): add old target reuse as fallback

tiagolobocastro:
It's possible for a connection to fail after we've republished. In this case, if there's no other nodes available we'll get stuck here forever. This change allows us to reuse the previously failed node if it's available again. Adds a bdd test for this scenario.

### DIFF
--- a/control-plane/agents/src/bin/core/volume/operations.rs
+++ b/control-plane/agents/src/bin/core/volume/operations.rs
@@ -448,10 +448,26 @@ impl ResourcePublishing for OperationGuardArc<VolumeSpec> {
 
         let mut older_nexus = specs.nexus(target_cfg.target().nexus()).await?;
         let mut move_nexus = true;
-
+        let mut nexus_node = None;
         match healthy_volume_replicas(&spec, &older_nexus.as_ref().node, registry).await {
             Ok(_) => {
-                if request.reuse_existing
+                let reuse_existing =
+                    match request.reuse_existing_fallback && !request.reuse_existing {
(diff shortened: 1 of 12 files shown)


tiagolobocastro:
bors try

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -473,7 +486,10 @@ impl ResourcePublishing for OperationGuardArc<VolumeSpec> {
         }
 
         // Get the newer target node for the new nexus creation.
-        let nexus_node = self.next_target_node(registry, request, true).await?;
+        let nexus_node = match nexus_node {
+            Some(node) => node,
+            None => self.next_target_node(registry, request, true).await?,
```
Abhinandan-Purkait: Just wondering, can two times calling `next_target_node` can produce different results(if something moved/changed in background), even if it does should it matter? 

Do we need to recall the `next_target_node` here? Can we simply error out on the `None` arm?

Abhinandan-Purkait (approved):
LGTM. This looks great. Thanks

control-plane/agents/src/bin/core/volume/operations.rs
tiagolobocastro: hmm good point.
No we can't error out on None, but I've reworked the code allowing us to fail on error, PTAL

Abhinandan-Purkait: Awesome. Thanks.

tiagolobocastro:
bors merge

PR #511, opened 2023-05-09
chore: wait for no shutdown nexuses

chriswldenyer:
Before setting a node to state drained, verify that none of the volumes hosted on that node has shutdown nexuses.


tiagolobocastro (approved):

chriswldenyer:
bors merge

PR #514, opened 2023-05-09
feat(snapshots): add snapshot creation endpoint to openapi

Abhinandan-Purkait:


control-plane/rest/openapi-specs/v0_api_spec.yaml
```diff
@@ -2094,6 +2133,17 @@ components:
         - replicas
         - size
         - thin
+    CreateSnapshotBody:
+      example:
+        parameters: {"param1": "value1"}
+      description: Create Snapshot Body
+      type: object
+      properties:
```
mtzaurus: Does it mean that we pass input body as an opaque key:value object without specifying its exact properties ?

Abhinandan-Purkait: No, this is as per the csi specs, the CreateSnapshotRequest has, 3 fields the snapshot name, the source and the parameters. We get the first 2 from url and thus I have put the parameters in the body. `parameters` is basically one of the fields in the payload.

control-plane/rest/openapi-specs/v0_api_spec.yaml
```diff
@@ -1575,6 +1575,41 @@ paths:
           $ref: '#/components/responses/ServerError'
       security:
         - JWT: []
+  '/volumes{volume_id}/snapshots/{snapshot_name}':
```
```suggestion
  '/volumes/{volume_id}/snapshots/{snapshot_name}':
```

control-plane/rest/openapi-specs/v0_api_spec.yaml
```diff
@@ -1575,6 +1575,41 @@ paths:
...
+    put:
+      tags:
+        - Volumes
+      operationId: put_volume_snapshots
```
```suggestion
      operationId: put_volume_snapshot
```

control-plane/rest/openapi-specs/v0_api_spec.yaml
```diff
@@ -2978,6 +3028,37 @@ components:
...
+        Snapshot Information
+      type: object
+      properties:
+        uuid:
+          $ref: '#/components/schemas/SnapshotId'
+        name:
+          description: Name of the snapshot
+          type: string
+        size:
+          description: Size that the snapshot should be.
```
```suggestion
          description: |-
            Size in bytes taken up by the snapshot in the backend.
            This size may grow larger as the original volume gets written to, up to the volume size.
```

control-plane/rest/openapi-specs/v0_api_spec.yaml
```diff
@@ -2978,6 +3028,37 @@ components:
...
+          type: integer
+          format: int64
+          minimum: 0
+        source:
+          $ref: '#/components/schemas/VolumeId'
+        creation_timestamp:
+          description: Timestamp when the point-in-time snapshot is taken on the storage system.
```
tiagolobocastro: We don't seem to be using this term "point-in-time" anywhere?

```suggestion
          description: Timestamp when snapshot is taken on the storage system.
```

control-plane/rest/openapi-specs/v0_api_spec.yaml
```diff
@@ -2978,6 +3028,37 @@ components:
...
+          type: string
+          format: date-time
+        ready:
+          description: Indicates if a snapshot is ready to use as a `volume_content_source` in a `CreateVolumeRequest`.
```
tiagolobocastro: Until we support such thing at rest, maybe simply:
```suggestion
          description: Indicates if a snapshot is ready to be used as a clone.
```

control-plane/rest/openapi-specs/v0_api_spec.yaml
```diff
@@ -2978,6 +3028,37 @@ components:
...
+          $ref: '#/components/schemas/SnapshotId'
+        name:
+          description: Name of the snapshot
+          type: string
+        size:
+          description: Size that the snapshot should be.
+          type: integer
+          format: int64
+          minimum: 0
+        source:
```
```suggestion
        source_volume:
```
Just to make it clear?

control-plane/rest/openapi-specs/v0_api_spec.yaml
Abhinandan-Purkait: Oops, how come this did not generate any error?

tiagolobocastro: it's still a valid path I guess, just not the one we want

Abhinandan-Purkait: Well, I am not sure if this qualifies as valid path, because without the `/` the url becomes variable right. Anyways that's a different discussion altogether.

control-plane/rest/openapi-specs/v0_api_spec.yaml
tiagolobocastro: Do we need to pass these at all for the moment? We don't pass them down in the volume create IIRC so maybe leave it out until it's needed?

control-plane/rest/openapi-specs/v0_api_spec.yaml
tiagolobocastro: It's a valid path :) 
/volumes6a666979-c222-4278-9aff-72053835abdc/snapshots/my-snap-1

Abhinandan-Purkait: No, what I mean is individually its an valid path but based on the value of `{volume_id}` the whole url would keep changing.

I had tested that with swagger and it was working for different values as well.

control-plane/rest/openapi-specs/v0_api_spec.yaml
Abhinandan-Purkait: We can remove it but, since its part of csi-spec it must have some significance that's why I had added it.

tiagolobocastro: Same as the create volume, they're opaque parameters from a class, but in this case from the snapshot class rather than the storage class AFAIK.
We may want to "de-opaque" some parameters to specific arguments like we do with create volume.

Abhinandan-Purkait: That's right. So shall I remove it for now?

tiagolobocastro: Yes I think so, we can always add it later

control-plane/rest/openapi-specs/v0_api_spec.yaml
```diff
@@ -2978,6 +3028,39 @@ components:
...
         - state
+    Snapshot:
+      description: |-
+        Snapshot Information
+      type: object
+      properties:
+        uuid:
+          $ref: '#/components/schemas/SnapshotId'
+        name:
+          description: Name of the snapshot
```
tiagolobocastro: Do we end these with `.` ? Others below have it, though I don't know if we have such practice already or not.

tiagolobocastro (approved):

dsharma-dc (approved):

control-plane/rest/openapi-specs/v0_api_spec.yaml
Abhinandan-Purkait: I think we should follow this. To be consistent. I will add wherever needed in my changes.

control-plane/rest/openapi-specs/v0_api_spec.yaml
```diff
@@ -2978,6 +3028,39 @@ components:
...
+          type: string
+        size:
+          description: |-
+            Size in bytes taken up by the snapshot in the backend.
+            This size may grow larger as the original volume gets written to, up to the volume size.
```
dsharma-dc: `This size may grow larger as the original volume gets written to`
This sounds like size of the snapshot being taken varies, which isn't true. I guess this intends to say that `subsequent snapshots keep getting bigger if the volume is being written to`

Abhinandan-Purkait: Okay. How about.
`Size of subsequent snapshots may grow larger as the original volume gets written to, up to the volume size.`

dsharma-dc: My suggestion:
`As the original volume keeps getting written to, the subsequent snapshots have a larger size compared to older point in time snapshots, maximum up to the volume size.`

Abhinandan-Purkait: Okay. Will update to this then.

Abhinandan-Purkait:
bors merge

tiagolobocastro:
bors cancel

control-plane/rest/openapi-specs/v0_api_spec.yaml
tiagolobocastro: Subsequent snapshot sizes do not change, it's always the same size, what changes is volume itself which becomes thin, and so as we write to the volume, the snapshot "allocation" is effectively growing in size.

I think we better have 2 different sizes, one is the thin snapshot size, and the other is the "allocated" size of the snapshot (not really of the snapshot but the extra allocation that it's done as result of writing to the volume). So initially "allocated" is 0, and it grows up to the volume size.

tiagolobocastro: Also, should we have a spec/state kind of object here, similar to most other resources?

Abhinandan-Purkait: Have split it, as per discussed.

tiagolobocastro (approved):

Abhinandan-Purkait:
bors merge

PR #515, opened 2023-05-09
Snaphot operation stubs

tiagolobocastro:


control-plane/agents/src/bin/core/controller/resources/operations.rs
```diff
@@ -149,3 +149,34 @@ pub(crate) trait ResourceShutdownOperations {
...
+#[async_trait::async_trait]
+pub(crate) trait ResourceSnapshotting {
+    type Create: Sync + Send;
+    type CreateOutput: Sync + Send + Sized;
+    type Destroy: Sync + Send;
+    type List: Sync + Send;
+    type ListOutput: Sync + Send + Sized;
+
+    /// Create a snapshot for the `Self` resource.
+    async fn create_snap(
```
Abhinandan-Purkait, line 165: Do we need the `_snap` suffix ?

chriswldenyer: IMHO keeping the suffix makes it easier to grep for the function if we have to. Otherwise we'd have to search through hundreds of create() functions.

tiagolobocastro: Not having the suffix can also make it more difficult to call if you have 2 traits with the same create name implemented for the same type, which happens here.

Abhinandan-Purkait: Right. Thanks

Abhinandan-Purkait (approved):
LGTM

tiagolobocastro:
bors merge

PR #519, opened 2023-05-12
fix(replica-topology): add child information even when replica node is down

tiagolobocastro:
We can still add child information when its replica node is down if the nexus is still up.


Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

PR #523, opened 2023-05-15
feat(snapshots/pstor): add initial snapshot definitions

tiagolobocastro:
These are the spec-like definitions that are stored in the pstor (etcd). As a reminder, we use a log-based approach where we first write-out what the operation is, eg: create, and upon success we complete by writing the final object. This allows us to recover in case of a crash for example.

It may be that we have only Create&Delete operations for a snapshot, it's not clear to me yet if we need an Update operation when using clones, but that is TBD.


control-plane/stor-port/src/types/v0/store/snapshots/mod.rs
```diff
@@ -0,0 +1,11 @@
+pub mod replica;
+pub mod volume;
+
+use serde::{Deserialize, Serialize};
+
+/// User specification of a snapshot.
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Default)]
+pub struct SnapshotSpec<SourceId> {
+    source_id: SourceId,
+    uuid: String,
```
tiagolobocastro: This will be SnapshotId which @dsharma-dc has in PR IIRC.

control-plane/stor-port/src/types/v0/store/snapshots/replica.rs
```diff
@@ -0,0 +1,52 @@
...
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Default)]
+pub struct ReplicaSnapshotMeta {
+    #[serde(skip)]
+    sequencer: OperationSequence,
+    /// Record of the operation in progress.
+    operation: Option<ReplicaSnapshotOperationState>,
+
+    /// Creation timestamp of the snapshot (set after creation time).
+    creation_timestamp: Option<DateTime<Utc>>,
```
tiagolobocastro, line 33: How should we store timestamp in etcd, epoch based counter or using formatted strings like DateTime ?

dsharma-dc: I don't think there would be a difference in terms of the how much space it takes to store in whatever way. Considering that, human readable format can be preferred.

control-plane/stor-port/src/types/v0/store/snapshots/volume.rs
```diff
@@ -0,0 +1,188 @@
...
+
+    /// Creation timestamp of the snapshot (set after creation time).
+    creation_timestamp: Option<DateTime<Utc>>,
+    /// Size of the snapshot (typically follows source size).
+    size: u64,
+    /// Transaction Id that defines this snapshot when it is created.
+    txn_id: String,
+    /// Replicas which "reference" to this snapshot as its parent, indexed by the transaction
+    /// id when they were attempted.
+    /// The "actual" snapshots can be accessed by the key `txn_id`.
```
Abhinandan-Purkait, line 43: Should the key be `txn_id` ? In that case what does the above `txn_id` field represent, the last transaction id?

control-plane/stor-port/src/types/v0/store/volume.rs
```diff
@@ -172,6 +172,18 @@ pub struct VolumeSpec {
     /// Affinity Group related information.
     #[serde(default)]
     pub affinity_group: Option<AffinityGroup>,
+    // /// Runtime volume information, useful to quick checks without having to read out from PSTOR
+    // /// or any other control-plane related registry.
+    // /// reviewer: is this something useful to keep?
+    // #[serde(skip)]
+    // runtime_info: VolumeRuntimeMetadata,
```
Abhinandan-Purkait, line 179: I think so, we can keep this. Helpful for list snapshot on volume level?

control-plane/stor-port/src/types/v0/store/snapshots/volume.rs
tiagolobocastro: Yes, the idea I had was the following: when create snapshot fails, we don't delete the snapshot. Next time we get called to create we generate new txn_id and create a new set of replicas. If this succeeds then this will be the final txn_id.

Abhinandan-Purkait: Cool. Thanks.

Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

PR #524, opened 2023-05-16
feat: allow for a range of 2 controller ids when publishing

tiagolobocastro:
Previously we had a strict limit of only 1 controller range. This effectively means only 1 initiator can connect to our volume target.
However it seems that when an initiator times out before the target "notices" this we end up getting a new connection whilst the old one is still in place, and thus the new connect call fails!

Why were we so strict? Because the current code in the dataplane can not guarantee that older initiators connected with allow_any set to true will be disconnected when we set a new allowed list!
So is it safe to bump this? We believe so, because currently CSI sets allowed list of 1 for all published volumes, so when updating this list the previous one should be booted out!

todo: update dataplane to boot out any connected host!


mtzaurus (approved):

tiagolobocastro:
bors merge

PR #531, opened 2023-05-17
fix(capacity/thin): don't remove enospc nexuses with no space in replicas

tiagolobocastro:
There's no point in faulting these nexuses because we can't write to the replicas anyway. Ensure that there's some slack left before attempting to remove these nexuses.

### DIFF
--- a/control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
+++ b/control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
@@ -536,44 +536,65 @@ pub(super) async fn faulted_nexus_remover(
     nexus: &mut OperationGuardArc<NexusSpec>,
     context: &PollContext,
 ) -> PollResult {
-    let nexus_uuid = nexus.uuid();
+    let Ok(nexus_state) = context.registry().nexus(nexus.uuid()).await else {
+        return PollResult::Ok(PollerState::Idle);
+    };
 
-    if let Ok(nexus_state) = context.registry().nexus(nexus_uuid).await {
-        if nexus_state.status == NexusStatus::Faulted {
-            let nexus = nexus.lock().clone();
-            let healthy_children = healthy_nexus_children(&nexus, context.registry()).await?;
-            let node = context.registry().node_wrapper(&nexus.node).await?;
-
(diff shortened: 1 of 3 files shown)


control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
```diff
@@ -536,44 +536,65 @@ pub(super) async fn faulted_nexus_remover(
...
+    let nexus_spec = nexus.as_ref();
+    let healthy_children = healthy_nexus_children(nexus_spec, context.registry()).await?;
+    let node = context.registry().node_wrapper(&nexus_spec.node).await?;
 
-            let node_online = node.read().await.is_online();
-            // only remove the faulted nexus when the children are available again
-            if node_online && !healthy_children.candidates().is_empty() {
-                faulted_nexus_remover(nexus, node).await?;
+    #[tracing::instrument(skip(nexus, node), fields(nexus.uuid = %nexus.uuid(), request.reconcile = true))]
+    async fn faulted_nexus_remover(
```
abhilashshetty04, line 552: This fn is defined twice in same file?
https://github.com/openebs/mayastor-control-plane/pull/531/files#diff-f9973091d7cb435f94abe5c9948848fcbde7f79e84952f4feb674a2774df0054R535

tiagolobocastro: It's not the same function, this was already like so. It allows us to attach a span only when we perform the actual removal

abhilashshetty04: But do you see this fn getting called during reconcilation. i dont see this fn getting called in nexus_reconciler.
Here: 
https://github.com/openebs/mayastor-control-plane/blob/fix-enospc-nx/control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs#L127

tiagolobocastro: It's called from the volume reconciler

abhilashshetty04 (approved):

tiagolobocastro:
bors merge

Abhinandan-Purkait (approved):

PR #532, opened 2023-05-18
Cherry-pick #531

tiagolobocastro:
fix(capacity/thin): don't remove enospc nexuses with no space in replicas

There's no point in faulting these nexuses because we can't write to the replicas anyway. Ensure that there's some slack left before attempting to remove these nexuses.


Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

PR #534, opened 2023-05-18
fix(core/io-engine): implement snapshot trait for the rpc client

tiagolobocastro:
Add trait to the rpc client rather than the resources, matching the existing infra. Also add generic parameter to allow implementing the same trait with different parameters.

### DIFF
--- a/control-plane/agents/src/bin/core/node/wrapper.rs
+++ b/control-plane/agents/src/bin/core/node/wrapper.rs
@@ -26,7 +26,9 @@ use stor_port::{
 };
 
 use crate::controller::{
-    io_engine::{NexusChildActionApi, NexusChildApi, NexusShareApi},
+    io_engine::{
+        NexusChildActionApi, NexusChildApi, NexusShareApi, NexusSnapshotApi, ReplicaSnapshotApi,
+    },
     registry::Registry,
     states::Either,
 };
@@ -34,6 +36,7 @@ use async_trait::async_trait;
 use parking_lot::RwLock;
 use std::{ops::DerefMut, sync::Arc};
 use stor_port::types::v0::transport::{
+    CreateNexusSnapshot, CreateNexusSnapshotResp, CreateReplicaSnapshot, CreateReplicaSnapshotResp,
     ImportPool, NexusChildAction, NexusChildActionContext, NexusChildActionKind,
 };
 use tracing::{debug, trace, warn};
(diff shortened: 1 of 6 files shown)


control-plane/agents/src/bin/core/node/wrapper.rs
```diff
@@ -1377,6 +1378,32 @@ impl NexusChildActionApi<NexusChildActionContextNode> for Arc<tokio::sync::RwLoc
...
+    async fn create_snapshot(
+        &self,
+        _request: &Self::CreateIn,
+    ) -> Result<Self::CreateOutput, SvcError> {
+        todo!()
+    }
+}
+
+#[async_trait]
+impl SnapshotApi<Replica> for Arc<tokio::sync::RwLock<NodeWrapper>> {
```
tiagolobocastro: Even so, it becomes quite difficult to call this... I'll push another commit where we split this into 2 traits.

tiagolobocastro: I've pushed an alternative approach with 2 snapshot traits with different method names, allowing us to call them in a nicer way, example:
```
node.create_repl_snapshot(...) vs
SnapshotApi::<Replica>::create_snapshot(&node)
```
We can also have 2 traits, same method
```
node.create_repl_snapshot(...) vs
SnapshotApi::<Replica>::create_snapshot(&node) vs
ReplicaSnapshot::create_snapshot(&node)
```
Let me know your thoughts @dsharma-dc @Abhinandan-Purkait

dsharma-dc: Agree about implementing for RpcClient.
The first approach (b10d9f44e517b4d3a0c71b1472856ee8376944e4) is exactly how I intended it in the beginning, but couldn't figure out generic parameters definition/usage owing to my language awareness barrier. :) 
What is the difficulty that you mention in calling this? Is it about the using full templated name to call?
Second approach is a bit more scattered though.
Another question: Is it mandatory to impl the trait for node wrapper, irrespective of approach?

tiagolobocastro: The difficulty is that you can't just call it directly, 
`node.create_snapshot()`
 on a module which imports both traits, because the compiler can't figure out which one to call. You have to disambiguate like so:
`SnapshotApi::<Replica>::create_snapshot(&node)`
which makes it a bit complicated to use..

dsharma-dc (approved):

tiagolobocastro:
bors merge

PR #539, opened 2023-05-22
feat(snapshot/create): add core agent create volume snapshot logic

tiagolobocastro:
Implements the core logic for creating a volume snapshot. A snapshot resource is created for creation or retry and a new txn id is allocated for this operation.
todo: should we start destroying previous txn id here?

If the volume is published then creation goes via the nexus rather than the replica, but in either case we prepare a ReplicaSnapshot so we can keep track of the operation. This allows us, for example, to clean up later if we fail/crash/etc.

todo: on retries, check if the txn_id has already been created! (in case the ctrl-plane crashes).

### DIFF
--- a/control-plane/agents/src/bin/core/controller/io_engine/v1/translation.rs
+++ b/control-plane/agents/src/bin/core/controller/io_engine/v1/translation.rs
@@ -403,8 +403,8 @@ impl AgentToIoEngine for CreateNexusSnapReplDescr {
     fn to_rpc(&self) -> Self::IoEngineMessage {
         Self::IoEngineMessage {
             replica_uuid: self.replica.to_string(),
-            snapshot_uuid: self.snap_uuid.as_ref().map(|u| u.to_string()),
-            skip: self.skip,
+            snapshot_uuid: Some(self.snap_uuid.to_string()),
+            skip: false,
         }
     }
 }
--- a/control-plane/agents/src/bin/core/volume/operations_helper.rs
+++ b/control-plane/agents/src/bin/core/volume/operations_helper.rs
@@ -169,7 +169,7 @@ impl OperationGuardArc<VolumeSpec> {
         state: VolumeState,
         spec_clone: VolumeSpec,
     ) -> Result<Volume, SvcError> {
-        // Create a ag guard to prevent candidate collision.
+        // Create an ag guard to prevent candidate collision.
         let _ag_guard = match registry.specs().get_or_create_affinity_group(&spec_clone) {
             Some(ag) => Some(ag.operation_guard_wait().await?),
             _ => None,
@@ -227,7 +227,7 @@ impl OperationGuardArc<VolumeSpec> {
                 .await?;
 
             // todo: we could ignore it here, since we've already removed it from the nexus
-            // now remove the replica from the pool
+            //  now remove the replica from the pool
             let result = self.destroy_replica(registry, remove.spec()).await;
 
             self.complete_update(registry, result, spec_clone).await?;
--- a/control-plane/agents/src/bin/core/volume/snapshot_helpers.rs
+++ b/control-plane/agents/src/bin/core/volume/snapshot_helpers.rs
@@ -2,21 +2,40 @@ use crate::controller::{
     registry::Registry,
     resources::{
         operations_helper::{GuardedOperationsHelper, SpecOperationsHelper},
-        OperationGuardArc, ResourceUid,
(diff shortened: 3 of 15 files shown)


control-plane/agents/src/bin/core/controller/io_engine/v1/translation.rs
```diff
@@ -403,8 +403,8 @@ impl AgentToIoEngine for CreateNexusSnapReplDescr {
     fn to_rpc(&self) -> Self::IoEngineMessage {
         Self::IoEngineMessage {
             replica_uuid: self.replica.to_string(),
-            snapshot_uuid: self.snap_uuid.as_ref().map(|u| u.to_string()),
-            skip: self.skip,
+            snapshot_uuid: Some(self.snap_uuid.to_string()),
+            skip: false,
```
Abhinandan-Purkait, line 407: Why are we hardcoding `false` here?

control-plane/agents/src/bin/core/volume/operations_helper.rs
```diff
@@ -169,7 +169,7 @@ impl OperationGuardArc<VolumeSpec> {
         state: VolumeState,
         spec_clone: VolumeSpec,
     ) -> Result<Volume, SvcError> {
-        // Create a ag guard to prevent candidate collision.
+        // Create an ag guard to prevent candidate collision.
```
Abhinandan-Purkait, line 172: Question: Do we need any special attention for `ag` vols? I guess we don't.

control-plane/agents/src/bin/core/volume/snapshot_helpers.rs
```diff
@@ -89,3 +89,25 @@ impl SpecOperationsHelper for VolumeSnapshot {
...
+        &GetPersistedNexusChildren::new_snapshot(volume),
+        registry,
+    )
+    .await?;
+
+    volume.trace(&format!("Snapshoteable replicas for volume: {children:?}"));
+
+    match children.candidates().as_slice() {
+        [item] => Ok(item.clone()),
+        [] => Err(SvcError::NoHealthyReplicas {
```
Abhinandan-Purkait, line 108: Is it always due to `NoHealthyReplicas` or do we need a new `NoSnapshotableReplicas` error type?

control-plane/agents/src/bin/core/volume/snapshot_operations.rs
```diff
@@ -31,16 +47,24 @@ impl ResourceSnapshotting for OperationGuardArc<VolumeSpec> {
...
-        _request: &Self::Create,
+        request: &Self::Create,
     ) -> Result<Self::CreateOutput, SvcError> {
-        OperationGuardArc::<VolumeSnapshot>::create(
+        let state = registry.volume_state(request.source_id()).await?;
+
+        let operation = VolumeOperation::CreateSnapshot(request.uuid().clone());
+        let spec_clone = self.start_update(registry, &state, operation).await?;
+
+        let snap_result = OperationGuardArc::<VolumeSnapshot>::create(
```
Abhinandan-Purkait, line 57: Do we not need the operation guard wait before this?

control-plane/agents/src/bin/core/controller/io_engine/v1/translation.rs
tiagolobocastro: We're only doing 1 replica at the moment, we don't need to consider multi-replica scenarios yet

control-plane/agents/src/bin/core/volume/operations_helper.rs
tiagolobocastro: We might be overlooking some scenario, but can't think of any atm

control-plane/agents/src/bin/core/volume/snapshot_helpers.rs
tiagolobocastro: For the moment it's simply no healthy replicas, we're not doing any checks to ascertain whether a replica is snapshoteable or not.

control-plane/agents/src/bin/core/volume/snapshot_operations.rs
tiagolobocastro: I don't think so, otherwise it defeats the point of doing operations inside the "parent's" update?

Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

PR #541, opened 2023-05-22
feat(snapshot/rpc): add service transport rpc for create&get

tiagolobocastro:


control-plane/grpc/proto/v1/misc/common.proto
```diff
@@ -110,61 +110,67 @@ enum ResourceKind {
...
 }
 
 // Filter by Node and Nexus id
-message NodeNexusFilter{
+message NodeNexusFilter {
   string node_id = 1;
   string nexus_id = 2;
 }
 
+// Filter by Snapshot id
```
Abhinandan-Purkait: `Filter by Volume Id and Snapshot Id` ?

Abhinandan-Purkait (approved):

control-plane/grpc/src/operations/volume/traits_snapshots.rs
```diff
@@ -0,0 +1,270 @@
...
+};
+
+use std::{collections::HashMap, convert::TryFrom};
+use stor_port::types::v0::{
+    store::SpecStatus,
+    transport::{SnapshotId, SnapshotTxId},
+};
+
+/// Volume snapshot information.
+pub trait ICreateVolumeSnapshot: Send + Sync + std::fmt::Debug {
```
Abhinandan-Purkait: Can we please rename this to something generic, then we can use the same for DeleteSnapshot as well. Or should we use a new one for that?

tiagolobocastro: yep, done.

control-plane/grpc/proto/v1/misc/common.proto
tiagolobocastro: Done

tiagolobocastro:
bors merge

PR #547, opened 2023-05-24
feat(snapshot): list snapshots grpc call

dsharma-dc:


control-plane/agents/src/bin/core/controller/io_engine/v1/replica.rs
```diff
@@ -131,4 +133,24 @@ impl crate::controller::io_engine::ReplicaSnapshotApi for super::RpcClient {
...
+                replica_uuid: request.replica.to_string(),
+            })
+            .await
+            .context(GrpcRequestError {
+                resource: ResourceKind::Replica,
+                request: "list_snapshots",
+            })?;
+
+        let ret = response.into_inner();
+        let ret_vec = ret.snapshots.iter().map(|s| s.try_to_agent()).collect();
```
dsharma-dc, line 153: Do we need explicit error handling here?

control-plane/stor-port/src/types/v0/transport/snapshot.rs
```diff
@@ -105,3 +105,9 @@ pub struct DestroyReplicaSnapshot {
     /// Incoming snapshot id to be deleted.
     pub snap_id: SnapshotId,
 }
+
+/// The request type to list the snapshots of a replica.
+pub struct ListReplicaSnapshots {
+    /// Incoming replica id to list snapshots for.
+    pub replica: ReplicaId,
```
tiagolobocastro: Should be optional to support listing from all replicas

control-plane/agents/src/bin/core/controller/io_engine/v1/replica.rs
tiagolobocastro: What do you mean by that?

tiagolobocastro (approved):
Please update to latest rpc to support optional replica id.

control-plane/agents/src/bin/core/controller/io_engine/v1/replica.rs
dsharma-dc: I was wondering if `try_to_agent()`' s return should be handled by `?` or it is automoatically propagated up since this function also has Result return type.

tiagolobocastro: It's probably being propagated via collect via return type inference to `Result<Vec<_>, _>`

control-plane/stor-port/src/types/v0/transport/snapshot.rs
dsharma-dc: Made optional and updated api submodule hash.

dsharma-dc:
bors merge

PR #548, opened 2023-05-24
fix(partial-rebuild/log): don't wait for replica if we can't partial rebuild

tiagolobocastro:
Update rpc api to access new rebuild log flag.
If there is no IO log in the child, this means we can't partial rebuild, so there's no point waiting for the replica to return, simply fault and full rebuild.


dsharma-dc (approved):

abhilashshetty04 (approved):

tiagolobocastro:
bors merge

tiagolobocastro:
bors cancel

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

PR #554, opened 2023-05-25
feat(snapshot/create): list replica snapshot on published vol snap

tiagolobocastro:
When snapping a published volume, list the individual replicas from their nodes. Fix bug which caused the operation create to not be updated on the snapshot.

### DIFF
--- a/control-plane/agents/src/bin/core/controller/io_engine/v1/translation.rs
+++ b/control-plane/agents/src/bin/core/controller/io_engine/v1/translation.rs
@@ -198,7 +198,7 @@ impl TryIoEngineToAgent for v1::replica::ReplicaSnapshot {
             snap_uuid: SnapshotId::try_from(self.snapshot_uuid.as_str()).map_err(|_| {
                 SvcError::InvalidUuid {
                     uuid: self.snapshot_uuid.to_owned(),
-                    kind: ResourceKind::VolumeSnapshot,
+                    kind: ResourceKind::ReplicaSnapshot,
                 }
             })?,
             snap_name: self.snapshot_name.clone(),
@@ -208,6 +208,7 @@ impl TryIoEngineToAgent for v1::replica::ReplicaSnapshot {
                 .timestamp
                 .clone()
                 .and_then(|t| std::time::SystemTime::try_from(t).ok())
+                // only valid on create, not on list! should we handle this as None ?
                 .unwrap_or(UNIX_EPOCH),
             replica_uuid: ReplicaId::try_from(self.replica_uuid.as_str()).map_err(|_| {
                 SvcError::InvalidUuid {
@@ -218,6 +219,7 @@ impl TryIoEngineToAgent for v1::replica::ReplicaSnapshot {
             replica_size: self.replica_size,
             entity_id: self.entity_id.clone(),
             txn_id: self.txn_id.clone(),
+            valid: self.valid_snapshot,
         })
     }
 }
--- a/control-plane/agents/src/bin/core/node/wrapper.rs
+++ b/control-plane/agents/src/bin/core/node/wrapper.rs
@@ -1,44 +1,41 @@
 use crate::{
     controller::{
-        io_engine::{GrpcClient, GrpcClientLocked, GrpcContext, NexusApi, PoolApi, ReplicaApi},
+        io_engine::{
+            GrpcClient, GrpcClientLocked, GrpcContext, NexusApi, NexusChildActionApi,
+            NexusChildApi, NexusShareApi, NexusSnapshotApi, PoolApi, ReplicaApi,
+            ReplicaSnapshotApi,
+        },
+        registry::Registry,
         resources::ResourceUid,
-        states::{ResourceStates, ResourceStatesLocked},
+        states::{Either, ResourceStates, ResourceStatesLocked},
     },
     node::{service::NodeCommsTimeout, watchdog::Watchdog},
     pool::wrapper::PoolWrapper,
     NumRebuilds,
 };
-
 use agents::errors::SvcError;
+use grpc::operations::snapshot::SnapshotInfo;
 use stor_port::{
     transport_api::{Message, MessageId, ResourceKind},
     types::v0::{
         store,
         store::{nexus::NexusState, replica::ReplicaState},
         transport::{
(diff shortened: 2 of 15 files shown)


control-plane/agents/src/bin/core/controller/io_engine/v1/translation.rs
```diff
@@ -208,7 +208,8 @@ impl TryIoEngineToAgent for v1::replica::ReplicaSnapshot {
                 .timestamp
                 .clone()
                 .and_then(|t| std::time::SystemTime::try_from(t).ok())
-                .unwrap_or(UNIX_EPOCH),
+                // only valid on create, not on list! should we handle this as None ?
+                .unwrap_or(std::time::SystemTime::now()),
```
dsharma-dc: If we haven't got a valid timestamp in ReplicaSnapshot as part of create snapshot grpc response, then is it right to mark it as now() in control-plane, which will look like a legitimate timestamp?

tiagolobocastro: No, but UNIX_EPOCH doesn't seem right either, which is why I've added a comment there; 
we might need to set timestamp as option on our side as well...

tiagolobocastro: I reverted back for UNIX_EPOCH for now, but I think we should probably handle this properly rather than fake it..

dsharma-dc: sure, will modify this in a follow-up change.
Unrelated to this directly, but with a placeholder timestamp (UNIX_EPOCH or None) what'll be the interpretation on control-plane about this snapshot, is it anyway useful? Should `list` operation enumerate this, if yes how would a user decide if a clone should be created using this snapshot since we don't know when was this created?

dsharma-dc (approved):

control-plane/agents/src/bin/core/controller/io_engine/v1/translation.rs
dsharma-dc: One thing we can perhaps do is, as soon as we create a new replica snapshot having valid timestamp, delete any existing snapshot for that replica having placeholder timestamp.

control-plane/agents/src/bin/core/controller/io_engine/v1/translation.rs
```diff
@@ -208,6 +208,7 @@ impl TryIoEngineToAgent for v1::replica::ReplicaSnapshot {
                 .timestamp
                 .clone()
                 .and_then(|t| std::time::SystemTime::try_from(t).ok())
+                // only valid on create, not on list! should we handle this as None ?
```
dsharma-dc, line 211: github doesn't allow to comment on unchanged lines :\

Should the ResourceKind above be modified to Resourcekind::ReplicaSnapshot as you have added new kind now.

```
SnapshotId::try_from(self.snapshot_uuid.as_str()).map_err(|_| {
                SvcError::InvalidUuid {
                    uuid: self.snapshot_uuid.to_owned(),
                    kind: ResourceKind::VolumeSnapshot,
                }
            })?,
```

control-plane/stor-port/src/types/v0/store/snapshots/volume.rs
```diff
@@ -218,16 +218,18 @@ impl SpecTransaction<VolumeSnapshotOperation> for VolumeSnapshot {
...
                     // replace-in-place the logged replica specs.
                     self.metadata
                         .transactions
-                        .insert(info.txn_id, vec![result.replica]);
+                        .insert(info.txn_id, vec![result.replica.clone()]);
                     self.status = SpecStatus::Created(());
+                } else {
+                    // means we've restarted with the op in progress... and the snapshot was not
+                    // successful!
+                    tracing::error!(?self, "Snapshot Create completion without the result");
```
Abhinandan-Purkait, line 231: How do we reach here? Should we not bail out in the operations itself if the `VolumeSnapshotCompleter` is empty/invalid.

control-plane/agents/src/bin/core/node/wrapper.rs
```diff
@@ -830,9 +843,32 @@ impl NodeStateFetcher {
         &self,
         _client: &mut GrpcClient,
     ) -> Result<Vec<ReplicaSnapshot>, SvcError> {
-        // todo: implement this
+        // client
+        //     .list_repl_snapshots(&ListReplicaSnapshots::default())
+        //     .await
```
Abhinandan-Purkait, line 848: Should this be uncommented?

tiagolobocastro: Seems to cause the dataplane to crash..

control-plane/stor-port/src/types/v0/store/snapshots/volume.rs
tiagolobocastro: We may reach there due to a bug (the one you saw before!) or if we call commit_op after container restart for whatever reason.

control-plane/agents/src/bin/core/node/wrapper.rs
Abhinandan-Purkait: Oohh okay.

Abhinandan-Purkait (approved):

control-plane/agents/src/bin/core/controller/io_engine/v1/translation.rs
tiagolobocastro: Yeah I know, it has some limitations..
Done :)

tiagolobocastro:
bors merge

tiagolobocastro:
bors cancel

control-plane/agents/src/bin/core/controller/io_engine/v1/translation.rs
tiagolobocastro: CSI says timestamp is REQUIRED so we can't simply make it optional...
Also, it seems that dataplane always returns the timestamp... perhaps this is just an artifact of the grpc protobuf which causes timestamp to be an Option.. so maybe we can keep this unwrap to epoch since it should never happen.

tiagolobocastro:
bors merge

PR #556, opened 2023-05-30
fix(nexus-cleanup): test fix

abhilashshetty04:


abhilashshetty04:
bors try

control-plane/agents/src/bin/core/tests/volume/switchover.rs
```diff
@@ -162,6 +162,20 @@ async fn old_nexus_delete_after_vol_destroy() {
         .start(cluster.node(0).as_str())
         .await
         .expect("failed to start container");
+    let mut count = 0;
```
tiagolobocastro: this kind of thing could be useful, please add a function to helpers.rs
See existing code where we panic based on timeout rather than iteration count.

control-plane/agents/src/bin/core/tests/volume/switchover.rs
```diff
@@ -173,6 +187,19 @@ async fn old_nexus_delete_after_vol_destroy() {
...
 }
 
+/// Checks if node is online, returns true if yes.
+async fn is_node_online(node_client: &impl NodeOperations, node_id: NodeId) -> bool {
+    let node = node_client
+        .get(Filter::Node(node_id), None)
+        .await
+        .expect("Cant get node object");
+    let node = node.0.get(0).unwrap();
+    if node.state().unwrap().status == NodeStatus::Online {
```
tiagolobocastro: probably don't want to unwrap here if we use this within a "retry" function.

abhilashshetty04: state will always be availble right? Even if node is offline. But will handle

control-plane/agents/src/bin/core/tests/volume/switchover.rs
abhilashshetty04: Sure

control-plane/agents/src/bin/core/tests/volume/switchover.rs
tiagolobocastro: I'm not sure if we can guarantee it's always there, if we can then we should probably make it non-optional at some point..

abhilashshetty04:
bors try

Abhinandan-Purkait (approved):

abhilashshetty04:
bors try

control-plane/agents/src/bin/core/tests/volume/switchover.rs
abhilashshetty04: Handled unwrap for now. Im sure node object will be there so using expect while fetching node.

control-plane/agents/src/bin/core/tests/volume/switchover.rs
abhilashshetty04: Done

abhilashshetty04:
bors try

abhilashshetty04:
bors try

abhilashshetty04:
bors cancel

tiagolobocastro (approved):
Please squash the commits

abhilashshetty04:
bors merge

PR #560, opened 2023-06-01
Get snapshot state and add pool destroy validation

tiagolobocastro:
fix(snapshot/pool-destroy): don't destroy pool if it has snapshots
    
    Block pool destruction if there are snapshots on it.
    This can happen if we delete a volume but not its snapshots.
    
---

    feat(snapshot/state): retrieve the replica snapshot states
    
    Add pool id to the replica snapshot spec. Why? Consider what happens when the source volume is
    deleted, we no longer have the replica so we cannot determine where the snapshot resides?!
    Using the pool id allows to determine where the snapshot resides.
    todo: Should we use the pool uuid instead?
    todo: refactor VolumeSnapshotState to be able to represent a snapshot state
    
### DIFF
--- a/control-plane/agents/src/bin/core/controller/states.rs
+++ b/control-plane/agents/src/bin/core/controller/states.rs
@@ -180,11 +180,20 @@ impl ResourceStates {
         self.replicas.get(id)
     }
 
+    /// Get a snapshot with the given ID.
+    pub(crate) fn snapshot_state(
+        &self,
+        id: &SnapshotId,
+    ) -> Option<&ResourceMutex<ReplicaSnapshotState>> {
+        self.snapshots.get(id)
+    }
+
     /// Clear all state information.
     pub(crate) fn clear_all(&mut self) {
         self.nexuses.clear();
         self.pools.clear();
         self.replicas.clear();
+        self.snapshots.clear();
     }
 
     /// Takes an iterator of resources resourced by an 'Arc' and 'Mutex' and returns a vector of
--- a/control-plane/agents/src/bin/core/node/wrapper.rs
+++ b/control-plane/agents/src/bin/core/node/wrapper.rs
@@ -554,6 +554,12 @@ impl NodeWrapper {
             .replica_state(replica_id)
             .map(|r| r.lock().replica.clone())
     }
+    /// Get snapshot from `snapshot_id`.
+    pub(crate) fn snapshot(&self, snapshot_id: &SnapshotId) -> Option<ReplicaSnapshot> {
+        self.resources()
+            .snapshot_state(snapshot_id)
+            .map(|r| r.lock().snapshot.clone())
+    }
     /// Is the node online.
     pub(crate) fn is_online(&self) -> bool {
         self.status() == NodeStatus::Online
@@ -947,6 +953,8 @@ pub(crate) trait GetterOps {
     async fn nexuses(&self) -> Vec<Nexus>;
     async fn nexus(&self, nexus_id: &NexusId) -> Option<Nexus>;
     async fn volume_nexus(&self, volume_id: &VolumeId) -> Option<Nexus>;
+
+    async fn snapshot(&self, snapshot: &SnapshotId) -> Option<ReplicaSnapshot>;
 }
 
 #[async_trait]
@@ -986,6 +994,11 @@ impl GetterOps for Arc<tokio::sync::RwLock<NodeWrapper>> {
     async fn volume_nexus(&self, volume_id: &VolumeId) -> Option<Nexus> {
         let node = self.read().await;
         node.volume_nexus(volume_id)
+    }
+
+    async fn snapshot(&self, snapshot: &SnapshotId) -> Option<ReplicaSnapshot> {
+        let node = self.read().await;
+        node.snapshot(snapshot)
     }
 }
 
--- a/control-plane/agents/src/bin/core/volume/registry.rs
+++ b/control-plane/agents/src/bin/core/volume/registry.rs
@@ -1,15 +1,27 @@
 use crate::controller::registry::Registry;
 use agents::errors::SvcError;
 use stor_port::types::v0::transport::{
-    uri_with_hostnqn, Nexus, NexusStatus, ReplicaStatus, ReplicaTopology, Volume, VolumeId,
-    VolumeState, VolumeStatus, VolumeUsage,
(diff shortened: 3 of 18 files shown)


control-plane/agents/src/bin/core/node/wrapper.rs
```diff
@@ -554,6 +554,12 @@ impl NodeWrapper {
             .replica_state(replica_id)
             .map(|r| r.lock().replica.clone())
     }
+    /// Get snapshot from `snapshot_id`.
+    pub(crate) fn snapshot(&self, snapshot_id: &SnapshotId) -> Option<ReplicaSnapshot> {
```
Abhinandan-Purkait, line 558: `snapshot` --> `replica_snapshot` ?

control-plane/agents/src/bin/core/controller/states.rs
```diff
@@ -180,6 +180,14 @@ impl ResourceStates {
...
+    /// Get a snapshot with the given ID.
+    pub(crate) fn snapshot_state(
+        &self,
+        id: &SnapshotId,
+    ) -> Option<&ResourceMutex<ReplicaSnapshotState>> {
+        self.snapshots.get(id)
+    }
+
     /// Clear all state information.
     pub(crate) fn clear_all(&mut self) {
```
niladrih, line 192: Does clear_all need to include snapshots?

niladrih (commented):
Left one comment..

control-plane/rest/service/src/v0/snapshots.rs
```diff
@@ -78,7 +78,8 @@ impl apis::actix_server::Snapshots for RestApi {
             ),
             state: models::VolumeSnapshotState::new_all(
                 snap.state().uuid(),
-                snap.state().size(),
+                // todo: should be optional?
+                snap.state().size().unwrap_or_default(),
```
Abhinandan-Purkait, line 82: Optional Should be better I guess, rather than showing 0 ?

control-plane/agents/src/bin/core/volume/registry.rs
```diff
@@ -197,4 +209,44 @@ impl Registry {
...
+        snapshot: &VolumeSnapshot,
+    ) -> Result<VolumeSnapshotState, SvcError> {
+        let mut replica_snaps = vec![];
+        let mut size = None;
+        if let Some(meta_snapshots) = snapshot.metadata().replica_snapshots() {
+            replica_snaps = Vec::with_capacity(meta_snapshots.len());
+            for replica_snap in meta_snapshots {
+                replica_snaps.push(
+                    if let Ok(state) = self.snapshot_replica(replica_snap.spec()).await {
+                        size = Some(state.snap_size);
```
Abhinandan-Purkait, line 241: Thought: Should we compare and pick the largest size?

control-plane/grpc/src/operations/volume/traits_snapshots.rs
```diff
@@ -132,24 +134,74 @@ pub struct ReplicaSnapshot {
...
 
+/// Volume replica snapshot state information.
+#[derive(Debug)]
+#[allow(unused)]
+pub enum VolumeReplicaSnapshotState {
+    /// When the snapshot is available.
+    Online {
+        /// The pool hosting the snapshot.
+        pool_id: PoolId,
+        /// The snapshot state.
```
Abhinandan-Purkait: `The snapshot state.` --> `The replica snapshot state.` Just to be exactly clear while hovering in IDEs. Same in below?

control-plane/stor-port/src/types/v0/store/snapshots/volume.rs
```diff
@@ -61,7 +61,7 @@ impl VolumeSnapshot {
                 self.spec().uuid(),
                 self.spec.source_id.to_string(),
                 self.metadata.prepare(),
-                "what's-my-name?",
+                self.spec.uuid.to_string(),
```
Abhinandan-Purkait: `self.spec.uuid.to_string()` -> `self.spec().uuid().to_string()` ?

Thought: Should we rather pass the volume id as name  as name would be stored as meta in spdk and could be useful?

Abhinandan-Purkait (approved):
LGTM. Thanks

control-plane/agents/src/bin/core/node/wrapper.rs
tiagolobocastro: For a node a snapshot is a replica snapshot, volume is a higher-level abstraction

control-plane/rest/service/src/v0/snapshots.rs
tiagolobocastro: Yeah I think so to

control-plane/agents/src/bin/core/volume/registry.rs
tiagolobocastro: I was going to do that but then realized they should be exactly the same, maybe doesn't matter?

control-plane/grpc/src/operations/volume/traits_snapshots.rs
tiagolobocastro: Indeed, done

control-plane/stor-port/src/types/v0/store/snapshots/volume.rs
tiagolobocastro: I don't think we can because this is the bdev name for the snapshot, it has to be unique AIUI, if it's the volume name it would collide with the nexus?

control-plane/agents/src/bin/core/controller/states.rs
tiagolobocastro: Fixed, good catch!

control-plane/stor-port/src/types/v0/store/snapshots/volume.rs
Abhinandan-Purkait: Right.

control-plane/agents/src/bin/core/volume/registry.rs
Abhinandan-Purkait: Yeah.

control-plane/agents/src/bin/core/node/wrapper.rs
Abhinandan-Purkait: Okay.

niladrih (approved):

tiagolobocastro:
bors merge

PR #561, opened 2023-06-02
feat(snapshots): cleanup snapshot resources in pending creation when source is deleted

Abhinandan-Purkait:


control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -222,6 +223,27 @@ impl ResourceLifecycle for OperationGuardArc<VolumeSpec> {
...
+            let snapshot_user_spec = snapshot.lock().spec().clone();
+            let result = snapshot_guard
+                .destroy(
+                    registry,
+                    &DestroyVolumeSnapshotRequest::new(snapshot, snapshot_user_spec),
+                )
+                .await;
+            if let Err(error) = result {
+                tracing::warn!(snapshot.uuid=%snapshot_guard.uuid(), error=%error,
+                    "Snapshot destruction failed. This will be garbage collected later"
```
```suggestion
                    "Snapshot destruction failed. It will be garbage collected later"
```

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -222,6 +223,27 @@ impl ResourceLifecycle for OperationGuardArc<VolumeSpec> {
             }
         }
 
+        // Destroy all the snapshots that are in creating state as the source is getting destroyed.
+        let pending_creation_snapshots = specs.get_creating_snapshots_by_volume(self.uuid());
+        for snapshot in pending_creation_snapshots {
+            let mut snapshot_guard = match snapshot.operation_guard_wait().await {
+                Ok(snapshot_guard) => snapshot_guard,
+                Err(_) => continue,
```
tiagolobocastro, line 231: I think this implies we need a GC to destroy creating snapshots with missing owner :)

control-plane/agents/src/bin/core/volume/specs.rs
```diff
@@ -629,6 +629,22 @@ impl ResourceSpecsLocked {
...
+        volume_id: &VolumeId,
+    ) -> Vec<ResourceMutex<VolumeSnapshot>> {
+        let specs = self.read();
+        // reviewer: Would this be a very costly filtration? How can we optimise this?
+        specs
+            .volume_snapshots
+            .to_vec()
+            .iter()
+            .cloned()
+            .filter(|s| s.lock().status().creating() && s.lock().spec().source_id() == volume_id)
```
tiagolobocastro: get s.lock() once rather than locking twice

control-plane/agents/src/bin/core/volume/specs.rs
```diff
@@ -629,6 +629,22 @@ impl ResourceSpecsLocked {
...
+        &self,
+        volume_id: &VolumeId,
+    ) -> Vec<ResourceMutex<VolumeSnapshot>> {
+        let specs = self.read();
+        // reviewer: Would this be a very costly filtration? How can we optimise this?
+        specs
+            .volume_snapshots
+            .to_vec()
+            .iter()
+            .cloned()
```
tiagolobocastro: seems excessive to do so (cleaning the entire values and then cloning inner again), specs.volume_snapshots.values() would suffice I think (and cloned() at the end only so we only clone the ones that match)

control-plane/agents/src/bin/core/volume/operations.rs
Abhinandan-Purkait: Yes, right. Anyhow this would be marked as `deleting` and the garbage collector we have would clean this up. Should be a problem right?

control-plane/agents/src/bin/core/volume/specs.rs
Abhinandan-Purkait: Okay.

control-plane/agents/src/bin/core/volume/specs.rs
Abhinandan-Purkait: Yeah, that is much better. Thank you.

control-plane/agents/src/bin/core/volume/operations.rs
tiagolobocastro: No it won't be marked as deleting because we failed to get the guard

control-plane/agents/src/bin/core/volume/specs.rs
```diff
@@ -629,6 +629,23 @@ impl ResourceSpecsLocked {
         specs.volume_snapshots.get(snapshot_id).cloned()
     }
 
+    /// Get the list of snapshots that are in creating state by its source.
+    pub(crate) fn get_creating_snapshots_by_volume(
```
tiagolobocastro:
```suggestion
    pub(crate) fn creating_snapshots_by_volume(
```

control-plane/agents/src/bin/core/volume/operations.rs
Abhinandan-Purkait: Right so we would need one more reconciler to handle that. Would add that in the next PR.

tiagolobocastro (approved):

control-plane/agents/src/bin/core/volume/specs.rs
Abhinandan-Purkait: Done.

Abhinandan-Purkait:
bors merge

PR #564, opened 2023-06-05
fix(snapshot/grpc): add missing snapshot state conversions

tiagolobocastro:


control-plane/stor-port/src/types/v0/store/snapshots/replica.rs
```diff
@@ -117,6 +121,14 @@ impl ReplicaSnapshotMeta {
     pub fn size(&self) -> u64 {
         self.size
     }
+    /// Get the snapshot size.
```
dsharma-dc: "/// Get the snapshot timestamp."

control-plane/stor-port/src/types/v0/transport/replica.rs
```diff
@@ -174,10 +178,30 @@ impl ReplicaSnapshotDescr {
         &self.snap_uuid
     }
 
+    /// Get a reference to snapshot id.
```
dsharma-dc: "/// Get a reference to pool id."

dsharma-dc (approved):

Abhinandan-Purkait (approved):

control-plane/stor-port/src/types/v0/store/snapshots/replica.rs
tiagolobocastro: Thanks, there were a few more too, fixed all I think.

tiagolobocastro:
bors merge

PR #567, opened 2023-06-06
feat(rebuild_history): resource map implementation for rebuild history

abhilashshetty04:
We have added rebuild history resource map in the Nodewrapper. This uses list_rebuild_record rpc to periodically update rebuild history of all nexus in the node.

This PR also adds test for rebuild history in control plane.

### DIFF
--- a/control-plane/agents/src/bin/core/controller/states.rs
+++ b/control-plane/agents/src/bin/core/controller/states.rs
@@ -1,12 +1,12 @@
 use crate::controller::resources::{resource_map::ResourceMap, ResourceMutex};
 use stor_port::types::v0::{
     store::{nexus::NexusState, pool::PoolState, replica::ReplicaState},
-    transport::{self, Nexus, NexusId, PoolId, Replica, ReplicaId},
+    transport::{self, Nexus, NexusId, PoolId, RebuildHistory, Replica, ReplicaId},
 };
 
 use indexmap::map::Values;
 use parking_lot::RwLock;
-use std::{ops::Deref, sync::Arc};
+use std::{collections::HashMap, ops::Deref, sync::Arc};
 use stor_port::types::v0::{
     store::snapshots::ReplicaSnapshotState,
     transport::{ReplicaSnapshot, SnapshotId},
@@ -37,6 +37,7 @@ pub(crate) struct ResourceStates {
     pools: ResourceMap<PoolId, PoolState>,
     replicas: ResourceMap<ReplicaId, ReplicaState>,
     snapshots: ResourceMap<SnapshotId, ReplicaSnapshotState>,
+    rebuild_history: ResourceMap<NexusId, RebuildHistory>,
 }
 
 /// Add/Update or remove resource from the registry.
@@ -55,11 +56,13 @@ impl ResourceStates {
         replicas: Vec<Replica>,
         nexuses: Vec<Nexus>,
         snapshots: Vec<ReplicaSnapshot>,
+        rebuild_histories: HashMap<NexusId, RebuildHistory>,
     ) {
         self.update_replicas(replicas);
         self.update_pools(pools);
         self.update_nexuses(nexuses);
         self.update_snapshots(snapshots);
+        self.update_rebuild_history(rebuild_histories);
     }
 
     /// Update nexus states.
@@ -188,12 +191,25 @@ impl ResourceStates {
         self.snapshots.get(id)
     }
 
+    /// This updates rebuild history resource map.
+    pub(crate) fn update_rebuild_history(&mut self, histories: HashMap<NexusId, RebuildHistory>) {
+        self.rebuild_history.clear();
+        let l: Vec<RebuildHistory> = histories.into_values().collect();
+        self.rebuild_history.populate(l);
+    }
+
+    /// Get a rebuild history with the given ID.
+    pub(crate) fn rebuild_history(&self, id: &NexusId) -> Option<&ResourceMutex<RebuildHistory>> {
+        self.rebuild_history.get(id)
+    }
+
     /// Clear all state information.
     pub(crate) fn clear_all(&mut self) {
         self.nexuses.clear();
         self.pools.clear();
         self.replicas.clear();
         self.snapshots.clear();
+        self.rebuild_history.clear();
     }
 
     /// Takes an iterator of resources resourced by an 'Arc' and 'Mutex' and returns a vector of
--- a/control-plane/agents/src/bin/core/nexus/service.rs
+++ b/control-plane/agents/src/bin/core/nexus/service.rs
@@ -1,10 +1,13 @@
-use crate::controller::{
-    registry::Registry,
-    resources::{
-        operations::{ResourceLifecycle, ResourceOffspring, ResourceSharing},
-        operations_helper::ResourceSpecsLocked,
-        OperationGuardArc,
+use crate::{
+    controller::{
+        registry::Registry,
+        resources::{
+            operations::{ResourceLifecycle, ResourceOffspring, ResourceSharing},
+            operations_helper::ResourceSpecsLocked,
+            OperationGuardArc,
+        },
     },
+    node::wrapper::GetterOps,
 };
 use agents::errors::SvcError;
 use grpc::{
@@ -15,7 +18,7 @@ use grpc::{
     },
 };
 use stor_port::{
-    transport_api::{v0::Nexuses, ReplyError, ResourceKind},
+    transport_api::{v0::Nexuses, ReplyError},
     types::v0::{
         store::nexus::NexusSpec,
         transport::{
@@ -24,7 +27,6 @@ use stor_port::{
         },
     },
 };
-use tonic::Status;
 
 #[derive(Debug, Clone)]
 pub(super) struct Service {
@@ -202,10 +204,12 @@ impl Service {
         &self,
         request: &GetRebuildRecord,
     ) -> Result<RebuildHistory, SvcError> {
-        Err(SvcError::Unimplemented {
-            resource: ResourceKind::Nexus,
-            request: "get_rebuild_history".to_string(),
-            source: Status::unimplemented("api still unimplemented"),
-        })
+        let node_id = self.registry.nexus(&request.nexus).await?.node;
+        let node = self.registry.node_wrapper(&node_id).await?;
+        node.rebuild_history(&request.nexus)
+            .await
+            .ok_or(SvcError::RebuildHistoryNotFound {
+                nexus_id: request.nexus.to_string(),
+            })
     }
 }
--- a/control-plane/agents/src/bin/core/node/wrapper.rs
+++ b/control-plane/agents/src/bin/core/node/wrapper.rs
@@ -24,18 +24,18 @@ use stor_port::{
             AddNexusChild, ApiVersion, Child, CreateNexus, CreateNexusSnapshot,
             CreateNexusSnapshotResp, CreatePool, CreateReplica, CreateReplicaSnapshot,
             DestroyNexus, DestroyPool, DestroyReplica, DestroyReplicaSnapshot, FaultNexusChild,
-            GetRebuildRecord, ImportPool, ListReplicaSnapshots, MessageIdVs, Nexus,
-            NexusChildAction, NexusChildActionContext, NexusChildActionKind, NexusId, NodeId,
-            NodeState, NodeStatus, PoolId, PoolState, RebuildHistory, Register, RemoveNexusChild,
-            Replica, ReplicaId, ReplicaName, ReplicaSnapshot, ShareNexus, ShareReplica,
-            ShutdownNexus, SnapshotId, UnshareNexus, UnshareReplica, VolumeId,
+            GetRebuildRecord, ImportPool, ListRebuildRecord, ListReplicaSnapshots, MessageIdVs,
+            Nexus, NexusChildAction, NexusChildActionContext, NexusChildActionKind, NexusId,
+            NodeId, NodeState, NodeStatus, PoolId, PoolState, RebuildHistory, Register,
+            RemoveNexusChild, Replica, ReplicaId, ReplicaName, ReplicaSnapshot, ShareNexus,
+            ShareReplica, ShutdownNexus, SnapshotId, UnshareNexus, UnshareReplica, VolumeId,
         },
     },
 };
 
 use async_trait::async_trait;
 use parking_lot::RwLock;
-use std::{ops::DerefMut, sync::Arc};
+use std::{collections::HashMap, ops::DerefMut, sync::Arc};
 use tracing::{debug, trace, warn};
 
 type NodeResourceStates = (
@@ -43,10 +43,13 @@ type NodeResourceStates = (
     Vec<PoolState>,
     Vec<Nexus>,
     Vec<ReplicaSnapshot>,
+    HashMap<NexusId, RebuildHistory>,
 );
 
 /// Default timeout for GET* gRPC requests (ex: GetPools, GetNexuses, etc..)
 const GETS_TIMEOUT: MessageId = MessageId::v0(MessageIdVs::Default);
+/// Maximum number of rebuild histories control plane requests per nexus.
+const MAX_HISTORY_PER_NEXUS: u32 = 8;
 
 enum ResourceType {
     All(
@@ -54,6 +57,7 @@ enum ResourceType {
         Vec<Replica>,
         Vec<Nexus>,
         Vec<ReplicaSnapshot>,
+        HashMap<NexusId, RebuildHistory>,
     ),
     Nexuses(Vec<Nexus>),
     Nexus(Either<Nexus, NexusId>),
@@ -541,6 +545,14 @@ impl NodeWrapper {
     fn nexus(&self, nexus_id: &NexusId) -> Option<Nexus> {
         self.resources().nexus_state(nexus_id).map(|s| s.nexus)
     }
+
+    /// Gets rebuild history for a nexus from resource map.
+    pub(crate) fn rebuild_history(&self, nexus_id: &NexusId) -> Option<RebuildHistory> {
+        self.resources()
+            .rebuild_history(nexus_id)
+            .map(|rs| rs.lock().clone())
+    }
+
     /// Get nexus for the given volume.
     fn volume_nexus(&self, volume_id: &VolumeId) -> Option<Nexus> {
         self.resources()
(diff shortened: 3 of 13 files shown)


abhilashshetty04:
bors try

control-plane/agents/src/bin/core/nexus/service.rs
```diff
@@ -14,8 +14,9 @@ use grpc::{
         NexusOperations, RemoveNexusChildInfo, ShareNexusInfo, UnshareNexusInfo,
     },
 };
+use std::ops::Deref;
```
tiagolobocastro: move this one below on its own

control-plane/agents/src/bin/core/nexus/service.rs
```diff
@@ -202,10 +202,13 @@ impl Service {
...
         request: &GetRebuildRecord,
     ) -> Result<RebuildHistory, SvcError> {
-        Err(SvcError::Unimplemented {
-            resource: ResourceKind::Nexus,
-            request: "get_rebuild_history".to_string(),
-            source: Status::unimplemented("api still unimplemented"),
-        })
+        let node_id = self.registry.nexus(&request.nexus).await?.node;
+        let node = self.registry.node_wrapper(&node_id).await?;
+        let wrapper = node.read().await.deref().clone();
```
tiagolobocastro: Rather than this you need to expand GetterOps with rebuild history.

control-plane/agents/src/bin/core/node/wrapper.rs
```diff
@@ -841,7 +857,23 @@ impl NodeStateFetcher {
...
-        Ok((replicas, pools, nexuses, snapshots))
+        let rebuild_history = self.fetch_rebuild_history(client).await?;
+        Ok((replicas, pools, nexuses, snapshots, rebuild_history))
+    }
+
+    pub(crate) async fn fetch_rebuild_history(
+        &self,
+        client: &mut GrpcClient,
+    ) -> Result<Vec<RebuildHistory>, SvcError> {
+        let nexuses = client.list_nexuses(self.id()).await?;
```
tiagolobocastro: This doesn't seem a good idea... potentially can make 1000x rpc calls!
Seems like we have to update the rebuild history api to make nexus optional and then query all in a single call. While we're at it we can add the cap of entries, this is why I suggested that first you raise the api/dataplane changes yesterday...

control-plane/agents/src/bin/core/tests/rebuild/mod.rs
```diff
@@ -169,14 +177,92 @@ async fn replica_wait_and_full_rebuild() {
...
+    // Stop the container and don't restart.
+    cluster
+        .composer()
+        .stop(&testrep_node.to_string())
+        .await
+        .expect("container stop failure");
+
+    // Check the pool on restarted node shouldn't be online.
+    // This must expire the wait timeout.
+    wait_till_pool_online(&cluster, testrep, Duration::from_secs(20)).await;
```
tiagolobocastro: Seems like we want the opposite instead, wait_till_pool_offline.

control-plane/agents/src/bin/core/tests/rebuild/mod.rs
```diff
@@ -169,14 +177,92 @@ async fn replica_wait_and_full_rebuild() {
...
+                HashMap::new(),
+                None,
+                vol_target.clone().to_string(),
+                models::VolumeShareProtocol::Nvmf,
+                None,
+                cluster.csi_node(0),
+            ),
+        )
+        .await
+        .expect("Failed tp publish volume");
```
tiagolobocastro:
```suggestion
        .expect("Failed to publish volume");
```

control-plane/agents/src/bin/core/tests/rebuild/mod.rs
```diff
@@ -214,4 +300,213 @@ async fn replica_wait_and_full_rebuild() {
...
+    wait_nexus_online(&nexus_client, nexus_id.clone())
+        .await
+        .expect("Rebuild didnt finish in 30 secs");
+    let history = nexus_client
+        .get_rebuild_history(&GetRebuildRecord::new(nexus_id.clone()), None)
+        .await
+        .expect("Failed to get rebuild record");
+    assert_eq!(
+        nexus_id, history.uuid,
+        "Cant match nexus id in rebuild history"
```
tiagolobocastro:
```suggestion
        "Can't match nexus id in rebuild history"
```

control-plane/agents/src/bin/core/tests/rebuild/mod.rs
```diff
@@ -214,4 +300,213 @@ async fn replica_wait_and_full_rebuild() {
...
+
+    assert!(
+        history.records.get(0).unwrap().is_partial,
+        "Rebuild type is not Partial rebuild"
+    )
+}
+
+const POOL_SIZE_BYTES: u64 = 128 * 1024 * 1024;
+#[tokio::test]
+async fn validate_empty_rebuild_record() {
```
tiagolobocastro: We're adding a lot of new clusters here, let's try to reuse clusters to keep  test runs down..

control-plane/agents/src/bin/core/node/wrapper.rs
abhilashshetty04: Next task will be to add limit to number of rebuild histories / nexus. I will start working on it from tomorrow. However, if we want to populated history of all apis the we have to call them one by one right?

tiagolobocastro: No... as I said, make the nexus id optional, and in that case all history would be returned (with some limits per nexus and perhaps total as well).

abhilashshetty04: Dataplane rpc changes will be required. Ill check.

control-plane/agents/src/bin/core/controller/states.rs
```diff
@@ -188,12 +191,36 @@ impl ResourceStates {
...
+    pub(crate) fn update_rebuild_history(&mut self, histories: Vec<RebuildHistory>) {
+        self.rebuild_history.clear();
+        self.rebuild_history.populate(histories);
+    }
+
+    #[allow(dead_code)]
+    /// Returns a vector of cloned rebuild history.
+    pub(crate) fn rebuild_histories_cloned(&self) -> Vec<RebuildHistory> {
+        Self::cloned_inner_states(self.rebuild_history.values())
+    }
```
Abhinandan-Purkait: Should we remove if its not needed rather than dead code?

control-plane/agents/src/bin/core/controller/states.rs
```diff
@@ -188,12 +191,36 @@ impl ResourceStates {
...
+    /// Get a rebuild history with the given ID.
+    pub(crate) fn rebuild_history(&self, id: &NexusId) -> Option<&ResourceMutex<RebuildHistory>> {
+        self.rebuild_history.get(id)
+    }
+
+    #[allow(dead_code)]
+    /// Returns an iterator of rebuild history.
+    pub(crate) fn rebuild_histories(&self) -> Values<NexusId, ResourceMutex<RebuildHistory>> {
+        self.rebuild_history.values()
+    }
```
Abhinandan-Purkait: Same here. Let's add when needed.

control-plane/agents/src/bin/core/nexus/service.rs
Abhinandan-Purkait: Yeah. I too think so.

control-plane/agents/src/bin/core/tests/rebuild/mod.rs
abhilashshetty04: This i can test in existing test before rebuild starts.

control-plane/agents/src/bin/core/node/wrapper.rs
abhilashshetty04: Incorporated

control-plane/agents/src/bin/core/tests/rebuild/mod.rs
abhilashshetty04: added validation on the returned result. we werent processing them before. changed fn name to `is_pool_online ` .

control-plane/agents/src/bin/core/tests/rebuild/mod.rs
abhilashshetty04: checking empty history for nexus in existing rebuild history case. We have 4 cases now.

abhilashshetty04:
bors try

control-plane/agents/src/bin/core/node/wrapper.rs
```diff
@@ -841,7 +857,16 @@ impl NodeStateFetcher {
         let pools = self.fetch_pools(client).await?;
         let nexuses = self.fetch_nexuses(client).await?;
         let snapshots = self.fetch_snapshots(client).await?;
-        Ok((replicas, pools, nexuses, snapshots))
+        let rebuild_history = self.fetch_rebuild_history(client).await?;
```
tiagolobocastro, line 862: just realized there may be an issue here, also exists with snapshots. if the dataplane does not support this call, we'll get grpc error..
We need to ignore this error, but where, here, or in the node wrapper? @Abhinandan-Purkait wdyt?

control-plane/agents/src/bin/core/tests/rebuild/mod.rs
```diff
@@ -12,8 +12,9 @@ use stor_port::types::v0::{
             ReplicaState, VolumePolicy, VolumeStatus,
         },
     },
-    transport::VolumeId,
+    transport::{Filter, GetRebuildRecord, NexusId, NexusStatus, VolumeId},
 };
+const REBUILD_WAIT_TIME: u64 = 40;
```
tiagolobocastro: wait time in what?

control-plane/agents/src/bin/core/tests/rebuild/mod.rs
```diff
@@ -116,7 +113,12 @@ async fn replica_wait_and_partial_rebuild() {
         .expect("container stop failure");
 
     // Check the pool on restarted node is imported back.
-    wait_till_pool_online(&cluster, testrep, Duration::from_secs(20)).await;
+    if is_pool_online(&cluster, testrep, Duration::from_secs(20))
+        .await
+        .is_none()
+    {
+        panic!("Pool is expected to be online");
+    }
```
tiagolobocastro:
```suggestion
    is_pool_online(&cluster, testrep, Duration::from_secs(20))
        .await
        .expect("Pool is expected to be online");
```

control-plane/agents/src/bin/core/tests/rebuild/mod.rs
```diff
@@ -169,14 +176,118 @@ async fn replica_wait_and_full_rebuild() {
...
+
+    // Stop the container and don't restart.
+    cluster
+        .composer()
+        .stop(&testrep_node.to_string())
+        .await
+        .expect("container stop failure");
+
+    // Check the pool on restarted node shouldn't be online.
+    // This must expire the wait timeout.
```
tiagolobocastro: Instead, why not simply wait until the pool is offline?

control-plane/agents/src/bin/core/tests/rebuild/mod.rs
```diff
@@ -169,14 +176,118 @@ async fn replica_wait_and_full_rebuild() {
...
+                cluster.csi_node(0),
+            ),
+        )
+        .await
+        .expect("Failed to publish volume");
+
+    let volume_state = volume.state.clone();
+    let nexus = volume_state.target.unwrap();
+    let nexus_id = NexusId::from(nexus.uuid);
+    sleep(Duration::from_secs(5));
```
tiagolobocastro: What's this sleep for?

control-plane/agents/src/bin/core/node/wrapper.rs
Abhinandan-Purkait: But we already return unimplemented for v0 right? So do we need to do anything more? 

Are you talking about the older versions on v1? For that I think we should do that in the node wrapper, it would be easy to manage from one place?

control-plane/agents/src/bin/core/tests/rebuild/mod.rs
abhilashshetty04: I have seen that as sson as we publish volume rebuild history was not available for the nexus. Hence thought of sleep here. Probably should check interval for state update and use that for sleep?

control-plane/agents/src/bin/core/tests/rebuild/mod.rs
abhilashshetty04: We wanted to wait till child wait timeout elapses. That needs to be done anyway for Full rebuild to start right? So we are checking throughout the interval which was 15s pool is not back online.

tiagolobocastro: Then we're waiting for the wrong thing here, we should wait until the child is rebuilt using rebuild history?

control-plane/agents/src/bin/core/tests/rebuild/mod.rs
tiagolobocastro: I'm confused, we're just publishing the nexus here right, so there will be no rebuild history?

control-plane/agents/src/bin/core/node/wrapper.rs
tiagolobocastro: For both cases... otherwise we'll fail this procedure here with `?`.
I'm not sure, the disadvantage of doing in the node wrapper is that we'd be hiding the errors from the caller.
Tempted to say for now let's handle it here, ignore not implemented error inline.

control-plane/agents/src/bin/core/tests/rebuild/mod.rs
abhilashshetty04: This is the wait time we have for nexus to be Online again. its reduced to 10s now.

control-plane/agents/src/bin/core/tests/rebuild/mod.rs
abhilashshetty04: Waiting for 500ms now. 100ms is cache period

abhilashshetty04: There is an assertion in full rebuild case where we assert child removal from nexus. That comes right after this check. So waiting through wait innterval (6s now, wait time is 5s)

control-plane/agents/src/bin/core/tests/rebuild/mod.rs
abhilashshetty04: There will be, we will not have any records in the vector. hence matching if nexus uuid is of newly published nexus and record len is 0.

control-plane/agents/src/bin/core/node/wrapper.rs
Abhinandan-Purkait: Right, that's true. Then in that case this place is better right. The other callers will still see the errors which is fine as it's unimplemented. And here we handle it.

@tiagolobocastro A question, what happens if any of the above calls fail for whatever reason, then the below ones would not get executed right? Should we not attempt all calls and return an aggregated error at the end?

tiagolobocastro: hmm don't think that would help, we mostly just care if it succeeded or not. If any fails we just bail out and set the node as bad.

Abhinandan-Purkait: But is this behaviour correct then? What if it was some interim failure and there is a chance that the other calls would have passed? Also should we not set the node as bad only on `transport error not route to host` something like that rather than for any error.

control-plane/agents/src/bin/core/tests/rebuild/mod.rs
abhilashshetty04: We will be polling for rebuild history in full rebuild case. I have also removed existing cases. So there are 2 cases which will test rebuild as well as history.

abhilashshetty04:
bors try

abhilashshetty04:
bors try

control-plane/agents/src/bin/core/node/wrapper.rs
tiagolobocastro: Yes, because we don't want to keep half states, let's say you were able to list replicas but not pools, control-plane could get very confused.

Abhinandan-Purkait: Got it. Thanks for explaining.

control-plane/agents/src/bin/core/tests/rebuild/mod.rs
```diff
@@ -184,34 +209,131 @@ async fn replica_wait_and_full_rebuild() {
...
+                break;
+            }
+            tokio::time::sleep(Duration::from_millis(250)).await;
+        }
+    }
+    Err(())
+}
+
+/// Checks if pool is online. Returns pool if yes, None if not.
+async fn is_pool_online(cluster: &Cluster, replica: &Replica, timeout: Duration) -> Option<Pool> {
```
tiagolobocastro:
```suggestion
async fn wait_pool_online(cluster: &Cluster, replica: &Replica, timeout: Duration) -> Option<Pool> {
```

control-plane/agents/src/bin/core/tests/rebuild/mod.rs
```diff
@@ -184,34 +209,131 @@ async fn replica_wait_and_full_rebuild() {
...
+            return Some(pool);
+        }
+
+        if std::time::Instant::now() > (start + timeout)
+            && (pool.state.is_none() || pool.state.as_ref().unwrap().status != PoolStatus::Online)
+        {
+            tracing::info!("Timeout waiting for pool to be Online - {:?}", pool);
+            return None;
+        }
+        tokio::time::sleep(Duration::from_secs(1)).await;
```
tiagolobocastro:
```suggestion
        tokio::time::sleep(Duration::from_millis(250)).await;
```

control-plane/agents/src/bin/core/tests/rebuild/mod.rs
```diff
@@ -184,34 +209,131 @@ async fn replica_wait_and_full_rebuild() {
...
+        if std::time::Instant::now() > (start + timeout) {
+            break;
+        }
+        tokio::time::sleep(Duration::from_millis(250)).await;
+    }
+    Err(())
+}
+
+/// Checks if nexus has rebuild record for predefined interval.
+async fn poll_rebuild_history(
```
tiagolobocastro, line 294:
```suggestion
async fn wait_rebuild_history_count( count: u8,
```
And check that we have `count` records ?

control-plane/stor-port/src/types/v0/transport/nexus.rs
```diff
@@ -746,6 +746,25 @@ pub struct GetRebuildRecord {
     pub nexus: NexusId,
 }
 
+/// List rebuild history request.
+#[derive(Serialize, Deserialize, Default, Debug, Clone)]
+pub struct ListRebuildRecord {
+    /// Maximum number of rebuild records per nexus.
+    pub count: Option<u32>,
```
tiagolobocastro: doesn't need to be pub?

tiagolobocastro (approved):

control-plane/agents/src/bin/core/tests/rebuild/mod.rs
abhilashshetty04: We validate number of records in test itself. After rebuild in both cases it should be 1. We return the record from here which will get validated.

abhilashshetty04:
bors merge

PR #568, opened 2023-06-06
chore(protobuf/volume): rearrange protobuf fields for get snapshots

dsharma-dc:


control-plane/plugin/src/resources/snapshot.rs
```diff
@@ -73,23 +73,18 @@ async fn get_snapshots(
...
     {
         Ok(snaps) => {
             let s = snaps.into_body();
-            snapshots.extend(s.entries);
+            Some(s.entries)
         }
         Err(e) => {
             println!("Failed to list volume snapshots. Error {e}");
-            return None;
+            None
```
tiagolobocastro, line 87: hmm why is this returning an Option? Shouldn't we simply return an error here?

tiagolobocastro (approved):

control-plane/plugin/src/resources/snapshot.rs
dsharma-dc: This is a helper function, whose return value is checked to decide whether to print table or not.
Again, I just followed existing CLI get operation semantics here :)

dsharma-dc:
bors merge

PR #570, opened 2023-06-08
feat(list_snapshots): pagination support for listing snapshots

dsharma-dc:


control-plane/agents/src/bin/core/volume/specs.rs
```diff
@@ -373,6 +373,43 @@ impl ResourceSpecs {
...
+    pub(crate) fn paginated_snapshots(
+        &self,
+        pagination: &Pagination,
+        vol_id: Option<&VolumeId>,
+    ) -> PaginatedResult<VolumeSnapshot> {
+        let mut last_result = false;
+        if let Some(vol_id) = vol_id {
+            // We need to filter the resource map based on volume id and apply the pagination
+            // parameters on that.
+            let filtered_snapshots = self.volume_snapshots.paginate_selected(
```
dsharma-dc: Here is the usage of new paginate function.  @tiagolobocastro @Abhinandan-Purkait

dsharma-dc:
bors try

control-plane/agents/src/bin/core/controller/resources/resource_map.rs
```diff
@@ -91,4 +91,38 @@ where
             .map(|v| v.lock().clone())
             .collect()
     }
+
+    /// Apply the closure on the elements and return a subset of the map (i.e. a paginated result),
+    /// containing the elements for which closure returns true.
+    pub(crate) fn paginate_selected<F>(
+        &self,
+        max_entries: u64,
```
tiagolobocastro: why not reuse existing offset/len as the previous paginate?
Then simply filter from closure and apply same skip/take? You might even be able to rewrite things so both paginate functions reuse same code.

control-plane/agents/src/bin/core/main.rs
```diff
@@ -1,6 +1,6 @@
 //! The Core Agent.
 //! todo: document.
-
+#![feature(iter_advance_by)]
```
tiagolobocastro: hmm don't think we want to use experimental features..

dsharma-dc: The usage seemed pretty harmless because even if advance_by fails, we will still get the snapshot list just that from beginning index. But if still no consensus on using this, we can replace it with a for loop and do iter.next() within.

Abhinandan-Purkait: We do have avoided experimental features, and this seems to be open in rust-lang for a long time. So probably better to avoid but I don't have any hard preferences here.

control-plane/agents/src/bin/core/controller/resources/resource_map.rs
dsharma-dc: Have modified to reuse existing paginate code as discussed. Pretty sleek, thank you !

dsharma-dc:
bors try

Abhinandan-Purkait (approved):

control-plane/agents/src/bin/core/controller/resources/resource_map.rs
```diff
@@ -84,8 +84,20 @@ where
...
+        self.paginate_filter(offset, len, |_| true)
+    }
+
+    /// Apply the closure on the elements and return a subset of the map (i.e. a paginated result),
+    /// containing the elements for which closure returns true.
+    pub(crate) fn paginate_filter<F: Fn(&&ResourceMutex<S>) -> bool>(
+        &self,
+        offset: u64,
+        len: u64,
+        f: F,
```
tiagolobocastro:
```suggestion
        filter: F,
```

control-plane/plugin/src/resources/tests.rs
```diff
@@ -161,13 +161,51 @@ async fn get_volume() {
...
+    let max_entries = 2;
+    let starting_token = Some(0);
+    let snapshots = test_cluster
         .rest_v00()
         .snapshots_api()
-        .get_volumes_snapshots(None, None)
+        .get_volumes_snapshots(max_entries, None, None, starting_token)
         .await
         .expect("Listing the snapshots failed");
+    assert_eq!(snapshots.entries.len(), max_entries as usize);
```
tiagolobocastro: Can we extend this test to "walk" to the end?

control-plane/grpc/src/operations/volume/traits_snapshots.rs
```diff
@@ -270,13 +270,20 @@ impl VolumeSnapshotState {
 pub struct VolumeSnapshots {
     /// Snapshot entries.
     pub entries: Vec<VolumeSnapshot>,
+    /// Next token for the paginated response.
+    pub next_token: Option<u64>,
```
tiagolobocastro, line 292: Does it need to be pub if we have a getter?

tiagolobocastro (approved):

control-plane/agents/src/bin/core/controller/resources/resource_map.rs
dsharma-dc: `filter` might seemingly confuse with keyword `filter`. Our usage is `.filter(filter)`
Naming it `seive` or `closure` perhaps.

control-plane/plugin/src/resources/tests.rs
dsharma-dc: done.

control-plane/grpc/src/operations/volume/traits_snapshots.rs
dsharma-dc: It's currently required by core/volume/service to instantiate in grpc response. I kept it without a `new` constructor following the existing templates.

dsharma-dc:
bors merge

dsharma-dc:
bors try

dsharma-dc:
bors merge

PR #574, opened 2023-06-13
fix(snapshots): fix source id in replica snapshot state

dsharma-dc:
The source id was incorrectly set as snap id itself. Fixed it to represent the replica uuid.


tiagolobocastro (approved):

dsharma-dc:
bors merge

PR #575, opened 2023-06-14
test(snapshots): bdd feature file and tests for snapshot list

dsharma-dc:


tests/bdd/features/snapshot/list/list.feature
```diff
@@ -0,0 +1,29 @@
...
+    And we have created more than one snapshot for the volume
+    Then the snapshots should be listed by <filter>
+    Examples:
+      | publish_status | filter        |
+      | published      | by_none        |
+      | published      | by_volume      |
+      | published      | by_volume_snap |
+      | unpublished    | by_none        |
+      | unpublished    | by_volume      |
+      | unpublished    | by_volume_snap |
```
dsharma-dc, line 20: Actually, I should better merge these two scenario when I look at them again from a 1000ft.

tests/bdd/features/snapshot/list/test_list.py
```diff
@@ -0,0 +1,116 @@
...
+]
+
+
+@scenario("list.feature", "Volume snapshots list")
+def test_volume_snapshots_list():
+    """Volume snapshots list."""
+
+
+@given("a deployer cluster")
+def a_deployer_cluster():
```
tiagolobocastro: let's mark this as a module fixture (might need to split impl into separate fixture perhaps) to avoid recreating cluster for every example

tests/bdd/features/snapshot/list/test_list.py
```diff
@@ -0,0 +1,116 @@
...
+            thin=False,
+        ),
+    )
+    if publish_status == "published":
+        ApiClient.volumes_api().put_volume_target(
+            VOLUME1_UUID,
+            publish_volume_body=PublishVolumeBody(
+                {}, Protocol("nvmf"), node=NODE1, frontend_node="app-node-1"
+            ),
+        )
```
tiagolobocastro: let's yield and delete volume

tests/bdd/features/snapshot/list/test_list.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+@given("we have created more than one snapshot for the volume")
+def we_have_created_more_than_one_snapshot_for_the_volume():
+    """we have created more than one snapshot for the volume."""
+    for snapid in VOLUME1_SNAP_IDS:
+        ApiClient.snapshots_api().put_volume_snapshot(VOLUME1_UUID, snapid)
```
tiagolobocastro: let's yield and delete snapshot

tiagolobocastro (commented):
Should we add tests for paging as well?

dsharma-dc:
> Should we add tests for paging as well?

Yes I guess. I'll add a scenario for that.

tests/bdd/features/snapshot/list/list.feature
```diff
@@ -0,0 +1,21 @@
+Feature: Volume Snapshot listing
+
+  Background:
+    Given a deployer cluster
+
+  # LISTING BY DIFFERENT CRITERIAS
+  Scenario Outline: Volume snapshots list
+    Given we have a single replica <publish_status> volume
+    And we have created more than one snapshot for the volume
+    Then the snapshots should be listed by <filter>
```
Abhinandan-Purkait, line 10: Would be worth it adding  another bdd similar to this, but after we have disruption, like node killed and brought back or restarted ? cc @tiagolobocastro

dsharma-dc: I think we are doing that kind of validation in create bdd tests.

Abhinandan-Purkait: We are not right? We only use the `get by snapshot id` call in all other tests. We need to test all paths.

dsharma-dc: Yes I was referring to that get by snapshot id after a node disruption. Is it worth running all filters for listing after a node reboot? The existence of snapshot is to be established primarily.

tiagolobocastro: Yeah might be worth it to have one scenario where we stop/kill a node and list, and then we bring it back and list. 
Just to be on the safe side all is as expected and no conversion missing.

tiagolobocastro: I think we also should add here checks for all snapshot fields, example:
```
    And the snapshot status ready to clone should be true
    And the snapshot source size should be equal to the volume size
    And the snapshot size should be equal to the volume size
    And the snapshot timestamp should be in the RFC 3339 format
    And the snapshot timestamp should be within 1 minute of creation (UTC)
    And the volume snapshot state has a single online replica snapshot
    And the replica snapshot's timestamp should be in the RFC 3339 format
    And the replica snapshot's timestamp should be within 1 minute of creation (UTC)
```

dsharma-dc: Shall we rather have a single line in bdd saying:
`And the snapshot fields must be valid`

Then in the test implementation, keep an array or dictionary of all the fields that are to be validated, and validate them. This would avoid changing bdd if any new field is added/removed in future.

tiagolobocastro: One can have that if we have multiple scenarios but we should have one scenario with that explicitly so it's clear. However  I've added that on created, so maybe that's enough to keep it explicit there and simple here CC @Abhinandan-Purkait

tests/bdd/features/snapshot/list/test_list.py
```diff
@@ -0,0 +1,382 @@
...
+        pytest.fail("Timestamp is in the future?")
+    age = now - timestamp
+    if age > datetime.timedelta(minutes=minutes):
+        pytest.fail("Timestamp is older than 1 minute")
+
+
+def del_volume():
+    try:
+        ApiClient.volumes_api().del_volume(VOLUME1_UUID)
+        ApiClient.volumes_api().del_volume(VOLUME2_UUID)
```
tiagolobocastro: probably need to coalesce, otherwise if first fails, second is not deleted?

tests/bdd/features/snapshot/list/test_list.py
```diff
@@ -0,0 +1,382 @@
...
+    wait_node_ready(name)
+
+
+@retry(wait_fixed=100, stop_max_attempt_number=20)
+def wait_node_ready(name):
+    node = ApiClient.nodes_api().get_node(name)
+    assert node.state.status == NodeStatus("Online")
+
+
+@retry(wait_fixed=500, stop_max_attempt_number=20)
```
tiagolobocastro: Keep the same as other bdds
```suggestion
@retry(wait_fixed=100, stop_max_attempt_number=20)
```

tests/bdd/features/snapshot/list/test_list.py
```diff
@@ -0,0 +1,382 @@
...
+def wait_volume_replica_offline(volume):
+    volume = ApiClient.volumes_api().get_volume(volume.spec.uuid)
+    replicas = list(volume.state.replica_topology.values())
+    assert len(replicas) == 1
+    assert replicas[0].state == ReplicaState("Unknown")
+
+
+@retry(wait_fixed=300, stop_max_attempt_number=20)
```
tiagolobocastro:
```suggestion
@retry(wait_fixed=100, stop_max_attempt_number=20)
```

tiagolobocastro (approved):

tests/bdd/features/snapshot/list/test_list.py
```diff
@@ -0,0 +1,382 @@
...
+
+def restart_node(name):
+    Docker.restart_container(name)
+    wait_node_ready(name)
+
+
+@retry(wait_fixed=100, stop_max_attempt_number=20)
+def wait_node_ready(name):
+    node = ApiClient.nodes_api().get_node(name)
+    assert node.state.status == NodeStatus("Online")
```
tiagolobocastro, line 351: I'm updating create test with this too

```suggestion
    assert node.state.status == NodeStatus("Online")
    for pool in ApiClient.pools_api().get_node_pools(name):
        assert pool.state.status == PoolStatus("Online")
```

tests/bdd/features/snapshot/list/test_list.py
Abhinandan-Purkait: @tiagolobocastro Should this be atleast equal to the cache_period?

Abhinandan-Purkait (approved):

tests/bdd/features/snapshot/list/test_list.py
dsharma-dc: This should be ok

```
def del_volume():
    for vol in [VOLUME1_UUID, VOLUME2_UUID]:
        try:
            ApiClient.volumes_api().del_volume(vol)
        except NotFoundException:
            pass
```

dsharma-dc:
bors try

tests/bdd/features/snapshot/list/test_list.py
tiagolobocastro: Looks good to me

tests/bdd/features/snapshot/list/test_list.py
tiagolobocastro: That would imply we are synced with the cache reload, we're not. We only know next poll is <= "cache_period" away, so could be "cache_period", could be 40ms, could be 0ms.

dsharma-dc:
bors merge

PR #577, opened 2023-06-14
test(snapshot/bdd/create): add feature file for snapshot creation

tiagolobocastro:


tiagolobocastro:
@Abhinandan-Purkait @dsharma-dc 
I'll start coding the tests meanwhile but would be good to get feedback on the feature file itself

tests/bdd/features/snapshot/create/feature.feature
```diff
@@ -0,0 +1,78 @@
...
+
+  Scenario Outline: Snapshot creation of a single replica volume
+    Given we have a single replica <publish_status> volume where the replica is <replica_location> to the target
+    When we try to create a snapshot for the volume
+    Then the snapshot creation should be successful
+    And the snapshot status ready to clone should be true
+    And the snapshot source size should be equal to the volume size
+    And the snapshot size should be equal to the volume size
+    And the snapshot timestamp should be in the RFC 3339 format
+    And the snapshot timestamp should be within 1 minute of creation (UTC)
```
dsharma-dc, line 14: We don't really need to mention timezone UTC here specifically, as we are talking about a delta.

dsharma-dc: What do we consider "creation" time here? The timestamp at which test will call put_volume_snapshot?

tests/bdd/features/snapshot/create/feature.feature
```diff
@@ -0,0 +1,78 @@
...
+      | unpublished    | remote           |
+      | unpublished    | local            |
+
+  Scenario: Snapshot creation for a single replica volume fails when the replica is offline
+    Given we have a single replica volume
+    And the replica is offline
+    And volume is in faulted state
+    When we try to create a snapshot for the volume
+    Then the snapshot creation should fail with preconditions failed
+    And no snapshot should be listed
```
dsharma-dc: Rather `the snapshot uuid used in create snapshot request shouldn't get listed` , because we may have possibly existing snapshots for the replica.

tests/bdd/features/snapshot/create/feature.feature
```diff
@@ -0,0 +1,78 @@
...
+    When the node A is brought back
+    And we retry the same snapshot creation
+    Then the snapshot creation should be successful
+
+  Scenario: Snapshot creation of a single replica volume after its source is deleted
+    Given we have a single replica volume
+    When a snapshot is created for the volume
+    When the volume is deleted
+    And we retry the same snapshot creation
+    Then it should fail with already exists
```
dsharma-dc, line 64: This one looks confusing to me. Are we testing that `same snapshot/id creation should fail with eexist` i.e. the last scenario down in this file, or we are testing that `any snapshot creation after volume deletion should fail` ?

tests/bdd/features/snapshot/create/feature.feature
tiagolobocastro: > We don't really need to mention timezone UTC here specifically, as we are talking about a delta.

We might do because if your computer is not on UTC then you can't just compare time as the snapshot time is in utc?

> What do we consider "creation" time here? The timestamp at which test will call put_volume_snapshot?

It doesn't matter too much if it's right before we issue the call or after, 1 minute will be a large enough slack for work for both cases

tests/bdd/features/snapshot/create/feature.feature
tiagolobocastro: we're testing that it fails with eexists even when the volume has been deleted

tests/bdd/features/snapshot/create/feature.feature
```diff
@@ -0,0 +1,78 @@
...
+
+  Scenario: Snapshot creation for multi-replica volume is not supported
+    Given we have a multi-replica volume
+    When we try to create a snapshot for the volume
+    Then the snapshot creation should be fail with preconditions failed
+
+  Scenario: Existence of snapshot after node restart
+    Given we have a single replica volume
+    And a successfully created snapshot
+    When we stop the node hosting the snapshot
```
Abhinandan-Purkait: Instead of stopping and starting can we just restart?

Abhinandan-Purkait: And if we want to stop and start then probably should we kill and start, i.e ungracefully.

tests/bdd/features/snapshot/create/feature.feature
```diff
@@ -0,0 +1,78 @@
...
+    Then the replica snapshot should be reported as offline
+    When we resume the node hosting the snapshot
+    Then the replica snapshot should be reported as online
+
+  Scenario: Retrying snapshot creation until it succeeds
+    Given we have a single replica volume
+    And the volume is published on node A
+    And the node A is brought down
```
Abhinandan-Purkait: pause rather than stop ? Otherwise we might hit precondition? :-p

tests/bdd/features/snapshot/create/feature.feature
tiagolobocastro: why does it matter whether we stop or kill here?

tests/bdd/features/snapshot/create/feature.feature
tiagolobocastro: Right :D

tests/bdd/features/snapshot/create/feature.feature
Abhinandan-Purkait: Stop does things gracefully, which might not be the case always.

tiagolobocastro: Vice versa :) upgrades etc will use stop

tiagolobocastro: Maybe we could use outline and test both

Abhinandan-Purkait: Right.

control-plane/rest/openapi-specs/v0_api_spec.yaml
```diff
@@ -3305,7 +3308,6 @@ components:
         - uuid
         - size
         - source_volume
-        - timestamp
```
Abhinandan-Purkait, line 3308: Do we even need the timestamp for VolumeSnapshotState? Csi will consume from metadata. And in state we have the individual replica snapshot state time stamp.

tests/bdd/features/snapshot/create/test_feature.py
```diff
@@ -0,0 +1,527 @@
...
+    yield request
+
+
+def snapshot_create_attempt(snapshot_create_request):
+    pytest.exception = None
+    pytest.snapshot = None
+    try:
+        pytest.snapshot = snapshot_create_request()
+    except openapi.exceptions.ApiException as e:
+        print(ApiClient.snapshots_api().get_volumes_snapshots())
```
Abhinandan-Purkait: Don't need the print?

Abhinandan-Purkait (approved):

control-plane/rest/openapi-specs/v0_api_spec.yaml
tiagolobocastro: hmm we might not even need it yes

tests/bdd/features/snapshot/create/test_feature.py
tiagolobocastro: No, it's from debugging :D

tiagolobocastro:
bors merge

PR #578, opened 2023-06-16
Multi tag fix

pchandra19:
Fix for build failure as below , when multiple tags are on HEAD 
Building mayadata/mcp-core:v1.0.6 v1.0.7 ...
building '/nix/store/alzvq957g8xky34yz9bqfqqs74yx2dld-mayastor-version.drv'...
error: The path name 'control-plane-v1.0.6 v1.0.7' is invalid: the ' ' character is invalid. Path names are alphanumeric and can include the symbols +-._?= and must not begin with a period. Note: If 'control-plane-v1.0.6 v1.0.7' is a source file and you cannot rename it on disk, builtins.path { name = ... } can be used to give it an alternative name.
(use '--show-trace' to show detailed location information)

Also, modified the GA to display tag version on kubectl binary

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -16,7 +16,7 @@ dockerhub_tag_exists() {
 
 # Get the tag at the HEAD
 get_tag() {
-  vers=`git tag --points-at HEAD`
+  vers=`git tag --points-at HEAD | tail -n 1`
   echo -n $vers
 }
 get_hash() {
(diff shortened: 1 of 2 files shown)


scripts/release.sh
```diff
@@ -16,7 +16,7 @@ dockerhub_tag_exists() {
 
 # Get the tag at the HEAD
 get_tag() {
-  vers=`git tag --points-at HEAD`
+  vers=`git tag --points-at HEAD | tail -n 1`
```
tiagolobocastro: Btw this is how dataplane is doing it:
```suggestion
  vers=`git describe --exact-match 2>/dev/null || echo ""`
```

tiagolobocastro (approved):

PR #581, opened 2023-06-19
feat(snapshot): mark snapshot ready to csi, reject creation with unready snapshot source

Abhinandan-Purkait:


Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors cancel

Abhinandan-Purkait:
bors try-

Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors try-

Abhinandan-Purkait:
bors cancel

Abhinandan-Purkait:
bors try

control-plane/csi-driver/src/bin/controller/controller.rs
```diff
@@ -174,11 +175,41 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
         tracing::trace!(request = ?args);
         let _permit = self.create_volume_permit().await?;
 
-        if args.volume_content_source.is_some() {
-            return Err(Status::invalid_argument(
-                "Source for create volume is not supported",
-            ));
-        }
+        if let Some(source) = args.volume_content_source {
```
tiagolobocastro: I don't think this is quite right, because you're allowing the create call to proceed when `ready_as_source` is true, but the current csi-controller code doesn't support it.
I think it's better to keep as is, just return error is content is some, and we'll change this when we start supporting the feature.

control-plane/csi-driver/src/bin/controller/controller.rs
```diff
@@ -817,7 +848,7 @@ fn snapshot_to_csi(snapshot: models::VolumeSnapshot) -> Snapshot {
             .metadata
             .timestamp
             .and_then(|t| prost_types::Timestamp::from_str(&t).ok()),
-        // What about when snapshot is in deleted phase?
-        ready_to_use: snapshot.definition.metadata.status == models::SpecStatus::Created,
+        // Seems like csi, doesn't really care what the state is after successful creation.
+        ready_to_use: (snapshot.definition.metadata.status == models::SpecStatus::Created),
```
tiagolobocastro: are the parenthesis needed?

control-plane/grpc/src/operations/volume/traits_snapshots.rs
```diff
@@ -276,13 +276,12 @@ impl VolumeSnapshotState {
...
         self.timestamp.as_ref()
     }
-    /// Get the volume snapshot readiness to clone.
-    pub fn clone_ready(&self) -> bool {
-        match self.repl_snapshots.as_slice() {
-            [VolumeReplicaSnapshotState::Online { state, .. }] => state.clone_ready(),
-            // todo: handle more than one replica for multi-replica snapshots
+    /// Get the volume snapshot readiness to be used as a source.
+    pub fn ready_as_source(&self) -> bool {
+        self.repl_snapshots.iter().any(|snapshot| match snapshot {
```
tiagolobocastro: please keep this as it was, we don't support more than 1 replica snapshot clone even when clone drops

tests/bdd/features/snapshot/csi/controller/operations.feature
```diff
@@ -19,3 +19,9 @@ Feature: Snapshot - CSI Controller Operations
         And a snapshot is created for that volume
         When a ListSnapshotRequest request is sent to the CSI controller
         Then the list should succeed
+
+    Scenario: Create Volume Operation with snapshot source
+        Given we have a single replica volume
+        And a snapshot is created for that volume
+        When a CreateVolumeRequest request with snapshot as source is sent to the CSI controller
+        Then the volume creation should fail
```
tiagolobocastro:
```suggestion
        Then the volume creation should fail with invalid argument
```

tiagolobocastro (approved):

control-plane/csi-driver/src/bin/controller/controller.rs
Abhinandan-Purkait: But the ready_as_source would not be true right? We are setting it as false.

control-plane/csi-driver/src/bin/controller/controller.rs
Abhinandan-Purkait: No, they are not.

control-plane/grpc/src/operations/volume/traits_snapshots.rs
Abhinandan-Purkait: Okay.

control-plane/csi-driver/src/bin/controller/controller.rs
Abhinandan-Purkait: And since its coming all the way from dataplane I don't think it would be `true` right?

tiagolobocastro: It may be during upgrade, if for some reason core gets updated before csi-controller. Then csi-controller will be completely broken and create a "normal" volume as we saw yesterday, which is really bad!

Abhinandan-Purkait: Okay.

Abhinandan-Purkait:
bors merge

PR #584, opened 2023-06-20
fix(snapshots): only list created snapshots, and not creating

dsharma-dc:


control-plane/rest/service/src/v0/snapshots.rs
```diff
@@ -90,7 +90,15 @@ impl apis::actix_server::Snapshots for RestApi {
             entries: snaps
                 .entries()
                 .iter()
-                .map(to_models_volume_snapshot)
+                .filter_map(|s| {
+                    if models::SpecStatus::from(s.meta().status().clone())
```
tiagolobocastro: This doesn't seem right, I think this should be done in the CSI side and not here

dsharma-dc: @tiagolobocastro I've fixed to make changes to the csi now.

dsharma-dc:
A snapshot in Creating status:

`[VolumeSnapshot { definition: VolumeSnapshotDefinition { metadata: VolumeSnapshotMetadata { status: Creating, timestamp: None, txn_id: "1", transactions: {"1": [ReplicaSnapshot { uuid: b60bdab5-3809-4408-a24b-d360f51b387c, source_id: 494efaa3-ba63-4307-aa2e-7ae2b0cac1d7, status: Deleting }]} }....`

Without this change - simulated test for snapshot as Creating. The list call from csi operations test still receives Creating snapshot.

```
***LISTED***
entries {
  snapshot {
    snapshot_id: "3f49d30d-a446-4b40-b3f6-f439345f1ce9"
    source_volume_id: "d01b8bfb-0116-47b0-a03a-447fcbdc0e99"
  }
}
```

With this change - simulated test for snapshot as Creating. The list call from csi operations test does not receive Creating snapshot.

```
***LISTED***
<>
```

dsharma-dc:
bors try

control-plane/csi-driver/src/bin/controller/client.rs
```diff
@@ -408,6 +408,16 @@ impl IoEngineApiClient {
...
 
-        Ok(snapshots.into_body())
+        let next_token = snapshots.body().next_token;
+        // Return only the snapshots that are Created, leaving out other statuses if any.
+        Ok(models::VolumeSnapshots {
+            entries: snapshots
+                .into_body()
+                .entries
+                .into_iter()
+                .filter(|s| s.definition.metadata.status == models::SpecStatus::Created)
```
tiagolobocastro: hmm, perhaps it should instead filter out Creating state, otherwise we won't list snapshots in the delete process? Guess depends what behaviour we want here CC @Abhinandan-Purkait

dsharma-dc: I think it might be confusing to list the ones in delete process because this status isn't visible to CSI(or is it?), so consumer won't know it's a Created one or Deleting one.

tiagolobocastro: No, but if it's deleting and not deleted it might also be confusing, because the snapshot is still there, it hasn't been deleted and the storage usage is still there.
Also, currently the delete call fails if we fail to delete the snapshot, so for consistency we should list deleting as well.

dsharma-dc: tweaked the filter, please take a look.

control-plane/csi-driver/src/bin/controller/client.rs
```diff
@@ -408,6 +408,16 @@ impl IoEngineApiClient {
             )
             .await?;
 
-        Ok(snapshots.into_body())
+        let next_token = snapshots.body().next_token;
+        // Return only the snapshots that are Created, leaving out other statuses if any.
```
tiagolobocastro:
```suggestion
        // Return only the snapshots that have been created, leaving out other statuses if any.
```

tiagolobocastro (approved):

control-plane/csi-driver/src/bin/controller/client.rs
dsharma-dc: I'd rather modify this comment now as `Don't return snapshots that are still in Creating state`

dsharma-dc:
bors merge

control-plane/csi-driver/src/bin/controller/client.rs
tiagolobocastro: sounds good to me

dsharma-dc:
bors cancel

control-plane/csi-driver/src/bin/controller/client.rs
dsharma-dc: @tiagolobocastro Another thought around this - if we are listing a 'deleting' one, and the consumer picks that one up for next steps e.g. creating a clone, the clone create will fail possibly failing an application. There'll always be more successful deletions compared to sporadic deletion failures.
I'd say we don't want that kind of angst?

tiagolobocastro: If CSI has already tried to delete it it's because the CR has been deleted.

dsharma-dc:
bors merge

dsharma-dc:
bors merge

PR #586, opened 2023-06-20
fix(snapshot/retry): add txnid to the snapshot name

tiagolobocastro:
This allows for safer retries as the snapshot name will not clash. Add timeout test which leads to a snapshot in the dataplane, but not in the control-plane. The retry on the control-plane should then succeed creating a new snapshot in the data-plane. todo: clean-up previous snapshot transaction in creating status?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -252,6 +252,7 @@ dependencies = [
  "hyper",
  "indexmap",
  "itertools",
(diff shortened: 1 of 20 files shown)


Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

PR #593, opened 2023-06-26
refactor(rebuild-history): update rpc to use timestamp filter

tiagolobocastro:
Only fetch rebuilds which the control-plane is not already aware of, since once the records are created they never change.


sinhaashish (approved):
LGTM

tiagolobocastro:
bors merge

PR #594, opened 2023-06-27
feat(callhome): add publisher lib, volume and pool events

datacore-vvarakantham:
Publisher library. Create and delete events for volume and pool. Deployer for testing events on nats mbus.

### DIFF
--- a/control-plane/agents/src/bin/core/main.rs
+++ b/control-plane/agents/src/bin/core/main.rs
@@ -99,6 +99,10 @@ pub(crate) struct CliArgs {
 
     #[clap(flatten)]
     thin_args: ThinArgs,
+
+    /// Events message-bus endpoint url.
+    #[clap(long, short)]
+    events_url: Option<url::Url>,
 }
 impl CliArgs {
     fn args() -> Self {
@@ -141,10 +145,11 @@ async fn main() {
     let cli_args = CliArgs::args();
     utils::print_package_info!();
     println!("Using options: {cli_args:?}");
-    utils::tracing_telemetry::init_tracing(
-        "core-agent",
+    utils::tracing_telemetry::init_tracing_with_eventing(
+        "agent-core",
         cli_args.tracing_tags.clone(),
         cli_args.jaeger.clone(),
+        cli_args.events_url.clone(),
     );
     server(cli_args).await;
 }
--- a/control-plane/grpc/src/operations/mod.rs
+++ b/control-plane/grpc/src/operations/mod.rs
@@ -1,6 +1,7 @@
 #![warn(missing_docs)]
 
 use crate::common;
+use events_api::event::EventMessage;
 use stor_port::types::v0::store::SpecStatus;
 
 /// Module for all corresponding client, server, traits for nexus rpc transport.
@@ -136,4 +137,10 @@ impl From<common::SpecStatus> for SpecStatus<()> {
             common::SpecStatus::Deleted => SpecStatus::Deleted,
         }
     }
+}
+
+/// Event trait definition for creating events.
+pub(crate) trait Event {
+    /// Create event message.
+    fn event(&self) -> EventMessage;
 }
--- a/control-plane/grpc/src/operations/pool/server.rs
+++ b/control-plane/grpc/src/operations/pool/server.rs
@@ -1,5 +1,5 @@
 use crate::{
-    operations::pool::traits::PoolOperations,
+    operations::{pool::traits::PoolOperations, Event},
     pool,
     pool::{
         create_pool_reply, get_pools_reply,
@@ -37,9 +37,12 @@ impl PoolGrpc for PoolServer {
     ) -> Result<tonic::Response<pool::CreatePoolReply>, tonic::Status> {
         let req: CreatePoolRequest = request.into_inner();
         match self.service.create(&req, None).await {
-            Ok(pool) => Ok(Response::new(CreatePoolReply {
-                reply: Some(create_pool_reply::Reply::Pool(pool.into())),
-            })),
+            Ok(pool) => {
+                let _ = pool.event().generate();
+                Ok(Response::new(CreatePoolReply {
+                    reply: Some(create_pool_reply::Reply::Pool(pool.into())),
+                }))
+            }
             Err(err) => Ok(Response::new(CreatePoolReply {
                 reply: Some(create_pool_reply::Reply::Error(err.into())),
             })),
@@ -52,7 +55,10 @@ impl PoolGrpc for PoolServer {
     ) -> Result<tonic::Response<DestroyPoolReply>, tonic::Status> {
         let req = request.into_inner();
         match self.service.destroy(&req, None).await {
-            Ok(()) => Ok(Response::new(DestroyPoolReply { error: None })),
+            Ok(()) => {
+                let _ = req.event().generate();
+                Ok(Response::new(DestroyPoolReply { error: None }))
+            }
             Err(e) => Ok(Response::new(DestroyPoolReply {
                 error: Some(e.into()),
             })),
--- a/control-plane/grpc/src/operations/pool/traits.rs
+++ b/control-plane/grpc/src/operations/pool/traits.rs
@@ -1,9 +1,11 @@
 use crate::{
     common,
     context::Context,
+    operations::Event,
     pool,
     pool::{get_pools_request, CreatePoolRequest, DestroyPoolRequest},
 };
+use events_api::event::{EventAction, EventCategory, EventMessage, EventMeta, EventSource};
 use std::convert::TryFrom;
 use stor_port::{
     transport_api::{v0::Pools, ReplyError, ResourceKind},
@@ -382,6 +384,40 @@ impl From<PoolSpecStatus> for common::SpecStatus {
             PoolSpecStatus::Created(_) => Self::Created,
             PoolSpecStatus::Deleting => Self::Deleting,
             PoolSpecStatus::Deleted => Self::Deleted,
+        }
+    }
+}
+
+// Create pool creation event message from pool data.
+impl Event for Pool {
+    fn event(&self) -> EventMessage {
+        let event_meta = EventMeta::new();
+        let event_source = EventSource::new(self.node().to_string());
+        EventMessage {
+            category: EventCategory::Pool as i32,
+            action: EventAction::Create as i32,
+            target: self.id().to_string(),
+            metadata: Some(EventMeta {
+                source: Some(event_source),
+                ..event_meta
+            }),
+        }
+    }
+}
+
+// Create pool deletion event message.
+impl Event for DestroyPoolRequest {
+    fn event(&self) -> EventMessage {
+        let event_meta = EventMeta::new();
+        let event_source = EventSource::new(self.node_id().to_string());
+        EventMessage {
+            category: EventCategory::Pool as i32,
+            action: EventAction::Delete as i32,
+            target: self.pool_id().to_string(),
+            metadata: Some(EventMeta {
+                source: Some(event_source),
+                ..event_meta
+            }),
         }
     }
 }
--- a/control-plane/grpc/src/operations/volume/traits.rs
+++ b/control-plane/grpc/src/operations/volume/traits.rs
@@ -4,7 +4,7 @@ use crate::{
     context::Context,
     misc::traits::{StringValue, ValidateRequestTypes},
     nexus,
-    operations::Pagination,
+    operations::{Event, Pagination},
     replica, volume,
     volume::{
         get_volumes_request, CreateVolumeRequest, DestroyShutdownTargetRequest,
@@ -12,6 +12,7 @@ use crate::{
         SetVolumeReplicaRequest, ShareVolumeRequest, UnpublishVolumeRequest, UnshareVolumeRequest,
     },
 };
+use events_api::event::{EventAction, EventCategory, EventMessage, EventMeta, EventSource};
 use stor_port::{
     transport_api::{v0::Volumes, ReplyError, ResourceKind},
     types::v0::{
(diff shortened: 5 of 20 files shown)


control-plane/agents/src/bin/core/main.rs
```diff
@@ -99,6 +99,19 @@ pub(crate) struct CliArgs {
 
     #[clap(flatten)]
     thin_args: ThinArgs,
+
+    /// Trace events to the mbus.
+    #[clap(long, short, default_value = "false")]
+    eventing: bool,
+
+    /// Mbus server url.
+    /// If eventing is true default value can be used.
```
tiagolobocastro: Why not leave as None by default and use mbus_url itself, if set then eventing is enabled

control-plane/grpc/src/operations/mod.rs
```diff
@@ -137,3 +137,8 @@ impl From<common::SpecStatus> for SpecStatus<()> {
         }
     }
 }
+
+pub(crate) trait Event<T> {
```
tiagolobocastro: What does this do?

control-plane/grpc/src/operations/pool/server.rs
```diff
@@ -52,7 +56,10 @@ impl PoolGrpc for PoolServer {
     ) -> Result<tonic::Response<DestroyPoolReply>, tonic::Status> {
         let req = request.into_inner();
         match self.service.destroy(&req, None).await {
-            Ok(()) => Ok(Response::new(DestroyPoolReply { error: None })),
+            Ok(()) => {
+                tracing::event!(target: "target-mbus", tracing::Level::INFO, event_data = &EventMessage::get_event(&req));
```
tiagolobocastro: Due to the way we have idempotency these should be on dataplane, otherwise we might not even receive this event was a pool is deleted.

control-plane/grpc/src/operations/mod.rs
```diff
@@ -137,3 +137,8 @@ impl From<common::SpecStatus> for SpecStatus<()> {
...
+    /// Create event message from value.
+    fn get_event(value: &T) -> String;
```
tiagolobocastro:
```suggestion
    fn event(value: &T) -> String;
```

tiagolobocastro: Also, why do we need the T, can't we just have:
```rust
pub(crate) trait Event { fn event(&self) -> String; }
```
Also, should the output be an "Event" type ? And let the "user" of this event convert to String or whichever is required?

control-plane/grpc/src/operations/pool/traits.rs
```diff
@@ -385,3 +387,43 @@ impl From<PoolSpecStatus> for common::SpecStatus {
...
+impl Event<Pool> for EventMessage {
+    fn get_event(value: &Pool) -> String {
+        let event_meta = EventMeta::new();
+        let event = EventMessage {
+            category: Category::Pool as i32,
+            action: Action::CreateEvent as i32,
+            target: value.id().to_string(),
+            metadata: Some(EventMeta {
+                source: Some(EventSource {
+                    component: "core-agent".to_string(),
```
tiagolobocastro: this is not correct, this is a transport crate, any component could be using these. Maybe should be set by the caller?

utils/event-publisher/src/event_handler.rs
```diff
@@ -0,0 +1,22 @@
...
+impl EventHandle {
+    /// Initilize the buffer, starts the event publisher and return the layer for handling the event
+    /// traces.
+    pub fn init(mbus_url: String) -> EventLayer {
+        let (send, recv) = tokio::sync::mpsc::channel::<EventMessage>(MAX_BUFFER_MSGS);
+
+        let layer = EventLayer::new(send);
+        tokio::spawn(async move {
+            let publisher = MbusPublisher::new(&mbus_url).await;
+            publisher.publish_events(recv).await;
```
tiagolobocastro: missing shutdown handler, or is that handled by the publisher already?

utils/event-publisher/src/event_layer.rs
```diff
@@ -0,0 +1,58 @@
+use mbus_api::event::EventMessage;
+use std::fmt::Debug;
+use tokio::sync::mpsc::{error::TrySendError, Sender};
+use tracing::field::{Field, Visit};
+use tracing_subscriber::{layer::Context, Layer};
+
+pub struct EventLayer {
```
tiagolobocastro: missing doc comments throughout

utils/event-publisher/src/event_layer.rs
```diff
@@ -0,0 +1,58 @@
...
+        let mut visitor = EventVisitor::default();
+
+        event.record(&mut visitor);
+
+        let output = visitor.0;
+
+        if let Err(err) = self.sender.try_send(output) {
+            match err {
+                TrySendError::Full(_) => {
+                    tracing::warn!("Buffer is full");
```
tiagolobocastro: when can this happen, can we limit these warnings to once every X seconds?

utils/event-publisher/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "event-publisher"
```
tiagolobocastro: Should this be in the dependencies repo?

utils/event-publisher/src/publisher.rs
```diff
@@ -0,0 +1,39 @@
...
+            mbus: mbus_init(mbus_url).await,
+        }
+    }
+
+    /// Receives even messages from buffer through receiver handler of the buffer and publishes the
+    /// messages to the mbus.
+    pub async fn publish_events(mut self, mut recv: tokio::sync::mpsc::Receiver<EventMessage>) {
+        loop {
+            let msg = recv.recv().await;
+            if let Some(msg1) = msg {
```
tiagolobocastro: shouldn't we break out of the loop if msg is None?
```rust
let Some(msg) = msg else { break; };
```

control-plane/agents/src/bin/core/main.rs
datacore-vvarakantham: Should I pass default bus url directly though args, if eventing is enabled?

tiagolobocastro: Yes just pass it through args

control-plane/grpc/src/operations/mod.rs
datacore-vvarakantham: It's the trait definition for generating events. 
impl From<Volume> for EventMessage {} gives error that Volume and EventMessage are from external crates. So I defined a trait and using that to get event message.

control-plane/grpc/src/operations/mod.rs
datacore-vvarakantham: The event(value) takes different value types to get the event message. Can we have a single trait without T for different data types?

The tracing::event() is accepting Json string. So I directly converted it to the String type. We can also do the conversion before calling tracing::event. Should I change the code to that?

utils/event-publisher/src/event_handler.rs
datacore-vvarakantham: Currently I am just printing out the publish errors without closing the receiver. 

If the receiver is disconnected due to some error, the channel should be recreated and the sender handle in event layer should also be replaced with the sender from the recreated one. Should I handle the disconnection this way?

control-plane/grpc/src/operations/mod.rs
tiagolobocastro: Please document all that here as it's not clear to me from just looking at the code.
hmm but Event is local to this crate so it should work?
Ah right so that's why you have Event here, in this case output of event should be `EventMessage`.

utils/event-publisher/src/event_layer.rs
datacore-vvarakantham: Whenever there is a new message to be sent to the buffer and the buffer is full this error occurs. I am just printing out the error without interrupting the code with retries. Should I limit the warnings based on timings?

utils/event-publisher/src/publisher.rs
datacore-vvarakantham: Yes we should. 
This method returns None if the channel has been closed and there are no remaining messages in the channel’s buffer. The channel is closed when all senders have been dropped, or when close is called.

I will handle that case.

control-plane/grpc/src/operations/mod.rs
tiagolobocastro: Yes we can have different Self type, so you can impl Event for Pool or impl Event for Volume for example.

Yes I think we should return a specific EventMessage type and have the conversion to json/string at the caller side; 
In tracing there's a new trait coming (not stable yet) which would allow us to impl the tracing type for Event;

For now I suggest adding a .to_string() for EventMessage so you could to
pool.event().to_string()

control-plane/grpc/src/operations/pool/server.rs
tiagolobocastro: As discussed, let's keep it in until we have it on dataplane

utils/deployer-cluster/src/lib.rs
```diff
@@ -746,6 +746,12 @@ impl ClusterBuilder {
         }
         self
     }
+    // /// Enable eventing.
```
tiagolobocastro: if not needed can delete?

utils/event-publisher/src/event_handler.rs
tiagolobocastro: I mean shutdown as in cint or sigterm being received? 
We have a utils shutdown crate which let's us select futures with shutdown so we can exit gracefully, and for example flush any remaining events, but if we don't need to flush anything as long as this doesn't block the process from exiting then it's fine.

utils/event-publisher/src/event_layer.rs
tiagolobocastro: Yes I think so, to avoid flooding stdout with errors if nats/buffer get full.

utils/utils-lib/src/constants.rs
```diff
@@ -129,3 +129,5 @@ pub const SNAPSHOT_TRANSACTION_PRUNE_LIMIT: usize = 10;
 
 /// Maximum number of snapshot transactions allowed.
 pub const SNAPSHOT_MAX_TRANSACTION_LIMIT: usize = 5;
+/// The default mbus server url.
+pub const DEFAULT_MBUS_SERVER_URL: &str = "nats://mayastor-nats:4222";
```
tiagolobocastro: This repo is kubernetes agnostic, shouldn't have a default here, or at least should be based on deployer cluster container name, eg: nats://nats:4222

tiagolobocastro (commented):
Can we also add new containers for deployer so we can test this? 
And add some basic testing

utils/event-publisher/src/event_handler.rs
niladrih: And it should be able to fail without returning a non-zero error code, because the bus might be unreachable (e.g. helm chart uninstall).

control-plane/grpc/src/operations/pool/server.rs
```diff
@@ -37,9 +37,12 @@ impl PoolGrpc for PoolServer {
     ) -> Result<tonic::Response<pool::CreatePoolReply>, tonic::Status> {
         let req: CreatePoolRequest = request.into_inner();
         match self.service.create(&req, None).await {
-            Ok(pool) => Ok(Response::new(CreatePoolReply {
-                reply: Some(create_pool_reply::Reply::Pool(pool.into())),
-            })),
+            Ok(pool) => {
+                tracing::event!(target: "target-mbus", tracing::Level::INFO, event_data = &pool.event().to_string());
```
niladrih: nit: Could this be put inside a macro or a function (which accepts a dyn ToString), because of how often this is used?

control-plane/grpc/src/operations/volume/traits.rs
```diff
@@ -1630,3 +1631,40 @@ impl TryFrom<volume::AffinityGroupSpec> for AffinityGroupSpec {
...
+                    ..event_source
+                }),
+                ..event_meta
+            }),
+        }
+    }
+}
+
+// Create volume delete event message.
+impl Event for ValidatedDestroyVolumeRequest {
```
niladrih: Why are the types implementing 'Event' in case of vol creation and vol deletion named so differently (Volume vs ValidateDestroyVolumeRequest)

datacore-vvarakantham: I could get the volume creation data from 'volume' and deletion data from 'ValidateDestroyVolumeRequest'. So I implemented event trait on these types to get the EventMessage. 

The 'volume' is the volume data that we receive as a response from the volumeserver in the create_volume fn.
And we don't get any data as a response from the volumeserver in the delete_volume fn. So I am using the data from the request data - 'ValidateDestroyVolumeRequest'.

utils/event-publisher/src/event_handler.rs
datacore-vvarakantham: Even if the nats connection is lost or nats is uninstalled, the thread will continue to try to connect to nats.

sinhaashish (approved):
LGTM

control-plane/agents/src/bin/core/main.rs
```diff
@@ -99,6 +99,10 @@ pub(crate) struct CliArgs {
 
     #[clap(flatten)]
     thin_args: ThinArgs,
+
+    /// Mbus server url.
+    #[clap(long, short)]
+    mbus_url: Option<String>,
```
tiagolobocastro: Should this be events_url rather than mbus_url?
Also should we use url::Url instead of String?

control-plane/agents/src/bin/core/main.rs
```diff
@@ -141,10 +145,12 @@ async fn main() {
     let cli_args = CliArgs::args();
     utils::print_package_info!();
     println!("Using options: {cli_args:?}");
-    utils::tracing_telemetry::init_tracing(
+    utils::tracing_telemetry::init_tracing_with_eventing(
         "core-agent",
         cli_args.tracing_tags.clone(),
         cli_args.jaeger.clone(),
+        "agent-core",
```
tiagolobocastro: Why is there core-agent and agent-core twice?

deployer/src/infra/nats.rs
```diff
@@ -0,0 +1,40 @@
...
+*/
+
+#[async_trait]
+impl ComponentAction for Nats {
+    fn configure(&self, options: &StartOptions, cfg: Builder) -> Result<Builder, Error> {
+        Ok(if !options.eventing {
+            cfg
+        } else {
+            cfg.add_container_spec(
+                ContainerSpec::from_image("nats", "nats:latest")
```
tiagolobocastro: If we add the nats binary to the shell and use the binary instead of the container, will make it easier for debug?
Alternatively we should at least pin nats to a particular version rather than latest.

deployer/src/infra/nats.rs
```diff
@@ -0,0 +1,40 @@
...
+        if options.eventing {
+            cfg.start("nats").await?;
+        }
+        if options.eventing || cfg.container_exists("nats").await {
+            let _ = message_bus_init("localhost:4222", Some(1)).await;
+        }
+        Ok(())
+    }
+    async fn wait_on(&self, _options: &StartOptions, _cfg: &ComposeTest) -> Result<(), Error> {
+        Ok(())
```
tiagolobocastro, line 45: We need to wait here till nats is ready

deployer/src/lib.rs
```diff
@@ -300,6 +300,9 @@ pub struct StartOptions {
     /// This can be used for userspace io against a volume's nvmf target using the spdk ioengine.
     #[clap(long)]
     pub(crate) fio_spdk: bool,
+
+    #[clap(long, default_value = "false")]
```
tiagolobocastro, line 304: add doc comments

tiagolobocastro:
We're still missing tests?

control-plane/agents/src/bin/core/main.rs
datacore-vvarakantham: "core-agent" is used for service name for tracing. 

Passing the source component "agent-core" from here. There was a review comment to use the source_component "agent-core" in from_str trait, because the app label is "agent-core" for the core-agent

tiagolobocastro: Can we use the same argument, we can rename core-agent to agent-core

utils/utils-lib/src/tracing_telemetry.rs
```diff
@@ -98,7 +118,20 @@ pub fn init_tracing_ext<T: std::net::ToSocketAddrs>(
         FmtLayer::None => (None, None),
     };
 
-    let subscriber = Registry::default().with(filter).with(stdout).with(stderr);
+    // Get the optional eventing layer.
+    let events_layer = match events_url {
+        Some(url) => {
+            let events_filter = filter::Targets::new().with_target("target-mbus", Level::INFO);
```
tiagolobocastro: Hmm not sure about this name, doesn't have events, what about `mbus-events-target` ?

control-plane/agents/src/bin/core/main.rs
```diff
@@ -99,6 +99,10 @@ pub(crate) struct CliArgs {
 
     #[clap(flatten)]
     thin_args: ThinArgs,
+
+    /// Mbus server url.
```
tiagolobocastro:
```suggestion
    /// Events message-bus endpoint url.
```

tiagolobocastro (commented):
Still missing a test using deployer or did I miss it?

tiagolobocastro (approved):

datacore-vvarakantham:
bors merge

datacore-vvarakantham:
bors merge

sinhaashish:
bors merge

PR #599, opened 2023-06-27
fix(snapshots/thin): a thick volume becomes thin after snapshots are created

tiagolobocastro:
test: tests should clean themselves
    
---

    fix(snapshot): update replica information
    
    After taking a snapshot the replica info changes, so update it as well.
    
---

    feat(snapshot/usage): add snapshot usage for the replica usage
    
    In preparation of pending dataplane rpc changes.
    
---

    fix(snapshots/thin): a thick volume becomes thin after snapshots are created
    
    After first snapshot is created we mark the volume as thin and this is sticky.
    

control-plane/stor-port/src/types/v0/store/volume.rs
```diff
@@ -428,9 +470,15 @@ impl SpecTransaction<VolumeOperation> for VolumeSpec {
 
     fn log_op(&self, operation: &VolumeOperation) -> (bool, bool) {
         match operation {
-            VolumeOperation::CreateSnapshot(_) | VolumeOperation::DestroySnapshot(_) => {
-                (false, false)
+            VolumeOperation::CreateSnapshot(_) => {
+                if !self.thin && !self.as_thin() {
```
dsharma-dc: Is this check to be interpreted as `We are thick currently, and this is the first snapshot being created` ?

control-plane/stor-port/src/types/v0/store/volume.rs
```diff
@@ -172,10 +172,44 @@ pub struct VolumeSpec {
...
+                snapshot_as_thin: as_thin,
+            },
+            runtime: Default::default(),
+        }
+    }
+    /// Insert snapshot in the list.
+    fn insert_snapshot(&mut self, snapshot: SnapshotId) {
+        self.runtime.snapshots.insert(snapshot);
+        // we become thin provisioned!
+        self.persisted.snapshot_as_thin = true;
```
dsharma-dc: I think we'll be doing this for each snapshot creation but effectively only first snapshot is doing this? Shall we make it more intuitive like:
```
if !self.persisted.snapshot_as_thin
    self.persisted.snapshot_as_thin = true;
```

control-plane/stor-port/src/types/v0/store/volume.rs
tiagolobocastro: Yes correct, do you think self.thin && !self.has_snapshots() would be clearer?

control-plane/stor-port/src/types/v0/store/volume.rs
tiagolobocastro: hmm I don't see that as more intuitive, it's less to me before it's redundant.

Abhinandan-Purkait (approved):
Do we need plugin changes?

tiagolobocastro:
bors merge

control-plane/stor-port/src/types/v0/store/volume.rs
tiagolobocastro: Used has_snapshots check.

tiagolobocastro:
bors cancel

tiagolobocastro:
bors merge

PR #615, opened 2023-07-04
test(nvme/connect): wait till the device is ready

tiagolobocastro:
We had a hardcoded sleep of 1s, which sometimes is not enough. Instead use retrying decorator to wait up to 2s for the device to be ready.


Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

PR #621, opened 2023-07-05
fix(cordon-drain): return correct updated spec

tiagolobocastro:
Outdated spec was being returned as part of the cordon/drain operations.


chriswldenyer (approved):
LGTM

tiagolobocastro:
bors merge

PR #625, opened 2023-07-06
feat(snapshot-clone): add rest api endpoint for create volume from snapshot

Abhinandan-Purkait:


control-plane/rest/openapi-specs/v0_api_spec.yaml
```diff
@@ -2721,6 +2763,8 @@ components:
           type: array
           items:
             $ref: '#/components/schemas/HostNqn'
+        kind:
+          $ref: '#/components/schemas/ReplicaKind'
```
tiagolobocastro, line 2769: should this be optional, or add a default?

tiagolobocastro (approved):

control-plane/rest/openapi-specs/v0_api_spec.yaml
Abhinandan-Purkait: Do we need a default, as the grpc layer has this as non-optional? Is it because now we add `serde(default)`?

The kind in ReplicaSpec is optional which I will add once your PR is merged.

Abhinandan-Purkait: Added the `Regular` as default

Abhinandan-Purkait:
bors merge

PR #634, opened 2023-07-12
fix(pool_controller): mark cr state as created when pool spec already present

abhilashshetty04:


abhilashshetty04:
bors try

control-plane/agents/src/bin/core/pool/pool_operations.rs
```diff
@@ -36,6 +39,15 @@ impl ResourceLifecycle for OperationGuardArc<PoolSpec> {
             });
         }
 
+        if let Ok(pool) = registry.specs().pool(&request.id) {
```
tiagolobocastro, line 42: this validation is already done in start_create?

tiagolobocastro: Ah because if the node is not online we return that error instead of already exists..

abhilashshetty04: yes @tiagolobocastro

abhilashshetty04: Its idempotent for Online Pool during operator update because of that.

tiagolobocastro (approved):

abhilashshetty04:
bors merge

PR #639, opened 2023-07-14

tiagolobocastro:
A user hit a very weird situation where there were 2 created nexuses containing the same replica. How can this happen? A potential situation is fixed where we now collect volume state AFTER we get the volume guard, though it's a very tight race so I suspect something else might still be at play here..

Irregardless of how this can happen we now plug the hole by ensuring we always check wit the volume replica removal logic before attempting to disown and destroy a replica.

### DIFF
--- a/.github/workflows/pr-submodule-branch.yml
+++ b/.github/workflows/pr-submodule-branch.yml
@@ -1,0 +1,16 @@
+name: Submodule Branch Check
+on:
+  pull_request:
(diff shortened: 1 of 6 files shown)


tiagolobocastro:
bors try

Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

PR #641, opened 2023-07-17

tiagolobocastro:
When removing a nexus child, don't attempt to destroy the equivalent replica, let the other reconcilers do this.
Adds a test for this using a fake nexus volume.


tiagolobocastro:
bors try

Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

PR #642, opened 2023-07-17
Cherry-pick #641

tiagolobocastro:

When removing a nexus child, don't attempt to destroy the equivalent replica, let the other reconcilers do this.
Adds a test for this using a fake nexus volume.


Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

PR #645, opened 2023-07-20
fix(snapshot/timeout): increase nexus and replica snapshot timeout

tiagolobocastro:
Increase timeouts for nexus and replica snapshots as it seems sometimes the subsystem pause can take a few seconds. On a particular test it took 11s to pause.
If we increase the IO timeout this might take even longer? So we probably need a way to undo snapshot creation because when we timeout the snapshot might not be created yet so the control-plane can't undo it!


Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

PR #650, opened 2023-07-26
feat(clone): enable option for fs id change to clone uuid

Abhinandan-Purkait:
- `cloneFsIdAsVolumeId` -- Parameter to enable changing the filesystem uuid to clone uuid to tackle duplicate id issue on xfs.
- If Parameter is not enabled `nouuid` to be used as a default behaviour.

### DIFF
--- a/control-plane/csi-driver/src/bin/controller/controller.rs
+++ b/control-plane/csi-driver/src/bin/controller/controller.rs
@@ -91,6 +91,7 @@ impl From<ApiClientError> for Status {
             ApiClientError::Conflict(reason) => Status::aborted(reason),
             ApiClientError::Aborted(reason) => Status::aborted(reason),
             ApiClientError::Unavailable(reason) => Status::unavailable(reason),
+            ApiClientError::InvalidArgument(reason) => Status::invalid_argument(reason),
             // TODO: Revisit the error mapping. Currently handled specifically for snapshot create.
             // ApiClientError::PreconditionFailed(reason) => Status::resource_exhausted(reason),
             // ApiClientError::ResourceExhausted(reason) => Status::resource_exhausted(reason),
@@ -257,6 +258,8 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
             None => false,
         };
 
+        let mut volume_context = args.parameters.clone();
+
         // First check if the volume already exists.
         match IoEngineApiClient::get_client()
             .get_volume_for_create(&parsed_vol_uuid)
@@ -281,7 +284,7 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
 
                 let sts_affinity_group_name = context.sts_affinity_group();
 
-                match volume_content_source {
+                let volume = match volume_content_source {
                     Some(snapshot_uuid) => {
                         IoEngineApiClient::get_client()
                             .create_snapshot_volume(
@@ -293,7 +296,7 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
                                 thin,
                                 sts_affinity_group_name.clone().map(AffinityGroup::new),
                             )
-                            .await?;
+                            .await?
                     }
                     None => {
                         IoEngineApiClient::get_client()
@@ -305,8 +308,16 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
                                 thin,
                                 sts_affinity_group_name.clone().map(AffinityGroup::new),
                             )
-                            .await?;
+                            .await?
                     }
+                };
+
+                // Append the 'fsId' : 'volume id' to the context if change was requested for the
+                // clone.
+                if volume.spec.content_source.is_some()
+                    && context.clone_fs_id_as_volume_id().unwrap_or(false)
+                {
+                    volume_context.insert("fsId".to_string(), volume_uuid.clone());
                 }
 
                 if let Some(ag_name) = sts_affinity_group_name {
@@ -325,7 +336,7 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
         let volume = rpc::csi::Volume {
             capacity_bytes: size as i64,
             volume_id: volume_uuid,
-            volume_context: args.parameters.clone(),
+            volume_context,
             content_source: args.volume_content_source,
             accessible_topology: vt_mapper.volume_accessible_topology(),
         };
@@ -387,9 +398,7 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
         let _guard = csi_driver::limiter::VolumeOpGuard::new(volume_id)?;
 
         match args.volume_capability {
-            Some(c) => {
-                check_volume_capabilities(&[c])?;
-            }
+            Some(c) => check_volume_capabilities(&[c])?,
             None => {
                 return Err(Status::invalid_argument("Missing volume capability"));
             }
--- a/control-plane/csi-driver/src/bin/node/filesystem_ops.rs
+++ b/control-plane/csi-driver/src/bin/node/filesystem_ops.rs
@@ -1,0 +1,311 @@
+//! This module consists of the various filesystem specific operations utility. Including creation
+//! of filesystem, changing the parameter of filesystem like uuid, default mount options for
+//! specific filesystem, repairing of the filesystem, retrieving specific properties of the
+//! filesystem.
+
+use crate::mount;
+use csi_driver::filesystem::FileSystem as Fs;
+use devinfo::{blkid::probe::Probe, DevInfoError};
+use std::process::Output;
+use tokio::process::Command;
+use tonic::async_trait;
+use tracing::{debug, trace};
+use uuid::Uuid;
+
+/// `nouuid` mount flag, to allow duplicate fs uuid.
+const XFS_NO_UUID_FLAG: &str = "nouuid";
+
+/// Error type filesystem operations.
+type Error = String;
+
+/// Ext4 filesystem type.
+pub(crate) struct Ext4Fs;
+/// XFS filesystem type.
+pub(crate) struct XFs;
+
+/// Filesystem type for csi node ops, wrapper over the parent Filesystem enum.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub(crate) struct FileSystem(Fs);
+
+impl From<Fs> for FileSystem {
+    fn from(value: Fs) -> Self {
+        Self(value)
+    }
+}
+
+impl AsRef<str> for FileSystem {
+    fn as_ref(&self) -> &str {
+        self.0.as_ref()
+    }
+}
+
+impl std::fmt::Display for FileSystem {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.0.as_ref())
+    }
+}
+
+impl FileSystem {
+    /// Convert the top level enumeration to specific filesystem types to access the supported
+    /// operations by them.
+    pub(crate) fn fs_ops(&self) -> Result<&dyn FileSystemOps, Error> {
+        static EXT4FS: Ext4Fs = Ext4Fs {};
+        static XFS: XFs = XFs {};
+        match self.0 {
+            Fs::Ext4 => Ok(&EXT4FS),
+            Fs::Xfs => Ok(&XFS),
+            _ => Err(format!("Unsupported filesystem {self}")),
+        }
+    }
+    /// Get a specific filesystem property by specifying the device path.
+    pub(crate) fn property(device: &str, property: &str) -> Result<String, DevInfoError> {
+        let probe = Probe::new_from_filename(device)?;
+        probe.do_probe()?;
+        probe.lookup_value(property)
+    }
+}
+
+#[async_trait]
+pub(crate) trait FileSystemOps: Send + Sync {
+    /// Create the filesystem using its fs util.
+    async fn create(&self, device: &str) -> Result<(), Error>;
+    /// Get the default mount options along with the user passed options for specific filesystems.
+    fn mount_flags(&self, mount_flags: Vec<String>) -> Vec<String>;
+    /// Unmount the filesystem if the filesystem uuid and the provided uuid differ.
+    fn unmount_on_fs_id_diff(
+        &self,
+        device_path: &str,
+        fs_staging_path: &str,
+        volume_uuid: &Uuid,
+    ) -> Result<(), Error>;
+    /// Repair the filesystem with specific filesystem utility.
+    async fn repair(
+        &self,
+        device: &str,
+        staging_path: &str,
+        options: &[String],
+        volume_uuid: &Uuid,
+    ) -> Result<(), Error>;
+    /// Set the filesystem uuid.
+    async fn set_uuid(&self, device: &str, volume_uuid: &Uuid) -> Result<(), Error>;
+    /// Set the filesystem uuid with repair if needed.
+    async fn set_uuid_with_repair(
+        &self,
+        device: &str,
+        staging_path: &str,
+        options: &[String],
+        volume_uuid: &Uuid,
+    ) -> Result<(), Error> {
+        if self.set_uuid(device, volume_uuid).await.is_err() {
+            self.repair(device, staging_path, options, volume_uuid)
+                .await?;
+            self.set_uuid(device, volume_uuid).await?;
+        }
+        Ok(())
+    }
+}
+
+#[async_trait]
+impl FileSystemOps for Ext4Fs {
+    async fn create(&self, device: &str) -> Result<(), Error> {
+        let binary = format!("mkfs.{}", "ext4");
+        let output = Command::new(&binary)
+            .arg(device)
+            .output()
+            .await
+            .map_err(|error| format!("failed to execute {binary}: {error}"))?;
+        ack_command_output(output, binary)
+    }
+
+    fn mount_flags(&self, mount_flags: Vec<String>) -> Vec<String> {
+        mount_flags
+    }
+
+    fn unmount_on_fs_id_diff(
+        &self,
+        _device_path: &str,
+        _fs_staging_path: &str,
+        _volume_uuid: &Uuid,
+    ) -> Result<(), Error> {
+        Ok(())
+    }
+
+    async fn repair(
+        &self,
+        device: &str,
+        _staging_path: &str,
+        _options: &[String],
+        _volume_uuid: &Uuid,
+    ) -> Result<(), Error> {
+        let binary = "e2fsck".to_string();
+        let output = Command::new(&binary)
+            .arg("-y")
+            .arg("-f")
+            .arg(device)
+            .output()
+            .await
+            .map_err(|error| format!("failed to execute {binary}: {error}"))?;
+
+        trace!(
+            "Output from {} command: {}, status code: {:?}",
+            binary,
+            String::from_utf8(output.stdout.clone()).unwrap(),
+            output.status.code()
+        );
+
+        // For ext4 fs repair using e2fsck the success condition is when status code is 0 or 1.
+        if !output.status.success() && output.status.code() != Some(1) {
+            return Err(format!(
+                "{} command failed: {}",
+                binary,
+                String::from_utf8(output.stderr).unwrap()
+            ));
+        }
+        Ok(())
+    }
+
+    async fn set_uuid(&self, device: &str, volume_uuid: &Uuid) -> Result<(), Error> {
+        if let Ok(probed_uuid) = FileSystem::property(device, "UUID") {
+            if probed_uuid == volume_uuid.to_string() {
+                return Ok(());
+            }
+        }
+
+        let binary = "tune2fs".to_string();
+        let output = Command::new(&binary)
+            .arg("-U")
+            .arg(volume_uuid.to_string())
+            .arg(device)
+            .output()
+            .await
+            .map_err(|error| format!("failed to execute {binary}: {error}"))?;
+
+        if !output.status.success() {
+            return Err(format!(
+                "{} command failed: {}",
+                binary,
+                String::from_utf8(output.stderr).unwrap()
+            ));
+        }
+
+        let probe_uuid = FileSystem::property(device, "UUID")
+            .map_err(|error| format!("Failed to get UUID of device {device}: {error}"))?;
+
+        if volume_uuid.to_string() != probe_uuid {
+            return Err(format!(
+                "failed to set filesystem uuid using : {binary}, {}",
+                String::from_utf8(output.stdout).unwrap_or_default()
+            ));
+        }
+        debug!("Changed filesystem uuid to {volume_uuid} for {device}");
+        Ok(())
+    }
+}
+
+#[async_trait]
+impl FileSystemOps for XFs {
+    async fn create(&self, device: &str) -> Result<(), Error> {
+        let binary = format!("mkfs.{}", "xfs");
+        let output = Command::new(&binary)
+            .arg(device)
+            .output()
+            .await
+            .map_err(|error| format!("failed to execute {binary}: {error}"))?;
+        ack_command_output(output, binary)
+    }
+
+    fn mount_flags(&self, mount_flags: Vec<String>) -> Vec<String> {
+        let mut mount_flags = mount_flags;
+        if !mount_flags.contains(&XFS_NO_UUID_FLAG.to_string()) {
+            mount_flags.push(XFS_NO_UUID_FLAG.to_string())
+        }
+        mount_flags
+    }
+
+    fn unmount_on_fs_id_diff(
+        &self,
+        device_path: &str,
+        fs_staging_path: &str,
+        volume_uuid: &Uuid,
+    ) -> Result<(), Error> {
+        mount::unmount_on_fs_id_diff(device_path, fs_staging_path, volume_uuid)
+    }
+
+    /// Xfs filesystem needs an unmount to clear the log, so that the parameters can be changed.
+    /// Mount the filesystem to a defined path and then unmount it.
+    async fn repair(
+        &self,
+        device: &str,
+        staging_path: &str,
+        options: &[String],
+        volume_uuid: &Uuid,
+    ) -> Result<(), Error> {
+        mount::filesystem_mount(device, staging_path, &FileSystem(Fs::Xfs), options).map_err(|error| {
+            format!(
+                "(xfs repairing) Failed to mount device {device} onto {staging_path} for {volume_uuid} : {error}",
+            )
+        })?;
+        mount::filesystem_unmount(staging_path).map_err(|error| {
+            format!(
+                "(xfs repairing) Failed to unmount device {device} from {staging_path} for {volume_uuid} : {error}",
+            )
+        })
+    }
+
+    async fn set_uuid(&self, device: &str, volume_uuid: &Uuid) -> Result<(), Error> {
+        if let Ok(probed_uuid) = FileSystem::property(device, "UUID") {
+            if probed_uuid == volume_uuid.to_string() {
+                return Ok(());
+            }
+        }
+
+        let binary = "xfs_admin".to_string();
+        let output = Command::new(&binary)
+            .arg("-U")
+            .arg(volume_uuid.to_string())
+            .arg(device)
+            .output()
+            .await
+            .map_err(|error| format!("failed to execute {binary}: {error}"))?;
+
+        if !output.status.success() {
+            return Err(format!(
+                "{} command failed: {}",
+                binary,
+                String::from_utf8(output.stderr).unwrap()
+            ));
+        }
+
+        let probe_uuid = FileSystem::property(device, "UUID")
+            .map_err(|error| format!("Failed to get UUID of device {device}: {error}"))?;
+
+        if volume_uuid.to_string() != probe_uuid {
+            return Err(format!(
+                "failed to set filesystem uuid using : {binary}, {}",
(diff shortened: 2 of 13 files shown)


control-plane/csi-driver/src/bin/controller/controller.rs
```diff
@@ -402,6 +400,9 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
 
         let params = PublishParams::try_from(&args.volume_context)?;
 
+        // Extract the change fs id from the params if present.
+        let change_fs_id = *params.change_fs_id();
```
tiagolobocastro: hmm should we process this at all for the moment?

control-plane/csi-driver/src/bin/node/filesystem_vol.rs
```diff
@@ -20,11 +20,15 @@ macro_rules! failure {
...
 
+/// `nouuid` mount flag, to allow duplicate fs uuid.
+const NO_UUID_FLAG: &str = "nouuid";
+
 pub(crate) async fn stage_fs_volume(
     msg: &NodeStageVolumeRequest,
     device_path: &str,
     mnt: &MountVolume,
     filesystems: &[FileSystem],
+    clone_uuid: Option<&String>,
```
tiagolobocastro: Should this be a uuid rather than String?

control-plane/csi-driver/src/bin/node/format.rs
```diff
@@ -2,32 +2,89 @@
...
+    /// Change the filesystem uuid using its fs util.
+    fn change_fs_id(&self, device: &str, uuid: &str) -> Result<(), String>;
+}
 
-    if let Err(error) = probe.do_probe() {
-        return Err(format!("probe failed: {error}"));
+impl FileSystemOps for FileSystem {
+    fn create(&self, device: &str) -> Result<(), String> {
+        match self {
+            FileSystem::Ext4 | FileSystem::Xfs => {
```
tiagolobocastro: Shouldn't we have specific types for all filesystem, and in turn they implement FileSystemOps?

control-plane/csi-driver/src/context.rs
```diff
@@ -70,6 +70,8 @@ pub enum Parameters {
     PvcNamespace,
     #[strum(serialize = "stsAffinityGroup")]
     StsAffinityGroup,
+    #[strum(serialize = "changefsid")]
```
```suggestion
    #[strum(serialize = "changeFsId")]
```

tiagolobocastro: I wonder if this is clear enough?
Shouldn't it be more like `changeFsIdToPvc` or something like that?

control-plane/csi-driver/src/bin/controller/controller.rs
Abhinandan-Purkait: Shall I return an error here if this is Not None?

Abhinandan-Purkait: Or just make it false even if its true for now?

control-plane/csi-driver/src/bin/node/filesystem_vol.rs
Abhinandan-Purkait: Yeah could be, but since comes from the publish context, we would need a round of conversion to uuid. That's why kept as string.

control-plane/csi-driver/src/bin/node/format.rs
Abhinandan-Purkait: We can, but again they would not have any fields. So an enum seemed straight forward.

control-plane/csi-driver/src/context.rs
Abhinandan-Purkait: `changeCloneFsId` rather?

control-plane/csi-driver/src/bin/node/filesystem_vol.rs
tiagolobocastro: At least should validate that it's a uuid before calling into fs utility

control-plane/csi-driver/src/bin/node/format.rs
tiagolobocastro: Yes they won't have any fields but it decouples all this code so it won't have to be glued together.

control-plane/csi-driver/src/bin/node/filesystem_vol.rs
Abhinandan-Purkait: Okay, sure. Would make that change.

control-plane/csi-driver/src/context.rs
tiagolobocastro: Ah we only do it for clones right, but stll need to clarify that it changes it to the Pvc uuid

control-plane/csi-driver/src/bin/node/filesystem_vol.rs
```diff
@@ -26,7 +29,21 @@ pub(crate) async fn stage_fs_volume(
     mnt: &MountVolume,
     filesystems: &[FileSystem],
 ) -> Result<(), Status> {
-    let volume_id = &msg.volume_id;
+    let volume_uuid = Uuid::parse_str(&msg.volume_id).map_err(|error| {
```
avishnu, line 32: Since we are changing the variable name to volume_uuid, should this be applicable in other functions as well? Eg. In unpublish_fs_volume

control-plane/csi-driver/src/bin/controller/controller.rs
```diff
@@ -1,6 +1,6 @@
 use crate::{ApiClientError, CreateVolumeTopology, CsiControllerConfig, IoEngineApiClient};
 
-use csi_driver::context::{CreateParams, PublishParams};
+use csi_driver::context::{CreateParams, PublishParams, IS_CLONE};
```
avishnu: Do we need the IS_CLONE? Can this not be deciphered from the CreateParams?

control-plane/csi-driver/src/bin/node/filesystem_vol.rs
Abhinandan-Purkait: In the other methods, the conversion to `uuid` type is not needed so have not changed them there. In this method it was needed so, have removed the other one and used this.

control-plane/csi-driver/src/bin/controller/controller.rs
Abhinandan-Purkait: No, CreateParams are the ones that get passed from StorageClass. There is no storageclass parameter that tells us the volume would be a clone. So have added this to publish context in the publish flow.

avishnu (approved):
LGTM.
If we are going to change volume clone to volume restore, do we need a change in the code like variable names, function names, file names, etc?
If yes, we can still do those changes in a new PR.

control-plane/csi-driver/src/bin/controller/controller.rs
avishnu: Instead of IS_CLONE, should we consider passing the VolumeSource here? 
- If the VolumeSource is empty, it's a regular volume.
- If the VolumeSource is a snapshot, it's a volume restore.
- If the VolumeSource is another volume, it's a volume clone.

Abhinandan-Purkait:
> LGTM. If we are going to change volume clone to volume restore, do we need a change in the code like variable names, function names, file names, etc? If yes, we can still do those changes in a new PR.

Yes, I will change them to the `restore`.

control-plane/csi-driver/src/bin/controller/controller.rs
Abhinandan-Purkait: Yes, we can do that as well.

control-plane/csi-driver/src/bin/node/filesystem_vol.rs
```diff
@@ -26,7 +29,21 @@ pub(crate) async fn stage_fs_volume(
...
+        failure!(
+            Code::Internal,
```
tiagolobocastro: Should be InvalidArgument ?

control-plane/csi-driver/src/bin/controller/controller.rs
```diff
@@ -484,6 +482,11 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
 
         publish_context.insert("uri".to_string(), uri);
 
+        // Append `IS_CLONE` to the the publish context if the volume has a content source.
+        if volume.spec.content_source.is_some() {
+            publish_context.insert(IS_CLONE.to_string(), "true".to_string());
```
tiagolobocastro: I think here we can put content_source then

control-plane/csi-driver/src/bin/controller/controller.rs
tiagolobocastro: Another way around this: make change fs uuid common, and not just for restores or clones.

control-plane/csi-driver/src/bin/node/filesystem_vol.rs
```diff
@@ -70,23 +87,46 @@ pub(crate) async fn stage_fs_volume(
...
+        if is_clone && change_fs_id {
+            continue_stage = continue_after_unmount_on_fs_id_diff(
+                fstype,
+                device_path,
+                fs_staging_path,
+                &volume_uuid,
+            )
+            .map_err(|error| {
+                failure!(
+                    Code::Aborted,
```
tiagolobocastro: What type is error here?

control-plane/csi-driver/src/bin/node/format.rs
```diff
@@ -1,46 +1,66 @@
...
 
-    if let Err(error) = probe.do_probe() {
-        return Err(format!("probe failed: {error}"));
-    }
-
-    if let Ok(fs) = probe.lookup_value("TYPE") {
+    if let Ok(fs) = fs {
         debug!("Found existing filesystem ({}) on device {}", fs, device);
+        if change_fs_id {
+            match fstype {
```
tiagolobocastro: Why not simply use the trait here?

control-plane/csi-driver/src/bin/node/format.rs
```diff
@@ -1,46 +1,66 @@
...
-    }
-
-    Err(format!(
-        "{} command failed: {}",
-        binary,
-        String::from_utf8(output.stderr).unwrap()
-    ))
+/// Xfs filesystem needs an unmount to clear the log, so that the parameters can be changed.
+/// Mount the filesystem to a defined path and then unmount it.
+fn repair_xfs_filesystem(
```
tiagolobocastro: Shouldn't this be in the trait?

control-plane/csi-driver/src/context.rs
```diff
@@ -1,43 +1,22 @@
...
+pub const IS_CLONE: &str = "is_clone";
 
-// Implement Display for the filesystem
-impl std::fmt::Display for FileSystem {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "{}", self.as_ref())
-    }
-}
+/// `nouuid` mount flag, to allow duplicate fs uuid.
+pub const NO_UUID_FLAG: &str = "nouuid";
```
tiagolobocastro: this should be in the filesystem rather than here

control-plane/csi-driver/src/context.rs
```diff
@@ -104,15 +85,23 @@ impl Parameters {
...
 
 /// Volume publish parameters.
 #[allow(dead_code)]
+#[derive(Debug)]
 pub struct PublishParams {
     io_timeout: Option<u32>,
     ctrl_loss_tmo: Option<u32>,
     keep_alive_tmo: Option<u32>,
     fs_type: Option<FileSystem>,
+    change_fs_id: Option<bool>,
```
tiagolobocastro: Why not just do this, and let the csi-controller do the checks for clone/restore?
Then csi-node just knows if it has to ensure fs needs to have this id or not, and doesn't need to know about clones/restores.

```suggestion
    fs_id: Option<String>,
```

control-plane/csi-driver/src/filesystem.rs
```diff
@@ -0,0 +1,217 @@
...
+
+/// Ext4 filesystem type.
+struct Ext4Fs;
+/// XFS filesystem type.
+struct XFs;
+
+impl FileSystem {
+    /// Convert the top level enumeration to specific filesystem types to access the supported
+    /// operations by them.
+    pub fn to_fs_ops(&self) -> Result<&dyn FileSystemOps, Error> {
```
tiagolobocastro:
```suggestion
    pub fn fs_ops(&self) -> Result<&dyn FileSystemOps, Error> {
```
You can also not return an error if you implement traits which return unsupported for all operators for the other filesystems.

control-plane/csi-driver/src/filesystem.rs
```diff
@@ -0,0 +1,217 @@
...
+    }
+    /// Get a specific filesystem property by specifying the device path.
+    pub fn get_property(device: &str, property: &str) -> Result<String, DevInfoError> {
+        let probe = Probe::new_from_filename(device)?;
+        probe.do_probe()?;
+        probe.lookup_value(property)
+    }
+    /// Get the filesystem specific mount flags that should be used as an default along with used
+    /// defined flags.
+    pub fn default_mount_flags(&self, mount_flags: Vec<String>) -> Result<Vec<String>, Error> {
```
tiagolobocastro: Do we need this? Should we let calling code create the ops and then just call default_mount_flags?

control-plane/csi-driver/src/bin/node/filesystem_vol.rs
Abhinandan-Purkait: String

control-plane/csi-driver/src/bin/node/format.rs
Abhinandan-Purkait: I was not able to put the `xfs_repair` in the trait because the mount module is not accessible in the FileSystem lib.

control-plane/csi-driver/src/bin/node/format.rs
Abhinandan-Purkait: I was not able to put the xfs_repair in the trait because the mount module is not accessible in the FileSystem lib.

control-plane/csi-driver/src/context.rs
Abhinandan-Purkait: This sounds good. Thanks.

control-plane/csi-driver/src/bin/controller/controller.rs
Abhinandan-Purkait: Have removed it as discussed.

control-plane/csi-driver/src/bin/node/filesystem_ops.rs
```diff
@@ -0,0 +1,276 @@
...
+use csi_driver::filesystem::FileSystem as Fs;
+use devinfo::{blkid::probe::Probe, DevInfoError};
+use std::process::Output;
+use tokio::process::Command;
+use tonic::async_trait;
+use tracing::trace;
+use uuid::Uuid;
+
+/// `nouuid` mount flag, to allow duplicate fs uuid.
+const NO_UUID_FLAG: &str = "nouuid";
```
tiagolobocastro: Prefix this with XFS ?

control-plane/csi-driver/src/bin/node/filesystem_ops.rs
```diff
@@ -0,0 +1,276 @@
...
+        static XFS: XFs = XFs {};
+        static UNSUPPORTED: Unsupported = Unsupported {};
+        match self.0 {
+            Fs::Ext4 => &EXT4FS,
+            Fs::Xfs => &XFS,
+            _ => &UNSUPPORTED,
+        }
+    }
+    /// Get a specific filesystem property by specifying the device path.
+    pub(crate) fn get_property(device: &str, property: &str) -> Result<String, DevInfoError> {
```
tiagolobocastro:
```suggestion
    pub(crate) fn property(device: &str, property: &str) -> Result<String, DevInfoError> {
```

control-plane/csi-driver/src/bin/node/filesystem_ops.rs
```diff
@@ -0,0 +1,276 @@
...
+        probe.lookup_value(property)
+    }
+}
+
+#[async_trait]
+pub(crate) trait FileSystemOps: Send + Sync {
+    /// Create the filesystem using its fs util.
+    async fn create(&self, device: &str) -> Result<(), Error>;
+    /// Get the default mount options along with the user passed options for specific filesystems.
+    fn default_mount_flags(&self, mount_flags: Vec<String>) -> Vec<String>;
```
tiagolobocastro:
```suggestion
    fn mount_flags(&self, user_flags: Vec<String>) -> Vec<String>;
```

control-plane/csi-driver/src/bin/node/filesystem_ops.rs
```diff
@@ -0,0 +1,276 @@
...
+                binary,
+                String::from_utf8(output.stderr).unwrap()
+            ));
+        }
+        Ok(())
+    }
+
+    /// Change the uuid of the ext4 filesystem after repair.
+    pub(crate) async fn set_uuid(device: &str, uuid: &str) -> Result<(), Error> {
+        Self::repair(device).await?;
```
tiagolobocastro: First we should check if the current uuid is already the same, and if so don't bother doing anything.

control-plane/csi-driver/src/bin/node/filesystem_ops.rs
```diff
@@ -0,0 +1,276 @@
...
+                "{} command failed: {}",
+                binary,
+                String::from_utf8(output.stderr).unwrap()
+            ));
+        }
+        Ok(())
+    }
+
+    /// Change the uuid of the ext4 filesystem after repair.
+    pub(crate) async fn set_uuid(device: &str, uuid: &str) -> Result<(), Error> {
```
tiagolobocastro: set_uuid should be part of the trait

control-plane/csi-driver/src/bin/node/filesystem_ops.rs
```diff
@@ -0,0 +1,276 @@
...
+    /// Convert the top level enumeration to specific filesystem types to access the supported
+    /// operations by them.
+    pub(crate) fn fs_ops(&self) -> &dyn FileSystemOps {
+        static EXT4FS: Ext4Fs = Ext4Fs {};
+        static XFS: XFs = XFs {};
+        static UNSUPPORTED: Unsupported = Unsupported {};
+        match self.0 {
+            Fs::Ext4 => &EXT4FS,
+            Fs::Xfs => &XFS,
+            _ => &UNSUPPORTED,
```
tiagolobocastro: hmm maybe I take it back, perhaps it was better to have a result here?

control-plane/csi-driver/src/bin/node/filesystem_ops.rs
```diff
@@ -0,0 +1,276 @@
...
+
+/// Ext4 filesystem type.
+pub(crate) struct Ext4Fs;
+/// XFS filesystem type.
+pub(crate) struct XFs;
+/// Unsupported filesystem type.
+struct Unsupported;
+
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub(crate) struct FileSystem(pub(crate) Fs);
```
tiagolobocastro: missing doc comment

make inner field private and add from_fs|from_str method instead?

control-plane/csi-driver/src/bin/node/format.rs
```diff
@@ -1,46 +1,42 @@
...
-    }
+    let fs_ops = fstype.fs_ops();
 
-    if let Ok(fs) = probe.lookup_value("TYPE") {
+    if let Ok(fs) = fs {
         debug!("Found existing filesystem ({}) on device {}", fs, device);
-        return Ok(());
-    }
+        if let Some(fs_id) = fs_id {
+            match fstype.0 {
```
tiagolobocastro: this should be part of the trait

control-plane/csi-driver/src/bin/node/mount.rs
```diff
@@ -351,3 +353,22 @@ async fn wait_file_removal(
...
     }
 }
+
+/// If the filesystem uuid doesn't match with the provided uuid, unmount the device.
+pub(crate) fn unmount_on_fs_id_diff(
+    device_path: &str,
+    fs_staging_path: &str,
+    volume_uuid: &Uuid,
+) -> Result<(), String> {
+    let probed_uuid = FileSystem::get_property(device_path, "UUID").map_err(|error| {
```
tiagolobocastro: should we catch this error and umount it?

control-plane/csi-driver/src/context.rs
```diff
@@ -104,15 +81,27 @@ impl Parameters {
...
         Self::parse_bool(value)
     }
+    /// Parse the value for `Self::CloneFsAsIdVolumeId`
+    pub fn clone_fs_id_as_volume_id(
+        value: Option<&String>,
+    ) -> Result<Option<bool>, ParseBoolError> {
+        Self::parse_bool(value)
+    }
+    /// Parse the value for `Self::FsId`
+    pub fn fs_id(value: Option<&String>) -> Option<String> {
```
tiagolobocastro: Should this be uuid::Uuid ?

control-plane/csi-driver/src/context.rs
```diff
@@ -104,15 +81,27 @@ impl Parameters {
...
 
 /// Volume publish parameters.
 #[allow(dead_code)]
+#[derive(Debug)]
 pub struct PublishParams {
     io_timeout: Option<u32>,
     ctrl_loss_tmo: Option<u32>,
     keep_alive_tmo: Option<u32>,
     fs_type: Option<FileSystem>,
+    fs_id: Option<String>,
```
tiagolobocastro: uuid::Uuid ?

control-plane/csi-driver/src/bin/node/format.rs
Abhinandan-Purkait: The parameters for both are different. That's why was not able to put in a trait. Now if we separate the repair from this, then repair again cannot be a part of trait.

control-plane/csi-driver/src/bin/node/mount.rs
Abhinandan-Purkait: You mean unmount on both error and fs id diff?

tiagolobocastro: Yes

control-plane/csi-driver/src/bin/node/filesystem_ops.rs
```diff
@@ -0,0 +1,300 @@
...
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub(crate) struct FileSystem(Fs);
+
+impl From<Fs> for FileSystem {
+    fn from(value: Fs) -> Self {
+        Self(value)
+    }
+}
+
+// Implement as ref for the FileSystem.
```
tiagolobocastro: These kind of comments above impls are redundant

control-plane/csi-driver/src/bin/node/filesystem_ops.rs
```diff
@@ -0,0 +1,300 @@
...
+    ) -> Result<(), Error>;
+    /// Repair the filesystem with specific filesystem utility.
+    async fn repair(
+        &self,
+        device: &str,
+        staging_path: &str,
+        options: &[String],
+        volume_uuid: &Uuid,
+    ) -> Result<(), Error>;
+    /// Set the filesystem uuid after repair.
```
tiagolobocastro:
```suggestion
    /// Set the filesystem uuid.
```

control-plane/csi-driver/src/bin/node/filesystem_ops.rs
```diff
@@ -0,0 +1,300 @@
...
+            return Err(format!(
+                "{} command failed: {}",
+                binary,
+                String::from_utf8(output.stderr).unwrap()
+            ));
+        }
+        Ok(())
+    }
+
+    /// Change the uuid of the ext4 filesystem after repair.
```
tiagolobocastro:
```suggestion
    /// Change the uuid of the ext4 filesystem.
```

control-plane/csi-driver/src/bin/node/filesystem_ops.rs
```diff
@@ -0,0 +1,300 @@
...
+        staging_path: &str,
+        options: &[String],
+        volume_uuid: &Uuid,
+    ) -> Result<(), Error> {
+        if let Ok(probed_uuid) = FileSystem::property(device, "UUID") {
+            if probed_uuid == volume_uuid.to_string() {
+                return Ok(());
+            }
+        }
+        self.repair(device, staging_path, options, volume_uuid)
```
tiagolobocastro: We should first try to change it, and only if it fails then repair and change.

control-plane/csi-driver/src/bin/node/filesystem_ops.rs
```diff
@@ -0,0 +1,300 @@
...
+            )
+        })?;
+        mount::filesystem_unmount(staging_path).map_err(|error| {
+            format!(
+                "(xfs repairing) Failed to unmount device {device} from {staging_path} for {volume_uuid} : {error}",
+            )
+        })
+    }
+
+    /// Change the uuid of the xfs filesystem after repair.
```
tiagolobocastro:
```suggestion
    /// Change the uuid of the xfs filesystem.
```

control-plane/csi-driver/src/bin/node/filesystem_ops.rs
```diff
@@ -0,0 +1,300 @@
...
+        staging_path: &str,
+        options: &[String],
+        volume_uuid: &Uuid,
+    ) -> Result<(), Error> {
+        if let Ok(probed_uuid) = FileSystem::property(device, "UUID") {
+            if probed_uuid == volume_uuid.to_string() {
+                return Ok(());
+            }
+        }
+        self.repair(device, staging_path, options, volume_uuid)
```
tiagolobocastro: same comment as above for ext4

control-plane/csi-driver/src/bin/node/filesystem_ops.rs
tiagolobocastro: Maybe we can separate set_uuid, in 2, set_uuid and set_uuid_with_repair? And implement this logic
set_uuid_with_repair:
   if set_uuid.ok return early
   repair
   set_uuid
set_uuid:
    if uuid.same return early
    set_uuid
    match uuid

tiagolobocastro (approved):

control-plane/csi-driver/src/bin/node/filesystem_ops.rs
```diff
@@ -0,0 +1,332 @@
...
+    ) -> Result<(), Error>;
+    /// Repair the filesystem with specific filesystem utility.
+    async fn repair(
+        &self,
+        device: &str,
+        staging_path: &str,
+        options: &[String],
+        volume_uuid: &Uuid,
+    ) -> Result<(), Error>;
+    /// Set the filesystem.
```
tiagolobocastro:
```suggestion
    /// Set the filesystem uuid.
```

control-plane/csi-driver/src/bin/node/filesystem_ops.rs
```diff
@@ -0,0 +1,332 @@
...
+        if volume_uuid.to_string() != probe_uuid {
+            return Err(format!(
+                "failed to set filesystem uuid using : {binary}, {}",
+                String::from_utf8(output.stdout).unwrap_or_default()
+            ));
+        }
+        Ok(())
+    }
+
+    async fn set_uuid_with_repair(
```
tiagolobocastro: Does this need specific or can we have a default impl?

control-plane/csi-driver/src/bin/node/format.rs
```diff
@@ -1,46 +1,33 @@
...
-    if let Err(error) = probe.do_probe() {
-        return Err(format!("probe failed: {error}"));
-    }
+    let fs_ops = fstype.fs_ops()?;
 
-    if let Ok(fs) = probe.lookup_value("TYPE") {
+    if let Ok(fs) = fs {
         debug!("Found existing filesystem ({}) on device {}", fs, device);
+        if let Some(fs_id) = fs_id {
+            debug!("Attempting filesystem repair and set uuid for {fs_id}, device: {device}");
```
tiagolobocastro:
```suggestion
            debug!("Attempting to set uuid for filesystem {fs_id}, device: {device}");
```

control-plane/csi-driver/src/bin/node/format.rs
```diff
@@ -1,46 +1,33 @@
...
+            fs_ops
+                .set_uuid_with_repair(device, staging_path, options, fs_id)
+                .await?;
+            debug!("Changed {fstype} filesystem uuid to {fs_id} for {device}");
```
tiagolobocastro: probably set_uuid should return bool so avoid adding this debug message when it's already set?

tiagolobocastro (approved):

control-plane/csi-driver/src/bin/node/filesystem_ops.rs
Abhinandan-Purkait: Default sounds good.

control-plane/csi-driver/src/bin/node/format.rs
Abhinandan-Purkait: I would rather move it inside the `set_uuid()`

tiagolobocastro: That sounds good too

Abhinandan-Purkait:
bors merge

PR #655, opened 2023-08-07
feat(snapshots): add garbage collector to cleanup leaked snapshots unknown to control-plane

Abhinandan-Purkait:


control-plane/agents/src/bin/core/controller/reconciler/node/snapshot.rs
```diff
@@ -0,0 +1,105 @@
...
+        .await
+    {
+        for snap in snaps {
+            if let Ok((snap_uuid, txn_id)) = extract_uuid_txn_id(snap.snap_name()) {
+                // Don't cleanup if the snapshot is known, i.e part of present snapshot's
+                // transaction.
+                if let Some(snap_rsc) = context.registry().specs().volume_snapshot_rsc(&snap_uuid) {
+                    let snapshot_spec = snap_rsc.lock().clone();
+                    if snapshot_spec.metadata().transactions().contains_key(txn_id) {
+                        continue;
```
tiagolobocastro: Rather than rsc, we should take the operation guard here and check if it's != txn_id

control-plane/agents/src/bin/core/controller/reconciler/node/snapshot.rs
```diff
@@ -0,0 +1,105 @@
...
+                // Attempt destroy of the snapshot if it qualifies as unknown.
+                if node_wrapper
+                    .destroy_repl_snapshot(&DestroyReplicaSnapshot::new(
+                        snap.snap_uuid().clone(),
+                        snap.pool_uuid().clone(),
+                    ))
+                    .await
+                    .is_ok()
+                {
+                    info!(
```
tiagolobocastro: Make it warn and add snap information as tracing values rather than just a string.

control-plane/agents/src/bin/core/controller/reconciler/node/snapshot.rs
```diff
@@ -0,0 +1,105 @@
...
+                }
+            }
+        }
+    }
+
+    Ok(PollerState::Idle)
+}
+
+/// Extracts the uuid and txn id from the snapshot name of format <snap_uuid>/<txn_id>.
+fn extract_uuid_txn_id(input: &str) -> Result<(SnapshotId, &str), SvcError> {
```
tiagolobocastro: Just make it an Option here since it's not an error if snapshots were created outside of control-plane.

control-plane/agents/src/bin/core/tests/volume/snapshot.rs
```diff
@@ -417,3 +417,151 @@ async fn snapshot_timeout() {
...
 }
+
+#[tokio::test]
+async fn unknown_snapshot_garbage_collector() {
+    let cluster = ClusterBuilder::builder()
+        .with_rest(false)
+        .with_agents(vec!["core"])
+        .with_io_engines(1)
+        .with_pools(1)
+        .with_cache_period("500ms")
```
tiagolobocastro: this would be "slow", make it very small ms, especially since we're running at 5x here.

control-plane/agents/src/bin/core/tests/volume/snapshot.rs
```diff
@@ -417,3 +417,151 @@ async fn snapshot_timeout() {
...
+        )
+        .await
+        .unwrap();
+    let snaps = rpc_handle
+        .list_replica_snaps(None, None)
+        .await
+        .expect("Should get replica snaps");
+    assert_eq!(snaps.snapshots.len(), 5);
+
+    cluster.composer().start("core").await.unwrap();
```
tiagolobocastro: Isn't there a restart core in the cluster? Which does the wait as well

control-plane/agents/src/bin/core/controller/reconciler/node/snapshot.rs
Abhinandan-Purkait: Okay.

control-plane/agents/src/bin/core/tests/volume/snapshot.rs
Abhinandan-Purkait: Indeed. maybe 50ms?

control-plane/agents/src/bin/core/tests/volume/snapshot.rs
Abhinandan-Purkait: restart_core seems to restart only and not wait.

control-plane/agents/src/bin/core/controller/reconciler/node/snapshot.rs
```diff
@@ -0,0 +1,99 @@
...
+/// Identify unknown snapshots and attempt to clean them up.
+#[tracing::instrument(skip(node_spec, context), level = "trace", fields(request.reconcile = true))]
+async fn delete_unknown_volume_snapshot_reconciler(
+    context: &PollContext,
+    node_spec: &mut OperationGuardArc<NodeSpec>,
+) -> PollResult {
+    let node_wrapper = match context.registry().node_wrapper(node_spec.uid()).await {
+        Ok(node_wrapper) => node_wrapper,
+        Err(error) => return PollResult::Err(error),
+    };
```
tiagolobocastro: Can't we do just this?

```suggestion
    let node_wrapper = context.registry().node_wrapper(node_spec.uid()).await?;
```

control-plane/agents/src/bin/core/tests/volume/snapshot.rs
tiagolobocastro: Ah, it removes the lock, we should use restart followed by the liveness then.
Maybe we can add a helper method that does both?

control-plane/agents/src/bin/core/controller/reconciler/node/snapshot.rs
```diff
@@ -0,0 +1,99 @@
...
+        .await
+    {
+        for snap in snaps {
+            if let Some((snap_uuid, txn_id)) = extract_uuid_txn_id(snap.snap_name()) {
+                // Don't cleanup if the snapshot is known, i.e part of present snapshot's
+                // transaction.
+                if let Ok(snap_guard) = context.registry().specs().volume_snapshot(&snap_uuid).await
+                {
+                    let snapshot_spec = snap_guard.lock().clone();
+                    if snapshot_spec.metadata().transactions().contains_key(txn_id) {
```
tiagolobocastro: Do we need to clone?

```suggestion
                    let snapshot_spec = snap_guard.as_ref();
                    if snapshot_spec.metadata().transactions().contains_key(txn_id) {
```

tiagolobocastro (approved):

control-plane/agents/src/bin/core/tests/volume/snapshot.rs
Abhinandan-Purkait: Done. Added the helper.

Abhinandan-Purkait:
bors merge

control-plane/agents/src/bin/core/controller/reconciler/node/snapshot.rs
```diff
@@ -0,0 +1,96 @@
...
+
+    if let Ok(snaps) = node_wrapper
+        .list_repl_snapshots(&ListReplicaSnapshots::All)
+        .await
+    {
+        for snap in snaps {
+            if let Some((snap_uuid, txn_id)) = extract_uuid_txn_id(snap.snap_name()) {
+                // Don't cleanup if the snapshot is known, i.e part of present snapshot's
+                // transaction.
+                if let Ok(snap_guard) = context.registry().specs().volume_snapshot(&snap_uuid).await
```
tiagolobocastro: There's a bug here, if the guard is taken we ignore and destroy, which is wrong

tiagolobocastro (changes requested):

control-plane/agents/src/bin/core/controller/reconciler/node/snapshot.rs
Abhinandan-Purkait: Yeah. Do we need the guard? I think we can use the rsc only.

tiagolobocastro: We *do* need the guard, but we can do one of two:
1. delete on not_found error, skip/continue on any other error
2. get rsc first, then try to upgrade to guard and skip/continue on error

Abhinandan-Purkait: Done.

Abhinandan-Purkait:
bors try

control-plane/agents/src/bin/core/controller/reconciler/node/snapshot.rs
```diff
@@ -0,0 +1,100 @@
...
+                };
+                if node_wrapper
+                    .destroy_repl_snapshot(&DestroyReplicaSnapshot::new(
+                        snap.snap_uuid().clone(),
+                        snap.pool_uuid().clone(),
+                    ))
+                    .await
+                    .is_ok()
+                {
+                    tracing::warn!(snapshot.uuid=%snap.snap_uuid(), snapshot.pool_uuid=%snap.pool_uuid(), "Deleted unidentified replica snapshot from pool");
```
tiagolobocastro:
```suggestion
                    tracing::warn!(snapshot.uuid=%snap.snap_uuid(), snapshot.pool_uuid=%snap.pool_uuid(), "Deleted unknown replica snapshot from pool");
```

tiagolobocastro (approved):

Abhinandan-Purkait:
bors merge

PR #662, opened 2023-08-17
Pin volume target to replicas for 1 replica volumes

tiagolobocastro:
fix(csi/controller): publish idempotency with 1r volumes
    
    With 1r volumes the target node may not be the same as app node. Moreover this could
    already happen if HA kicked in after publish (though probaboly not likely).
    Add frontend configuration to the volume target config, which allows us to check
    if a given node is allowed access. If it is then we can allow idempotent publish.
    
---

    fix(node/drain): delete shutdown targets on timeout
    
    For deployer clusters, if the drain operation times out, delete the shutdown targets.
    
---

    fix(volume/drain): during node drain don't move target away from replicas
    
    For 1-r volumes, don't move the target away from their replicas.
    
---

    fix: pin volume target to replicas (1-r volumes)
    
    Pin the nexus together with the replica for 1-r volumes.
    
---

    test(rust): allow specifying options and arguments
    
    Now we can run like so for example:
    ./scripts/rust/test.sh -- -Zunstable-options --report-time
    
### DIFF
--- a/control-plane/agents/src/bin/core/controller/reconciler/node/nexus.rs
+++ b/control-plane/agents/src/bin/core/controller/reconciler/node/nexus.rs
@@ -1,7 +1,7 @@
 use crate::controller::{
     reconciler::{PollContext, TaskPoller},
     resources::{
-        operations::{ResourceDrain, ResourcePublishing},
+        operations::{ResourceDrain, ResourcePublishing, ResourceShutdownOperations},
         operations_helper::OperationSequenceGuard,
         OperationGuard, OperationGuardArc, ResourceMutex,
     },
@@ -10,15 +10,20 @@ use crate::controller::{
 use agents::errors::SvcError;
 use itertools::Either;
 use std::{collections::HashSet, time::Duration};
-use stor_port::types::v0::{
-    store::{
-        node::{DrainingVolumes, NodeSpec},
-        volume::VolumeSpec,
+use stor_port::{
+    platform,
+    types::v0::{
+        store::{
+            node::{DrainingVolumes, NodeSpec},
+            volume::VolumeSpec,
+        },
+        transport::{
+            DestroyShutdownTargets, NodeId, RepublishVolume, VolumeId, VolumeShareProtocol,
+        },
     },
-    transport::{NodeId, RepublishVolume, VolumeId, VolumeShareProtocol},
 };
 
-const DRAINING_VOLUME_TIMEOUT_SECONDS: u64 = 120;
+const DRAINING_VOLUME_TIMEOUT: Duration = Duration::from_secs(120);
 
 /// Node drain reconciler.
 #[derive(Debug)]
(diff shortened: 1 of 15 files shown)


Abhinandan-Purkait (approved):

scripts/rust/test.sh
```diff
@@ -1,22 +1,42 @@
...
+ARGS=""
+OPTS=""
+DO_ARGS=
+while [ "$#" -gt 0 ]; do
+  case $1 in
+    --)
+      DO_ARGS="y"
+      shift;;
+    *)
+      if [ -n "$DO_ARGS" ]; then
```
chriswldenyer: IMO it's clearer as:
if [ "$DO_ARGS" == "y" ]; then

control-plane/csi-driver/src/bin/controller/controller.rs
```diff
@@ -467,7 +467,8 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
                     Ok(node) if node.state.as_ref().map(|n| n.status).unwrap_or(NodeStatus::Unknown) != NodeStatus::Online || cordon_check(node.spec.as_ref()) => {
                         Ok(None)
                     },
-                    Ok(_) => Ok(Some(node_id.as_str()))
+                    Ok(_) if volume.spec.num_replicas > 1 => Ok(Some(node_id.as_str())),
```
chriswldenyer: This could do with a comment explaining what's happening here.

control-plane/agents/src/bin/core/controller/reconciler/node/nexus.rs
```diff
@@ -266,3 +229,111 @@ async fn check_and_drain_node(
...
+    node_spec: &OperationGuardArc<NodeSpec>,
+    mut guarded_vol_spec: OperationGuardArc<VolumeSpec>,
+) -> Either<bool, VolumeId> {
+    let Some(target) = guarded_vol_spec.as_ref().target() else {
+        return Either::Left(false);
+    };
+
+    let drain_node_id = node_spec.as_ref().id();
+    let Some(target_node) =
+        drain_target_node(context, target.node(), drain_node_id, &guarded_vol_spec)
```
chriswldenyer: (grumble) IMO I'd prefer it to be called get_drain_target_node() because at first glance the function name makes you think it is doing the drain.
But I suppose we have to stick to the rustacean way...

chriswldenyer (approved):
Looks reasonable

control-plane/agents/src/bin/core/controller/reconciler/node/nexus.rs
tiagolobocastro: I agree, reading it now sounds confusing.
I'll rename to select_drain_target_node.

control-plane/csi-driver/src/bin/controller/controller.rs
tiagolobocastro: Done

scripts/rust/test.sh
tiagolobocastro: Done

tiagolobocastro:
bors merge

tiagolobocastro:
bors cancel

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

PR #668, opened 2023-08-25
fix(dsp): mark cr state as created when control plane is not responsive

abhilashshetty04:


k8s/operators/src/pool/diskpool/v1beta1.rs
```diff
@@ -107,6 +107,7 @@ impl DiskPoolStatus {
     /// Set when Pool is not found for some reason.
     pub fn not_found() -> Self {
         Self {
+            cr_state: CrPoolState::Created,
```
tiagolobocastro: What if the pool is not created yet?
Should we check what the current cr_state is and keep ip?

abhilashshetty04: We enter this part of the code if cr_state is Created. If pool was never created it has a different fn in reconciler match arm. I might have to handle Terminating cr state.

abhilashshetty04:
```match dsp.status {
        Some(DiskPoolStatus {
            cr_state: CrPoolState::Creating,
            ..
        }) => dsp.create_or_import().await,
        Some(DiskPoolStatus {
            cr_state: CrPoolState::Created,
            ..
        })
        | Some(DiskPoolStatus {
            cr_state: CrPoolState::Terminating,
            ..
        }) => dsp.pool_check().await,
        // We use this state to indicate its a new CRD however, we could (and
        // perhaps should) use the finalizer callback.
        None => dsp.init_cr().await,
    }```

abhilashshetty04: If pool was never created it will be in `CrPoolState::Creating` state. If no field at all for the new create event its `init_cr()`

abhilashshetty04: checking if deletion timestamp is set. If yes, Will mark it as Teminating

tiagolobocastro: Seems overcomplicated, not not simply pass self to not_found and have not found keep existing cr states and clear status?
EDIT: or even pass the state or something like that

abhilashshetty04: Done. We are passsing current state now.

Abhinandan-Purkait (approved):

abhilashshetty04:
bors merge

PR #669, opened 2023-08-28
fix: error for multi-replica volume from snapshot

Abhinandan-Purkait:
FIX: The error variant for multi replica snapshot clone/restore creation was the same as multi replica snapshot operations. Add a specific variant for the same.

### DIFF
--- a/control-plane/agents/src/common/errors.rs
+++ b/control-plane/agents/src/common/errors.rs
@@ -298,6 +298,10 @@ pub enum SvcError {
     },
     #[snafu(display("Snapshots for multi-replica volumes are not allowed"))]
     NReplSnapshotNotAllowed {},
+    #[snafu(display(
+        "Cannot create a multi-replica volume from a snapshot of a single-replica volume"
+    ))]
+    NReplSnapshotCloneCreationNotAllowed {},
     #[snafu(display("Replica's {} snapshot was unexpectedly skipped", replica))]
     ReplicaSnapSkipped { replica: String },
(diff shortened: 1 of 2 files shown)


control-plane/agents/src/common/errors.rs
```diff
@@ -298,6 +298,8 @@ pub enum SvcError {
     },
     #[snafu(display("Snapshots for multi-replica volumes are not allowed"))]
     NReplSnapshotNotAllowed {},
+    #[snafu(display("Multi-replica volume is not allowed from snapshot"))]
```
avishnu: Cannot create a multi-replica volume from a snapshot of a single-replica volume

Abhinandan-Purkait: Are you suggesting the belongs to be added at the end?

avishnu: > Are you suggesting the belongs to be added at the end?

Just suggesting a modified text.

Abhinandan-Purkait: Sounds good. Done. Thanks

avishnu: Just wondering if the message should be generic, something like "Not enough sources to create the volume replicas".
We can probably take this up later.

avishnu (approved):
Lgtm

control-plane/agents/src/common/errors.rs
Abhinandan-Purkait: Sure.

Abhinandan-Purkait:
bors merge

PR #681, opened 2023-10-12
fix(pool/topology): ensure code is as per design and docs

tiagolobocastro:
If a pool has label:
A: D

Then it should match a volume with:
A: ""


Abhinandan-Purkait (approved):
LGTM.

Abhinandan-Purkait:
@tiagolobocastro Question: This ensures that if the value for a key is not specified we would consider any value for that key to be true right? Sort of like, `anything that has the key`

tiagolobocastro:
> @tiagolobocastro Question: This ensures that if the value for a key is not specified we would consider any value for that key to be true right? Sort of like, `anything that has the key`

Yes correct, example, anything that has that key matches the criteria.

tiagolobocastro:
bors merge

PR #683, opened 2023-10-25
feat(eventing): add switch over and nvme path events

datacore-vvarakantham:
Added switch over and nvme path events.

### DIFF
--- a/control-plane/agents/src/bin/ha/cluster/switchover.rs
+++ b/control-plane/agents/src/bin/ha/cluster/switchover.rs
@@ -1,6 +1,10 @@
 use crate::{core_grpc, etcd::EtcdStore, nodes::NodeList};
+use agents::Event;
 use anyhow::anyhow;
 use chrono::Utc;
+use events_api::event::{
+    EventAction, EventCategory, EventMessage, EventMeta, EventSource, SwitchOverStatus,
+};
 use grpc::operations::{
     ha_node::{client::NodeAgentClient, traits::NodeAgentOperations},
     volume::traits::VolumeOperations,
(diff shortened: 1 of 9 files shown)


control-plane/agents/src/bin/ha/cluster/switchover.rs
```diff
@@ -647,6 +663,32 @@ impl From<Operation> for Stage {
...
+        let status = match self.stage() {
+            Stage::Init => SwitchOverStatus::SwitchOverStarted,
+            Stage::Successful => SwitchOverStatus::SwitchOverCompleted,
+            Stage::Errored => SwitchOverStatus::SwitchOverFailed,
+            _ => SwitchOverStatus::UnknownSwitchOverStatus,
+        };
+
+        let event_source = EventSource::new(node_id.to_string()).with_switch_over_data(
+            status,
+            self.timestamp,
```
tiagolobocastro, line 678: Is this when the path down message is received from agent-ha-node?

tiagolobocastro (approved):
Are we adding agent-ha-node events?

control-plane/agents/src/bin/ha/cluster/switchover.rs
datacore-vvarakantham: Yes it is.

No I was adding just the agent-ha-cluster events. What agent-ha-node events could be added?

tiagolobocastro: In Node agent there are things we could pick up on, path suspected, fail or "fixed". These would be very useful to have.

datacore-vvarakantham: I added nvme path suspect and fail events now.

tiagolobocastro: Missing one for when the path is "fixed" ?

datacore-vvarakantham: Can we get the fixed from node agent? Can we differentiate between the fixed ones and the new ones when a new path is added?

datacore-vvarakantham: Added NvmePathConnect event.

chriswldenyer (approved):
Looks reasonable

datacore-vvarakantham:
bors merge

PR #692, opened 2023-11-28
feat(drain): disallow node draining if ha is disabled

Abhinandan-Purkait:
Node Drain relies on the HA feature to move and reconnect the nexus. So if HA is disabled node drain will not work and can leave the volumes in an inconsistent state. 

This PR restricts node draining if the HA is not enabled.

### DIFF
--- a/control-plane/agents/src/bin/core/controller/registry.rs
+++ b/control-plane/agents/src/bin/core/controller/registry.rs
@@ -98,6 +98,8 @@ pub(crate) struct RegistryInner<S: Store> {
     legacy_prefix_present: bool,
     /// Thin provisioning parameters.
     thin_args: ThinArgs,
+    /// Check if the HA feature is enabled.
+    ha_disabled: bool,
 }
 
 impl Registry {
@@ -118,6 +120,7 @@ impl Registry {
         create_volume_limit: usize,
         host_acl: Vec<HostAccessControl>,
         thin_args: ThinArgs,
+        ha_enabled: bool,
     ) -> Result<Self, SvcError> {
         let store_endpoint = Self::format_store_endpoint(&store_url);
         tracing::info!("Connecting to persistent store at {}", store_endpoint);
@@ -167,6 +170,7 @@ impl Registry {
                 host_acl,
                 legacy_prefix_present,
                 thin_args,
+                ha_disabled: ha_enabled,
             }),
         };
         registry.init().await?;
@@ -191,6 +195,11 @@ impl Registry {
         }
 
         Ok(registry)
+    }
+
+    /// Check if the HA feature is enabled.
+    pub(crate) fn ha_disabled(&self) -> bool {
+        self.ha_disabled
     }
 
     /// Formats the store endpoint with a default port if one isn't supplied.
--- a/control-plane/agents/src/bin/core/main.rs
+++ b/control-plane/agents/src/bin/core/main.rs
@@ -103,6 +103,11 @@ pub(crate) struct CliArgs {
     /// Events message-bus endpoint url.
     #[clap(long, short)]
     events_url: Option<url::Url>,
+
+    /// Disable the HA/Failover feature.
+    /// This is useful when the frontend nodes do not support the NVMe ANA feature.
+    #[clap(long, env = "HA_DISABLED")]
+    pub(crate) disable_ha: bool,
 }
 impl CliArgs {
     fn args() -> Self {
@@ -172,6 +177,7 @@ async fn server(cli_args: CliArgs) -> anyhow::Result<()> {
             cli_args.hosts_acl.clone()
         },
         cli_args.thin_args,
+        cli_args.disable_ha,
     )
     .await?;
 
--- a/control-plane/agents/src/bin/core/node/service.rs
+++ b/control-plane/agents/src/bin/core/node/service.rs
@@ -361,6 +361,11 @@ impl Service {
 
     /// Apply a drain label to the specified node. The reconciler will perform the drain.
     async fn drain(&self, id: NodeId, label: String) -> Result<Node, SvcError> {
+        // Don't allow draining if HA_ENABLED is false. If it is undefined we treat it as true.
+        if self.registry.ha_disabled() {
+            return Err(SvcError::DrainNotAllowedWhenHAisDisabled {});
+        }
+
         let mut guarded_node = self.specs().guarded_node(&id).await?;
 
         let spec = guarded_node.drain(&self.registry, label.clone()).await?;
--- a/control-plane/agents/src/common/errors.rs
+++ b/control-plane/agents/src/common/errors.rs
@@ -330,6 +330,10 @@ pub enum SvcError {
     ClonedSnapshotVolumeRepl {},
     #[snafu(display("The source snapshot is not created"))]
     SnapshotNotCreated {},
+    #[snafu(display("Draining is not allowed without HA"))]
+    DrainNotAllowedWhenHAisDisabled {},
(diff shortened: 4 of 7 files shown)


control-plane/agents/src/bin/core/node/service.rs
```diff
@@ -361,6 +361,11 @@ impl Service {
 
     /// Apply a drain label to the specified node. The reconciler will perform the drain.
     async fn drain(&self, id: NodeId, label: String) -> Result<Node, SvcError> {
+        // Don't allow draining if HA_ENABLED is false or undefined.
```
avishnu: does undefined wrap to false? Shouldn't undefined be defaulted as HA is enabled?

Abhinandan-Purkait: It's upto us what we want undefined to unwrap to. I have made it false because if we let it be true then we have the same problem again i.e we might drain even if HA might be disabled.
For example for a case, someone removes the env variable from the deployment manually.

avishnu: Ok, does that mean we always have the helm variable for HA_ENABLED with default set to true? I was hinting if the variable HA_ENABLED itself was optional and default behavior was true...

Abhinandan-Purkait: Yes, we will always have `HA_ENABLED=true` on the core-agent deployment as a default helm installation. 

Now if someone does a `helm install .... --set agents.ha.enabled=false` then `HA_ENABLED=false` will be set on the core-agent deployment.

We will get a undefined only when someone goes and edits the  core-agent deployment and removes the `HA_ENABLED` manually.

tiagolobocastro: Should it be the other way around, be enabled by default and disallowed when agents.ha.enabled=false ?
Then we don't need to keep env var unless it's disabled

Abhinandan-Purkait: @tiagolobocastro Yes we can do that as well. What if its removed manually, should we allow the drain assuming its all fine?

avishnu: Treat missing ha_enabled field same as ha_enabled = true

Abhinandan-Purkait: Okay. Sure. I will make that change.

avishnu (approved):
Lgtm

control-plane/agents/src/bin/core/node/service.rs
```diff
@@ -361,6 +361,11 @@ impl Service {
 
     /// Apply a drain label to the specified node. The reconciler will perform the drain.
     async fn drain(&self, id: NodeId, label: String) -> Result<Node, SvcError> {
+        // Don't allow draining if HA_ENABLED is false. If it is undefined we treat it as true.
+        if std::env::var("HA_ENABLED").unwrap_or("true".to_string()) == "false" {
```
tiagolobocastro: This won't be used internally, only externally.
Let's add an option to the cli args, and pass them to the registry. 
We can make the cli option: `--disable-ha` (you can also add DISABLE_HA env variable as you have here)

Then we can make the check here for service calls and also within the node drain itself just in case it ever gets called some other way.

control-plane/agents/src/common/errors.rs
```diff
@@ -918,6 +920,12 @@ impl From<SvcError> for ReplyError {
                 source,
                 extra,
             },
+            SvcError::DrainNotAllowedWhenHAisDisabled {} => ReplyError {
+                kind: ReplyErrorKind::Unavailable,
```
tiagolobocastro: Is unavailable the right error kind here? Or should it be something more like forbidden or preconditions failed or invalid?

tiagolobocastro (commented):
I think we should also ensure that on republish we never move the nexus?
Otherwise if we don't want to do that then we should change this config from disable_ha to disable_drain ?

control-plane/agents/src/common/errors.rs
Abhinandan-Purkait: Ahh, failed precondition might be a better fit here. Forbidden is not present. I would go with failed precondition.

Abhinandan-Purkait: Done.

control-plane/agents/src/bin/core/node/service.rs
Abhinandan-Purkait: Done. Btw @tiagolobocastro I did not disable republish as core-agent ha tests use republish without HA components.

Abhinandan-Purkait: Although, since we decided the default value would be true, it would work. Should I disable republish what do you think?

control-plane/agents/src/bin/core/controller/registry.rs
```diff
@@ -193,6 +197,11 @@ impl Registry {
         Ok(registry)
     }
 
+    /// Flag to track whether HA is enabled or not.
```
```suggestion
    /// Check if the HA feature is enabled.
```

control-plane/agents/src/bin/core/main.rs
```diff
@@ -103,6 +103,10 @@ pub(crate) struct CliArgs {
     /// Events message-bus endpoint url.
     #[clap(long, short)]
     events_url: Option<url::Url>,
+
+    /// Flag to track whether HA is enabled or not.
+    #[clap(long, default_value = "true", env = "HA_ENABLED")]
+    pub(crate) ha_enabled: bool,
```
tiagolobocastro: It's bit confusing to have enabled bool by default IMHO, so suggest something like this:

```suggestion
    /// Disable the HA/Failover feature. 
    /// This is useful when the frontend nodes do not support the NVMe ANA feature.
    #[clap(long, env = "HA_DISABLED")]
    pub(crate) disable_ha: bool,
```

Abhinandan-Purkait: The problem is then we would need to negate and pass the value from helm template. I kept it so that we can pass the value as it is.

tiagolobocastro: What's the problem with negating the value in helm?

control-plane/agents/src/bin/core/node/service.rs
tiagolobocastro: > Done. Btw @tiagolobocastro I did not disable republish as core-agent ha tests use republish without HA components.

That's fine I think, because we wouldn't  disable ha on the core agent

tiagolobocastro: > Although, since we decided the default value would be true, it would work. Should I disable republish what do you think?

Yes exactly, I think we should disable it.

control-plane/agents/src/bin/core/main.rs
Abhinandan-Purkait: No there is no problem, it's just that helm follows the enabled keyword flow, so I did that way initially.

tiagolobocastro: Anyway, not a deal breaker to me, just a nit, we can keep it but please enhance doc comment.

Abhinandan-Purkait: No, if we look from the cli args perspective I think the disable makes more sense here. I will make the change.

control-plane/agents/src/bin/core/node/service.rs
Abhinandan-Purkait: Done.

tiagolobocastro (approved):

Abhinandan-Purkait:
bors merge

PR #694, opened 2023-12-04
fix(kubectl-plugin): non-zero exit code on error

datacore-vvarakantham:
Returning non zero exit code with an error from plugin with kubectl command.

### DIFF
--- a/control-plane/plugin/src/bin/rest-plugin/main.rs
+++ b/control-plane/plugin/src/bin/rest-plugin/main.rs
@@ -58,7 +58,7 @@ async fn execute(cli_args: CliArgs) {
     }
 
     // Perform the operations based on the subcommand, with proper output format.
-    match &cli_args.operations {
+    let result = match &cli_args.operations {
         Operations::Drain(resource) => match resource {
             DrainResources::Node(drain_node_args) => {
                 node::Node::drain(
@@ -132,5 +132,10 @@ async fn execute(cli_args: CliArgs) {
                 node::Node::uncordon(id, label, &cli_args.output).await
             }
         },
+    };
+
+    if let Err(error) = result {
+        eprintln!("{error}");
+        std::process::exit(1);
     };
 }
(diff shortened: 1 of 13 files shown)


Abhinandan-Purkait (approved):
Should it be rather a wrapper function which logs an error message and returns an exit code, rather than calling exit everywhere. In that way we can reuse the same method? @tiagolobocastro wdyt?

tiagolobocastro:
> Should it be rather a wrapper function which logs an error message and returns an exit code, rather than calling exit everywhere. In that way we can reuse the same method? @tiagolobocastro wdyt?

hmm yeah...
I think we can take one step further, and simply return an error up the call stack

control-plane/plugin/src/operations.rs
```diff
@@ -1,6 +1,11 @@
-use crate::resources::{utils, CordonResources, DrainResources, GetResources, ScaleResources};
+use crate::resources::{
+    error::Error, utils, CordonResources, DrainResources, GetResources, ScaleResources,
+};
 use async_trait::async_trait;
 
+/// Result wrapper for kubectl commands.
+pub type KubectlResult = Result<(), Error>;
```
tiagolobocastro: there's no Kubectl here
```suggestion
/// Result wrapper for plugin commands.
pub type PluginResult = Result<(), Error>;
```

control-plane/plugin/src/bin/rest-plugin/main.rs
```diff
@@ -133,4 +133,9 @@ async fn execute(cli_args: CliArgs) {
             }
         },
     };
+
+    let _ = result.map_err(|error| {
+        eprintln!("{error}");
```
tiagolobocastro, line 138: if output json/yaml then please change print accordingly

control-plane/plugin/src/bin/rest-plugin/main.rs
```diff
@@ -133,4 +133,9 @@ async fn execute(cli_args: CliArgs) {
             }
         },
     };
+
+    let _ = result.map_err(|error| {
```
tiagolobocastro:
```suggestion
    if let Err(error) = result {
```

tiagolobocastro (approved):

datacore-vvarakantham:
bors merge

datacore-vvarakantham:
bors merge

PR #699, opened 2023-12-11
refactor(nix/build): replace builtins.filterSource and builtins.path

tiagolobocastro:
Not only is filterSource deprecated but also path is compatible with nix-illegal characters in their names, like @.


niladrih (approved):

dsharma-dc (approved):

tiagolobocastro:
bors merge

PR #702, opened 2023-12-21
feat(fs consistent snapshot): use ioctl for issuing FIFREEZE and FITHAW

Abhinandan-Purkait:


control-plane/csi-driver/src/bin/node/main_.rs
```diff
@@ -166,8 +167,46 @@ pub(super) async fn main() -> anyhow::Result<()> {
                     Example:\n --node-selector key=value --node-selector key2=value2",
                 ),
         )
+        .subcommand(
+            clap::Command::new("fsfreeze")
+                .arg(
+                    Arg::new("volume-id")
+                        .short('v')
+                        .long("volume-id")
+                        .value_name("UUID")
```
tiagolobocastro, line 182: add .required(true) here and below

control-plane/csi-driver/src/bin/node/main_.rs
```diff
@@ -166,8 +167,46 @@ pub(super) async fn main() -> anyhow::Result<()> {
...
+                        .help("Volume Id")
```
tiagolobocastro:
```suggestion
                        .help("Uuid of the volume to freeze")
```

control-plane/csi-driver/src/bin/node/nodeplugin_grpc.rs
```diff
@@ -46,18 +50,50 @@ impl NodePlugin for NodePluginSvc {
     ) -> Result<Response<FreezeFsReply>, Status> {
         let volume_id = request.into_inner().volume_id;
         debug!("freeze_fs({})", volume_id);
-        freeze_volume(&volume_id).await?;
-        Ok(Response::new(FreezeFsReply {}))
+        let _guard = VolumeFsfreezeGuard::new_str(&volume_id)?;
+        let output = Command::new("csi-node")
+            .arg("--node-name")
+            .arg("dummy")
```
tiagolobocastro: I think you can make node-name required only for when there's no subcommand?

control-plane/csi-driver/src/bin/node/nodeplugin_grpc.rs
```diff
@@ -46,18 +50,50 @@ impl NodePlugin for NodePluginSvc {
...
+            .arg("fsfreeze")
+            .arg("-v")
+            .arg(volume_id.clone())
+            .output()
+            .await
+            .map_err(|error| Status::aborted(error.to_string()))?;
```
tiagolobocastro: So an error from freeze/unfreeze is always converted to aborted?

control-plane/csi-driver/src/bin/node/nodeplugin_grpc.rs
```diff
@@ -46,18 +50,50 @@ impl NodePlugin for NodePluginSvc {
     ) -> Result<Response<FreezeFsReply>, Status> {
         let volume_id = request.into_inner().volume_id;
         debug!("freeze_fs({})", volume_id);
-        freeze_volume(&volume_id).await?;
-        Ok(Response::new(FreezeFsReply {}))
+        let _guard = VolumeFsfreezeGuard::new_str(&volume_id)?;
+        let output = Command::new("csi-node")
```
tiagolobocastro: maybe hide the implementation of calling freeze/unfreeze?
Also, do we want to add any timeouts? (Can't remember if this was covered on design or not)

control-plane/csi-driver/src/bin/node/nodeplugin_svc.rs
```diff
@@ -23,104 +26,100 @@ pub(crate) enum ServiceError {
...
+
+#[derive(EnumString, Clone, Debug, Eq, PartialEq, Display)]
+#[strum(serialize_all = "lowercase")]
+pub(crate) enum FsFreezeOpt {
+    Freeze,
+    Unfreeze,
+}
+
+// Nix macro to issue FIFREEZE ioctl.
+ioctl_readwrite!(fs_freeze, 'X', 119, i32);
```
tiagolobocastro: freeze/unfreeze "low-level" code should be in separate module

control-plane/csi-driver/src/bin/node/main_.rs
Abhinandan-Purkait: Done.

control-plane/csi-driver/src/bin/node/nodeplugin_grpc.rs

control-plane/csi-driver/src/bin/node/nodeplugin_grpc.rs
Abhinandan-Purkait: Have refactored it, it would now contain the Errno and accordingly we can return specific errors.

control-plane/csi-driver/src/bin/node/nodeplugin_grpc.rs
Abhinandan-Purkait: Have added executor with timeout.

control-plane/csi-driver/src/bin/node/nodeplugin_svc.rs
Abhinandan-Purkait: Moved.

control-plane/csi-driver/src/limiter.rs
```diff
@@ -30,32 +31,46 @@ impl From<Error> for tonic::Status {
...
+            })?,
+            serializer_type,
+        )
     }
 
     /// Tries to take a new volume reservation and returns error if an existing
     /// operation is already in progress.
-    pub fn new(uuid: uuid::Uuid) -> Result<Self, Error> {
-        let mut inventory = volume_serializer().lock().unwrap();
+    pub fn new(uuid: uuid::Uuid, serializer_type: VolumeSerializerType) -> Result<Self, Error> {
```
tiagolobocastro: let's add separate functions, rather than another parameter?
I've also been wondering btw, since destaging with the fs freeze call stuck doesn't work, should we use the same guard then?

control-plane/csi-driver/src/filesystem/mod.rs
```diff
@@ -0,0 +1,57 @@
...
+/// Filesystem freeze low level code.
+pub mod fsfreeze;
+
+use std::time::Duration;
+use tokio::{process::Command, time::timeout};
+use tonic::Status;
+use tracing::debug;
+
+// Name of the csi-node binary to be used to make a call.
+const CSI_NODE_BINARY: &str = "csi-node";
```
tiagolobocastro: Is this needed? Can't you find out the path from std::env? Maybe args[0]..

control-plane/csi-driver/src/filesystem/mod.rs
```diff
@@ -0,0 +1,57 @@
...
+        .arg(FREEZE_VOLUME_ID_ARG_NAME)
+        .arg(volume_id)
+        .spawn()
+        .map_err(|error| {
+            Status::failed_precondition(format!(
+                "Failed to spawn child process for {freeze_op} for {volume_id}, {error}"
+            ))
+        })?;
+
+    let execute_with_timeout_result = timeout(timeout_duration, command.wait()).await;
```
tiagolobocastro: Shouldn't the timeout be set only at the calling side, and allow the future to run even after the timeout?
Otherwise we'll keep stacking freeze ioctl?

control-plane/csi-driver/src/bin/node/nodeplugin_svc.rs
```diff
@@ -177,3 +104,67 @@ pub(crate) async fn find_mount(
         Ok(None)
     }
 }
+
+/// /// Issue FIFREEZE and FITHAW ioctl on the mount_path derived from the volume_id based on the
+/// command.
```
tiagolobocastro:
```suggestion
/// Issue FIFREEZE and FITHAW ioctl on the mount_path derived from the volume_id based on the command.
```

control-plane/csi-driver/src/bin/node/nodeplugin_svc.rs
```diff
@@ -177,3 +104,67 @@ pub(crate) async fn find_mount(
...
+    let uuid = Uuid::parse_str(volume_id).context(InvalidVolumeId {
+        volume_id: volume_id.to_string(),
+    })?;
+
+    if let Some(device) = Device::lookup(&uuid).await.context(InternalFailure {
+        volume_id: volume_id.to_string(),
+    })? {
+        let device_path = device.devname();
+        if let Some(mnt) = mount::find_mount(Some(&device_path), None) {
+            // Make a preflight check, to ensure subsystem has atleast one live path.
```
tiagolobocastro:
```suggestion
            // Make a preflight check, to ensure subsystem has at least one live path.
```

control-plane/csi-driver/src/bin/node/nodeplugin_svc.rs
```diff
@@ -177,3 +104,67 @@ pub(crate) async fn find_mount(
...
+            fsfreeze_preflight_check(volume_id, &device_nqn).map_err(|errno| {
+                ServiceError::FsfreezeFailed {
+                    volume_id: volume_id.to_string(),
+                    errno,
+                }
+            })?;
+
+            // Execute the fsfreeze call.
+            let dest = mnt.dest.display().to_string();
+            return fsfreeze_ioctl(command, dest).await.map_err(|errno| {
```
tiagolobocastro: fsfreeze_ioctl should take a &Path I guess?

tiagolobocastro:
```suggestion
            fsfreeze_ioctl(command, dest).await.map_err(|errno| {
```

control-plane/csi-driver/src/bin/node/nodeplugin_svc.rs
```diff
@@ -177,3 +104,67 @@ pub(crate) async fn find_mount(
...
+                });
+            }
+            debug!(
+                "{} for volume_id :{} : failed, cannot find volume",
+                command, volume_id
+            );
+        }
+        return Err(ServiceError::VolumeNotMounted {
+            volume_id: volume_id.to_string(),
+        });
```
tiagolobocastro:
```suggestion
        Err(ServiceError::VolumeNotMounted {
            volume_id: volume_id.to_string(),
        })
```

control-plane/csi-driver/src/bin/node/main_.rs
```diff
@@ -166,8 +170,48 @@ pub(super) async fn main() -> anyhow::Result<()> {
                     Example:\n --node-selector key=value --node-selector key2=value2",
                 ),
         )
+        .subcommand(
+            clap::Command::new("fsfreeze")
```
tiagolobocastro: nit:
```suggestion
            clap::Command::new("fs-freeze")
```

control-plane/csi-driver/src/bin/node/main_.rs
```diff
@@ -166,8 +170,48 @@ pub(super) async fn main() -> anyhow::Result<()> {
...
+                    Arg::new("volume-id")
+                        .short('v')
+                        .long("volume-id")
+                        .value_name("UUID")
+                        .required(true)
+                        .help("Uuid of the volume to freeze")
+                )
+        )
+        .subcommand(
+            clap::Command::new("fsunfreeze")
```
tiagolobocastro: nit:
```suggestion
            clap::Command::new("fs-unfreeze")
```

control-plane/csi-driver/src/bin/node/nodeplugin_svc.rs
Abhinandan-Purkait: This is needed right?

Abhinandan-Purkait: Have tried that, but seems it causing lifetime issue.

control-plane/csi-driver/src/bin/node/nodeplugin_svc.rs
Abhinandan-Purkait: This is needed ?

tiagolobocastro: Nevermind, I misread the diff on this function, ignore the other one too...
If you're up to it I suggest changing code to let Some else return to make it easier to read.

Abhinandan-Purkait: Right, done.

control-plane/csi-driver/src/bin/node/fsfreeze/bin/ioctl.rs
```diff
@@ -0,0 +1,61 @@
...
+/// Preflight checks to ensure that fsfreeze call would be made in a desirable condition.
+pub fn fsfreeze_preflight_check(volume_id: &str, nqn: &str) -> Result<(), Errno> {
+    let subsystems =
+        nvmeadm::nvmf_subsystem::Subsystem::try_from_nqn(nqn).map_err(|_| Errno::ENODATA)?;
+    if subsystems.iter().any(|subsystem| subsystem.state == "live") {
+        return Ok(());
+    }
+    error!(
+        volume_id = volume_id,
+        nqn = nqn,
```
tiagolobocastro:
```suggestion
        volume.id = volume_id,
        subsystem.nqn = nqn,
```
or 
```suggestion
        volume_id, nqn,
```

control-plane/csi-driver/src/bin/node/fsfreeze/bin/ioctl.rs
```diff
@@ -0,0 +1,61 @@
...
+}
+
+/// Issue FIFREEZE and FITHAW ioctl on the mount_path based on the command.
+pub async fn fsfreeze_ioctl(command: FsFreezeOpt, mount_path: PathBuf) -> Result<(), Errno> {
+    let blocking_task = tokio::task::spawn_blocking(move || {
+        let file = fs::File::open(&mount_path).map_err(|error| {
+            error!(
+                "Failed to get file descriptor path: {}, {}",
+                mount_path.display(),
+                error.to_string()
```
tiagolobocastro:
```suggestion
                error,
                mount = mount_path.display(),
                "Failed to open volume mount path"
```

control-plane/csi-driver/src/bin/node/fsfreeze/bin/ioctl.rs
```diff
@@ -0,0 +1,61 @@
...
+                Ok(_) => Ok(()),
+                Err(Errno::EBUSY) => Ok(()),
+                Err(errno) => Err(errno),
+            },
+            FsFreezeOpt::Unfreeze => loop {
+                match unsafe { fs_thaw(file.as_raw_fd(), &mut 0) } {
+                    Ok(val) if val != 0 => return Ok(()),
+                    Err(Errno::EINVAL) => return Ok(()),
+                    Err(errno) => return Err(errno),
+                    _ => {}
```
tiagolobocastro: nit: might be clearer here
```suggestion
                    Ok(val) => { if val != 0 return Ok(()); },
                    Err(Errno::EINVAL) => return Ok(()),
                    Err(errno) => return Err(errno),
```
or 
```suggestion
                    Ok(val) if val != 0 => return Ok(()),
                    Ok(val) => {},
                    Err(Errno::EINVAL) => return Ok(()),
                    Err(errno) => return Err(errno),
```

control-plane/csi-driver/src/bin/node/error.rs
```diff
@@ -145,3 +148,27 @@ impl From<DeviceError> for tonic::Status {
         }
     }
 }
+
+/// Error for the fsfreeze operations.
+#[derive(Debug, Snafu)]
+#[snafu(visibility(pub(crate)), context(suffix(false)))]
+pub(crate) enum FsfreezeError {
+    #[snafu(display("Cannot find volume: volume ID: {}", volume_id))]
```
tiagolobocastro: Can do this throughout

```suggestion
    #[snafu(display("Cannot find volume: volume ID: {volume_id}"))]
```

control-plane/csi-driver/src/bin/node/error.rs
```diff
@@ -145,3 +148,27 @@ impl From<DeviceError> for tonic::Status {
...
+    VolumeNotFound { volume_id: String },
+    #[snafu(display("Volume is not mounted: volume ID: {}", volume_id))]
+    VolumeNotMounted { volume_id: String },
+    #[snafu(display("Invalid volume ID: {}, {}", volume_id, source))]
+    InvalidVolumeId {
+        source: uuid::Error,
```
tiagolobocastro: Depending on how errors are returned up, ensure source error doesn't appear twice in the error message.

control-plane/csi-driver/src/bin/node/dev.rs
```diff
@@ -62,6 +62,7 @@ pub(crate) trait Attach: Sync + Send {
 pub(crate) trait Detach: Sync + Send {
     async fn detach(&self) -> Result<(), DeviceError>;
     fn devname(&self) -> DeviceName;
+    fn devnqn(&self) -> String;
```
tiagolobocastro: Consider returning &str

control-plane/csi-driver/src/bin/node/fsfreeze/bin/mod.rs
```diff
@@ -0,0 +1,90 @@
...
+            })?
+    else {
+        return Err(FsfreezeError::VolumeNotFound {
+            volume_id: volume_id.to_string(),
+        });
+    };
+
+    let device_path = device.devname();
+    if let Some(mnt) = mount::find_mount(Some(&device_path), None) {
+        // Make a preflight check, to ensure subsystem has at least one live path.
```
tiagolobocastro, line 45: wonder if we be a bit paranoid, and ensure only 1 path, and live ofcourse.

control-plane/csi-driver/src/bin/node/fsfreeze/bin/mod.rs
```diff
@@ -0,0 +1,90 @@
...
+    } else {
+        // mount::find_mount does not return any matches,
+        // for mounts which are bind mounts to block devices
+        // (raw block volume).
+        // It would be incorrect to return the VolumeNotFound error,
+        // if the volume is mounted as a raw block volume on this node.
+        // Use findmnt to work out if volume is mounted as a raw
+        // block, i.e. we get some matches, and return the
+        // BlockDeviceMount error.
+        let mountpaths = findmnt::get_mountpaths(&device_path).map_err(|error| {
```
tiagolobocastro, line 70: what's the difference between these two calls then? find_mount and get_mountpaths

control-plane/csi-driver/src/bin/node/fsfreeze/mod.rs
```diff
@@ -0,0 +1,47 @@
...
+// Volume name arg for fs freeze and unfreeze subcommand.
+const FREEZE_VOLUME_ID_ARG_NAME: &str = "--volume-id";
+
+/// Creates a child process that calls up the csi-node binary to execute FIFREEZE and FITHAW using
+/// the volume id.
+pub async fn fsfreeze(volume_id: &str, freeze_op: FsFreezeOpt) -> Result<(), Status> {
+    let volume_id = volume_id.to_string();
+    let binary_name = std::env::args().next().unwrap_or(CSI_NODE_BINARY.into());
+    tokio::spawn(async move {
+        debug!(volume_id = volume_id, "{freeze_op}");
```
tiagolobocastro: same comment as before

control-plane/csi-driver/src/bin/node/fsfreeze/mod.rs
```diff
@@ -0,0 +1,47 @@
...
+// Default csi-node binary name.
+const CSI_NODE_BINARY: &str = "csi-node";
+// Volume name arg for fs freeze and unfreeze subcommand.
+const FREEZE_VOLUME_ID_ARG_NAME: &str = "--volume-id";
+
+/// Creates a child process that calls up the csi-node binary to execute FIFREEZE and FITHAW using
+/// the volume id.
+pub async fn fsfreeze(volume_id: &str, freeze_op: FsFreezeOpt) -> Result<(), Status> {
+    let volume_id = volume_id.to_string();
+    let binary_name = std::env::args().next().unwrap_or(CSI_NODE_BINARY.into());
```
tiagolobocastro: Looks like you can use `current_exe`, might be better than `std::env::args`

control-plane/csi-driver/src/bin/node/fsfreeze/mod.rs
```diff
@@ -0,0 +1,47 @@
...
+            .arg(&volume_id)
+            .output()
+            .await
+        {
+            Ok(output) => {
+                if output.status.success() {
+                    Ok(())
+                } else {
+                    Err(Status::internal(format!(
+                        "{freeze_op} for {volume_id}, failed with {}, error: {}",
```
tiagolobocastro: I might be reading this wrong, but this means we don't get any meaningful error back?

control-plane/csi-driver/src/bin/node/main_.rs
```diff
@@ -103,7 +108,7 @@ pub(super) async fn main() -> anyhow::Result<()> {
                 .long("node-name")
                 .value_name("NAME")
                 .help("Unique node name where this instance runs")
-                .required(true)
+                .required(true),
```
tiagolobocastro: spurious?

```suggestion
                .required(true)
```

control-plane/csi-driver/src/bin/node/nodeplugin_svc.rs
```diff
@@ -2,150 +2,43 @@
...
-pub(crate) async fn unfreeze_volume(volume_id: &str) -> Result<(), ServiceError> {
-    fsfreeze(volume_id, "--unfreeze").await
-}
-
 /// Lookup the device by its volume id.
-pub(crate) async fn lookup_device(volume_id: &str) -> Result<Box<dyn Detach>, ServiceError> {
-    let uuid = Uuid::parse_str(volume_id).context(InvalidVolumeId {
-        volume_id: volume_id.to_string(),
-    })?;
+pub(crate) async fn lookup_device(volume_id: &str) -> Result<Box<dyn Detach>, tonic::Status> {
```
tiagolobocastro, line 19: What's the reason for giving up the service error in favour of direct status?

control-plane/csi-driver/src/bin/node/fsfreeze/bin/ioctl.rs
Abhinandan-Purkait:
```
error!(
                error = error.to_string(),
                mount = mount_path.display().to_string(),
                "Failed to open volume mount path"
            );
```

We need to convert to string other wise doesn't work like the above.

control-plane/csi-driver/src/bin/node/error.rs
Abhinandan-Purkait: I don't think they are, this is being returned at one place only.

control-plane/csi-driver/src/bin/node/fsfreeze/bin/mod.rs
Abhinandan-Purkait: There can be multiple live paths as well right, the duplicate thing that we saw the other day :-p ? 

But should we fail the freeze even if we have multiple live paths?

control-plane/csi-driver/src/bin/node/fsfreeze/bin/mod.rs
Abhinandan-Purkait: find_mount takes either of the source or destination and returns that if present.

get_mountpaths returns all mount paths for a device.

control-plane/csi-driver/src/bin/node/fsfreeze/mod.rs
Abhinandan-Purkait: `debug!(volume.id = volume_id, "{freeze_op}");` Changed to this

control-plane/csi-driver/src/bin/node/fsfreeze/mod.rs
Abhinandan-Purkait: We do get the error back, the error contains the Errno

Abhinandan-Purkait: Now since this is an error from the stderr from the Command output we cannot have very nice mapping from Stderr string to tonic status. That's why most of it is internal or aborted.

For example below gets returned to the client when volume is not found/ preflight check fails.

`fs-freeze for ec4e66fd-3b33-4439-b504-d49aba53da26, failed with exit status: 1, error: Error: Cannot find volume: volume ID: ec4e66fd-3b33-4439-b504-d49aba53da26`

`fs-freeze for ec4e66fd-3b33-4439-b504-d49aba53da26, failed with exit status: 1, error: Error: fsfreeze failed: volume ID: ec4e66fd-3b33-4439-b504-d49aba53da26, EHOSTUNREACH: No route to host`

control-plane/csi-driver/src/bin/node/nodeplugin_svc.rs
Abhinandan-Purkait: To reduce another layer of conversion from service error to status.

control-plane/csi-driver/src/bin/node/fsfreeze/mod.rs
tiagolobocastro: Stderr? But we get the actual error back via status right?

control-plane/csi-driver/src/bin/node/fsfreeze/mod.rs
tiagolobocastro: probably volume.uuid better as I noticed we already use that elsewhere in this csi-node

control-plane/csi-driver/src/bin/node/fsfreeze/bin/mod.rs
tiagolobocastro: how did we get duplicate live paths again?

control-plane/csi-driver/src/bin/node/fsfreeze/bin/ioctl.rs
tiagolobocastro: Just add the formatter,
```suggestion
error!(
                %error,
                mount = %mount_path.display(),
                "Failed to open volume mount path"
            );
```

control-plane/csi-driver/src/bin/node/nodeplugin_svc.rs
tiagolobocastro: I guess that's fine, though means you can't reuse this code for other internal/non-tonic things

tiagolobocastro:
There's a refactor in commit lint so `feat(fs consistent snapshot): refactor` should be `reactor(xx): xyz`
Though if you're adding this refactor to code which you've added in your PR don't keep the rafactor and squash into original commit instead, unless there's a specific reason to add the commits in stages, then it's fine.

control-plane/csi-driver/src/bin/node/fsfreeze/bin/mod.rs
Abhinandan-Purkait: By adding a `duplicate` param to the nvme connect. Not sure if doing duplicate connect has any purpose.

control-plane/csi-driver/src/bin/node/fsfreeze/mod.rs
Abhinandan-Purkait: Yes we get, but what I mean is in this method where we call using the `Command` we get the error from the `Stderr` of `output`, and we return a tonic status internal/aborted to grpc client using the stderr string. 

But since `Command` method gives error form of `Stderr` string we don't have a very good way of mapping the particular `fsfreeze error` to specific `tonic status`.

control-plane/csi-driver/src/bin/node/fsfreeze/bin/ioctl.rs
Abhinandan-Purkait: Ahh right, I forgot about this, thanks. :-;

control-plane/csi-driver/src/bin/node/fsfreeze/mod.rs
Abhinandan-Purkait: Alright, would change to volume.uuid

Abhinandan-Purkait:
> There's a refactor in commit lint so `feat(fs consistent snapshot): refactor` should be `reactor(xx): xyz` Though if you're adding this refactor to code which you've added in your PR don't keep the rafactor and squash into original commit instead, unless there's a specific reason to add the commits in stages, then it's fine.

I will squash the commits and spilt them into logical commits.

control-plane/csi-driver/src/bin/node/fsfreeze/bin/mod.rs
Abhinandan-Purkait: @tiagolobocastro Should I make this change or is it fine for now?

`
Make a preflight check, to ensure subsystem has at exactly one live path. Can't be 0 can't be more than 1.
`

Abhinandan-Purkait:
```
let live_subsystems: Vec<nvmeadm::nvmf_subsystem::Subsystem> =
        nvmeadm::nvmf_subsystem::Subsystem::try_from_nqn(nqn)
            .map_err(|_| Errno::ENODATA)?
            .into_iter()
            .filter(|subsystem| subsystem.state == "live")
            .collect();
    match live_subsystems.len() {
        0 => {
            error!(
                volume.id = volume_id,
                subsystem.nqn = nqn,
                "Preflight check for fsfreeze failed, subsystem has no live path"
            );
            Err(Errno::EHOSTUNREACH)
        }
        1 => Ok(()),
        _ => {
            error!(
                volume.id = volume_id,
                subsystem.nqn = nqn,
                "Preflight check for fsfreeze failed, subsystem has more than one live path"
            );
            Err(Errno::EMLINK)
        }
    }
```

tiagolobocastro: > By adding a `duplicate` param to the nvme connect. Not sure if doing duplicate connect has any purpose.

Ah right.. that was a "bug" though right? Though I guess we can simply use .all and ensure all paths are live then?

control-plane/csi-driver/src/bin/node/fsfreeze/mod.rs
Abhinandan-Purkait: Done.

control-plane/csi-driver/src/bin/node/error.rs
```diff
@@ -145,3 +148,43 @@ impl From<DeviceError> for tonic::Status {
...
+    BlockDeviceMount { volume_id: String },
+    #[snafu(display("Not a valid fsfreeze command"))]
+    InvalidFreezeCommand,
+}
+
+impl From<FsfreezeError> for ExitCode {
+    fn from(value: FsfreezeError) -> Self {
+        match value {
+            FsfreezeError::VolumeNotFound { .. } => ExitCode::from(Errno::ENODEV as u8),
+            FsfreezeError::VolumeNotMounted { .. } => ExitCode::from(Errno::EINVAL as u8),
```
tiagolobocastro:
```suggestion
            FsfreezeError::VolumeNotMounted { .. } => ExitCode::from(Errno::ENOENT as u8),
```

control-plane/csi-driver/src/bin/node/error.rs
```diff
@@ -145,3 +148,43 @@ impl From<DeviceError> for tonic::Status {
...
+            FsfreezeError::InvalidVolumeId { .. } => ExitCode::from(Errno::EINVAL as u8),
+            FsfreezeError::FsfreezeFailed { errno, .. } => ExitCode::from(errno as u8),
+            FsfreezeError::InternalFailure { .. } => ExitCode::from(Errno::ENXIO as u8),
```
tiagolobocastro: When do we get Internal Failure again?

control-plane/csi-driver/src/bin/node/fsfreeze/bin/ioctl.rs
```diff
@@ -0,0 +1,68 @@
...
+        .all(|subsystem| subsystem.state == "live")
+    {
+        return Ok(());
+    }
+    error!(
+        volume.id = volume_id,
+        subsystem.nqn = nqn,
+        "Preflight check for fsfreeze failed, nvmf subsystem not in desired state"
+    );
+    Err(Errno::EPERM)
```
tiagolobocastro:
```suggestion
    Err(Errno::EOPNOTSUPP)
```

control-plane/csi-driver/src/bin/node/fsfreeze/bin/ioctl.rs
```diff
@@ -0,0 +1,68 @@
...
+}
+
+/// Issue FIFREEZE and FITHAW ioctl on the mount_path based on the command.
+pub async fn fsfreeze_ioctl(command: FsFreezeOpt, mount_path: PathBuf) -> Result<(), Errno> {
+    let blocking_task = tokio::task::spawn_blocking(move || {
+        let file = fs::File::open(&mount_path).map_err(|error| {
+            error!(
+                %error,
+                mount = %mount_path.display().to_string(),
```
tiagolobocastro: Does this work?

```suggestion
                mount = %mount_path.display(),
```

control-plane/csi-driver/src/bin/node/fsfreeze/bin/mod.rs
```diff
@@ -0,0 +1,90 @@
...
+            FsfreezeError::InternalFailure {
+                source: error,
+                volume_id: volume_id.to_string(),
+            }
+        })?;
+        if !mountpaths.is_empty() {
+            debug!(
+                "{} for volume_id :{} : failed for block device",
+                command, volume_id
+            );
```
tiagolobocastro: This debug message is not very clear

control-plane/csi-driver/src/bin/node/main_.rs
```diff
@@ -166,8 +172,46 @@ pub(super) async fn main() -> anyhow::Result<()> {
...
+        )
+        .subcommand(
+            clap::Command::new("fs-unfreeze")
+                .arg(
+                    Arg::new("volume-id")
+                        .short('v')
+                        .long("volume-id")
+                        .value_name("UUID")
+                        .required(true)
+                        .help("Uuid of the volume to freeze")
```
tiagolobocastro:
```suggestion
                        .help("Uuid of the volume to unfreeze")
```

utils/utils-lib/src/tracing_telemetry.rs
```diff
@@ -155,6 +155,34 @@ pub fn init_tracing_ext<T: std::net::ToSocketAddrs>(
     };
 }
 
+/// Init tracing for utility binary, ex csi-node for fsfreeze.
+pub fn init_tracing_for_utility_binary(stderr: bool) {
```
tiagolobocastro: Let's create a Builder type instead:
```rust
struct FmtStyle { Compact, Pretty };
struct TracingTel {
  writer: FmtLayer,
  style: FmtStyle,
  colours: bool
}
```
 allowing the code in the csi to be written like so:
```rust
TracingTel::builder().with_writer(FmtLayer::Stdout).with_style(FmtStyle::Compact).with_colours(false)
```

control-plane/csi-driver/src/bin/node/error.rs
Abhinandan-Purkait: All the device specific errors are mapped to internal

Abhinandan-Purkait: ENXIO is no such device or address, that's the closest I think is an Errno

control-plane/csi-driver/src/bin/node/fsfreeze/bin/ioctl.rs
Abhinandan-Purkait: Right, this would be better

control-plane/csi-driver/src/bin/node/fsfreeze/bin/mod.rs
Abhinandan-Purkait: I will remove this, it's not really needed.

control-plane/csi-driver/src/bin/node/main_.rs
Abhinandan-Purkait: Oops, missed this. :-;

control-plane/csi-driver/src/bin/node/error.rs
Abhinandan-Purkait: ELIBACC --> Changed to this.

utils/utils-lib/src/tracing_telemetry.rs
Abhinandan-Purkait: Done.

tiagolobocastro (approved):

utils/utils-lib/src/tracing_telemetry.rs
```diff
@@ -155,7 +155,82 @@ pub fn init_tracing_ext<T: std::net::ToSocketAddrs>(
...
+            }
+            (FmtLayer::Stdout, FmtStyle::Pretty) => {
+                let stdout = tracing_subscriber::fmt::layer()
+                    .with_writer(std::io::stdout)
+                    .pretty()
+                    .with_ansi(self.colours);
+                let subscriber = Registry::default().with(stdout);
+                subscriber.init();
+            }
+            _ => {}
```
tiagolobocastro: Is this required?

Abhinandan-Purkait: Yeah, the `FmtLayer::None` is also there, so I added a catch all.

Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors merge

Abhinandan-Purkait:
PR got merged by did not get closed. Bors issue?

PR #705, opened 2023-12-26
feat(volume/resize): implement resource resize for volume

dsharma-dc:


control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -185,10 +185,62 @@ impl ResourceResize for OperationGuardArc<VolumeSpec> {
...
+                registry,
+                &state,
+                VolumeOperation::Resize(request.requested_size),
+            )
+            .await?;
+        // For the volume being resized, resize each replica of the volume. For a failure to
+        // any of the replica, the volume resize operation is to be deemed failure.
+        let replicas = specs.volume_replicas(&request.uuid);
+        for replica in replicas {
+            let mut replica = replica.operation_guard_wait().await?;
```
tiagolobocastro: you need to move this to a sub function, cannot return error while update has been started..

dsharma-dc: Do you mean something like this? `resize_volume_replicas` is new helper function.

```
        let spec_clone = volume.start_update().await?;
        match self.resize_volume_replicas(registry, &replicas, request.requested_size, &spec_clone).await {
            Ok(_) => {
                let volume = registry.volume(&request.uuid).await?;
                self.complete_update(registry, Ok(volume), spec_clone).await
            },
            Err(e) => Err(e),
        }
```

Abhinandan-Purkait: Should we not `continue` if we cannot take the guard? By continue I mean resize as many as replicas we can? Similar to what we do for volume delete

Abhinandan-Purkait: In line 214 I don't think you can return using the `?`, there are 2 things we can do, i.e if 214 gives an error `continue`, `ignore error silently/log and move ahead` or use a `validate_update_step` to return the error from 214.

dsharma-dc: Alright, I'm moving these lines to a helper function as mentioned above, will handle there.

dsharma-dc: If we do that, by design the `volume resize` operation is failure as a whole if even one replica is left out - and the added space from replicas that get resized will anyway be potentially reclaimed.

dsharma-dc: @Abhinandan-Purkait @tiagolobocastro PTAL if the diff looks ok now.

control-plane/agents/src/bin/core/pool/specs.rs
```diff
@@ -149,6 +149,8 @@ impl SpecOperationsHelper for ReplicaSpec {
             }
             ReplicaOperation::Unshare => Ok(()),
             ReplicaOperation::OwnerUpdate(_) => Ok(()),
+            // XXX: Shall we do something here with input size?
+            ReplicaOperation::Resize(_) => Ok(()),
```
Abhinandan-Purkait, line 153: Yeah, probably can add a validation here itself?

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -185,10 +185,41 @@ impl ResourceResize for OperationGuardArc<VolumeSpec> {
...
+        // resize any of the replica, the volume resize operation is to be deemed failure.
+        match self
+            .resize_volume_replicas(registry, &replicas, request.requested_size, &spec_clone)
+            .await
+        {
+            Ok(_) => {
+                let volume = registry.volume(&request.uuid).await?;
+                self.complete_update(registry, Ok(volume), spec_clone).await
+            }
+            Err(e) => Err(e),
```
Abhinandan-Purkait: Can we not pass the result directly to the complete_update?

dsharma-dc: > Can we not pass the result directly to the complete_update?

We can if we make `resize_volume_replicas` return a `Volume` object. Keeping it separate would be required in the case when we have to work on nexus after getting Ok() here.

Abhinandan-Purkait: We cannot return error from line 218, it's again the same issue.
Also we cannot simply return from 221 since update has started.

So rather we can do this ?

```suggestion
           let result = match self
            .resize_volume_replicas(registry, &replicas, request.requested_size, &spec_clone)
            .await
        {
            Ok(_) => {
                match registry.volume(&request.uuid).await {
                       Ok(volume) => Ok(volume),
                       Err(e) => Err(e)
                   }
            }
            Err(e) => Err(e),
        }
         self.complete_update(registry, result, spec_clone).await
```

dsharma-dc: ooh yike, I missed that line 218 call. 221 would've been ok I think because before that `validate` has been done internally by helper function. I'll fix based on suggestion.

dsharma-dc: Modified.

dsharma-dc:
bors try

control-plane/agents/src/bin/core/pool/specs.rs
dsharma-dc: > Yeah, probably can add a validation here itself?

Added on the VolumeOperation check step.

control-plane/agents/src/bin/core/controller/scheduling/volume.rs
```diff
@@ -738,3 +738,72 @@ impl ResourceFilter for CloneVolumeSnapshot {
         self.data.list
     }
 }
+
+/// The context to check pool capacity for volume replica resize feasibility.
+#[derive(Clone)]
+pub(crate) struct ReplicaResizePoolsContext {
+    registry: Registry,
+    spec: VolumeSpec,
+    pub required_capacity: u64,
```
Abhinandan-Purkait: Doesn't need to be pub?

Abhinandan-Purkait (approved):

control-plane/agents/src/bin/core/controller/scheduling/volume.rs
dsharma-dc: > Doesn't need to be pub?

Will make it private and add a getter function to be used from `PoolBaseFilters`.

control-plane/agents/src/bin/core/controller/scheduling/resources/mod.rs
```diff
@@ -101,6 +102,25 @@ impl PoolItemLister {
             None => vec![],
         }
     }
+    /// Get a list of pool items to resize the replicas that reside on them.
+    pub(crate) async fn list_for_resize(registry: &Registry, spec: &VolumeSpec) -> Vec<PoolItem> {
+        let pool_ids = registry.specs().volume_replica_pools(&spec.uuid);
+
+        let items = Self::nodes(registry)
```
tiagolobocastro: Rather than listing all nodes, you could just list the ones that that the pools

control-plane/agents/src/bin/core/controller/scheduling/resources/mod.rs
```diff
@@ -101,6 +102,25 @@ impl PoolItemLister {
...
+            .flat_map(|n| {
+                n.pool_wrappers()
+                    .iter()
+                    .filter(|p| registry.specs().pool(&p.id).is_ok() && pool_ids.contains(&p.id))
+                    .map(|p| PoolItem::new(n.clone(), p.clone(), None))
+                    .collect::<Vec<_>>()
+            })
+            .collect();
+
+        items
```
tiagolobocastro: nit: don't need variable, can just return

```suggestion
            .collect()
```

control-plane/agents/src/bin/core/controller/scheduling/volume.rs
```diff
@@ -738,3 +738,79 @@ impl ResourceFilter for CloneVolumeSnapshot {
...
+#[derive(Clone)]
+pub(crate) struct ReplicaResizePoolsContext {
+    registry: Registry,
+    spec: VolumeSpec,
+    required_capacity: u64,
+}
+
+impl ReplicaResizePoolsContext {
+    /// The additional capacity that we need.
+    pub fn required_capacity(&self) -> u64 {
```
tiagolobocastro:
```suggestion
    pub(crate) fn required_capacity(&self) -> u64 {
```

control-plane/agents/src/bin/core/controller/scheduling/volume.rs
```diff
@@ -738,3 +738,79 @@ impl ResourceFilter for CloneVolumeSnapshot {
...
+        self.required_capacity
+    }
+}
+
+pub(crate) struct ResizeVolumeReplicas {
```
tiagolobocastro, line 781: missing doc comment

control-plane/agents/src/bin/core/controller/scheduling/volume.rs
```diff
@@ -738,3 +738,79 @@ impl ResourceFilter for CloneVolumeSnapshot {
...
+    data: ResourceData<ReplicaResizePoolsContext, PoolItem>,
+}
+
+impl ResizeVolumeReplicas {
+    pub(crate) async fn builder(
```
tiagolobocastro:
```suggestion
    async fn builder(
```

control-plane/agents/src/bin/core/controller/scheduling/volume.rs
```diff
@@ -738,3 +738,79 @@ impl ResourceFilter for CloneVolumeSnapshot {
...
+    }
+    fn with_thick_policy(self) -> Self {
+        self.policy(ThickPolicy::new())
+    }
+    fn with_simple_policy(self) -> Self {
+        let simple = SimplePolicy::new(&self.data.context().registry);
+        self.policy(simple)
+    }
+
+    pub(crate) async fn builder_with_defaults(
```
tiagolobocastro, line 818: doc comment

control-plane/agents/src/bin/core/controller/scheduling/volume_policy/pool.rs
```diff
@@ -34,6 +37,15 @@ impl PoolBaseFilters {
             false => item.pool.free_space() > request.size,
         }
     }
+    /// Return true if the pool doesn't have enough capacity to resize the replica by requested
```
tiagolobocastro: Should be the opposite?

control-plane/agents/src/bin/core/pool/specs.rs
```diff
@@ -149,6 +149,8 @@ impl SpecOperationsHelper for ReplicaSpec {
             }
             ReplicaOperation::Unshare => Ok(()),
             ReplicaOperation::OwnerUpdate(_) => Ok(()),
+            // XXX: Shall we do something here with input size?
```
tiagolobocastro:
```suggestion
            // todo: Shall we do something here with input size?
```

control-plane/agents/src/bin/core/pool/specs.rs
tiagolobocastro: Guess could check if replica is online?

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -185,10 +185,32 @@ impl ResourceResize for OperationGuardArc<VolumeSpec> {
...
+
+        let spec_clone = self
+            .start_update(
+                registry,
+                &state,
+                VolumeOperation::Resize(request.requested_size),
+            )
+            .await?;
+        // For the volume being resized, resize each replica of the volume. For a failure to
+        // resize any of the replica, the volume resize operation is to be deemed failure.
```
```suggestion
        // Resize each replica of the volume. 
        // If any replica fails to be resized then the volume resize operation is deemed as a failure.
```

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -185,10 +185,32 @@ impl ResourceResize for OperationGuardArc<VolumeSpec> {
...
+        let result = match self
+            .resize_volume_replicas(registry, &replicas, request.requested_size, &spec_clone)
+            .await
+        {
+            Ok(_) => registry.volume(&request.uuid).await,
+            Err(e) => Err(e),
+        };
```
tiagolobocastro: another case where a simple map(||) would work if there was no async..

control-plane/agents/src/bin/core/volume/operations_helper.rs
```diff
@@ -527,6 +527,48 @@ impl OperationGuardArc<VolumeSpec> {
         Ok(created_replicas)
     }
 
+    pub(super) async fn resize_volume_replicas(
```
tiagolobocastro, line 531: missing doc

control-plane/agents/src/bin/core/volume/operations_helper.rs
```diff
@@ -527,6 +527,48 @@ impl OperationGuardArc<VolumeSpec> {
...
+    ) -> Result<(), SvcError> {
+        for replica in replicas {
+            let mut replica = match replica.operation_guard_wait().await {
+                Ok(r) => r,
+                Err(e) => {
+                    return self
+                        .validate_update_step(registry, Err(e), spec_clone)
+                        .await
+                }
+            };
```
tiagolobocastro:
```suggestion
            let mut replica = replica.operation_guard_wait().await?;
```

control-plane/agents/src/bin/core/volume/operations_helper.rs
```diff
@@ -527,6 +527,48 @@ impl OperationGuardArc<VolumeSpec> {
...
+                        ),
+                    )
+                    .await;
+                self.validate_update_step(registry, ret, spec_clone).await?;
+            } else {
+                let ecode = Err(SvcError::Internal {
+                    details: "Replica node not found".to_string(),
+                });
+                return self.validate_update_step(registry, ecode, spec_clone).await;
+            }
```
tiagolobocastro:
```suggestion
            let Some(node) = ResourceSpecsLocked::replica_node(registry, replica.as_ref()).await else {
              return Err(SvcError::Internal {
                    details: format!("Replica's {} node not found", replica.uuid()),
              });
            };
            let ret = replica
                .resize(
                    registry,
                    &ResizeReplica::new(
                        &node,
                        replica.as_ref().pool_name(),
                        None,
                        replica.uuid(),
                        requested_size,
                    ),
                )
                .await;
            Ok(())
```

control-plane/agents/src/bin/core/volume/scheduling.rs
```diff
@@ -126,3 +126,18 @@ pub(crate) async fn target_node_candidate(
         Some(node) => Ok(node.clone()),
     }
 }
+
+pub(crate) async fn pool_capacity_for_repl_resize(
```
tiagolobocastro: doc comments

control-plane/agents/src/bin/core/volume/scheduling.rs
```diff
@@ -126,3 +126,18 @@ pub(crate) async fn target_node_candidate(
...
+    req_capacity: u64,
+) -> Vec<PoolWrapper> {
+    let builder =
+        volume::ResizeVolumeReplicas::builder_with_defaults(registry, volume, req_capacity).await;
+    let pools_nocap = builder.collect();
+
+    pools_nocap
+        .iter()
+        .map(|p| p.pool().clone())
+        .collect::<Vec<_>>()
```
tiagolobocastro:
```suggestion
  builder.collect()
      .into_iter()
      .map(|item| item.collect())
      .collect()
```

control-plane/agents/src/bin/core/volume/specs.rs
```diff
@@ -320,6 +320,37 @@ pub(crate) async fn healthy_volume_replicas(
...
 }
 
+/// Check if the volume state is ok for resize operation.
+pub fn vol_status_ok_for_resize(spec: &VolumeSpec) -> bool {
+    // The volume should be unpublished/offline.
+    spec.active_config().is_none()
+        || spec
+            .config()
+            .as_ref()
+            .is_some_and(|c| c.active_target().is_none())
```
tiagolobocastro:
```suggestion
    spec.target().is_none()
```

control-plane/stor-port/src/types/v0/transport/replica.rs
```diff
@@ -619,15 +619,15 @@ impl ResizeReplica {
     pub fn new(
         node: &NodeId,
         pool_id: &PoolId,
-        name: &ReplicaName,
+        name: Option<&ReplicaName>,
```
tiagolobocastro, line 622: Why does it need to be optional?

control-plane/agents/src/bin/core/volume/operations_helper.rs
```diff
@@ -527,6 +527,48 @@ impl OperationGuardArc<VolumeSpec> {
...
+                            replica.as_ref().pool_name(),
+                            None,
+                            replica.uuid(),
+                            requested_size,
+                        ),
+                    )
+                    .await;
+                self.validate_update_step(registry, ret, spec_clone).await?;
+            } else {
+                let ecode = Err(SvcError::Internal {
```
tiagolobocastro: I think the scheduling should be reversed, and returned the replica states which are to be resized, and we don't need to handle this condition, it'd be handled in the scheduling which is what usually happens here.

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -185,10 +185,30 @@ impl ResourceResize for OperationGuardArc<VolumeSpec> {
...
+            .start_update(
+                registry,
+                &state,
+                VolumeOperation::Resize(request.requested_size),
+            )
+            .await?;
+        // For the volume being resized, resize each replica of the volume. For a failure to
+        // resize any of the replica, the volume resize operation is to be deemed failure.
+        let result = self
+            .resize_volume_replicas(registry, &replicas, request.requested_size, &spec_clone)
```
dsharma-dc: @Abhinandan-Purkait Here is modification we discussed.

control-plane/stor-port/src/types/v0/transport/replica.rs
dsharma-dc: > Why does it need to be optional?

Primarily following other types here. replica uuid is there as mandatory field in most, and name is optional.

control-plane/agents/src/bin/core/controller/scheduling/resources/mod.rs
dsharma-dc: > Rather than listing all nodes, you could just list the ones that that the pools

I could take this also: 
> let node_ids = registry.specs().volume_replica_nodes(&spec.uuid);

But subsequently, fetching node wrappers from registry is `async` call, which will be unclean and perhaps not allowed in the iterator to get required node wrappers.

control-plane/agents/src/bin/core/volume/operations.rs
Abhinandan-Purkait: Looks good.

control-plane/agents/src/bin/core/volume/operations_helper.rs
dsharma-dc: > I think the scheduling should be reversed, and returned the replica states which are to be resized, and we don't need to handle this condition, it'd be handled in the scheduling which is what usually happens here.

Modified, ptal. If the scheduling doesn't return all replicas as `resizable_replicas`, error will be thrown now, else we proceed to use the `resizable_replicas`

dsharma-dc:
```
E               openapi.exceptions.ServiceException: Status Code: 507
E               Reason: Insufficient Storage
E               HTTP response headers: HTTPHeaderDict({'content-length': '220', 'content-type': 'application/json', 'date': 'Tue, 09 Jan 2024 18:54:15 GMT'})
E               HTTP response body: {"details":"","message":"SvcError :: NoCapacityToResizeRepl: Replicas [\"4f791aeb-0f97-42fa-a43b-ceb375a88c96\"] can't be resized due to required capacity (52428810) not available on pool(s).","kind":"ResourceExhausted"}
```

dsharma-dc:
bors try

dsharma-dc:
bors try

control-plane/agents/src/bin/core/controller/scheduling/resources/mod.rs
```diff
@@ -101,6 +102,33 @@ impl PoolItemLister {
             None => vec![],
         }
     }
+    /// Get a list of replicas wrapped as ChildItem, for resize.
+    pub(crate) async fn list_for_resize(registry: &Registry, spec: &VolumeSpec) -> Vec<ChildItem> {
+        let replicas = registry.specs().volume_replicas(&spec.uuid);
+        let state_replicas = registry.replicas().await;
```
tiagolobocastro: A volume usually only has a few replicas, probably better to just get their states, ex:

```suggestion
        let mut state_replicas = Vec::with_capacity(replicas.len());
        for replica in &replicas {
            if let Ok(replica) = registry.replica(replica.uuid()).await {
                state_replicas.push(replica);
            }
        }
```

control-plane/agents/src/bin/core/controller/scheduling/resources/mod.rs
```diff
@@ -101,6 +102,33 @@ impl PoolItemLister {
...
+            .filter_map(|replica_spec| {
+                let replica_spec = replica_spec.lock().clone();
+                let replica_state = state_replicas
+                    .iter()
+                    .find(|state| state.uuid == replica_spec.uuid);
+
+                let pool_id = replica_spec.pool.pool_name().clone();
+                pool_wrappers
+                    .iter()
+                    .find(|p| p.id == pool_id)
```
tiagolobocastro: No need to clone

```suggestion
                let pool_id = replica_spec.pool.pool_name();
                pool_wrappers
                    .iter()
                    .find(|p| &p.id == pool_id)
```

control-plane/agents/src/bin/core/controller/scheduling/volume.rs
tiagolobocastro: Still missing

control-plane/agents/src/bin/core/controller/scheduling/volume_policy/pool.rs
```diff
@@ -34,6 +37,15 @@ impl PoolBaseFilters {
             false => item.pool.free_space() > request.size,
         }
     }
+    /// Return true if the pool has enough capacity to resize the replica by requested
```
tiagolobocastro:
```suggestion
    /// Return true if the pool has enough capacity to resize the replica by the requested
```

control-plane/agents/src/bin/core/controller/scheduling/volume_policy/pool.rs
```diff
@@ -34,6 +37,15 @@ impl PoolBaseFilters {
...
+    /// value.
+    pub(crate) fn min_free_space_repl_resize(
```
tiagolobocastro, line 51: I think we should modify min_free_space to take into account resize, otherwise seems like we're doing checks for commitment here?

control-plane/agents/src/bin/core/tests/volume/resize.rs
```diff
@@ -0,0 +1,190 @@
...
+                uuid: volume.uuid().clone(),
+            },
+            None,
+        )
+        .await;
+}
+
+// Resizing a published volume should throw error that volume is in-use.
+#[tokio::test]
+async fn resize_published() {
```
tiagolobocastro: use the same tokio test as above

control-plane/agents/src/bin/core/volume/operations_helper.rs
```diff
@@ -527,6 +527,55 @@ impl OperationGuardArc<VolumeSpec> {
...
+        replicas: &Vec<Replica>,
+        requested_size: u64,
+        spec_clone: &VolumeSpec,
+    ) -> Result<(), SvcError> {
+        for replica in replicas {
+            let repl_rsc = match registry.specs().replica_rsc(&replica.uuid) {
+                Some(rsc) => rsc,
+                None => {
+                    return self
+                        .validate_update_step(
```
tiagolobocastro: don't need the validates here, see my previous review

control-plane/agents/src/bin/core/volume/scheduling.rs
```diff
@@ -126,3 +126,20 @@ pub(crate) async fn target_node_candidate(
         Some(node) => Ok(node.clone()),
     }
 }
+
+/// List, filter and collect replicas for which the pools have the required capacity
+/// for resizing the replica.
+pub(crate) async fn pool_capacity_for_repl_resize(
```
tiagolobocastro: naming needs changing now I guess?
```suggestion
pub(crate) async fn resizeable_replicas(
```

control-plane/agents/src/bin/core/volume/specs.rs
```diff
@@ -320,6 +320,39 @@ pub(crate) async fn healthy_volume_replicas(
...
 
+/// Check if the volume state is ok for resize operation.
+pub fn vol_status_ok_for_resize(spec: &VolumeSpec) -> bool {
+    // The volume should be unpublished/offline.
+    spec.target().is_none()
+}
+/// Check if any replica is on a pool that doesn't have sufficient space for
+/// resize operation. If no such replica present, it means the volume is good
+/// to be resized and the returned vector will be of zero length.
+pub(crate) async fn pool_capacity_for_repl_resize(
```
tiagolobocastro:
```suggestion
pub(crate) async fn resizeable_replicas(
```

control-plane/agents/src/bin/core/volume/specs.rs
```diff
@@ -1055,7 +1088,20 @@ impl SpecOperationsHelper for VolumeSpec {
             }
             VolumeOperation::CreateSnapshot(_) => Ok(()),
             VolumeOperation::DestroySnapshot(_) => Ok(()),
-            VolumeOperation::Resize(_) => todo!(),
+            VolumeOperation::Resize(req_size) => {
+                if self.size > *req_size {
+                    return Err(SvcError::InvalidArguments {});
```
tiagolobocastro: Probably better to return a more descriptive error here?

```suggestion
                if self.size >= *req_size {
                    return Err(SvcError::InvalidArguments {});
```

control-plane/agents/src/bin/core/volume/specs.rs
```diff
@@ -1055,7 +1088,20 @@ impl SpecOperationsHelper for VolumeSpec {
...
+                }
+                // Do Pre-checks for volume resize.
+                if !vol_status_ok_for_resize(self) {
+                    return Err(SvcError::InUse {
+                        kind: ResourceKind::Volume,
+                        id: self.uuid.to_string(),
+                    });
+                }
```
tiagolobocastro: Should we move the error to the function itself?
```suggestion
                vol_status_ok_for_resize(self)?;
```

control-plane/agents/src/bin/core/controller/scheduling/volume_policy/mod.rs
```diff
@@ -59,4 +59,7 @@ impl DefaultBasePolicy {
             .filter(pool::PoolBaseFilters::capacity)
             .filter(pool::PoolBaseFilters::min_free_space)
     }
+    fn filter_resize(request: ResizeVolumeReplicas) -> ResizeVolumeReplicas {
+        request.filter(pool::PoolBaseFilters::min_free_space_repl_resize)
```
tiagolobocastro: missing volume overcommit check in min_free_space_repl_resize, probably we could instead modify existing min_free_space check which also checks for snapshots
also missing pool overcommit checks

control-plane/agents/src/bin/core/controller/scheduling/resources/mod.rs
dsharma-dc: This should rather be `p.id == *pool_id` then ? Comparing values, not addresses/references.

control-plane/agents/src/bin/core/volume/specs.rs
dsharma-dc: Let's keep it as is since we may have to modify function body for published volumes possibly later on, so we can do diverse error handling based on returned `bool`.

tiagolobocastro: The problem is that it doesn't add up, you're check if volume is ok for resize generically, but on the error case you return a specific "inUse", which means that actually the code is already assuming that ok for resize means in published/in use, see what I mean?

control-plane/agents/src/bin/core/controller/scheduling/resources/mod.rs
tiagolobocastro: It's the same right? In rust references aren't the same as other languages, so it's not compare pointers for example

control-plane/agents/src/bin/core/volume/specs.rs
dsharma-dc: This fits the most among existing error codes. Will have to add a new one otherwise. OR better have `InvalidArguments` take a custom message. But that'll require change in all existing usages of `InvalidArguments`

tiagolobocastro: You can add new custom error code for resize

dsharma-dc: Will add one named `VolumeShrinkUnsupported`

control-plane/agents/src/bin/core/volume/specs.rs
dsharma-dc: hmm, ok. Modified. Will update diff on PR soon.

control-plane/agents/src/bin/core/volume/operations_helper.rs
dsharma-dc: > don't need the validates here, see my previous review

Fixed. Will update diff on PR soon.

control-plane/agents/src/bin/core/controller/scheduling/resources/mod.rs
dsharma-dc: hmm, that's something. I had to try this toy code to see it in action :)
```
fn main() {
    let str1 = "myid1";
    let str2 = "myid1";
    
    if &str1 == &str2 {
        println!("EQUAL")
    } else {
        println!("NOT EQUAL")
    }
}
```

control-plane/agents/src/bin/core/controller/scheduling/volume_policy/mod.rs
dsharma-dc: You mean modify the  `SimplePolicy:: min_free_space_repl_resize` ? 
Need to write `ReplicaResizePoolsContext` versions of `SimplePolicy:: min_free_space_repl_resize` and `SimplePolicy::overcommit` perhaps, which will be applied after `PoolBaseFilters::min_free_space_repl_resize` ? Existing ones expect `GetSuitablePoolsContext` request.

control-plane/agents/src/bin/core/volume/operations_helper.rs
dsharma-dc: Since this block is now modified to take Replica states, the nodes is directly taken from the state.

control-plane/agents/src/bin/core/controller/scheduling/volume_policy/mod.rs
tiagolobocastro: Alright, then we need to add a new one, can add to min_free_space_repl_resize then; might need some refactor to be able to reuse code

control-plane/agents/src/bin/core/pool/specs.rs
dsharma-dc: While the `ReplicaStatus` isn't `Online`, there could be a possibility that underlying lvol is healthy/reachable and hence resizable? so perhaps let's not block the flow at this point ?

tiagolobocastro: hmm should be very small possibility that the lvol is healthy if the status is not online. This should actually be checked in the scheduling, the replicas should be Online.
Also we should also check if the node is cordoned?

dsharma-dc: > hmm should be very small possibility that the lvol is healthy if the status is not online. This should actually be checked in the scheduling, the replicas should be Online. Also we should also check if the node is cordoned?

@tiagolobocastro Not sure on cordon check. Cordon is to prevent scheduling new resources on a node right? So if a replica is already existing on that node we should let it resize I think.

tiagolobocastro: Yeah, makes sense

control-plane/agents/src/bin/core/controller/scheduling/volume_policy/mod.rs
dsharma-dc: PTAL now

dsharma-dc:
bors try

control-plane/agents/src/bin/core/controller/scheduling/volume_policy/pool.rs
```diff
@@ -13,17 +19,36 @@ impl PoolBaseFilters {
     pub(crate) fn capacity(request: &GetSuitablePoolsContext, item: &PoolItem) -> bool {
         item.pool.capacity > request.size
     }
+    /// Helper util for overcommit checks.
+    fn overcommit_util(allowed_commit_percent: u64, pool: &PoolWrapper, required: u64) -> bool {
```
tiagolobocastro: Why not leave this as overcommit and add the calling helper to the caller, example, to `GetSuitablePoolsContext` and to `ReplicaResizePoolsContext`.

```suggestion
    fn overcommit(allowed_commit_percent: u64, pool: &PoolWrapper, required: u64) -> bool {
```

control-plane/agents/src/bin/core/controller/scheduling/volume_policy/pool.rs
```diff
@@ -34,6 +59,14 @@ impl PoolBaseFilters {
             false => item.pool.free_space() > request.size,
         }
     }
+    /// Return true if the pool has enough capacity to resize the replica by the requested
+    /// value.
+    pub(crate) fn min_free_space_repl_resize(
+        request: &ReplicaResizePoolsContext,
+        item: &ChildItem,
+    ) -> bool {
+        item.pool().free_space() > request.required_capacity()
```
tiagolobocastro: for thin volumes, we can just check for some watermark: existing `min_free_space`. and have resize behave just like create

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -185,10 +185,39 @@ impl ResourceResize for OperationGuardArc<VolumeSpec> {
...
+                VolumeOperation::Resize(request.requested_size),
+            )
+            .await?;
+        // Resize each replica of the volume. If any replica fails to be resized then the
+        // volume resize operation is deemed as a failure.
+        let result = self
+            .resize_volume_replicas(registry, &resizeable_replicas, request.requested_size)
+            .await;
+
+        self.complete_update(registry, result, spec_clone).await?;
```
tiagolobocastro, line 214: Should we undo the replica resize here if we failed?

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -185,10 +185,39 @@ impl ResourceResize for OperationGuardArc<VolumeSpec> {
...
         &mut self,
-        _registry: &Registry,
-        _request: &Self::Resize,
+        registry: &Registry,
+        request: &Self::Resize,
     ) -> Result<Self::ResizeOutput, SvcError> {
-        unimplemented!()
+        let spec = self.as_ref().clone();
+        let state = registry.volume_state(&request.uuid).await?;
+        let spec_repl_ids = registry
```
tiagolobocastro: Why not just pass the replicas themselves?

control-plane/agents/src/bin/core/volume/operations_helper.rs
```diff
@@ -527,6 +527,42 @@ impl OperationGuardArc<VolumeSpec> {
...
+            let repl_rsc = match registry.specs().replica_rsc(&replica.uuid) {
+                Some(rsc) => rsc,
+                None => {
+                    return Err(SvcError::ReplicaNotFound {
+                        replica_id: replica.uuid.clone(),
+                    })
+                }
+            };
+
+            let mut replica_grd = repl_rsc.operation_guard_wait().await?;
```
tiagolobocastro:
```suggestion
            let mut replica_grd = registry.specs().replica(&replica.uuid).await?;
```

control-plane/agents/src/bin/core/volume/specs.rs
```diff
@@ -320,6 +320,53 @@ pub(crate) async fn healthy_volume_replicas(
...
+            id: spec.uuid_str(),
+        });
+    }
+
+    Ok(())
+}
+/// Check if any replica is on a pool that doesn't have sufficient space for
+/// resize operation. If no such replica present, it means the volume is good
+/// to be resized and the returned vector will be of zero length.
+pub(crate) async fn resizeable_replicas(
```
tiagolobocastro, line 338: Can't this added to operations_helper and take self of the guard?

control-plane/agents/src/bin/core/volume/specs.rs
```diff
@@ -320,6 +320,53 @@ pub(crate) async fn healthy_volume_replicas(
...
+    spec: &VolumeSpec,
+    registry: &Registry,
+    spec_repl_ids: Vec<ReplicaId>,
```
tiagolobocastro: Don't think we need this, can just get the volume replicas here, and avoid copying, especially since this is just needed in the error case.

control-plane/agents/src/bin/core/volume/specs.rs
```diff
@@ -1055,7 +1102,12 @@ impl SpecOperationsHelper for VolumeSpec {
             }
             VolumeOperation::CreateSnapshot(_) => Ok(()),
             VolumeOperation::DestroySnapshot(_) => Ok(()),
-            VolumeOperation::Resize(_) => todo!(),
+            VolumeOperation::Resize(_) => {
+                // Do Pre-checks for volume resize.
+                vol_status_ok_for_resize(self)?;
```
tiagolobocastro: Not that we have the scheduling taking place outside here, probably can move this to the resize call?

control-plane/agents/src/common/errors.rs
```diff
@@ -814,6 +834,18 @@ impl From<SvcError> for ReplyError {
...
                 extra,
             },
+            SvcError::ResizeReplError { .. } => ReplyError {
+                kind: ReplyErrorKind::FailedPrecondition,
+                resource: ResourceKind::Replica,
+                source,
+                extra,
+            },
+            SvcError::VolumeResizeArgsUnsupported { .. } => ReplyError {
+                kind: ReplyErrorKind::FailedPrecondition,
```
tiagolobocastro: If it's unsupported, should this be InvalidArg?

tiagolobocastro (approved):

control-plane/agents/src/bin/core/volume/operations.rs
dsharma-dc: removed this parameter altogether and fetching spec_replicas inside the function.

control-plane/agents/src/bin/core/volume/operations.rs
dsharma-dc: I'm planning to handle these cases as part of space reclaim and reconciler task.

control-plane/agents/src/bin/core/controller/scheduling/volume_policy/pool.rs
dsharma-dc: added as method in respective structs.

control-plane/agents/src/bin/core/volume/specs.rs
dsharma-dc: This seems consistent to other existing helpers that trigger scheduling from this file.

tiagolobocastro: Ah I'm not sure all the other ones belong here either tbh.
Let's keep then and I'll have a look later and move if necessary.

dsharma-dc:
bors merge

dsharma-dc:
hmm, compilation failure in CI, let's retry.
> error: could not compile `io-engine-tests` (test "reservations") due to previous error

>warning: build failed, waiting for other jobs to finish...
>error: could not compile `agents` (bin "core")

dsharma-dc:
bors merge

PR #707, opened 2024-01-04
feat(multi-replica-snapshot): Handle multireplica snapshot in control-plane

hrudaya21:
1. Add multi-replica snapshot capability in control-plane.
2. Limit snapshot creation based on new environment variable. (max_snapshots)

### DIFF
--- a/control-plane/agents/src/bin/core/volume/operations.rs
+++ b/control-plane/agents/src/bin/core/volume/operations.rs
@@ -811,14 +811,14 @@ impl ResourceLifecycleExt<CreateVolumeSource<'_>> for OperationGuardArc<VolumeSp
         request_src: &CreateVolumeSource,
     ) -> Result<Self::CreateOutput, SvcError> {
         request_src.pre_flight_check()?;
-        let request = request_src.source();
-
         let specs = registry.specs();
         let mut volume = specs
             .get_or_create_volume(request_src)?
             .operation_guard_wait()
             .await?;
-        let volume_clone = volume.start_create_update(registry, request).await?;
+        let volume_clone = volume
+            .start_create_update(registry, request_src.source())
+            .await?;
 
         // If the volume is a part of the ag, create or update accordingly.
         registry.specs().get_or_create_affinity_group(&volume_clone);
--- a/control-plane/agents/src/bin/core/volume/service.rs
+++ b/control-plane/agents/src/bin/core/volume/service.rs
@@ -291,7 +291,6 @@ impl Service {
             }
             filter => return Err(SvcError::InvalidFilter { filter }),
         };
-
         Ok(Volumes {
             entries: filtered_volumes,
             next_token: match last_result {
@@ -403,7 +402,6 @@ impl Service {
         volume.set_replica(&self.registry, request).await?;
         self.registry.volume(&request.uuid).await
     }
-
     /// Create a volume snapshot.
     #[tracing::instrument(level = "info", skip(self), err, fields(volume.uuid = %request.source_id, snapshot.source_uuid = %request.source_id, snapshot.uuid = %request.snap_id))]
     async fn create_snapshot(
@@ -426,6 +424,16 @@ impl Service {
             }
             Err(error) => Err(error),
         }?;
+
+        if let Some(max_snapshots) = volume.as_ref().max_snapshots {
+            if volume.as_ref().metadata.num_snapshots() as u32 >= max_snapshots {
+                return Err(SvcError::SnapshotMaxLimit {
+                    max_snapshots,
+                    volume_id: volume.as_ref().uuid.to_string(),
+                });
+            }
+        }
+
         let snapshot = volume
             .create_snap(
                 &self.registry,
--- a/control-plane/agents/src/bin/core/volume/snapshot_helpers.rs
+++ b/control-plane/agents/src/bin/core/volume/snapshot_helpers.rs
@@ -25,11 +25,13 @@ use stor_port::{
     },
 };
 
+use std::collections::HashSet;
+
 /// A request type for creating snapshot of a volume, which essentially
 /// means a snapshot of all(or selected) healthy replicas associated with that volume.
 pub(super) struct PrepareVolumeSnapshot {
     pub(super) parameters: SnapshotParameters<VolumeId>,
-    pub(super) replica_snapshot: (Replica, ReplicaSnapshot),
+    pub(super) replica_snapshot: Vec<(Replica, ReplicaSnapshot)>,
     pub(super) completer: VolumeSnapshotCompleter,
 }
 
@@ -95,34 +97,42 @@ impl SpecOperationsHelper for VolumeSnapshot {
 pub(crate) async fn snapshoteable_replica(
     volume: &VolumeSpec,
     registry: &Registry,
-) -> Result<ChildItem, SvcError> {
-    if volume.num_replicas != 1 {
+) -> Result<Vec<ChildItem>, SvcError> {
+    let children = super::scheduling::snapshoteable_replica(volume, registry).await?;
+
+    if children.candidates().len() != volume.num_replicas as usize {
+        return Err(SvcError::InsufficientHealthyReplicas {
+            id: volume.uuid_str(),
+        });
+    }
+
+    //todo: Remove this check once we support snapshotting with n-replicas.
+    if volume.num_replicas != 1 || children.candidates().len() != 1 {
         return Err(SvcError::NReplSnapshotNotAllowed {});
     }
 
-    let children = super::scheduling::snapshoteable_replica(volume, registry).await?;
-
     volume.trace(&format!("Snapshoteable replicas for volume: {children:?}"));
 
-    let item = match children.candidates().as_slice() {
-        [item] => Ok(item),
-        [] => Err(SvcError::NoHealthyReplicas {
+    if children.candidates().is_empty() {
+        return Err(SvcError::NoHealthyReplicas {
             id: volume.uuid_str(),
-        }),
-        _ => Err(SvcError::NReplSnapshotNotAllowed {}),
-    }?;
-
-    let pools = SnapshotVolumeReplica::builder_with_defaults(registry, volume, item)
-        .await
-        .collect();
-
-    match pools
-        .iter()
-        .any(|pool_item| pool_item.pool.id == item.pool().id)
-    {
-        true => Ok(item.clone()),
-        false => Err(SvcError::NotEnoughResources {
-            source: NotEnough::PoolFree {},
-        }),
+        });
     }
+
+    //todo: check for snapshot chain for all the replicas.
+
+    let pools =
+        SnapshotVolumeReplica::builder_with_defaults(registry, volume, children.candidates())
+            .await
+            .collect();
+    let pools: HashSet<_> = pools.iter().map(|item| item.pool.id.clone()).collect();
+
+    for item in children.candidates() {
+        if !pools.contains(&item.pool().id) {
+            return Err(SvcError::NotEnoughResources {
+                source: NotEnough::PoolFree {},
+            });
+        }
+    }
+    Ok(children.candidates().clone())
 }
--- a/control-plane/agents/src/bin/core/volume/snapshot_operations.rs
+++ b/control-plane/agents/src/bin/core/volume/snapshot_operations.rs
@@ -38,7 +38,10 @@ use stor_port::{
 };
 
 use chrono::{DateTime, Utc};
-use std::collections::{HashMap, HashSet};
+use std::{
+    collections::{HashMap, HashSet},
+    time::SystemTime,
+};
 
 #[async_trait::async_trait]
 impl ResourceSnapshotting for OperationGuardArc<VolumeSpec> {
@@ -66,7 +69,6 @@ impl ResourceSnapshotting for OperationGuardArc<VolumeSpec> {
             },
         )
         .await;
-
         self.complete_update(registry, snap_result, spec_clone)
             .await
     }
@@ -155,13 +157,13 @@ impl ResourceLifecycleWithLifetime for OperationGuardArc<VolumeSnapshot> {
         let volume = &request.volume;
         let request = &request.request;
 
-        let replica = snapshoteable_replica(volume.as_ref(), registry).await?;
+        let replicas = snapshoteable_replica(volume.as_ref(), registry).await?;
         let target_node = if let Some(target) = volume.as_ref().target() {
-            registry.node_wrapper(target.node()).await
+            let node = registry.node_wrapper(target.node()).await?;
+            Some(node)
         } else {
-            registry.node_wrapper(&replica.state().node).await
-        }?;
-
+            None
+        };
         let specs = registry.specs();
         let mut snapshot = specs
             .get_or_create_snapshot(request)
@@ -179,13 +181,17 @@ impl ResourceLifecycleWithLifetime for OperationGuardArc<VolumeSnapshot> {
         // Try to prune 1 stale transaction, if present..
         snapshot.prune(registry, Some(1)).await.ok();
 
-        let prepare_snapshot = snapshot.snapshot_params(&replica)?;
+        let prepare_snapshot = snapshot.snapshot_params(&replicas)?;
         snapshot
             .start_create_update(
                 registry,
                 &VolumeSnapshotCreateInfo::new(
                     prepare_snapshot.parameters.txn_id(),
-                    prepare_snapshot.replica_snapshot.1.clone(),
+                    prepare_snapshot
+                        .replica_snapshot
+                        .iter()
+                        .map(|(_, snapshot)| snapshot.clone())
+                        .collect(),
                     &prepare_snapshot.completer,
                 ),
             )
@@ -285,31 +291,39 @@ impl ResourcePruning for OperationGuardArc<VolumeSnapshot> {
 }
 
 impl OperationGuardArc<VolumeSnapshot> {
-    fn snapshot_params(&self, replica: &ChildItem) -> Result<PrepareVolumeSnapshot, SvcError> {
+    fn snapshot_params(
+        &self,
+        replicas: &Vec<ChildItem>,
+    ) -> Result<PrepareVolumeSnapshot, SvcError> {
         let Some(parameters) = self.as_ref().prepare() else {
             return Err(SvcError::AlreadyExists {
                 id: self.uuid().to_string(),
                 kind: ResourceKind::VolumeSnapshot,
             });
         };
+        let mut replica_snapshots = Vec::with_capacity(replicas.len());
         let volume = self.as_ref().spec().source_id();
         let generic_params = parameters.params().clone();
-        let snapshot_source = ReplicaSnapshotSource::new(
-            replica.spec().uid().clone(),
-            replica.state().pool_id.clone(),
-            replica.state().pool_uuid.clone().unwrap_or_default(),
-        );
-        let replica_snapshot = ReplicaSnapshot::new_vol(
-            ReplicaSnapshotSpec::new(&snapshot_source, SnapshotId::new()),
-            SnapshotParameters::new(volume, generic_params),
-            replica.state().size,
-            0,
-            replica.spec().size,
-        );
-        let replica = replica.state().clone();
+        for replica in replicas {
+            let snapshot_source = ReplicaSnapshotSource::new(
+                replica.spec().uid().clone(),
+                replica.state().pool_id.clone(),
+                replica.state().pool_uuid.clone().unwrap_or_default(),
+            );
+            let replica_snapshot = ReplicaSnapshot::new_vol(
+                ReplicaSnapshotSpec::new(&snapshot_source, SnapshotId::new()),
+                SnapshotParameters::new(volume, generic_params.clone()),
+                replica.state().size,
+                0,
+                replica.spec().size,
+            );
+            let replica = replica.state().clone();
+            replica_snapshots.push((replica, replica_snapshot));
+        }
+
         Ok(PrepareVolumeSnapshot {
             parameters,
-            replica_snapshot: (replica, replica_snapshot),
+            replica_snapshot: replica_snapshots,
             completer: VolumeSnapshotCompleter::default(),
         })
     }
@@ -318,13 +332,14 @@ impl OperationGuardArc<VolumeSnapshot> {
         volume: &OperationGuardArc<VolumeSpec>,
         prep_params: &PrepareVolumeSnapshot,
         registry: &Registry,
-        target_node: N,
+        target_node: Option<N>,
     ) -> Result<VolumeSnapshotCreateResult, SvcError> {
-        if let Some(target) = volume.as_ref().target() {
-            self.snapshot_nexus(prep_params, target, registry, target_node)
+        let target = volume.as_ref().target();
+        if target.is_some() && target_node.is_some() {
+            self.snapshot_nexus::<N>(prep_params, target.unwrap(), registry, target_node.unwrap())
                 .await
         } else {
-            self.snapshot_replica(prep_params, target_node).await
+            self.snapshot_replica::<N>(prep_params, registry).await
         }
     }
 
@@ -335,88 +350,106 @@ impl OperationGuardArc<VolumeSnapshot> {
         registry: &Registry,
         target_node: N,
     ) -> Result<VolumeSnapshotCreateResult, SvcError> {
-        let mut replica_snap = prep_params.replica_snapshot.1.clone();
-        let replica = &prep_params.replica_snapshot.0;
         let generic_params = prep_params.parameters.params();
-
-        let replica_id = replica_snap.spec().source_id().replica_id();
+        let nexus_snap_desc = prep_params
+            .replica_snapshot
+            .iter()
+            .map(|(_, snapshot)| {
+                CreateNexusSnapReplDescr::new(
+                    snapshot.spec().source_id().replica_id(),
+                    snapshot.spec().uuid().clone(),
+                )
+            })
+            .collect::<Vec<_>>();
         let response = target_node
             .create_nexus_snapshot(&CreateNexusSnapshot::new(
                 SnapshotParameters::new(target.nexus(), generic_params.clone()),
-                vec![CreateNexusSnapReplDescr::new(
-                    replica_id,
-                    replica_snap.spec().uuid().clone(),
-                )],
+                nexus_snap_desc,
             ))
             .await?;
 
-        if response.skipped.contains(replica_id) || !response.skipped.is_empty() {
+        if !response.skipped.is_empty() {
             return Err(SvcError::ReplicaSnapSkipped {
-                replica: replica_snap.spec().uuid().to_string(),
+                replica: response
+                    .skipped
+                    .iter()
+                    .map(|r| r.to_string())
+                    .collect::<Vec<String>>()
+                    .join(", "),
             });
         }
-
-        let snapped = match response.replicas_status.as_slice() {
-            [snapped] if &snapped.replica_uuid == replica_snap.spec().source_id().replica_id() => {
-                Ok(snapped)
-            }
-            _ => Err(SvcError::ReplicaSnapMiss {
-                replica: replica_snap.spec().uuid().to_string(),
-            }),
-        }?;
-
-        if let Some(error) = snapped.error {
+        let failed_replicas = response
+            .replicas_status
+            .iter()
+            .filter(|&snap| snap.error.is_some())
+            .collect::<Vec<_>>();
+        if !failed_replicas.is_empty() {
             return Err(SvcError::ReplicaSnapError {
-                replica: replica_snap.spec().uuid().to_string(),
-                error,
+                failed_replicas: failed_replicas
+                    .iter()
+                    .map(|&snap| (snap.replica_uuid.to_string(), snap.error.unwrap()))
+                    .collect::<Vec<_>>(),
             });
         }
-
         let timestamp = DateTime::<Utc>::from(response.snap_time);
         // What if snapshot succeeds but we can't fetch the replica snapshot, should we carry
         // on as following, or should we bail out?
-        let node = registry.node_wrapper(&replica.node).await?;
+        let mut replica_snapshots = prep_params.replica_snapshot.clone();
+        for (replica, replica_snap) in replica_snapshots.iter_mut() {
+            let node = registry.node_wrapper(&replica.node).await?;
+            let snapshot = NodeWrapper::fetch_update_snapshot_state(
+                &node,
+                ReplicaSnapshotInfo::new(
+                    replica_snap.spec().source_id().replica_id(),
+                    replica_snap.spec().uuid().clone(),
+                ),
+            )
+            .await?;
 
-        let snapshot = NodeWrapper::fetch_update_snapshot_state(
-            &node,
-            ReplicaSnapshotInfo::new(
-                replica_snap.spec().source_id().replica_id(),
-                replica_snap.spec().uuid().clone(),
-            ),
-        )
-        .await?;
-
-        replica_snap.complete_vol(
-            snapshot.timestamp().into(),
-            snapshot.replica_size(),
-            snapshot.allocated_size() + snapshot.predecessor_alloc_size(),
-        );
-        Ok(VolumeSnapshotCreateResult::new_ok(replica_snap, timestamp))
+            replica_snap.complete_vol(
+                snapshot.timestamp().into(),
+                snapshot.replica_size(),
+                snapshot.allocated_size() + snapshot.predecessor_alloc_size(),
+            );
+        }
+        let snapshots = replica_snapshots
+            .iter()
+            .map(|(_, replica_snapshot)| replica_snapshot.clone())
+            .collect::<Vec<_>>();
+        Ok(VolumeSnapshotCreateResult::new_ok(snapshots, timestamp))
     }
+
     async fn snapshot_replica<N: ReplicaSnapshotApi>(
         &self,
         prep_params: &PrepareVolumeSnapshot,
-        target_node: N,
+        registry: &Registry,
     ) -> Result<VolumeSnapshotCreateResult, SvcError> {
-        let mut replica_snap = prep_params.replica_snapshot.clone();
         let volume_params = prep_params.parameters.params().clone();
+        let mut timestamp = SystemTime::now();
 
-        let replica_params = volume_params.with_uuid(replica_snap.1.spec().uuid());
-        let response = target_node
-            .create_repl_snapshot(&CreateReplicaSnapshot::new(SnapshotParameters::new(
-                replica_snap.1.spec().source_id().replica_id(),
-                replica_params,
-            )))
-            .await?;
-        let timestamp = response.timestamp();
-        replica_snap.1.complete_vol(
-            timestamp.into(),
-            response.replica_size(),
-            response.allocated_size() + response.predecessor_alloc_size(),
-        );
+        let mut replica_snapshots = prep_params.replica_snapshot.clone();
+        for (replica, replica_snap) in replica_snapshots.iter_mut() {
+            let replica_params = volume_params.clone().with_uuid(replica_snap.spec().uuid());
+            let target_node = registry.node_wrapper(&replica.node).await?;
+            let response = target_node
+                .create_repl_snapshot(&CreateReplicaSnapshot::new(SnapshotParameters::new(
+                    replica_snap.spec().source_id().replica_id(),
+                    replica_params,
+                )))
+                .await?;
+            timestamp = response.timestamp();
+            replica_snap.complete_vol(
+                timestamp.into(),
+                response.replica_size(),
+                response.allocated_size() + response.predecessor_alloc_size(),
+            );
+        }
 
         Ok(VolumeSnapshotCreateResult::new_ok(
-            replica_snap.1,
+            replica_snapshots
+                .iter()
+                .map(|(_, replica_snapshot)| replica_snapshot.clone())
+                .collect::<Vec<_>>(),
             timestamp.into(),
         ))
     }
--- a/control-plane/agents/src/common/errors.rs
+++ b/control-plane/agents/src/common/errors.rs
@@ -339,10 +339,9 @@ pub enum SvcError {
     ReplicaSnapSkipped { replica: String },
     #[snafu(display("Replica's {} snapshot was unexpectedly not taken", replica))]
     ReplicaSnapMiss { replica: String },
-    #[snafu(display("Replica's {} snapshot failed with error {}", replica, error))]
+    #[snafu(display("Failed to snapshot replicas {:?}", failed_replicas))]
     ReplicaSnapError {
-        replica: String,
-        error: nix::errno::Errno,
+        failed_replicas: Vec<(String, nix::errno::Errno)>,
     },
     #[snafu(display("The service is busy, cannot process request"))]
     ServiceBusy {},
@@ -376,6 +375,18 @@ pub enum SvcError {
         cluster_capacity_limit: u64,
         excess: u64,
     },
+    #[snafu(display(
+        "The number of healthy replicas does not match the expected replica count of volume '{id}'"
+    ))]
+    InsufficientHealthyReplicas { id: String },
+    #[snafu(display(
+        "Reached maximum snapshots limit {max_snapshots} for volume {volume_id}, delete unused snapshots to continue"))]
+    SnapshotMaxLimit {
+        max_snapshots: u32,
+        volume_id: String,
+    },
+    #[snafu(display("Invalid property name '{property_name}' for the volume '{id}'"))]
+    InvalidSetProperty { property_name: String, id: String },
 }
 
 impl SvcError {
@@ -1014,6 +1025,24 @@ impl From<SvcError> for ReplyError {
             },
             SvcError::CapacityLimitExceeded { .. } => ReplyError {
                 kind: ReplyErrorKind::CapacityLimitExceeded,
+                resource: ResourceKind::Volume,
+                source,
+                extra,
+            },
+            SvcError::InsufficientHealthyReplicas { .. } => ReplyError {
+                kind: ReplyErrorKind::FailedPrecondition,
+                resource: ResourceKind::VolumeSnapshot,
+                source,
+                extra,
+            },
+            SvcError::SnapshotMaxLimit { .. } => ReplyError {
+                kind: ReplyErrorKind::OutOfRange,
+                resource: ResourceKind::Volume,
+                source,
+                extra,
+            },
+            SvcError::InvalidSetProperty { .. } => ReplyError {
+                kind: ReplyErrorKind::InvalidArgument,
                 resource: ResourceKind::Volume,
                 source,
                 extra,
--- a/control-plane/csi-driver/src/context.rs
+++ b/control-plane/csi-driver/src/context.rs
@@ -57,6 +57,8 @@ pub enum Parameters {
     PoolTopologyAffinity,
     #[strum(serialize = "poolTopologySpread")]
     PoolTopologySpread,
+    #[strum(serialize = "maxSnapshots")]
+    MaxSnapshots,
 }
 impl Parameters {
     fn parse_human_time(
@@ -160,6 +162,10 @@ impl Parameters {
         value: Option<&String>,
     ) -> Result<Option<HashMap<String, String>>, serde_json::Error> {
         Self::parse_map(value)
+    }
+    /// Parse the value for `Self::MaxSnapshots`.
+    pub fn max_snapshots(value: Option<&String>) -> Result<Option<u32>, ParseIntError> {
+        Self::parse_u32(value)
     }
 }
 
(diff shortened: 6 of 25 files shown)


control-plane/agents/src/bin/core/controller/registry.rs
```diff
@@ -100,6 +100,8 @@ pub(crate) struct RegistryInner<S: Store> {
     thin_args: ThinArgs,
     /// Check if the HA feature is enabled.
     ha_disabled: bool,
+    /// Max snapshots limit per volume.
```
Abhinandan-Purkait: Should we call it `Global max snapshot limit per volume` just to be clear that we fall back to this if not passed via params?

control-plane/agents/src/bin/core/controller/registry.rs
```diff
@@ -280,7 +284,10 @@ impl Registry {
     pub(crate) fn create_volume_limit(&self) -> usize {
         self.create_volume_limit
     }
-
+    /// Max snapshots limit per volume.
```
Abhinandan-Purkait: same doc comment as above?

control-plane/grpc/src/operations/volume/traits.rs
```diff
@@ -1656,6 +1675,27 @@ impl SetVolumeReplicaInfo for SetVolumeReplica {
...
+impl SetVolumePropInfo for SetVolumeProp {
+    fn uuid(&self) -> VolumeId {
+        self.uuid.clone()
+    }
+    fn prop_name(&self) -> String {
+        self.prop_name.clone()
+    }
+    fn prop_value(&self) -> String {
+        self.prop_value.clone()
+    }
```
Abhinandan-Purkait: As discussed probably should it be a map of params? Maybe for later

control-plane/agents/src/bin/core/controller/registry.rs
```diff
@@ -100,6 +100,8 @@ pub(crate) struct RegistryInner<S: Store> {
     thin_args: ThinArgs,
     /// Check if the HA feature is enabled.
     ha_disabled: bool,
+    /// Max snapshots limit per volume.
+    max_snapshots: i32,
```
tiagolobocastro: Something like this or similar

```suggestion
    max_snapshots_volume: i32,
```

control-plane/agents/src/bin/core/controller/resources/operations.rs
```diff
@@ -159,6 +159,19 @@ pub(crate) trait ResourceReplicas {
     ) -> Result<Self::MoveResp, SvcError>;
 }
 
+/// Volume resource operation.
+#[async_trait::async_trait]
+pub(crate) trait ResourceVolume {
```
tiagolobocastro:
```suggestion
pub(crate) trait ResourceProperties {
```

control-plane/agents/src/bin/core/controller/resources/operations.rs
```diff
@@ -159,6 +159,19 @@ pub(crate) trait ResourceReplicas {
     ) -> Result<Self::MoveResp, SvcError>;
 }
 
+/// Volume resource operation.
```
tiagolobocastro:
```suggestion
/// Property modification as resource operation.
```

control-plane/agents/src/bin/core/main.rs
```diff
@@ -108,6 +108,9 @@ pub(crate) struct CliArgs {
     /// This is useful when the frontend nodes do not support the NVMe ANA feature.
     #[clap(long, env = "HA_DISABLED")]
     pub(crate) disable_ha: bool,
+    /// Maximum snapshots allowed per volume.
+    #[clap(long, short, default_value = "1")]
+    pub(crate) max_snapshots_arg: i32,
```
tiagolobocastro:
```suggestion
    pub(crate) max_snapshots_volume: i32,
```

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -811,14 +836,16 @@ impl ResourceLifecycleExt<CreateVolumeSource<'_>> for OperationGuardArc<VolumeSp
         request_src: &CreateVolumeSource,
     ) -> Result<Self::CreateOutput, SvcError> {
         request_src.pre_flight_check()?;
-        let request = request_src.source();
-
+        let mut request = request_src.source().clone();
+        if request.max_snapshots().is_none() {
+            request.max_snapshots = Some(registry.max_snapshots());
```
tiagolobocastro: Would be good to set what the rules are here, seems like per-request overrides the global?

control-plane/agents/src/bin/core/volume/service.rs
```diff
@@ -426,6 +447,21 @@ impl Service {
...
         }?;
+
+        let max_snapshots = if let Some(max_limit) = volume.as_ref().max_snapshots {
+            max_limit
+        } else {
+            self.registry.max_snapshots()
+        };
+        if max_snapshots == 0
+            || (max_snapshots > 0
+                && volume.as_ref().metadata.num_snapshots() as i32 >= max_snapshots)
```
tiagolobocastro: Doesn't this cover it?

```suggestion
        if volume.as_ref().metadata.num_snapshots() as i32 >= max_snapshots)
```

control-plane/agents/src/bin/core/volume/snapshot_helpers.rs
```diff
@@ -95,34 +95,32 @@ impl SpecOperationsHelper for VolumeSnapshot {
...
-        [] => Err(SvcError::NoHealthyReplicas {
-            id: volume.uuid_str(),
-        }),
-        _ => Err(SvcError::NReplSnapshotNotAllowed {}),
-    }?;
+    if children.candidates().len() != volume.num_replicas as usize
+        || children.candidates().is_empty()
+    {
+        return Err(SvcError::AllReplicaNotHealthy {});
+    }
```
tiagolobocastro: Separate the errors out

```suggestion
    if children.candidates().is_empty() {
        return Err(SvcError::NoHealthyReplicas { id: volume.uuid_str() });
    }
    if children.candidates().len() != volume.num_replicas as usize {
        return Err(SvcError::AllReplicaNotHealthy { id: volume.uuid_str() });
    }
```

control-plane/agents/src/bin/core/volume/snapshot_operations.rs
```diff
@@ -155,11 +157,11 @@ impl ResourceLifecycleWithLifetime for OperationGuardArc<VolumeSnapshot> {
         let volume = &request.volume;
         let request = &request.request;
 
-        let replica = snapshoteable_replica(volume.as_ref(), registry).await?;
+        let replicas = snapshoteable_replica(volume.as_ref(), registry).await?;
         let target_node = if let Some(target) = volume.as_ref().target() {
             registry.node_wrapper(target.node()).await
         } else {
-            registry.node_wrapper(&replica.state().node).await
+            registry.node_wrapper(&replicas[0].state().node).await
```
tiagolobocastro: This probably doesn't make sense now, since we don't have a single replica node, but many

control-plane/csi-driver/src/context.rs
```diff
@@ -116,6 +123,10 @@ impl Parameters {
     pub fn fs_id(value: Option<&String>) -> Result<Option<Uuid>, UuidError> {
         Self::parse_uuid(value)
     }
+    /// Parse the value for `Self::MaxSnapshots`.
+    pub fn max_snapshots(value: Option<&String>) -> Result<Option<i32>, ParseIntError> {
+        Self::parse_i32(value)
```
tiagolobocastro: Why a signed value?

control-plane/grpc/proto/v1/volume/volume.proto
```diff
@@ -61,6 +61,8 @@ message VolumeSpec {
   optional VolumeContentSource content_source  = 11;
   // Number of snapshots taken on this volume.
   uint32 num_snapshots = 12;
+  // Max snapshots limit per volume.
+  optional int32 max_snapshots = 13;
```
tiagolobocastro: Why a signed value?

```suggestion
  optional uint32 max_snapshots = 13;
```

control-plane/grpc/proto/v1/volume/volume.proto
```diff
@@ -342,6 +346,16 @@ message SetVolumeReplicaRequest {
   // replica count
   uint32 replicas = 2;
 }
+// Set the volume property
+message SetVolumePropRequest {
+  // uuid of the volume
+  google.protobuf.StringValue uuid = 1;
```
```suggestion
  string uuid = 1;
```

control-plane/grpc/proto/v1/volume/volume.proto
```diff
@@ -342,6 +346,16 @@ message SetVolumeReplicaRequest {
   // replica count
   uint32 replicas = 2;
 }
+// Set the volume property
+message SetVolumePropRequest {
+  // uuid of the volume
```
tiagolobocastro: Capitalize these

control-plane/grpc/proto/v1/volume/volume.proto
```diff
@@ -591,6 +613,7 @@ service VolumeGrpc {
   rpc ShareVolume (ShareVolumeRequest) returns (ShareVolumeReply) {}
   rpc UnshareVolume (UnshareVolumeRequest) returns (UnshareVolumeReply) {}
   rpc SetVolumeReplica (SetVolumeReplicaRequest) returns (SetVolumeReplicaReply) {}
+  rpc SetVolumeProperty(SetVolumePropRequest) returns (SetVolumePropReply) {}
```
```suggestion
  rpc SetVolumeProperty (SetVolumePropRequest) returns (SetVolumePropReply) {}
```

control-plane/grpc/src/operations/volume/traits.rs
```diff
@@ -95,6 +95,12 @@ pub trait VolumeOperations: Send + Sync {
         req: &dyn SetVolumeReplicaInfo,
         ctx: Option<Context>,
     ) -> Result<Volume, ReplyError>;
+    /// Set volume prop
```
tiagolobocastro:
```suggestion
    /// Set volume property.
```

control-plane/grpc/src/operations/volume/traits.rs
```diff
@@ -1656,6 +1675,27 @@ impl SetVolumeReplicaInfo for SetVolumeReplica {
     }
 }
 
+/// Trait to be implemented for SetVolumeProp operation.
+pub trait SetVolumePropInfo: Send + Sync + std::fmt::Debug {
+    /// Uuid of the concerned volume
```
tiagolobocastro: End all with `.`

control-plane/grpc/src/operations/volume/traits.rs
tiagolobocastro: hmm, you mean like set properties rather than set property?

control-plane/plugin/src/bin/rest-plugin/main.rs
```diff
@@ -122,6 +122,16 @@ async fn execute(cli_args: CliArgs) {
...
         },
+        Operations::Set(resource) => {
+            match resource {
+                SetResources::Volume {
+                    id,
+                    prop_name,
+                    prop_value,
+                } => volume::Volume::set(id, prop_name, prop_value, &cli_args.output).await,
+            };
+            Ok(())
```
tiagolobocastro:
```suggestion
            }
```

control-plane/plugin/src/operations.rs
```diff
@@ -70,6 +74,13 @@ pub trait Scale {
     async fn scale(id: &Self::ID, replica_count: u8, output: &utils::OutputFormat) -> PluginResult;
 }
 
+/// Set trait.
+/// To be implemented by resources which support the 'set' operation.
+#[async_trait(?Send)]
+pub trait Set {
+    type ID;
+    async fn set(id: &Self::ID, prop_name: &str, prop_value: &str, output: &utils::OutputFormat);
```
tiagolobocastro: This should return an error..

control-plane/plugin/src/resources/utils.rs
```diff
@@ -22,7 +22,8 @@ lazy_static! {
         "THIN-PROVISIONED",
         "ALLOCATED",
         "SNAPSHOTS",
-        "SOURCE"
+        "SOURCE",
+        "MAX_SNAPSHOTS"
```
tiagolobocastro: Hmm is this really worth showing? I think we can do without to avoid cluttering the already long line

```suggestion
```

control-plane/plugin/src/resources/volume.rs
```diff
@@ -192,6 +193,32 @@ impl Scale for Volume {
...
+                    // Print json or yaml based on output format.
+                    utils::print_table(output, volume.into_body());
+                }
+                OutputFormat::None => {
+                    // In case the output format is not specified, show a success message.
+                    println!("Volume {id} max_snapshots set to {prop_value} successfully 🚀")
+                }
+            },
+            Err(e) => {
+                println!("Failed to scale volume {id}. Error {e}")
```
tiagolobocastro: should return an error now

control-plane/rest/openapi-specs/v0_api_spec.yaml
```diff
@@ -1463,6 +1463,43 @@ paths:
           $ref: '#/components/responses/ServerError'
       security:
         - JWT: []
+  '/volumes/{volume_id}/prop_name/{prop_name}/prop_value/{prop_value}':
```
```suggestion
  '/volumes/{volume_id}/properties/{property_name}/{property_value}':
```

control-plane/rest/openapi-specs/v0_api_spec.yaml
```diff
@@ -1463,6 +1463,43 @@ paths:
...
+          schema:
+            type: string
+        - in: path
+          name: prop_value
+          required: true
+          schema:
+            type: string
+            format: uint8
+            minimum: 0
+            maximum: 255
```
tiagolobocastro: If we want generic properties then I don't think we should be constraining the format here
```suggestion
```

control-plane/rest/openapi-specs/v0_api_spec.yaml
```diff
@@ -2334,6 +2371,7 @@ components:
         thin: false
         topology: null
         affinity_group: null
+        max_snapshots: -1
```
tiagolobocastro: I see now why the negative, though not too fond of doing it this way tbh..
Why not have null or be unlimited snapshots?

control-plane/rest/openapi-specs/v0_api_spec.yaml
```diff
@@ -3561,6 +3611,11 @@ components:
           type: integer
           format: int32
           minimum: 0
+        num_snapshot_replicas:
+          description: Number of snapshot replicas for a volumesnapshot.
```
tiagolobocastro, line 3770: What does this mean?

control-plane/stor-port/src/types/v0/store/volume.rs
```diff
@@ -18,7 +18,20 @@ use crate::{
...
 use std::collections::HashMap;
+use strum_macros::{EnumCount as EnumCountMacro, EnumIter};
+/// Volume properties.
+#[derive(Debug, EnumCountMacro, EnumIter)]
+pub enum VolumeAttr {
+    MaxSnapshots,
+}
 
+impl VolumeAttr {
+    pub fn name(&self) -> &'static str {
```
tiagolobocastro: missing doc comment

control-plane/stor-port/src/types/v0/store/volume.rs
```diff
@@ -18,7 +18,20 @@ use crate::{
 use pstor::ApiVersion;
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
+use strum_macros::{EnumCount as EnumCountMacro, EnumIter};
+/// Volume properties.
+#[derive(Debug, EnumCountMacro, EnumIter)]
+pub enum VolumeAttr {
+    MaxSnapshots,
```
tiagolobocastro, line 30: doc comment this too

control-plane/stor-port/src/types/v0/store/volume.rs
```diff
@@ -505,6 +521,14 @@ impl SpecTransaction<VolumeOperation> for VolumeSpec {
                 VolumeOperation::Resize(size) => {
                     self.size = size;
                 }
+                VolumeOperation::SetVolumeProperty(prop) => {
+                    let prop_name = prop.prop_name();
+                    if prop_name == VolumeAttr::MaxSnapshots.name() {
```
tiagolobocastro: Why not have the derivation with the value already set?

control-plane/stor-port/src/types/v0/store/volume.rs
```diff
@@ -630,6 +655,29 @@ impl PublishOperation {
     }
 }
 
+/// SetVolumeProperty Operation parameters.
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+pub struct VolumeProperty {
```
tiagolobocastro: I think this can be an enum with the explicit properties we want?

control-plane/agents/src/bin/core/volume/snapshot_operations.rs
```diff
@@ -335,88 +348,97 @@ impl OperationGuardArc<VolumeSnapshot> {
...
-                vec![CreateNexusSnapReplDescr::new(
-                    replica_id,
-                    replica_snap.spec().uuid().clone(),
-                )],
+                nexus_snap_desc,
             ))
             .await?;
 
-        if response.skipped.contains(replica_id) || !response.skipped.is_empty() {
+        if !response.skipped.is_empty() {
```
tiagolobocastro, line 371: Could we just check if skipped contains the replicas which we tried to snap?

control-plane/agents/src/bin/core/volume/snapshot_operations.rs
```diff
@@ -335,88 +348,97 @@ impl OperationGuardArc<VolumeSnapshot> {
...
-        )
-        .await?;
-
-        replica_snap.complete_vol(
-            snapshot.timestamp().into(),
-            snapshot.replica_size(),
-            snapshot.allocated_size() + snapshot.predecessor_alloc_size(),
-        );
-        Ok(VolumeSnapshotCreateResult::new_ok(replica_snap, timestamp))
+        for (replica, replica_snap) in replicas.iter().zip(replica_snaps.iter_mut()) {
```
tiagolobocastro: how do you ensure replica_snaps is associated to the equivalent replica?

control-plane/agents/src/bin/core/volume/snapshot_operations.rs
```diff
@@ -285,31 +287,42 @@ impl ResourcePruning for OperationGuardArc<VolumeSnapshot> {
...
+            );
+            replica_snapshots.push(replica_snapshot);
+            let replica = replica.state().clone();
+            repica_states.push(replica);
+        }
+
         Ok(PrepareVolumeSnapshot {
             parameters,
-            replica_snapshot: (replica, replica_snapshot),
+            replica_snapshot: (repica_states, replica_snapshots),
```
tiagolobocastro: hmm is this the correct way, or should it be a vec of (state, snapshot)? Then we already have our list matched correctly?

tiagolobocastro:
@hrudaya21 this is a big chunk to review, next please please split this into separate work items / pr's

hrudaya21:
> @hrudaya21 this is a big chunk to review, next please please split this into separate work items / pr's

Yes. I understand, next time, I will make different PR for easy review and rework. For current PR, what is your suggestion ? I am okay to split to multiple PRs now as well, as it has separate commit and not much effort to split.

tiagolobocastro:
> > @hrudaya21 this is a big chunk to review, next please please split this into separate work items / pr's
> 
> Yes. I understand, next time, I will make different PR for easy review and rework. For current PR, what is your suggestion ? I am okay to split to multiple PRs now as well, as it has separate commit and not much effort to split.

It's fine now that it's here, it's not so much about the 2 commits, but rather each commit actually should be spread into multiple commits.
For large pieces of work we tend to split into separate commits and/or pr's for rest, grpc, csi and core agent for example.

hrudaya21:
> > > @hrudaya21 this is a big chunk to review, next please please split this into separate work items / pr's
> > 
> > 
> > Yes. I understand, next time, I will make different PR for easy review and rework. For current PR, what is your suggestion ? I am okay to split to multiple PRs now as well, as it has separate commit and not much effort to split.
> 
> It's fine now that it's here, it's not so much about the 2 commits, but rather each commit actually should be spread into multiple commits. For large pieces of work we tend to split into separate commits and/or pr's for rest, grpc, csi and core agent for example.

oohh.. understood. make sense as well. Will takecare in coming PRs. Thank you.

control-plane/grpc/src/operations/volume/traits.rs
Abhinandan-Purkait: I mean yeah, accepting a map can set multiple properties at once. Like for example, max_snapshots and topology. Otherwise it would be multiple calls, which isn't bad either. Whatever is better ux.

control-plane/grpc/proto/v1/volume/volume.proto
Abhinandan-Purkait: @tiagolobocastro @hrudaya21 I think we should create a doc covering protobuf types sometime, otherwise we keep doing same mistakes again and are not consistent.

control-plane/agents/src/bin/core/volume/snapshot_operations.rs
hrudaya21: Currently dataplane skip all the replicas which are marked as skipped from control-plane explicitly. 
If any of the replica which is passed from control-plane and failed to take snapshot, it will be tracked as part of `NexusReplicaSnapshotStatus.status` field in gRPC response from dataplane.

control-plane/agents/src/bin/core/volume/snapshot_operations.rs
hrudaya21: It is one to one mapping from replica to replica_snapshot. But I am thinking better to use a map where key is replica and value is replica_snapshot. It will give a better clarity.

control-plane/stor-port/src/types/v0/store/volume.rs
hrudaya21: Currently value is updated in the commit stage. In the update stage, We are not changing the value. Similar to SetReplica flow. As it is just a single value update, done at the end of the operation.

control-plane/rest/openapi-specs/v0_api_spec.yaml
hrudaya21: It means, when We take snapshot for 3 replica volume, We created 3 snapshot as well, which is explained as `snapshot replica`. Any other suggestion ?

control-plane/rest/openapi-specs/v0_api_spec.yaml
hrudaya21: Changed to logic as follow.
1. If max_snapshot field is not passed in storage class, then it is considered as unlimited else it will consider the storage class parameter value to judge max snapshot limit.
2. Using plugin, We can change the value of max_snapshot field value.
3. Parameter added in core-agent argument is removed now.

control-plane/stor-port/src/types/v0/store/volume.rs
hrudaya21: done

control-plane/stor-port/src/types/v0/store/volume.rs
hrudaya21: Done

control-plane/agents/src/bin/core/controller/registry.rs
hrudaya21: This parameter is removed now.
1. We have the control from the storage class parameter.
2. If storage class parameter is not passed, then the max_snapshot limit is `unlimited`
3. We can use plugin command to set the new value of the max_snapshot.

control-plane/agents/src/bin/core/volume/snapshot_operations.rs
hrudaya21: Done as vec<(state, snapshot)> now to have one to one mapping.

control-plane/stor-port/src/types/v0/store/volume.rs
hrudaya21: Make it enum now.

control-plane/grpc/proto/v1/volume/volume.proto
hrudaya21: Done for `SetVolumePropRequest` and it's parameters. Seems We need to do for other parameters present in this file.

control-plane/grpc/src/operations/volume/traits.rs
hrudaya21: Currently created a new task for this. In the current PR, it is handled as property enum.
https://dcsw-bolt.atlassian.net/browse/GTM-1934

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -638,6 +644,34 @@ impl ResourceReplicas for OperationGuardArc<VolumeSpec> {
...
+
+    async fn set_property(
+        &mut self,
+        registry: &Registry,
+        request: &Self::Request,
+    ) -> Result<(), SvcError> {
+        let state = registry.volume_state(&request.uuid).await?;
+        let operation = match VolumeAttr::from_str(&request.prop_name) {
+            Ok(VolumeAttr::MaxSnapshots) => VolumeOperation::SetVolumeProperty(
+                VolumeProperty::new(VolumeAttr::MaxSnapshots, &request.prop_value),
```
tiagolobocastro: Wonder if this should be part of the grpc proto, so here we'd receive valid parameter

control-plane/agents/src/bin/core/volume/snapshot_operations.rs
```diff
@@ -318,13 +332,13 @@ impl OperationGuardArc<VolumeSnapshot> {
         volume: &OperationGuardArc<VolumeSpec>,
         prep_params: &PrepareVolumeSnapshot,
         registry: &Registry,
-        target_node: N,
+        target_node: Option<N>,
     ) -> Result<VolumeSnapshotCreateResult, SvcError> {
         if let Some(target) = volume.as_ref().target() {
-            self.snapshot_nexus(prep_params, target, registry, target_node)
+            self.snapshot_nexus(prep_params, target, registry, target_node.unwrap())
```
tiagolobocastro: no unwraps please :)

control-plane/agents/src/bin/core/volume/snapshot_operations.rs
```diff
@@ -335,88 +349,102 @@ impl OperationGuardArc<VolumeSnapshot> {
...
-            _ => Err(SvcError::ReplicaSnapMiss {
-                replica: replica_snap.spec().uuid().to_string(),
-            }),
-        }?;
-
-        if let Some(error) = snapped.error {
+        if let Some(snap) = response
+            .replicas_status
+            .iter()
+            .find(|&snap| snap.error.is_some())
```
tiagolobocastro: should we collect all the errors?

control-plane/agents/src/bin/core/volume/snapshot_operations.rs
```diff
@@ -335,88 +349,102 @@ impl OperationGuardArc<VolumeSnapshot> {
...
-        )
-        .await?;
-
-        replica_snap.complete_vol(
-            snapshot.timestamp().into(),
-            snapshot.replica_size(),
-            snapshot.allocated_size() + snapshot.predecessor_alloc_size(),
-        );
-        Ok(VolumeSnapshotCreateResult::new_ok(replica_snap, timestamp))
+        for (replica, replica_snap) in prep_params.replica_snapshot.clone().iter_mut() {
```
tiagolobocastro: Not sure if this is doing what you expect, you're cloning it so it's not modifying the list which you then return below

control-plane/agents/src/bin/core/volume/snapshot_operations.rs
```diff
@@ -335,88 +349,102 @@ impl OperationGuardArc<VolumeSnapshot> {
...
+            replica_snap.complete_vol(
+                snapshot.timestamp().into(),
+                snapshot.replica_size(),
+                snapshot.allocated_size() + snapshot.predecessor_alloc_size(),
+            );
+        }
+        let snapshots = prep_params
+            .replica_snapshot
+            .iter()
+            .map(|(_, snapshot)| snapshot.clone())
```
tiagolobocastro:
```suggestion
            .into_iter()
            .map(|(_, replica_snapshot)| replica_snapshot)
```

control-plane/agents/src/bin/core/volume/snapshot_operations.rs
```diff
@@ -335,88 +349,102 @@ impl OperationGuardArc<VolumeSnapshot> {
...
-            .await?;
-        let timestamp = response.timestamp();
-        replica_snap.1.complete_vol(
-            timestamp.into(),
-            response.replica_size(),
-            response.allocated_size() + response.predecessor_alloc_size(),
-        );
+        let mut timestamp = SystemTime::now();
+
+        for (replica, replica_snap) in prep_params.replica_snapshot.clone().iter_mut() {
```
tiagolobocastro: same potential bug as above

control-plane/agents/src/common/errors.rs
```diff
@@ -376,6 +376,19 @@ pub enum SvcError {
         cluster_capacity_limit: u64,
         excess: u64,
     },
+    #[snafu(display("All replicas are not healthy for volume '{}'", id))]
```
tiagolobocastro:
```suggestion
    #[snafu(display("All replicas are unhealthy for volume '{id}'"))]
```

tiagolobocastro: Don't we have similar error already?

control-plane/agents/src/common/errors.rs
```diff
@@ -376,6 +376,19 @@ pub enum SvcError {
...
+    AllReplicaNotHealthy { id: String },
+    #[snafu(display(
+        "Reached maximum snapshots limit {} for volume {}, delete unused snapshots to continue",
```
tiagolobocastro:
```suggestion
        "Reached maximum snapshots limit {max_snapshots} for volume {volume_id}, delete unused snapshots to continue",
```

control-plane/agents/src/common/errors.rs
```diff
@@ -1018,6 +1031,24 @@ impl From<SvcError> for ReplyError {
...
                 extra,
             },
+            SvcError::AllReplicaNotHealthy { .. } => ReplyError {
+                kind: ReplyErrorKind::FailedPrecondition,
+                resource: ResourceKind::VolumeSnapshot,
+                source,
+                extra,
+            },
+            SvcError::SnapshotMaxLimit { .. } => ReplyError {
+                kind: ReplyErrorKind::FailedPrecondition,
```
tiagolobocastro: wonder if there's a better error code for this

control-plane/agents/src/common/errors.rs
```diff
@@ -1018,6 +1031,24 @@ impl From<SvcError> for ReplyError {
...
+                resource: ResourceKind::Volume,
+                source,
+                extra,
+            },
+            SvcError::InvalidSetProperty { .. } => ReplyError {
+                kind: ReplyErrorKind::FailedPrecondition,
```
tiagolobocastro:
```suggestion
                kind: ReplyErrorKind::InvalidArgument,
```

control-plane/csi-driver/src/context.rs
```diff
@@ -353,12 +364,18 @@ impl TryFrom<&HashMap<String, String>> for CreateParams {
         )
         .map_err(|_| tonic::Status::invalid_argument("Invalid clone_fs_id_as_volume_id"))?;
 
+        let max_snapshots = Parameters::max_snapshots(args.get(Parameters::MaxSnapshots.as_ref()))
+            .map_err(|_| {
+                tonic::Status::invalid_argument("Invalid `maxSnapshots` value, expected an i32")
```
```suggestion
                tonic::Status::invalid_argument("Invalid `maxSnapshots` value, expected an u32")
```

control-plane/grpc/proto/v1/volume/volume.proto
```diff
@@ -259,12 +261,14 @@ message CreateVolumeRequest {
   VolumePolicy policy = 6;
   // replica placement topology for the volume creation only
   optional Topology topology = 7;
-  // flag indicating thin provisioning
+  // flag indicating thin provisioningcontrol-plane/grpc/proto/v1/volume/volume.proto
```
tiagolobocastro: assuming this was a mistake

```suggestion
  // flag indicating thin provisioning
```

control-plane/plugin/src/resources/volume.rs
```diff
@@ -192,6 +192,41 @@ impl Scale for Volume {
...
+            .await
+        {
+            Ok(volume) => match output {
+                OutputFormat::Yaml | OutputFormat::Json => {
+                    // Print json or yaml based on output format.
+                    utils::print_table(output, volume.into_body());
+                }
+                OutputFormat::None => {
+                    // In case the output format is not specified, show a success message.
+                    println!("Volume {id} max_snapshots set to {prop_value} successfully 🚀")
```
tiagolobocastro:
```suggestion
                    println!("Volume {id} max_snapshots set to {prop_value} successfully")
```

control-plane/rest/openapi-specs/v0_api_spec.yaml
tiagolobocastro: Ah I see, I think it's fine then

control-plane/stor-port/src/types/v0/store/volume.rs
tiagolobocastro: hmm don't see the doc comment for MaxSnapshots?

control-plane/agents/src/bin/core/volume/snapshot_operations.rs
hrudaya21: Avoided unwrap here.

control-plane/agents/src/bin/core/volume/snapshot_operations.rs
hrudaya21: Handled.

control-plane/agents/src/bin/core/volume/snapshot_operations.rs
hrudaya21: Fixed this issue. Thankyou.

control-plane/agents/src/common/errors.rs
hrudaya21: We have one error code which tells all the child are not healthy. This is the scenario, where some of the replicas are not healthy. Change the name of the error code and description. PTAL.

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -26,18 +26,24 @@ use crate::{
     },
 };
 use agents::errors::SvcError;
+use std::str::FromStr;
+
+// use grpc::operations::volume::traits::CreateVolumeInfo;
```
Abhinandan-Purkait: Remove commented code?

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -26,18 +26,24 @@ use crate::{
...
 use stor_port::{
```
Abhinandan-Purkait: Can you please arrange the imports, the rule is to keep out crates, libs on top followed by a blank line then 3rd party crates.

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -811,14 +845,13 @@ impl ResourceLifecycleExt<CreateVolumeSource<'_>> for OperationGuardArc<VolumeSp
         request_src: &CreateVolumeSource,
     ) -> Result<Self::CreateOutput, SvcError> {
         request_src.pre_flight_check()?;
-        let request = request_src.source();
-
+        let request = request_src.source().clone();
```
Abhinandan-Purkait: Don't need to clone here?

control-plane/agents/src/bin/core/volume/snapshot_helpers.rs
```diff
@@ -1,3 +1,5 @@
+use std::collections::HashSet;
```
Abhinandan-Purkait: Please arrange the import, as per the rule above.

control-plane/agents/src/bin/core/volume/snapshot_helpers.rs
```diff
@@ -95,34 +97,41 @@ impl SpecOperationsHelper for VolumeSnapshot {
...
     registry: &Registry,
-) -> Result<ChildItem, SvcError> {
-    if volume.num_replicas != 1 {
+) -> Result<Vec<ChildItem>, SvcError> {
+    let children = super::scheduling::snapshoteable_replica(volume, registry).await?;
+
+    //todo: Remove this check once we support snapshotting with n-replicas.
+    if volume.num_replicas != 1 || children.candidates().len() != 1 {
         return Err(SvcError::NReplSnapshotNotAllowed {});
     }
```
Abhinandan-Purkait: This check should be removed right now? We support n-replicas?

control-plane/agents/src/common/errors.rs
```diff
@@ -339,10 +339,9 @@ pub enum SvcError {
     ReplicaSnapSkipped { replica: String },
     #[snafu(display("Replica's {} snapshot was unexpectedly not taken", replica))]
     ReplicaSnapMiss { replica: String },
-    #[snafu(display("Replica's {} snapshot failed with error {}", replica, error))]
+    #[snafu(display("Snapshot failed for replicas {:?}", failed_replicas))]
```
Abhinandan-Purkait: maybe `Failed to snapshot replicas {:?}` ?

control-plane/agents/src/common/errors.rs
```diff
@@ -376,6 +375,19 @@ pub enum SvcError {
         cluster_capacity_limit: u64,
         excess: u64,
     },
+    #[snafu(display(
+        "The number of healthy replicas does not match the expected count for volume '{}'",
```
Abhinandan-Purkait: `expected replica count`

control-plane/rest/openapi-specs/v0_api_spec.yaml
```diff
@@ -1544,6 +1544,40 @@ paths:
           $ref: '#/components/responses/ServerError'
       security:
         - JWT: []
+  '/volumes/{volume_id}/properties/{property_name}/{property_value}':
```
Abhinandan-Purkait: `properties` --> `property` because we are changing only one?

Also should it be in body rather than path? Let's say in future we want to change multiple properties? @tiagolobocastro wdyt?

tiagolobocastro: hmm actually I think plurality is correct, think of: /volumes/{volume_id} <- single volume as well right?
For multiple values we can use put `/volumes/{volume_id}/properties/{property_name}` with body perhaps?

control-plane/agents/src/bin/core/volume/operations.rs
hrudaya21: moved plugin changes to a different PR.

control-plane/agents/src/bin/core/volume/snapshot_operations.rs
```diff
@@ -285,138 +291,168 @@ impl ResourcePruning for OperationGuardArc<VolumeSnapshot> {
...
+        &self,
+        replicas: &Vec<ChildItem>,
+    ) -> Result<PrepareVolumeSnapshot, SvcError> {
         let Some(parameters) = self.as_ref().prepare() else {
             return Err(SvcError::AlreadyExists {
                 id: self.uuid().to_string(),
                 kind: ResourceKind::VolumeSnapshot,
             });
         };
+        let mut replica_snapshots = vec![];
```
Abhinandan-Purkait:
```suggestion
        let mut replica_snapshots = Vec::with_capacity(replicas.len());
```

control-plane/agents/src/bin/core/volume/snapshot_operations.rs
```diff
@@ -285,138 +291,168 @@ impl ResourcePruning for OperationGuardArc<VolumeSnapshot> {
...
+        target_node: Option<N>,
     ) -> Result<VolumeSnapshotCreateResult, SvcError> {
         if let Some(target) = volume.as_ref().target() {
-            self.snapshot_nexus(prep_params, target, registry, target_node)
-                .await
+            if let Some(target_node) = target_node {
+                self.snapshot_nexus::<N>(prep_params, target, registry, target_node)
+                    .await
+            } else {
+                Err(SvcError::InvalidArguments {})
```
tiagolobocastro: How is this invalid arguments, how would we get here?
Perhaps we need to modify the code a bit, maybe doesn't make sense to use this function as is anymore.

control-plane/agents/src/bin/core/volume/snapshot_operations.rs
```diff
@@ -285,138 +291,168 @@ impl ResourcePruning for OperationGuardArc<VolumeSnapshot> {
...
-            .await?;
-        let timestamp = response.timestamp();
-        replica_snap.1.complete_vol(
-            timestamp.into(),
-            response.replica_size(),
-            response.allocated_size() + response.predecessor_alloc_size(),
-        );
+        let mut timestamp = SystemTime::now();
+
+        for (replica, replica_snap) in prep_params.replica_snapshot.iter_mut() {
```
tiagolobocastro: It's a bit confusing why we are modifying this input parameter. From what I can tell we don't use that after as we return via fn output. 
So I suggest instead you clone it beforehand:

```suggestion
        let mut replica_snapshots = prep_params.replica_snapshot.clone();
        for (replica, replica_snap) in replica_snapshots.iter_mut() {
```

control-plane/agents/src/common/errors.rs
```diff
@@ -376,6 +375,19 @@ pub enum SvcError {
         cluster_capacity_limit: u64,
         excess: u64,
     },
+    #[snafu(display(
+        "The number of healthy replicas does not match the expected replica count of volume '{}'",
+        id
```
tiagolobocastro:
```suggestion
        "The number of healthy replicas does not match the expected replica count of volume '{id}'"
```

control-plane/agents/src/common/errors.rs
```diff
@@ -376,6 +375,19 @@ pub enum SvcError {
...
+    ))]
+    InsufficientHealthyReplicas { id: String },
+    #[snafu(display(
+        "Reached maximum snapshots limit {max_snapshots} for volume {volume_id}, delete unused snapshots to continue"))]
+    SnapshotMaxLimit {
+        max_snapshots: u32,
+        volume_id: String,
+    },
+    #[snafu(display("Invalid property name '{}' for the volume '{}'", property_name, id))]
```
tiagolobocastro:
```suggestion
    #[snafu(display("Invalid property name '{property_name}' for the volume '{id}'"))]
```

control-plane/agents/src/common/errors.rs
```diff
@@ -1018,6 +1030,24 @@ impl From<SvcError> for ReplyError {
...
                 extra,
             },
+            SvcError::InsufficientHealthyReplicas { .. } => ReplyError {
+                kind: ReplyErrorKind::FailedPrecondition,
+                resource: ResourceKind::VolumeSnapshot,
+                source,
+                extra,
+            },
+            SvcError::SnapshotMaxLimit { .. } => ReplyError {
+                kind: ReplyErrorKind::ResourceExhausted,
```
tiagolobocastro: Wonder if this is the correct error code here, could be confused with no free space in the pool for taking the snapshot?

tiagolobocastro (approved):

Abhinandan-Purkait (approved):

control-plane/agents/src/bin/core/volume/snapshot_operations.rs
hrudaya21: fixed

control-plane/agents/src/bin/core/volume/snapshot_operations.rs
hrudaya21: tried differenet appraoch to avoid this error condition, which was never hitting. PTAL

control-plane/agents/src/common/errors.rs
Abhinandan-Purkait: /// The system is not in a state required for the operation's execution.
    FailedPrecondition = 9,
    
    /// Operation was attempted past the valid range.
    OutOfRange = 11,
    
    /// Some resource has been exhausted.
    ResourceExhausted = 8,
    
     /// The caller does not have permission to execute the specified operation.
    PermissionDenied = 7,

@tiagolobocastro These 4 seem to be a candidate. How about `OutOfRange` ?

hrudaya21: `PermissionDenied` seems to be more reasonable for me.

tiagolobocastro: I'm also leaning more towards `OutOfRange`

hrudaya21: okay, let me keep `OutOfRange` then.

hrudaya21:
bors try

hrudaya21:
bors try

hrudaya21:
bors try

hrudaya21:
bors merge

hrudaya21:
bors merge

PR #712, opened 2024-01-09
feat(pstor-proxy): add pstor proxy create, add frontend registration backend

Abhinandan-Purkait:
- Adds the new pstor-proxy binary.
- Adds the in memory cache for frontend registration.
- Adds the frontend registration grpc server and service.
- Adds the frontend grpc server and service.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -242,7 +242,7 @@ dependencies = [
  "anyhow",
  "async-trait",
  "chrono",
- "clap",
+ "clap 4.4.13",
  "crossbeam-queue",
  "deployer-cluster",
  "dyn-clonable",
@@ -335,6 +335,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "819e7219dbd41043ac279b19830f2efc897156490d7fd6ea916720117ee66311"
 dependencies = [
  "libc",
+]
+
+[[package]]
+name = "ansi_term"
+version = "0.12.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d52a9bb7ec0cf484c551830a7ce27bd20d67eac647e1befb56b0be4ee39a55d2"
+dependencies = [
+ "winapi",
 ]
 
 [[package]]
@@ -456,6 +465,17 @@ dependencies = [
  "proc-macro2",
  "quote",
  "syn 2.0.38",
+]
+
+[[package]]
+name = "atty"
+version = "0.2.14"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d9b39be18770d11421cdb1b9947a45dd3f37e93092cbf377614828a319d5fee8"
+dependencies = [
+ "hermit-abi 0.1.19",
+ "libc",
+ "winapi",
 ]
 
 [[package]]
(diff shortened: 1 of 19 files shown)


control-plane/pstor-proxy/proto/v1/frontend_node/registration.proto
```diff
@@ -0,0 +1,28 @@
...
+
+// Register the frontend node with control-plane
+message RegisterRequest {
+  // Id of the frontend node instance
+  string id = 1;
+  // Grpc endpoint of the frontend node instance
+  string grpc_endpoint = 2;
+  // Frontend node labels
+  optional StringMapValue labels = 3;
+}
```
dsharma-dc, line 17: How about importing the message type FrontendNodeSpec here and do:
```suggestion
message RegisterRequest {
  FrontendNodeSpec frontend_node = 1;
}
```

Cargo.lock
```diff
@@ -242,7 +242,7 @@ dependencies = [
  "anyhow",
  "async-trait",
  "chrono",
- "clap",
+ "clap 4.4.13",
```
tiagolobocastro, line 245: Please use the same clap version to avoid building twice

control-plane/pstor-proxy/proto/v1/frontend_node/frontend_node.proto
```diff
@@ -0,0 +1,65 @@
...
+  // Id of the frontend node instance
+  string id = 1;
+  // Grpc endpoint of the frontend node instance
+  string grpc_endpoint = 2;
+  // Frontend node labels
+  optional StringMapValue labels = 3;
+}
+
+// Frontend Node state
+message FrontendNodeState {
```
tiagolobocastro, line 30: State could be the grpc_endpoint and/or the status?

control-plane/pstor-proxy/proto/v1/frontend_node/frontend_node.proto
```diff
@@ -0,0 +1,65 @@
...
+// Request type for a GetFrontendNodes request
+message GetFrontendNodesRequest {
+  // Filter frontend nodes
+  oneof filter {
+    FrontendNodeFilter frontend_node = 1;
+  }
+  // Pagination to allow for multiple requests to get all frontend nodes
+  Pagination pagination = 2;
+  // Ignore 404 not found errors
+  bool ignore_notfound = 3;
```
tiagolobocastro, line 52: I know we've been doing it this way so far, but maybe consider separating into GetFrontendNode and ListFrontendNodes ? Then list would implicitly ignore 404

control-plane/pstor-proxy/proto/v1/frontend_node/frontend_node.proto
```diff
@@ -0,0 +1,65 @@
...
+}
+
+// Reply type for a GetFrontendNodes request
+message GetFrontendNodesReply {
+  oneof reply {
```
tiagolobocastro, line 57: Are the default gRPC status not enough for a get?

control-plane/pstor-proxy/proto/v1/frontend_node/registration.proto
Abhinandan-Purkait: Yes we can, but in future if we have more fields in frontend node spec, which might not be a part of registration.

Abhinandan-Purkait:
Closing this since we have resorted to an alternative approach.

PR #716, opened 2024-01-16
test(volume/resize): add bdd tests for offline volume resize

dsharma-dc:


tests/bdd/features/volume/resize/test_resize_offline.py
```diff
@@ -0,0 +1,306 @@
...
+
+@given(
+    parsers.parse(
+        "an unpublished volume with {repl_count:d} replicas and all are healthy"
+    )
+)
+def an_unpublished_volume_with_repl_count_replicas_and_all_are_healthy(repl_count):
+    """an unpublished volume with <repl_count> replicas and all are healthy."""
+    pytest.replica_count = repl_count
+    volume(create_volume)
```
dsharma-dc: I thought this was working bcz tests passed but this isn't correct. It worked because expand step has `create_volume` fixture.
@tiagolobocastro @Abhinandan-Purkait Looking for a way to take variable `repl_count` and pass it to fixture `create_volume`, any suggestions?

Abhinandan-Purkait (approved):

tests/bdd/features/volume/resize/test_resize_offline.py
Abhinandan-Purkait: There is indeed a way, let me try that and let you know.
`@pytest.mark.parametrize('tester_arg', [True, False])` using this thing.

tests/bdd/features/volume/resize/resize_offline.feature
```diff
@@ -0,0 +1,46 @@
...
+
+    Background:
+        Given a deployer cluster
+
+    Scenario Outline: expand an unpublished volume
+        Given an unpublished volume with <repl_count> replicas and all are healthy
+        When the volume is expanded
+        Then all the replicas of volume should be resized to new capacity
+        And the volume expand status should be success
+        And the volume is expanded to the new capacity
```
tiagolobocastro: try and use imperative rather than declarative in the control-plane bdd's, given these are not k8s specific

```suggestion
        When we issue a volume expand request
        Then the request should succeed
        And the volume is expanded to the new capacity
        And all the volume's replicas should be resized to the new capacity
```

tests/bdd/features/volume/resize/resize_offline.feature
```diff
@@ -0,0 +1,46 @@
+Feature: Volume resize
+
+    Background:
+        Given a deployer cluster
+
+    Scenario Outline: expand an unpublished volume
```
tiagolobocastro: nit: Start scenarios with upper case

```suggestion
    Scenario Outline: Expand an unpublished volume
```

tests/bdd/features/volume/resize/resize_offline.feature
```diff
@@ -0,0 +1,46 @@
...
+        Given a published volume with more than one replica and all are healthy
+        When the volume is expanded
+        Then the volume expand status should be failure
+        And the failure reason should be volume-in-use precondition
+
+
+    Scenario: expand an unpublished volume and then make it published
+        Given an unpublished volume with more than one replica and all are healthy
+        When the volume is expanded
+        Then all the replicas of volume should be resized to new capacity
```
tiagolobocastro: You can reuse the exact same wording as from scenario 1, this way you can reduce code!

tests/bdd/features/volume/resize/resize_offline.feature
```diff
@@ -0,0 +1,46 @@
...
+        When the volume is published
+        Then the volume should get published with expanded capacity
+        And the new capacity should be available for the application to use
+
+
+    Scenario: expand an unpublished volume having a not-online replica
```
tiagolobocastro:
```suggestion
    Scenario: Expand an unpublished volume with an offline replica
```

tests/bdd/features/volume/resize/resize_offline.feature
```diff
@@ -0,0 +1,46 @@
...
+        Given an unpublished volume with more than one replica
+        And one of the replica is not in online state
+        When the volume is expanded
+        Then the volume expand status should be failure
```
tiagolobocastro, line 57: Let's also ensure that when the replica becomes online it's rebuilt
Also would be good to check also what happens if we increase the replica count

tests/bdd/features/volume/resize/test_resize_offline.py
```diff
@@ -0,0 +1,308 @@
...
+@then("the volume should get published with expanded capacity")
+def the_volume_should_get_published_with_expanded_capacity():
+    """the volume should get published with expanded capacity."""
+    volume = ApiClient.volumes_api().get_volume(VOLUME_UUID)
+    assert hasattr(volume.spec, "target")
+    assert str(volume.spec.target.protocol) == str(Protocol("nvmf"))
+    assert volume.spec.size == VOLUME_NEW_SIZE
+
+
+def del_volume():
```
tiagolobocastro: you can use new api, which helps debug things on failure:
`Volume.cleanup(volume)`

tests/bdd/features/volume/resize/resize_offline.feature
dsharma-dc: But for unpublished volume rebuild won't be there after replica becomes online.

tests/bdd/features/volume/resize/resize_offline.feature
dsharma-dc: I think the order of checks would be better to check that replicas are resized first, then check the volume capacity

```
        And all the replicas of the volume should be resized to the new capacity
        And the volume is expanded to the new capacity
```

dsharma-dc:
bors try

dsharma-dc:
Failure in events test while expecting child removal event, unrelated. Retrying...
> thread 'event::events' panicked at control-plane/agents/src/bin/core/tests/event/mod.rs:258:10:
called `Result::unwrap()` on an `Err` value: Elapsed(())

bors try

tests/bdd/features/volume/resize/resize_offline.feature
tiagolobocastro: Sure, whichever way should work

tests/bdd/features/volume/resize/resize_offline.feature
tiagolobocastro: Yes we have to add a clause where the volume becomes republished.

tiagolobocastro (commented):
We're missing scenarios for interaction between snapshot and resize. This can be added on another PR I guess but I think it's important to check that the features work together.

tests/bdd/features/volume/resize/resize_offline.feature
dsharma-dc: Though it overlaps a bit with other existing scenarios, I'll add this here
```
        When the replica comes online again
        And we issue a volume expand request
        Then the request should succeed
        And all the replicas of the volume should be resized to the new capacity
        When the volume is published
        Then the onlined replica should be rebuilt
```

dsharma-dc: hmm, this doesn't seem to add up here cleanly @tiagolobocastro . The publish here has to happen after replica comes online because resize will fail for published volume, this means rebuild won't happen. If thinking about republish, I'll add a new scenario instead of modifying this one. The new one will be.

```
    Scenario: Expand a published volume after unpublishing it while having an offline replica
        Given a published volume with more than one replica and all are healthy
        And the volume is receiving IO
        And one of the replica goes offline
        And the volume is unpublished
        When the replica comes online again
        And we issue a volume expand request
        Then the request should succeed
        And all the replicas of the volume should be resized to the new capacity
        When the volume is republished
        Then the volume should get published with expanded capacity
        And the onlined replica should be rebuilt
```

dsharma-dc: I've made a note and I'll add this in next set of test changes, it's taking some trouble to work seamlessly. This PR can go ahead meanwhile.

dsharma-dc: ok, worked now. Added a new test scenario.

dsharma-dc:
bors try

dsharma-dc:
bors try

tests/bdd/features/volume/resize/resize_offline.feature
```diff
@@ -0,0 +1,65 @@
...
+    Scenario: Expand a published volume
+        Given a published volume with more than one replica and all are healthy
+        When we issue a volume expand request
+        Then the volume expand status should be failure
+        And the failure reason should be volume-in-use precondition
+
+
+    Scenario: Expand a published volume after unpublishing it while having an offline replica
+        Given a published volume with more than one replica and all are healthy
+        When the volume is receiving IO
```
tiagolobocastro, line 40: Is this step needed?

tests/bdd/features/volume/resize/test_resize_offline.py
```diff
@@ -0,0 +1,448 @@
...
+# This fixture will be automatically used by all tests.
+# It starts the deployer which launches all the necessary containers.
+# Pools are created for convenience such that they are available for use by the tests.
+@pytest.fixture(autouse=True, scope="module")
+def init(disks):
+    Deployer.start(
+        4,
+        faulted_child_wait_period="5s",
+        reconcile_period="1s",
+        cache_period="1s",
```
tiagolobocastro:
```suggestion
        faulted_child_wait_period="100ms",
        reconcile_period="100ms",
        cache_period="100ms",
```

tests/bdd/features/volume/resize/test_resize_offline.py
```diff
@@ -0,0 +1,448 @@
...
+    fio = Fio(name="fio-pre-resize", rw="write", uri=uri, extra_args="--loops=2")
+    pytest.fio = fio.open()
+
+
+@when("the volume is unpublished")
+def the_volume_is_unpublished():
+    """the volume is unpublished."""
+    # Ensure fio is still running till unpublish
+    time.sleep(1)
+    assert pytest.fio.poll() is None
```
tiagolobocastro, line 242:
```suggestion
try:
    code = pytest.fio.wait(timeout=2)
    assert code == 0, "FIO failed, exit code: %d" % code
except subprocess.TimeoutExpired:
    assert False, "FIO timed out"
```

tests/bdd/features/volume/resize/test_resize_offline.py
```diff
@@ -0,0 +1,448 @@
...
+        replica = ApiClient.replicas_api().get_replica(replica_uuid)
+        assert replica.size >= VOLUME_NEW_SIZE
+
+
+@then("the failure reason should be invalid arguments")
+def the_failure_reason_should_be_invalid_arguments():
+    """the failure reason should be invalid arguments."""
+    # Reason code: Invalid Arguments, mapped from
+    # SvcError::VolumeResizeArgsInvalid
+    assert pytest.exception.status == 400
```
tiagolobocastro:
```suggestion
    assert pytest.exception.status == http.HTTPStatus.BAD_REQUEST
```

tests/bdd/features/volume/resize/test_resize_offline.py
```diff
@@ -0,0 +1,448 @@
...
+
+
+@then("the failure reason should be volume-in-use precondition")
+def the_failure_reason_should_be_volumeinuse_precondition():
+    """the failure reason should be volume-in-use precondition."""
+    # Reason code: Conflict, mapped from SvcError::InUse
+    assert pytest.exception.status == 409
```
tiagolobocastro:
```suggestion
    assert pytest.exception.status == http.HTTPStatus.CONFLICT
```

tests/bdd/features/volume/resize/test_resize_offline.py
```diff
@@ -0,0 +1,448 @@
...
+@retry(wait_fixed=1000, stop_max_attempt_number=20)
+def wait_child_faulted():
+    vol = ApiClient.volumes_api().get_volume(VOLUME_UUID)
+    target = vol.state.target
+    childlist = target["children"]
+    faulted_child = None
+    for child in childlist:
+        if ChildState(child["state"]) == ChildState("Faulted"):
+            faulted_child = child["uri"]
+    assert faulted_child is not None, "Failed to find Faulted child!"
```
tiagolobocastro: Can use lambda here, can also check in case >1 are faulted

```suggestion
    faulted = list(
        filter(lambda child: child.get("state") == "Faulted", childlist)
    )
    assert len(degraded) == 1, "Failed to find 1 Faulted child"
```

tiagolobocastro (approved):

tests/bdd/features/volume/resize/test_resize_offline.py
dsharma-dc: `wait` will block till fio completion here for a timeout of 2sec. However, with `poll` I'm just making sure the fio process is still running in background when unpublish is done. `sleep` is optional though.

tests/bdd/features/volume/resize/resize_offline.feature
dsharma-dc: I added this to actually have some rebuild data buildup and be more realistic.

tests/bdd/features/volume/resize/test_resize_offline.py
tiagolobocastro: hmm but how can you unpublish the volume if fio is still running?

tiagolobocastro: Or is it because you don't care if fio fails?

dsharma-dc: Yeah in this case we don't care as long as some data was written to the volume.

tests/bdd/features/volume/resize/test_resize_offline.py
dsharma-dc: Kept faulted_child_wait_period at 2s to have `wait_faulted_child` check work reliably.

dsharma-dc:
bors merge

PR #718, opened 2024-01-22
feat(app node): add support for app node registration and operations

Abhinandan-Purkait:

Goals:
- App node resource is needed to issue commands to the node for operations like fsfreeze, unfreeze.
- App nodes will now register itself to the control-plane via the rest API.
- These app nodes would be used to issue fsfreeze, unfreeze as a part of snapshot creation to ensure filesystem consistency.

Changes in this PR:
- Adds the support for registration from rest api to control plane via grpc layer. This adds the grpc definitions and implementations for the same.
- Also adds the feature to get or list these app node from control-plane to rest via grpc, to be eventually used by the controller.


Abhinandan-Purkait:
I will add the cargo tests, in this PR. The review can be started independently.

control-plane/agents/src/bin/core/controller/registry.rs
```diff
@@ -501,4 +501,26 @@ impl Registry {
             false => vec![],
         }
     }
+
+    /// Register a frontend node(ex:- csi node) to control plane.
+    pub(crate) async fn register_frontend_node_spec(
```
tiagolobocastro: Does the `-` mean anything?

```suggestion
    /// Register a frontend node (ex: a csi node) with the control-plane.
    pub(crate) async fn register_frontend_node(
```

control-plane/agents/src/bin/core/controller/registry.rs
```diff
@@ -501,4 +501,26 @@ impl Registry {
...
+        frontend_node_spec: &RegisterFrontendNode,
+    ) -> Result<(), SvcError> {
+        self.specs()
+            .register_frontend_node_spec(self, frontend_node_spec)
+            .await?;
+        Ok(())
+    }
+
+    /// Deregister Register a frontend node(ex:- csi node) from control plane.
+    pub(crate) async fn deregister_frontend_node_spec(
```
tiagolobocastro: Name is quite long, probably can just say de/register frontend node without spec.

```suggestion
    /// Deregister a frontend node (ex: a csi node) from the control-plane.
    pub(crate) async fn deregister_frontend_node(
```

control-plane/agents/src/bin/core/controller/resources/operations_helper.rs
```diff
@@ -1063,13 +1064,82 @@ impl ResourceSpecsLocked {
...
             _ => {
                 // Not all spec types are persisted in the store.
                 unimplemented!("{} not persisted in store", spec_type);
             }
         };
         Ok(())
     }
+
+    /// Get a frontend node spec using id.
+    pub(crate) fn get_frontend_node_spec(&self, id: &FrontendNodeId) -> Option<FrontendNodeSpec> {
```
tiagolobocastro: These should not be here, see how it is done for volumes and nodes

control-plane/agents/src/bin/core/frontend_node/registry.rs
```diff
@@ -0,0 +1,45 @@
...
+        let mut frontend_nodes = Vec::with_capacity(frontend_node_specs.len());
+        let last = frontend_node_specs.last();
+        for spec in frontend_node_specs.result() {
+            let state = FrontendNodeState::from(spec.clone());
+            frontend_nodes.push(FrontendNode::new(spec, Some(state)));
+        }
+        PaginatedResult::new(frontend_nodes, last)
+    }
+    /// Gets the frontend node from the registry with the given id.
+    pub(crate) fn get_frontend_node(&self, id: &FrontendNodeId) -> Result<FrontendNode, SvcError> {
```
tiagolobocastro: As a rule of thumb don't use `get_` prefix for getters.

```suggestion
    pub(crate) fn frontend_node(&self, id: &FrontendNodeId) -> Result<FrontendNode, SvcError> {
```

control-plane/agents/src/bin/core/frontend_node/spec.rs
```diff
@@ -0,0 +1,48 @@
...
+impl ResourceSpecsLocked {
+    /// Get a subset of frontend nodes based on the pagination argument.
+    pub(crate) fn paginated_frontend_node_specs(
+        &self,
+        pagination: &Pagination,
+    ) -> PaginatedResult<FrontendNodeSpec> {
+        let specs = self.read();
+        specs.paginated_frontend_node_specs(pagination)
+    }
+    /// Get all frontend node specs.
```
tiagolobocastro:
```suggestion
    /// Get a copy of all frontend node specs.
```

control-plane/agents/src/bin/core/frontend_node/spec.rs
```diff
@@ -0,0 +1,48 @@
+use crate::controller::resources::operations_helper::{ResourceSpecs, ResourceSpecsLocked};
+use grpc::operations::{PaginatedResult, Pagination};
+use stor_port::types::v0::store::frontend_node::FrontendNodeSpec;
+
+impl ResourceSpecs {
+    /// Get all frontend node specs.
```
tiagolobocastro: Just to make it clear, and inline with existing ones.

```suggestion
    /// Get a copy of all frontend node specs.
```

control-plane/grpc/build.rs
```diff
@@ -16,6 +16,8 @@ fn main() {
                 "proto/v1/ha/node_agent.proto",
                 "proto/v1/ha/csi_node_nvme.proto",
                 "proto/v1/snapshot/snapshot.proto",
+                "proto/v1/frontend_node/frontend_node.proto",
```
tiagolobocastro: would it make sense to have frontend_node.proto inside node/ ?

control-plane/grpc/proto/v1/frontend_node/frontend_node.proto
```diff
@@ -0,0 +1,91 @@
...
+  string id = 1;
+  // Grpc endpoint of the frontend node instance.
+  string grpc_endpoint = 2;
+  // Frontend node status.
+  FrontendNodeStatus status = 3;
+}
+
+// Status of the FrontendNode
+enum FrontendNodeStatus {
+  // FrontendNode is deemed online if it has not missed the
```
tiagolobocastro: Should offline be 0 ?

control-plane/grpc/proto/v1/frontend_node/registration.proto
```diff
@@ -0,0 +1,28 @@
+syntax = "proto3";
+
+import "google/protobuf/empty.proto";
+import "google/protobuf/wrappers.proto";
+import "v1/misc/common.proto";
+
+package v1.frontend_node.registration;
+
+// Register the frontend node with control-plane.
```
tiagolobocastro:
```suggestion
// Register the frontend node with the control-plane.
```

control-plane/grpc/proto/v1/frontend_node/registration.proto
```diff
@@ -0,0 +1,28 @@
...
+message RegisterRequest {
+  // Id of the frontend node instance.
```
tiagolobocastro: Could this just be the spec?

control-plane/grpc/src/operations/frontend_node/client.rs
```diff
@@ -0,0 +1,122 @@
...
+            ctx,
+            MessageIdVs::GetFrontendNodes,
+        );
+        let response = self
+            .ops
+            .client()
+            .get_frontend_node(request)
+            .await?
+            .into_inner()
+            .frontend_node;
```
tiagolobocastro: Could the proto reply just be FrontendNode? or is it because we want to support returning other things?

control-plane/grpc/src/operations/frontend_node/server.rs
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    async fn list_frontend_nodes(
+        &self,
+        request: Request<ListFrontendNodesRequest>,
+    ) -> Result<Response<ListFrontendNodesReply>, Status> {
+        let req: ListFrontendNodesRequest = request.into_inner();
+        let filter = match req.filter {
+            None => Filter::None,
+            _ => {
```
tiagolobocastro: What are we doing here, seems only None is valid?

tiagolobocastro (commented):
Might have been better to have protos on separate PR, it's borderline too large to review as once (I'm quite bogged down with reviews atm..)
Ping me when you've added some tests

tiagolobocastro:
Let's actually explain what this is for in the PR description rather than just having one liners

control-plane/agents/src/bin/core/controller/registry.rs
Abhinandan-Purkait: frontend_node_spec and frontend_node are two seperate types, that is why I had to put the `spec` to specifically point on to it.

control-plane/agents/src/bin/core/controller/registry.rs
Abhinandan-Purkait: Same reason for putting the word spec.

control-plane/grpc/build.rs
Abhinandan-Purkait: Both are completely different things, not sure if we can bundle them together.

control-plane/grpc/proto/v1/frontend_node/frontend_node.proto
Abhinandan-Purkait: You mean Offline as default?

control-plane/grpc/proto/v1/frontend_node/registration.proto
Abhinandan-Purkait: I don't think we should do that, it's the same reason we have different types for NodeSpec and Register. If we add more fields to NodeSpec that doesn't need to be a part of registration.

control-plane/grpc/src/operations/frontend_node/client.rs
Abhinandan-Purkait: We can just return frontend node. But in case we want to bundle something else this would be flexible.

control-plane/grpc/src/operations/frontend_node/server.rs
Abhinandan-Purkait: Yes, because at the moment there is only None filter which can return multiple frontend nodes, this would be useful if we add more filters that can return multiple frontend nodes in future.

Abhinandan-Purkait:
> Might have been better to have protos on separate PR, it's borderline too large to review as once (I'm quite bogged down with reviews atm..) 

I tried to spilt into commits, although I agree the grpc layer PRs tend to get bigger. The protos are similar to the pstor-proxy PR you had reviewed, so I kept it here. Sorry :-;

> Ping me when you've added some tests

Have added the tests. PTAL

Abhinandan-Purkait:
> Let's actually explain what this is for in the PR description rather than just having one liners

I tried to add something. Does the explanation look fine?

control-plane/agents/src/bin/core/controller/registry.rs
tiagolobocastro: hmm here it's not a type, it's an entity, we're registering a frontend node? there's no register for frontend_node type?

control-plane/grpc/build.rs
tiagolobocastro: hmm they are both node protos, what I mean is like so:
>v1/node/frontend_node.proto
v1/node/target_node.proto

control-plane/grpc/proto/v1/frontend_node/frontend_node.proto
tiagolobocastro: Yes because 0 is the default

control-plane/grpc/proto/v1/frontend_node/registration.proto
tiagolobocastro: Right, good point.
In Proto is there any way to do type aliasing?

control-plane/grpc/src/operations/frontend_node/server.rs
tiagolobocastro: Then we don't need filter at all, we can add it later?

tiagolobocastro (approved):

control-plane/grpc/build.rs
Abhinandan-Purkait: Okay, should I bundle the `frontend_node_registration.proto` inside the `frontend_node.proto` otherwise maybe put that under seperate `registration` folder?

tiagolobocastro: possibly, or app_node_registration.proto?

control-plane/agents/src/bin/core/controller/registry.rs
Abhinandan-Purkait: Right, done.

control-plane/grpc/build.rs
Abhinandan-Purkait: Done.

control-plane/grpc/proto/v1/frontend_node/registration.proto
Abhinandan-Purkait: Unfortunately no, although what we can do is wrap a message inside another, but again that creates an optional field inside.

control-plane/grpc/src/operations/frontend_node/server.rs
Abhinandan-Purkait: Yep, removed.

Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors merge

PR #719, opened 2024-01-22
build(utils/version): add correct git HEAD paths

tiagolobocastro:
When this is used a dependency for extensions the build.rs for utils-lib ends up adding an invalid git path to the cargo re-runs. This actually causes continuous rebuilding on the extensions repo. Fixes this by ensure the path exists and by checking specifically for the extensions env var.
The latter one is, to put it bluntly, a hack, but atm I don't know else to do it in a quick way.


Abhinandan-Purkait (approved):

sinhaashish (approved):

niladrih (approved):

tiagolobocastro:
bors merge

tiagolobocastro:
bors cancel

tiagolobocastro:
bors merge

PR #720, opened 2024-01-22
refactor(rest-plugin): make it more modular

tiagolobocastro:
Adds common cli args to the library, which the binary then makes use of. 
Tidies up the error scenario with common error handling and implicit trace flushing.
Add execute trait to allow for tidier composition of operations, which will be especially 
useful to reduce code duplication in the kubectl plugin.


Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

PR #724, opened 2024-01-23
Delete Orphaned Volumes on a timer

tiagolobocastro:
fix(csi-controller/gc): list volumes with pagination
    
---

    chore: correct typo in the plaftform code
    
---

    refactor(csi-controller/gc): reduce excessive unwrap and nesting
    
---

    feat(csi-controller/orphaned-vols): gc orphaned volumes on a timer
    
    Due to a k8s bug, when pv deletion is attempted before pvc deletion, we may not
    receive the delete request, and thus leaking volumes, which cannot be deleted
    through any user facing api.
    
    Existing code to clean up orphaned volumes was being made use for the case when
    certain pv events might be missed.
    Therefore current WA is to delete the csi-controller pod.
    This change makes use of this existing logic but runs it on a given time period
    which is by default 5 minutes, which  allows us to automatically WA the bug.
    
### DIFF
--- a/control-plane/csi-driver/src/bin/controller/controller.rs
+++ b/control-plane/csi-driver/src/bin/controller/controller.rs
@@ -1,4 +1,6 @@
-use crate::{ApiClientError, CreateVolumeTopology, CsiControllerConfig, IoEngineApiClient};
+use crate::{
+    client::ListToken, ApiClientError, CreateVolumeTopology, CsiControllerConfig, IoEngineApiClient,
+};
 
 use csi_driver::context::{CreateParams, PublishParams};
 use rpc::csi::{volume_content_source::Type, Topology as CsiTopology, *};
@@ -652,7 +654,7 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
         let vt_mapper = VolumeTopologyMapper::init().await?;
 
         let volumes = IoEngineApiClient::get_client()
-            .list_volumes(max_entries, args.starting_token)
+            .list_volumes(max_entries, ListToken::String(args.starting_token))
             .await
             .map_err(|e| Status::internal(format!("Failed to list volumes, error = {e:?}")))?;
 
--- a/control-plane/csi-driver/src/bin/controller/pvwatcher.rs
+++ b/control-plane/csi-driver/src/bin/controller/pvwatcher.rs
@@ -16,24 +16,24 @@ use crate::IoEngineApiClient;
 #[derive(Clone)]
 pub(crate) struct PvGarbageCollector {
     pub(crate) pv_handle: Api<PersistentVolume>,
+    orphan_period: Option<humantime::Duration>,
 }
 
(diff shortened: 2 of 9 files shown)


control-plane/csi-driver/src/bin/controller/pvwatcher.rs
```diff
@@ -44,77 +41,99 @@ impl PvGarbageCollector {
...
-                        Err(error) => error!(%error, "Error while verifying if PV is present"),
+        if reclaim_policy == "Retain" && phase == "Released" {
+            debug!(pv.name = pv_name, "PV is a deletion candidate");
+            self.delete_volume(volume_uuid).await;
+        } else if phase == "Bound" {
+            match self.pv_handle.get_opt(&pv_name).await {
+                Ok(pvol) => match pvol {
+                    Some(_) => debug!(pv.name = pv_name, "PV present on API server"),
+                    None => {
+                        debug!(pv.name = pv_name, "PV is a deletion candidate");
```
tiagolobocastro: hmm what is the use case for deleting a bound pvc with no pv, can this happen?

tiagolobocastro:
Resolves https://github.com/openebs/mayastor/issues/1269

control-plane/csi-driver/src/bin/controller/controller.rs
```diff
@@ -11,6 +11,7 @@ use stor_port::types::v0::openapi::{
 };
 use utils::{CREATED_BY_KEY, DSP_OPERATOR};
 
+use crate::client::ListToken;
```
Abhinandan-Purkait: Should be up with line 1 ?

control-plane/csi-driver/src/bin/controller/server.rs
```diff
@@ -35,9 +34,10 @@ impl Connected for UnixStream {
 }
 
 #[derive(Clone, Debug)]
-pub struct UdsConnectInfo {
-    pub peer_addr: Option<Arc<tokio::net::unix::SocketAddr>>,
-    pub peer_cred: Option<tokio::net::unix::UCred>,
+#[allow(unused)]
```
Abhinandan-Purkait, line 38: Do we need this?

Abhinandan-Purkait (approved):

control-plane/csi-driver/src/bin/controller/controller.rs
tiagolobocastro: Good catch, yes

control-plane/csi-driver/src/bin/controller/server.rs
tiagolobocastro: Yes, before fields were public so linter didn't complain...
Marking it private and unused so it's clearer that they are not used, probably can remove them but I'm not sure if it's safe to do it or not.

tiagolobocastro:
bors merge

PR #725, opened 2024-01-25
feat(app node): add rest api endpoints and integrate control-plane grpc

Abhinandan-Purkait:

Goals:
- App node resource is needed to issue commands to the node for operations like fsfreeze, unfreeze.
- App nodes will now register itself to the control-plane via the rest API.
- These app nodes would be used to issue fsfreeze, unfreeze as a part of snapshot creation to ensure filesystem consistency.

Changes in this PR:
- This adds the rest api endpoints for registration and operations for app nodes

### DIFF
--- a/control-plane/rest/openapi-specs/v0_api_spec.yaml
+++ b/control-plane/rest/openapi-specs/v0_api_spec.yaml
@@ -2033,6 +2033,108 @@ paths:
           $ref: '#/components/responses/ServerError'
       security:
         - JWT: []
+  /app-nodes:
+    get:
+      tags:
+        - AppNodes
+      operationId: get_app_nodes
+      parameters:
+        - in: query
+          name: max_entries
+          description: The maximum number of results to return.
+          schema:
+            type: integer
+            default: 0
+          required: true
+        - in: query
+          name: starting_token
+          description: The offset to start pagination from.
+          schema:
+            type: integer
+      responses:
+        '200':
+          description: OK
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/AppNodes'
+        '4XX':
+          $ref: '#/components/responses/ClientError'
+          '5XX':
+            $ref: '#/components/responses/ServerError'
+      security:
+        - JWT: [ ]
+  '/app-nodes/{app_node_id}':
+    put:
+      tags:
+        - AppNodes
+      operationId: register_app_node
+      parameters:
+        - in: path
+          name: app_node_id
+          required: true
+          schema:
+            $ref: '#/components/schemas/AppNodeId'
+      requestBody:
+        description: Contents of the app node to be registered.
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/RegisterAppNode'
+        required: true
+      responses:
+        '204':
+          description: OK
+        '4XX':
+          $ref: '#/components/responses/ClientError'
+        '5XX':
+          $ref: '#/components/responses/ServerError'
+      security:
+        - JWT: []
+    delete:
+      tags:
+        - AppNodes
+      operationId: deregister_app_node
+      parameters:
+        - name: app_node_id
+          in: path
+          description: ID of the app node to be deregistered.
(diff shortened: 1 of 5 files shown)


control-plane/rest/openapi-specs/v0_api_spec.yaml
```diff
@@ -3816,6 +3922,87 @@ components:
...
         - drainlabels
+    RegisterAppNode:
+      example:
+        endpoint: '10.1.0.5:50055'
+        labels:
+          openebs.io/csi-node.nvme-ana: true,
+          openebs.io/zone: us-east-1a
+      type: object
+      properties:
+        endpoint:
```
tiagolobocastro, line 4019: missing descriptions here

control-plane/rest/openapi-specs/v0_api_spec.yaml
```diff
@@ -3816,6 +3922,87 @@ components:
...
+          type: string
+          format: url
+        labels:
+          type: object
```
tiagolobocastro, line 4027: Need to better define what these labels are, ex: are these control-plane labels, are these csi labels, k8s labels etc..
Though csi and kubernetes terms don't quite belong "here", so might need another type of abstraction, maybe a map of maps?

control-plane/rest/openapi-specs/v0_api_spec.yaml
```diff
@@ -3816,6 +3922,87 @@ components:
...
+          additionalProperties:
+            type: string
+      required:
+        - id
+        - endpoint
+    AppNode:
+      description: |-
+        Respresents an app node.
```
```suggestion
        Represents an application node, which connects to the target node via a share protocol (eg: nvmf)
```

control-plane/rest/openapi-specs/v0_api_spec.yaml
```diff
@@ -3816,6 +3922,87 @@ components:
...
+      type: object
+      properties:
+        id:
+          type: string
```
tiagolobocastro: Shouldn't this be AppNodeId?

control-plane/rest/openapi-specs/v0_api_spec.yaml
```diff
@@ -3816,6 +3922,87 @@ components:
...
+      required:
+        - id
+        - spec
+    AppNodeSpec:
+      description: |-
+        App node attributes.
+      type: object
+      properties:
+        id:
+          type: string
```
tiagolobocastro: AppNodeId here too

control-plane/rest/openapi-specs/v0_api_spec.yaml
```diff
@@ -3816,6 +3922,87 @@ components:
...
+    AppNodeState:
+      description: |-
+        Deemed state of the app node.
+      type: object
+      properties:
+        id:
+          type: string
+        endpoint:
+          type: string
+          format: hostname
```
tiagolobocastro: hmm is hostname correct for endpoint? or uri?

tiagolobocastro (commented):
lgtm though I wonder about the labels, wdyt?

control-plane/rest/openapi-specs/v0_api_spec.yaml
Abhinandan-Purkait: maybe two different label fields, like `labels` and `control_plane_labels`. We can go with map of maps but I think it would be better to have separate fields dedicated to it's own purpose, if that makes it easier to understand.

tiagolobocastro: the different labels fields work for 2, but not for 3.
let's say you also wanted to also keep k8s specific labels?

tiagolobocastro (approved):

control-plane/rest/openapi-specs/v0_api_spec.yaml
Abhinandan-Purkait: As discussed will, keep it for later.

Abhinandan-Purkait:
bors merge

PR #730, opened 2024-01-31
feat(app node): register csi nodes to control plane

Abhinandan-Purkait:
 
 Changes in this PR:
    - This adds the registration logic for csi-node to control-plane via rest api.
    - This csi-node info would be used to issue FIFREEZE and FITHAW via csi-controller.
    - Moves the rest api client to the common csi-driver library to be used by both csi-controller and csi-node.

### DIFF
--- a/control-plane/csi-driver/src/bin/controller/main.rs
+++ b/control-plane/csi-driver/src/bin/controller/main.rs
@@ -1,6 +1,5 @@
-use tracing::info;
-
-use clap::{Arg, ArgMatches};
+use client::{ApiClientError, CreateVolumeTopology, RestApiClient};
+use config::CsiControllerConfig;
 mod client;
 mod config;
 mod controller;
@@ -8,8 +7,8 @@ mod identity;
 mod pvwatcher;
 mod server;
 
-use client::{ApiClientError, CreateVolumeTopology, IoEngineApiClient};
-use config::CsiControllerConfig;
+use clap::{Arg, ArgMatches};
+use tracing::info;
 
 const CSI_SOCKET: &str = "/var/tmp/csi.sock";
 const CONCURRENCY_LIMIT: usize = 10;
@@ -18,7 +17,7 @@ const REST_TIMEOUT: &str = "30s";
 /// Initialize all components before starting the CSI controller.
 fn initialize_controller(args: &ArgMatches) -> anyhow::Result<()> {
     CsiControllerConfig::initialize(args)?;
-    IoEngineApiClient::initialize()
+    RestApiClient::initialize()
         .map_err(|error| anyhow::anyhow!("Failed to initialize API client, error = {error}"))?;
     Ok(())
 }
@@ -27,7 +26,7 @@ fn initialize_controller(args: &ArgMatches) -> anyhow::Result<()> {
 async fn ping_rest_api() {
     info!("Checking REST API endpoint accessibility ...");
 
-    match IoEngineApiClient::get_client().list_nodes().await {
+    match RestApiClient::get_client().list_nodes().await {
         Err(error) => tracing::error!(?error, "REST API endpoint is not accessible"),
         Ok(nodes) => {
             let names: Vec<String> = nodes.into_iter().map(|n| n.id).collect();
--- a/control-plane/csi-driver/src/bin/node/main_.rs
+++ b/control-plane/csi-driver/src/bin/node/main_.rs
@@ -21,7 +21,6 @@ use stor_port::platform;
 use utils::tracing_telemetry::{FmtLayer, FmtStyle};
 
 use crate::client::AppNodesClientWrapper;
-use anyhow::anyhow;
 use clap::Arg;
 use futures::TryFutureExt;
 use serde_json::json;
@@ -103,16 +102,15 @@ pub(super) async fn main() -> anyhow::Result<()> {
             Arg::new("rest-endpoint")
                 .long("rest-endpoint")
                 .env("ENDPOINT")
-                .default_value("http://ksnode-1:30011")
                 .help("A URL endpoint to the control plane's rest endpoint")
-                .required(true),
         )
         .arg(
-            Arg::new("instance-endpoint")
-                .long("instance-endpoint")
-                .env("MY_POD_IP")
-                .help("Endpoint of current instance")
-                .required(true)
+            Arg::new("enable-registration")
+                .long("enable-registration")
+                .action(clap::ArgAction::SetTrue)
+                .value_name("BOOLEAN")
+                .requires("rest-endpoint")
+                .help("Enable registration of the csi node with the control plane")
         )
         .arg(
             Arg::new("csi-socket")
@@ -133,10 +131,9 @@ pub(super) async fn main() -> anyhow::Result<()> {
             Arg::new("grpc-endpoint")
                 .short('g')
                 .long("grpc-endpoint")
-                .value_name("NAME")
+                .value_name("ENDPOINT")
                 .help("ip address where this instance runs, and optionally the gRPC port")
-                .default_value("0.0.0.0")
-                .required(false)
+                .required(true)
         )
         .arg(
             Arg::new("v")
@@ -299,32 +296,31 @@ pub(super) async fn main() -> anyhow::Result<()> {
     }
 
     // Initialize the rest api client.
-    let client =
-        AppNodesClientWrapper::initialize(matches.get_one::<String>("rest-endpoint").unwrap())?;
+    let client = AppNodesClientWrapper::initialize(matches.get_one::<String>("rest-endpoint"))?;
+
+    let registration_enabled = matches.get_flag("enable-registration");
 
     // Parse instance and grpc endpoints from the command line arguments and validate.
-    let (instance_sock_addr, grpc_sock_addr) = validate_endpoints(
-        matches.get_one::<String>("instance-endpoint").unwrap(),
+    let grpc_sock_addr = validate_endpoints(
         matches.get_one::<String>("grpc-endpoint").unwrap(),
+        registration_enabled,
     )?;
 
-    // Start the CSI server, node plugin grpc server and registration loop.
+    // Start the CSI server, node plugin grpc server and registration loop if registration is
+    // enabled.
     *crate::config::config().nvme_as_mut() = TryFrom::try_from(&matches)?;
-    tokio::select! {
-        result = CsiServer::run(csi_socket, &matches)? => {
-            result?;
-        }
-        result = NodePluginGrpcServer::run(grpc_sock_addr) => {
-            result?;
-        }
-        _ = run_registration_loop(node_name.clone(), instance_sock_addr.to_string(), Some(csi_labels), &client) => {}
-    }
-
-    // Deregister the node from the control plane on termination.
-    if let Err(error) = client.deregister_app_node(node_name).await {
-        error!("Failed to deregister node, {:?}", error);
-    }
-    Ok(())
+    let (csi, grpc, registration) = tokio::join!(
+        CsiServer::run(csi_socket, &matches)?,
+        NodePluginGrpcServer::run(grpc_sock_addr),
+        run_registration_loop(
+            node_name.clone(),
+            grpc_sock_addr.to_string(),
+            Some(csi_labels),
+            &client,
+            registration_enabled
+        )
+    );
+    vec![csi, grpc, registration].into_iter().collect()
 }
 
 struct CsiServer {}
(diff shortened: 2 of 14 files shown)


control-plane/csi-driver/src/client.rs
```diff
@@ -483,4 +463,42 @@ impl IoEngineApiClient {
 
         Ok(snapshot.into_body())
     }
+
+    /// Register a frontend node.
+    pub async fn register_app_node(
```
tiagolobocastro: I'd say leave this as is with pub(crate) in the csi-controller and add a new file for csi-node with just the bits we need - ie app-node stuff

control-plane/csi-driver/src/bin/node/registration.rs
```diff
@@ -0,0 +1,49 @@
+use csi_driver::client::RestApiClient;
+use snafu::Snafu;
+use std::{collections::HashMap, time::Duration};
+use tokio::task::JoinError;
+use tracing::trace;
+
+#[derive(Debug, Snafu)]
+#[snafu(visibility(pub(crate)), context(suffix(false)))]
```
tiagolobocastro: Don't seem to be making use of this

```suggestion
```

control-plane/csi-driver/src/bin/node/registration.rs
```diff
@@ -0,0 +1,49 @@
...
+    tokio::spawn(async move {
+        loop {
+            let interval_duration = match client.register_app_node(&id, &endpoint, &labels).await {
+                Ok(_) => {
+                    // If register_app_node is successful, set the interval to
+                    // REGISTRATION_INTERVAL_ON_SUCCESS.
+                    REGISTRATION_INTERVAL_ON_SUCCESS
+                }
+                Err(e) => {
+                    trace!("Failed to register app node: {:?}", e);
```
tiagolobocastro: Can we have similar to the io-engine, first error register, then retry silently?

control-plane/csi-driver/src/bin/node/registration.rs
```diff
@@ -0,0 +1,49 @@
...
+                    // If register_app_node fails, set the interval to
+                    // REGISTRATION_INTERVAL_ON_ERROR.
+                    REGISTRATION_INTERVAL_ON_ERROR
+                }
+            };
+
+            // Sleep for the interval duration.
```
tiagolobocastro: code is clear, comment doesn't add anything :) 
```suggestion
```

control-plane/csi-driver/src/bin/node/registration.rs
```diff
@@ -0,0 +1,49 @@
...
+    labels: Option<HashMap<String, String>>,
+) -> Result<(), RegistrationError> {
+    let client = RestApiClient::get_client();
+
+    tokio::spawn(async move {
+        loop {
+            let interval_duration = match client.register_app_node(&id, &endpoint, &labels).await {
+                Ok(_) => {
+                    // If register_app_node is successful, set the interval to
+                    // REGISTRATION_INTERVAL_ON_SUCCESS.
```
tiagolobocastro: code is clear, comment doesn't add anything :) 
same on error
```suggestion
```

control-plane/csi-driver/src/bin/node/registration.rs
```diff
@@ -0,0 +1,49 @@
...
+                    REGISTRATION_INTERVAL_ON_ERROR
+                }
+            };
+
+            // Sleep for the interval duration.
+            tokio::time::sleep(interval_duration).await;
+        }
+    })
+    .await
+    .map_err(|error| RegistrationError::TokioTaskWait { source: error })?;
```
tiagolobocastro: Why do we wait to await for it? there's no break out of the loop

control-plane/csi-driver/src/bin/controller/main.rs
```diff
@@ -1,24 +1,61 @@
...
+        .with_concurrency_limit(Some(concurrency_limit))
+        .build_url(url)
+        .map_err(|error| {
+            anyhow::anyhow!(
+                "Failed to create openapi configuration, Error: '{:?}'",
+                error
+            )
+        })?;
+
+    REST_CLIENT.get_or_init(|| RestApiClient {
```
tiagolobocastro: can we not use the oncecell now? or using it without self?

control-plane/csi-driver/src/bin/node/main_.rs
```diff
@@ -93,10 +97,26 @@ impl AsyncWrite for UnixStream {
...
+        .arg(
+            Arg::new("rest-endpoint")
+                .long("rest-endpoint")
+                .env("ENDPOINT")
+                .default_value("http://ksnode-1:30011")
+                .help("A URL endpoint to the control plane's rest endpoint")
+                .required(true),
+        )
+        .arg(
+            Arg::new("instance-endpoint")
```
tiagolobocastro: what's this?

control-plane/csi-driver/src/bin/node/registration.rs
tiagolobocastro: After reading the code on main, I think what you want instead is just to use the function as is, don't need to put the code on a task.

control-plane/csi-driver/src/client.rs
Abhinandan-Purkait: Not sure if we can do that, we would have to duplicate the RestApiClient structure at both places then.

control-plane/csi-driver/src/bin/controller/main.rs
Abhinandan-Purkait: I did not to change that bit of code, as it's add lot of unrelated to changes in this PR.

control-plane/csi-driver/src/bin/node/main_.rs
Abhinandan-Purkait: Ip of the pod, or node for a daemonset, which would be sent as a part of registration.

control-plane/csi-driver/src/bin/controller/main.rs
Abhinandan-Purkait: I added the csi-node code, without once cell. Do you want me to add the csi-controller without once cell in this PR?

control-plane/csi-driver/src/bin/node/registration.rs
Abhinandan-Purkait: As discussed would change that in openapi-generator.

control-plane/csi-driver/src/client.rs
Abhinandan-Purkait: Done.

control-plane/csi-driver/src/bin/node/main_.rs
tiagolobocastro: But we want to send a socket via the registration right?
So maybe we can use similar to other services:
> - "--grpc-endpoint=$(MY_POD_IP):50053"

control-plane/csi-driver/src/bin/node/main_.rs
```diff
@@ -93,10 +94,26 @@ impl AsyncWrite for UnixStream {
...
+    // Parse command line arguments.
     let matches = clap::Command::new(utils::package_description!())
         .about("k8s sidecar for IoEngine implementing CSI among others")
         .version(utils::version_info_str!())
         .subcommand_negates_reqs(true)
+        .arg(
+            Arg::new("rest-endpoint")
+                .long("rest-endpoint")
+                .env("ENDPOINT")
+                .default_value("http://ksnode-1:30011")
```
tiagolobocastro: Should we make this registration optional? Example: if not enabled we don't start these bits
This will allow for us to still test csi-node in isolation, otherwise we're making rest a requirement

control-plane/csi-driver/src/bin/node/main_.rs
```diff
@@ -271,21 +298,31 @@ pub(super) async fn main() -> anyhow::Result<()> {
...
-    } else {
-        format!("{endpoint}:{GRPC_PORT}")
-    }
-    .parse()?;
+    // Initialize the rest api client.
+    let client =
+        AppNodesClientWrapper::initialize(matches.get_one::<String>("rest-endpoint").unwrap())?;
+
+    // Parse instance and grpc endpoints from the command line arguments and validate.
+    let (instance_sock_addr, grpc_sock_addr) = validate_endpoints(
```
tiagolobocastro: Ah so now we have 2... can't we reuse the same endpoint?

control-plane/csi-driver/src/bin/node/registration.rs
```diff
@@ -0,0 +1,37 @@
...
+pub(crate) async fn run_registration_loop(
+    id: String,
+    endpoint: String,
+    labels: Option<HashMap<String, String>>,
+    client: &AppNodesClientWrapper,
+) {
+    let mut logged_error = false;
+    loop {
+        let interval_duration = match client.register_app_node(&id, &endpoint, &labels).await {
+            Ok(_) => REGISTRATION_INTERVAL_ON_SUCCESS,
```
tiagolobocastro: There's a bug here, logged_error needs to be reset on successful registration.
```suggestion
            Ok(_) => {
                if logged_error {
                    tracing::info!("Successfully re-registered the app_node");
                    logged_error = false;
                }
                REGISTRATION_INTERVAL_ON_SUCCESS
            }
```

control-plane/csi-driver/src/bin/node/main_.rs
```diff
@@ -345,37 +382,59 @@ impl CsiServer {
...
+                .ok_or(anyhow!("gRPC endpoint must have a port"))?
+        )
+    };
+
+    let instance_url = SocketAddr::from_str(&instance_endpoint)?;
+    let grpc_endpoint_url = SocketAddr::from_str(&grpc_endpoint)?;
+
+    if instance_url.port() != grpc_endpoint_url.port() {
+        return Err(anyhow!(
+            "instance endpoint and gRPC endpoint must have the same port"
```
tiagolobocastro: Seems a bit pointless to do this.
Also what if the ip is different?
Why can't we just reuse grpc_endpoint?

Abhinandan-Purkait: No we can't use the same endpoint, grpc endpoint can be 0.0.0.0:{port},  but the instance endpoint always needs to the ip of the instance. That's why grpc endpoint is a non required field.

control-plane/csi-driver/src/bin/node/registration.rs
Abhinandan-Purkait: Oops, right.

control-plane/csi-driver/src/bin/node/main_.rs
Abhinandan-Purkait: Yeah we can, but from an end user perspective that option would be not useful. Maybe we can add it when we want to have it.

control-plane/csi-driver/src/bin/node/main_.rs
tiagolobocastro: Why don't we change grpc endpoint to not be 0.0.0.0 and be MY_POD_IP?

control-plane/csi-driver/src/bin/controller/main.rs
tiagolobocastro: We can address this later then

control-plane/csi-driver/src/bin/node/main_.rs
tiagolobocastro: Not for the users, for our own testing, example we test csi-node in isolation

control-plane/csi-driver/src/bin/node/main_.rs
Abhinandan-Purkait: But we still need a default value right?

Abhinandan-Purkait: Or we can make it required with the MY_POD_IP maybe

control-plane/csi-driver/src/bin/node/main_.rs
Abhinandan-Purkait: Yeah, we can any naming suggestions for the cli arg?

Abhinandan-Purkait: Done.

control-plane/csi-driver/src/bin/node/main_.rs
Abhinandan-Purkait: Done.

control-plane/csi-driver/src/bin/node/main_.rs
tiagolobocastro: Rest should be optional, so no default value 

```suggestion
```

control-plane/csi-driver/src/bin/node/main_.rs
```diff
@@ -93,10 +93,26 @@ impl AsyncWrite for UnixStream {
...
+                .long("rest-endpoint")
+                .env("ENDPOINT")
+                .default_value("http://ksnode-1:30011")
+                .help("A URL endpoint to the control plane's rest endpoint")
+                .required(true),
+        )
+        .arg(
+            Arg::new("enable-registration")
+                .long("enable-registration")
+                .action(clap::ArgAction::SetTrue)
```
tiagolobocastro, line 110: we can add a requires("rest-endpoint")

tiagolobocastro (approved):

Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors merge

PR #731, opened 2024-02-01
feat(fs consistent snaps): integrate fs consistency code to snapshot creation flow

Abhinandan-Purkait:
File System Consistent Snapshot

Goals:
- Issue FIFREEZE and FITHAW before and after snapshot creation.
- Filesystem to be consistent before and after snapshot creation.

Changes in this PR:
- This PR integrates the fsfreeze and unfreeze before and after snapshot creation.
- This adds an parameter to VolumeSnapshotClass to enabled/disabled fs consistency.
- fs consistency would be enabled by default on the snapshot creation.

### DIFF
--- a/control-plane/csi-driver/src/bin/controller/controller.rs
+++ b/control-plane/csi-driver/src/bin/controller/controller.rs
@@ -1,7 +1,6 @@
 use crate::{
-    client::ListToken, ApiClientError, CreateVolumeTopology, CsiControllerConfig, IoEngineApiClient,
+    client::ListToken, ApiClientError, CreateVolumeTopology, CsiControllerConfig, RestApiClient,
 };
-
 use csi_driver::context::{CreateParams, PublishParams};
 use rpc::csi::{volume_content_source::Type, Topology as CsiTopology, *};
 use stor_port::types::v0::openapi::{
@@ -289,7 +288,7 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
         let mut volume_context = args.parameters.clone();
 
         // First check if the volume already exists.
-        match IoEngineApiClient::get_client()
+        match RestApiClient::get_client()
             .get_volume_for_create(&parsed_vol_uuid)
             .await
         {
@@ -318,7 +317,7 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
 
                 let volume = match volume_content_source {
                     Some(snapshot_uuid) => {
-                        IoEngineApiClient::get_client()
+                        RestApiClient::get_client()
                             .create_snapshot_volume(
                                 &parsed_vol_uuid,
                                 &snapshot_uuid,
@@ -331,7 +330,7 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
                             .await?
                     }
                     None => {
-                        IoEngineApiClient::get_client()
+                        RestApiClient::get_client()
                             .create_volume(
                                 &parsed_vol_uuid,
                                 replica_count,
@@ -388,7 +387,7 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
             Status::invalid_argument(format!("Malformed volume UUID: {}", args.volume_id))
         })?;
         let _guard = csi_driver::limiter::VolumeOpGuard::new(volume_uuid)?;
-        IoEngineApiClient::get_client()
+        RestApiClient::get_client()
             .delete_volume(&volume_uuid)
             .await
             .map_err(|e| {
@@ -437,9 +436,7 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
         }
 
         // Check if the volume is already published.
-        let volume = IoEngineApiClient::get_client()
-            .get_volume(&volume_id)
-            .await?;
+        let volume = RestApiClient::get_client().get_volume(&volume_id).await?;
 
         let params = PublishParams::try_from(&args.volume_context)?;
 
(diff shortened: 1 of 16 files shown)


control-plane/csi-driver/src/bin/controller/controller.rs
```diff
@@ -74,6 +78,66 @@ fn parse_protocol(proto: Option<&String>) -> Result<VolumeShareProtocol, Status>
...
+        None => None,
+    }
+}
+
+async fn issue_fs_freeze(endpoint: String, volume_id: String) -> Result<(), Status> {
+    trace!(
+        "Issuing fs freeze for volume {}, endpoint {}",
+        volume_id,
+        endpoint
+    );
```
tiagolobocastro: I think you can also add enter automatically

```suggestion
#[tracing::instrument]
async fn issue_fs_freeze(endpoint: String, volume_id: String) -> Result<(), Status> {
    trace!("Issuing fs freeze");
```

control-plane/csi-driver/src/bin/controller/controller.rs
```diff
@@ -74,6 +78,66 @@ fn parse_protocol(proto: Option<&String>) -> Result<VolumeShareProtocol, Status>
...
+    let mut client = NodePluginClient::connect(format!("http://{endpoint}"))
```
tiagolobocastro, line 97: Where does this come from? Does it have the tracing context support?

control-plane/csi-driver/src/bin/controller/controller.rs
```diff
@@ -74,6 +78,66 @@ fn parse_protocol(proto: Option<&String>) -> Result<VolumeShareProtocol, Status>
...
+        .map_err(|error| Status::failed_precondition(error.to_string()))?;
+    match client
+        .freeze_fs(Request::new(FreezeFsRequest {
+            volume_id: volume_id.clone(),
+        }))
+        .await
+    {
+        Ok(_) => Ok(()),
+        Err(status) if status.code() == tonic::Code::Unimplemented => {
+            trace!("fs unfreeze not supported for raw block volume: {volume_id}");
```
tiagolobocastro, line 108: If we add instrument above, then volume_id is automatically added.
```suggestion
            trace!("fs unfreeze not supported for raw block volume");
```
I wonder if not supported or something like that is a better error code here

control-plane/csi-driver/src/bin/controller/controller.rs
```diff
@@ -795,15 +859,73 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
...
             Status::invalid_argument(format!("Malformed snapshot ID: {}", request.name))
         })?;
+        let create_params = CreateSnapshotParams::try_from(&request.parameters)?;
+
+        // Get the volume object. Extract the app node endpoint if the quiesce is requested.
+        let volume = RestApiClient::get_client().get_volume(&volume_uuid).await?;
+        let app_node_endpoint_info = match get_volume_app_node(&volume) {
+            // Volume is not published, so no need to quiesce.
+            None => None,
+            Some(info) => match create_params.queisce().clone() {
```
```suggestion
            Some(info) => match create_params.queisce() {
```

tiagolobocastro: Also typo: it's quiesce

control-plane/csi-driver/src/bin/controller/controller.rs
```diff
@@ -795,15 +859,73 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
...
+                // Create the snapshot.
+                RestApiClient::get_client()
+                    .create_volume_snapshot(&volume_uuid, &snap_uuid)
+                    .await
+                    .map_err(|error| match error {
+                        ApiClientError::ResourceExhausted(reason) => {
+                            Status::resource_exhausted(reason)
+                        }
+                        ApiClientError::PreconditionFailed(reason) => {
+                            Status::resource_exhausted(reason)
```
tiagolobocastro, line 895: Why do this convert to resource exhausted?

control-plane/csi-driver/src/context.rs
```diff
@@ -392,6 +394,40 @@ fn generate_sts_affinity_group_name(
...
+
+/// Snapshot creation parameters.
+#[allow(dead_code)]
+#[derive(Debug)]
+pub struct CreateSnapshotParams {
+    queisce: Option<QuiesceCandidate>,
+}
+impl CreateSnapshotParams {
+    /// Get the `Parameters::queisce` value.
+    pub fn queisce(&self) -> &Option<QuiesceCandidate> {
```
```suggestion
    /// Get the `Parameters::quiesce` value.
    pub fn quiesce(&self) -> &Option<QuiesceCandidate> {
```

control-plane/csi-driver/src/bin/controller/controller.rs
Abhinandan-Purkait: This is pre-existing bit of code to handle the snapshot-controller. We cannot return precondition failed on snapshot failure as that would not allow for deletion.

control-plane/csi-driver/src/bin/controller/controller.rs
Abhinandan-Purkait:
```
    /// Operation is not implemented or not supported.
    Unimplemented = 12,
```
    
  unimplement itself is for not supported as per docs

control-plane/csi-driver/src/bin/controller/controller.rs
Abhinandan-Purkait: No it doesn't have tracing.

control-plane/csi-driver/src/bin/controller/controller.rs
Abhinandan-Purkait: Would this work, don't think telemetry is setup.

tiagolobocastro: doh yeah this won't work, it's not that telemetry is not setup, but tracing is not configured to output fields to stdout...

tiagolobocastro: My previous comment is incorrect, tracing is configured to output fields to the output so we can do what I suggested initially

control-plane/csi-driver/src/bin/controller/controller.rs
tiagolobocastro: Alright, we can add this later.

control-plane/csi-driver/src/bin/controller/controller.rs
tiagolobocastro: The problem is that unimplemented in CSI is usually for the operation itself, not for specifics, so this could be seen if you're running new csi-controller and old csi-node for example.

tiagolobocastro: Wonder if we should be doing a sync for the block volumes as well (not related to this PR)

control-plane/csi-driver/src/bin/controller/controller.rs
```diff
@@ -74,6 +78,66 @@ fn parse_protocol(proto: Option<&String>) -> Result<VolumeShareProtocol, Status>
     }
 }
 
+/// Get app node where volume frontend is situated.
+fn get_volume_app_node(volume: &Volume) -> Option<String> {
```
```suggestion
fn volume_app_node(volume: &Volume) -> Option<String> {
```

control-plane/csi-driver/src/bin/controller/controller.rs
```diff
@@ -74,6 +78,66 @@ fn parse_protocol(proto: Option<&String>) -> Result<VolumeShareProtocol, Status>
...
+    match volume.spec.target.as_ref() {
+        // todo: Might need to deal with multiple app nodes later?
+        Some(target) => target
+            .frontend_nodes
+            .as_ref()
+            .map(|nodes| nodes[0].name.clone()),
```
tiagolobocastro: This will crash if vector is empty vec, which can happen
```suggestion
            .and_then(|nodes| nodes.get(0))
            .map(|node| node.name.clone()),
```

control-plane/csi-driver/src/context.rs
```diff
@@ -392,6 +394,40 @@ fn generate_sts_affinity_group_name(
...
+    }
+}
+impl TryFrom<&HashMap<String, String>> for CreateSnapshotParams {
+    type Error = tonic::Status;
+
+    fn try_from(args: &HashMap<String, String>) -> Result<Self, Self::Error> {
+        let queisce = match args.get(Parameters::Quiesce.as_ref()) {
+            Some(fs) => QuiesceCandidate::from_str(fs.as_str())
+                .map(Some)
+                .map_err(|_| tonic::Status::invalid_argument("Invalid queisce type"))?,
```
```suggestion
                .map_err(|_| tonic::Status::invalid_argument("Invalid quiesce type"))?,
```

control-plane/csi-driver/src/context.rs
```diff
@@ -57,6 +57,8 @@ pub enum Parameters {
     PoolTopologyAffinity,
     #[strum(serialize = "poolTopologySpread")]
     PoolTopologySpread,
+    #[strum(serialize = "quiesce")]
+    Quiesce,
```
tiagolobocastro, line 61: I'm wondering if this should be two fold, example quiesceFs and quiesceBlock?

tiagolobocastro (approved):

control-plane/csi-driver/src/bin/controller/controller.rs
Abhinandan-Purkait: Yeah, any tonic status code that suits here?

Abhinandan-Purkait: Yeah we can add sync for block volumes. I will take that up as a separate feature request as that's not a part of this.

control-plane/csi-driver/src/context.rs
Abhinandan-Purkait: Or maybe just add a Block Variant to the Quiesce? That we can extend once we decide to do sync for block?

tiagolobocastro: yeah perhaps

control-plane/csi-driver/src/bin/controller/controller.rs
tiagolobocastro: Probably invalid argument?

PR #734, opened 2024-02-02
feat(volume/resize): reclaim capacity from replicas

dsharma-dc:
Volume replicas could have been possibly expanded as part of volume resize,
but the volume expansion failed overall leaving behind expanded replicas.
Since the overall operation failed, nexus size is still the original one and
hence the expanded capacity in replicas is wasteful. So we bring back the
replicas' size to what is expected by the volume spec.

### DIFF
--- a/control-plane/agents/src/bin/core/controller/reconciler/mod.rs
+++ b/control-plane/agents/src/bin/core/controller/reconciler/mod.rs
@@ -9,7 +9,7 @@ mod volume;
 
 pub(crate) use crate::controller::task_poller::PollTriggerEvent;
 use crate::controller::task_poller::{
-    squash_results, PollContext, PollEvent, PollResult, TaskPoller,
+    squash_results, PollContext, PollEvent, PollResult, PollerState, TaskPoller,
 };
 use poller::ReconcilerWorker;
 use std::fmt::Debug;
@@ -99,6 +99,11 @@ trait GarbageCollect {
     async fn disown_orphaned(&mut self, context: &PollContext) -> PollResult;
     /// Disown resources which have questionable existence, for example non reservable replicas.
     async fn disown_invalid(&mut self, context: &PollContext) -> PollResult;
+    /// Reclaim unused capacity - for example an expanded but unused replica, which may
+    /// happen as part of a failed volume expand operation.
+    async fn reclaim_space(&mut self, _context: &PollContext) -> PollResult {
+        PollResult::Ok(PollerState::Idle)
+    }
 }
 
 #[async_trait::async_trait]
--- a/control-plane/agents/src/bin/core/controller/reconciler/volume/garbage_collector.rs
+++ b/control-plane/agents/src/bin/core/controller/reconciler/volume/garbage_collector.rs
@@ -1,7 +1,7 @@
 use crate::controller::{
     reconciler::{GarbageCollect, PollContext, TaskPoller},
     resources::{
-        operations::ResourceLifecycle,
+        operations::{ResourceLifecycle, ResourceResize},
         operations_helper::{OperationSequenceGuard, SpecOperationsHelper},
         OperationGuardArc, TraceSpan, TraceStrLog,
     },
@@ -17,7 +17,7 @@ use stor_port::types::v0::{
     store::{
         nexus::NexusSpec, nexus_persistence::NexusInfo, replica::ReplicaSpec, volume::VolumeSpec,
     },
-    transport::{DestroyVolume, NexusOwners, ReplicaOwners, VolumeStatus},
+    transport::{DestroyVolume, NexusOwners, ReplicaOwners, ResizeReplica, VolumeStatus},
 };
 use tracing::Instrument;
 
@@ -71,6 +71,7 @@ impl GarbageCollect for OperationGuardArc<VolumeSpec> {
             self.destroy_deleting(context).await,
             self.disown_unused(context).await,
             self.disown_invalid(context).await,
+            self.reclaim_space(context).await,
         ])
     }
 
@@ -98,6 +99,13 @@ impl GarbageCollect for OperationGuardArc<VolumeSpec> {
     // disown them.
     async fn disown_invalid(&mut self, context: &PollContext) -> PollResult {
         disown_non_reservable_replicas(self, context).await
+    }
+
+    // Volume replicas could have been possibly expanded as part of volume resize,
+    // but the volume expansion failed overall leaving behind expanded replicas. So we
+    // bring back the replica size to what is expected by the volume spec.
+    async fn reclaim_space(&mut self, context: &PollContext) -> PollResult {
+        revert_expanded_replicas(self, context).await
     }
 }
 
@@ -248,6 +256,57 @@ async fn disown_unused_replicas(
     }
 
     GarbageCollector::squash_results(results)
+}
+
+/// Volume replicas could have been possibly expanded as part of volume resize,
+/// but the volume expansion failed overall leaving behind expanded replicas. So we
+/// bring back the replica size to what is expected by the volume spec.
+#[tracing::instrument(level = "debug", skip(context, volume), fields(volume.uuid = %volume.uuid(), request.reconcile = true))]
+async fn revert_expanded_replicas(
+    volume: &mut OperationGuardArc<VolumeSpec>,
+    context: &PollContext,
+) -> PollResult {
+    let target = context.specs().volume_target_nexus_rsc(volume.as_ref());
+    if let Some(tgt) = target {
+        let nexus_size = tgt.lock().size;
+        // panic only in debug environments e.g certain tests.
+        debug_assert_eq!(nexus_size, volume.as_ref().size);
+        if nexus_size != volume.as_ref().size {
+            // This should never happen, but just emit a warning.
+            volume.warn_span(|| tracing::info!(volume_size = %volume.as_ref().size, nexus_size = %nexus_size, "Nexus and Volume size mismatch!"));
+        }
+        // If there is a target present, we don't try to reclaim space from replicas because the
+        // attempt to downsize replica/lvol will potentially fail with -EBUSY if the lvol bdev
+        // has open descriptors.
+        return PollResult::Ok(PollerState::Idle);
+    }
+
+    for replica in context.specs().volume_replicas(volume.uuid()) {
+        let mut replica = match replica.operation_guard() {
+            Ok(guard) => guard,
+            Err(_) => continue,
+        };
+
+        if replica.as_ref().size > volume.as_ref().size {
+            let replica_state = context.registry().replica(replica.uuid()).await?;
+            volume.info_span(|| tracing::info!(replica = %replica.as_ref().uuid, replica_size = %replica.as_ref().size, "Reclaiming space from replica. Volume size {}", volume.as_ref().size));
+            // Resize replica back to volume spec size.
+            replica
+                .resize(
+                    context.registry(),
+                    &ResizeReplica::new(
+                        &replica_state.node,
+                        replica.as_ref().pool_name(),
+                        None,
+                        replica.uuid(),
+                        volume.as_ref().size,
+                    ),
+                )
+                .await?;
+        }
+    }
+
+    PollResult::Ok(PollerState::Idle)
 }
 
 /// After republishing if the nexus was not shutdown gracefully, we should disown its local


dsharma-dc:
bors try

dsharma-dc:
bors try

dsharma-dc:
> ## try
> Build failed:
> 
> * [continuous-integration/jenkins/branch](https://mayastor-ci.mayadata.io/job/Mayastor-Control-Plane/job/trying/8/display/redirect)

@tiagolobocastro There is still some unreliability in the test(always test_expand_a_published_volume_after_unpublishing_it_while_having_an_offline_replica), but not only this I see test_mixed.py also has a failure which seems unrelated here.
Any idea what can be improved here?

aland-zhang:
Volume online expansion k8s csi feature is very important，Hope function 'async fn controller_expand_volume' can be improved!
Thank you so much.

tiagolobocastro:
> > ## try
> > Build failed:
> > 
> > * [continuous-integration/jenkins/branch](https://mayastor-ci.mayadata.io/job/Mayastor-Control-Plane/job/trying/8/display/redirect)
> 
> @tiagolobocastro There is still some unreliability in the test(always test_expand_a_published_volume_after_unpublishing_it_while_having_an_offline_replica), but not only this I see test_mixed.py also has a failure which seems unrelated here. Any idea what can be improved here?

hmm please take a look at the expand test, I'll have a look at the mixed test

dsharma-dc:
bors try

dsharma-dc:
> > > ## try
> > > Build failed:
> > > 
> > > * [continuous-integration/jenkins/branch](https://mayastor-ci.mayadata.io/job/Mayastor-Control-Plane/job/trying/8/display/redirect)
> > 
> > 
> > @tiagolobocastro There is still some unreliability in the test(always test_expand_a_published_volume_after_unpublishing_it_while_having_an_offline_replica), but not only this I see test_mixed.py also has a failure which seems unrelated here. Any idea what can be improved here?
> 
> hmm please take a look at the expand test, I'll have a look at the mixed test

Test fixed. Total runtime is now 25-30secs.
 Ran 30-40 iterations and all passed. Let's see here with bors.

dsharma-dc:
bors try

dsharma-dc:
bors try

dsharma-dc:
> * continuous-integration/jenkins/branch

Need to reduce the retry interval , otherwise the rebuild is finishing fast and replica is visible as Online.

dsharma-dc:
bors try

control-plane/agents/src/bin/core/controller/reconciler/mod.rs
```diff
@@ -99,6 +99,11 @@ trait GarbageCollect {
     async fn disown_orphaned(&mut self, context: &PollContext) -> PollResult;
     /// Disown resources which have questionable existence, for example non reservable replicas.
     async fn disown_invalid(&mut self, context: &PollContext) -> PollResult;
+    /// Reclaim unused capacity - for example an expanded but unused replica, which may
+    /// hapen as part of a failed volume expand operation.
+    async fn reclaim_space(&mut self, _context: &PollContext) -> PollResult {
+        PollResult::Ok(PollerState::Idle)
+    }
```
tiagolobocastro: Wonder if this should be in a separate trait, ex: reclamation

```suggestion
    /// Reclaim unused capacity - for example an expanded but unused replica, which may
    /// happen as part of a failed volume expand operation.
    async fn reclaim_space(&mut self, _context: &PollContext) -> PollResult {
        PollResult::Ok(PollerState::Idle)
    }
```

control-plane/agents/src/bin/core/controller/reconciler/volume/garbage_collector.rs
```diff
@@ -250,6 +258,53 @@ async fn disown_unused_replicas(
...
+/// but the volume expansion failed overall leaving behind expanded replicas. So we
+/// bring back the replica size to what is expected by the volume spec.
+#[tracing::instrument(level = "debug", skip(context, volume), fields(volume.uuid = %volume.uuid(), request.reconcile = true))]
+async fn revert_expanded_replicas(
+    volume: &mut OperationGuardArc<VolumeSpec>,
+    context: &PollContext,
+) -> PollResult {
+    let target = context.specs().volume_target_nexus_rsc(volume.as_ref());
+    if let Some(tgt) = target {
+        let nexus_size = tgt.immutable_ref().size;
```
tiagolobocastro: immutable_ref is used for very specific cases where fields never change..
Also I suspect immutable_ref is probably no longer needed, needs to be checked.

```suggestion
        let nexus_size = tgt.as_ref().size;
```

control-plane/agents/src/bin/core/controller/reconciler/volume/garbage_collector.rs
```diff
@@ -250,6 +258,53 @@ async fn disown_unused_replicas(
...
+        debug_assert_eq!(nexus_size, volume.as_ref().size);
```
tiagolobocastro, line 273: Let's not crash...
```suggestion
```

control-plane/agents/src/bin/core/controller/reconciler/volume/garbage_collector.rs
```diff
@@ -250,6 +258,53 @@ async fn disown_unused_replicas(
...
+        }
+    }
+
+    for replica in context.specs().volume_replicas(volume.uuid()) {
+        let mut replica = match replica.operation_guard() {
+            Ok(guard) => guard,
+            Err(_) => continue,
+        };
+
+        if (replica.as_ref().size != volume.as_ref().size) && volume.as_ref().operation.is_none() {
```
tiagolobocastro: Isn't this check done up the call stack? volume.as_ref().operation.is_none()

```suggestion
        if (replica.as_ref().size != volume.as_ref().size) {
```

control-plane/agents/src/bin/core/tests/volume/resize.rs
```diff
@@ -179,3 +193,191 @@ async fn resize_on_no_capacity_pool() {
...
+                .await
+                .unwrap();
+        }
+    }
+}
+
+// Create lots of volume on the pool and then resize a volume.
+// We'll create single replica volumes and two 800 MiB pools are available, so
+// let's create about 60 volumes of 20MiB each so we have some space left for
+// expansion of one of the volume.
```
tiagolobocastro: Why not just 1 volume of 1.2GiB ?

control-plane/agents/src/bin/core/tests/volume/resize.rs
```diff
@@ -179,3 +193,191 @@ async fn resize_on_no_capacity_pool() {
...
+                    },
+                    None,
+                )
+                .await
+                .unwrap();
+        }
+    }
+}
+
+// Create lots of volume on the pool and then resize a volume.
```
tiagolobocastro: What are we testing here?

control-plane/agents/src/bin/core/volume/specs.rs
```diff
@@ -828,6 +829,53 @@ impl ResourceSpecsLocked {
         }
     }
 
+    pub(crate) fn check_capacity_limit_for_resize(
```
tiagolobocastro: How is this enforcing the limit when we have >1 concurrent calls?

tests/bdd/features/volume/resize/test_resize_offline.py
```diff
@@ -221,6 +280,7 @@ def one_of_the_replica_is_not_in_online_state():
     volume = ApiClient.volumes_api().get_volume(VOLUME_UUID)
     # Offline NODE3_NAME to make the replica Offline/Unknown
     Docker.stop_container(NODE3_NAME)
+    time.sleep(1)
```
tiagolobocastro: this should use a wait_replica_offline function

tests/bdd/features/volume/resize/test_resize_offline.py
```diff
@@ -238,9 +298,10 @@ def the_volume_is_receiving_io():
 def the_volume_is_unpublished():
     """the volume is unpublished."""
     # Ensure fio is still running till unpublish
-    time.sleep(1)
-    assert pytest.fio.poll() is None
-    unpublish_volume()
+    time.sleep(0.5)
+    if pytest.fio.poll() is None:
```
tiagolobocastro: What does poll None mean again, that it's still running?

tests/bdd/features/volume/resize/test_resize_offline.py
```diff
@@ -388,58 +451,32 @@ def the_volume_should_get_published_with_expanded_capacity():
...
-        publish_volume_body=PublishVolumeBody(
-            {},
-            Protocol("nvmf"),
-            republish=repub,
-            reuse_existing=reuse,
-            node=NODE2_NAME,
-            frontend_node="",
-        ),
-    )
+# Restriable functions - END
```
tiagolobocastro: Restriable ?

control-plane/agents/src/bin/core/tests/volume/resize.rs
dsharma-dc: The idea behind this test is to ensure we don't hang up cpu by holding locks. Hence more volumes' spec locks, plus the top level lock on all specs.

tests/bdd/features/volume/resize/test_resize_offline.py
dsharma-dc: Yes, `None` means running.

tests/bdd/features/volume/resize/test_resize_offline.py
dsharma-dc: `Retriable` :) will fix.

control-plane/agents/src/bin/core/controller/reconciler/volume/garbage_collector.rs
dsharma-dc: Is there a better way to definitively fail(during testing) if we are ever in this situation? Since it's debug assert it's harmless.

control-plane/agents/src/bin/core/controller/reconciler/volume/garbage_collector.rs
dsharma-dc: I don't see `as_ref` available with this type (ResourceMutex<T>). Throws error `method not found in ResourceMutex<NexusSpec>`. It looks to me here either we `lock and peek` using `tgt.lock().size` or use `immutable_ref` which is perhaps ok because the field `size` won't be modified underneath in this path. What do say would be preferred?

control-plane/agents/src/bin/core/volume/specs.rs
dsharma-dc: It's taking exclusive `write` lock on all specs just like `get_or_create_volume` does, while checking limits. Anything more needed here?

tiagolobocastro: That's just while checking it, but you could have 2 request come in a the same time, they each take turns checking, they both think they're in the resize usage, but the last one goes over the limit

control-plane/agents/src/bin/core/controller/reconciler/volume/garbage_collector.rs
tiagolobocastro: Ah right this is on the ResourceMutex... So you need to take the guard or lock it, should be fine here since we hold the volume lock.

control-plane/agents/src/bin/core/controller/reconciler/volume/garbage_collector.rs
tiagolobocastro: So is the intent here that it will fail during our testing only and not for users as they run debug?

control-plane/agents/src/bin/core/tests/volume/resize.rs
tiagolobocastro: Sorry I don't get it, where are we holding locks?

control-plane/agents/src/bin/core/controller/reconciler/volume/garbage_collector.rs
dsharma-dc: > So is the intent here that it will fail during our testing only and not for users as they run debug?

Yes to fail only in testing since users **won't** run debug(at least for all practical purposes).

control-plane/agents/src/bin/core/tests/volume/resize.rs
dsharma-dc: > Sorry I don't get it, where are we holding locks?

I'm referring to the lock during capacity limit checks. Below part:

```
                let specs = self.write();       <------------------- this
                let mut total: u64 = specs
                    .volumes
                    .values()
                    .filter_map(|v| {
                        let vol_locked = v.lock();      <------------------- and this
```

tests/bdd/features/volume/resize/test_resize_offline.py
dsharma-dc: added a function for that purpose.

control-plane/agents/src/bin/core/controller/reconciler/volume/garbage_collector.rs
dsharma-dc: Yea I guess it's redundant to check here because if there was a pending operation we won't be holding the volume guard here.

control-plane/agents/src/bin/core/controller/reconciler/volume/garbage_collector.rs
dsharma-dc: ok, used lock to peek.

control-plane/agents/src/bin/core/volume/specs.rs
dsharma-dc: hmm yeah, for that it seems to me that we need to seal the gap between this check and acquiring volumespec operation guard for resize. That can be achieved by acquiring volumespec operation guard from `check_capacity_limit_for_resize` before the RAII guard on specs gets dropped, if we are within capacity limit, and return that guard from this function. wdyt?

control-plane/agents/src/bin/core/controller/reconciler/mod.rs
dsharma-dc: This call won't be most likely implemented for anything else(perhaps may be Pool at some point in future)? Pretty much fits well here as well. We can separate later also if it tends to become more generic. Let me know...

dsharma-dc:
bors try

control-plane/agents/src/bin/core/tests/volume/resize.rs
tiagolobocastro: I still don't see how this test is testing that? Seems any resize would exercise that code anyway?

control-plane/agents/src/bin/core/volume/specs.rs
tiagolobocastro: Let's remove this and solve it in another PR

control-plane/agents/src/bin/core/controller/reconciler/mod.rs
```diff
@@ -99,6 +99,11 @@ trait GarbageCollect {
     async fn disown_orphaned(&mut self, context: &PollContext) -> PollResult;
     /// Disown resources which have questionable existence, for example non reservable replicas.
     async fn disown_invalid(&mut self, context: &PollContext) -> PollResult;
+    /// Reclaim unused capacity - for example an expanded but unused replica, which may
+    /// happen as part of a failed volume expand operation.
+    async fn reclaim_space(&mut self, _context: &PollContext) -> PollResult {
```
niladrih, line 104:
```suggestion
    async fn reclaim_space(&mut self, _: &PollContext) -> PollResult {
```

control-plane/agents/src/bin/core/controller/reconciler/volume/garbage_collector.rs
niladrih: Could you add a comment saying that this won't be executed for a regular run?

control-plane/agents/src/bin/core/controller/reconciler/volume/garbage_collector.rs
```diff
@@ -250,6 +258,53 @@ async fn disown_unused_replicas(
...
+            volume.warn_span(|| tracing::info!(volume_size = %volume.as_ref().size, nexus_size = %nexus_size, "Nexus and Volume size mismatch!"));
+            return PollResult::Ok(PollerState::Idle);
+        }
+    }
+
+    for replica in context.specs().volume_replicas(volume.uuid()) {
+        let mut replica = match replica.operation_guard() {
+            Ok(guard) => guard,
+            Err(_) => continue,
+        };
```
niladrih, line 288: Are we guaranteed to always succeed on `operation_guard()`? Do we skip replicas if it fails?

niladrih (commented):
Left 1 comment

control-plane/agents/src/bin/core/controller/reconciler/mod.rs
dsharma-dc: let's keep unused with a name, as conventionally followed everywhere in control-plane.

control-plane/agents/src/bin/core/controller/reconciler/volume/garbage_collector.rs
dsharma-dc: not guaranteed to always succeed, so in case of unable to acquire guard we just skip that replica and continue.

control-plane/agents/src/bin/core/volume/specs.rs
dsharma-dc: This PR now contains only reconciler changes.

dsharma-dc:
bors try

control-plane/agents/src/bin/core/controller/reconciler/volume/garbage_collector.rs
```diff
@@ -250,6 +258,53 @@ async fn disown_unused_replicas(
...
+                    context.registry(),
+                    &ResizeReplica::new(
+                        &replica_state.node,
+                        replica.as_ref().pool_name(),
+                        None,
+                        replica.uuid(),
+                        volume.as_ref().size,
+                    ),
+                )
+                .await?;
```
tiagolobocastro, line 305: hmm can this be done with ongoing IO with no disruption?

dsharma-dc: good point. There won't be any disruption to IO. However, the second part of resize lvol call where it updates bdev block count(after adjusting clusters) will keep running into EBUSY till bdev has open descriptors.
Wondering if that can be troublesome much.

dsharma-dc: Simplest would be skip doing this if we find target above at line 270

dsharma-dc: As discussed, not attempting reclaim in case a volume target is found.

control-plane/agents/src/bin/core/controller/reconciler/volume/garbage_collector.rs
```diff
@@ -250,6 +258,56 @@ async fn disown_unused_replicas(
...
+        return PollResult::Ok(PollerState::Idle);
+    }
+
+    for replica in context.specs().volume_replicas(volume.uuid()) {
+        let mut replica = match replica.operation_guard() {
+            Ok(guard) => guard,
+            Err(_) => continue,
+        };
+
+        if replica.as_ref().size != volume.as_ref().size {
```
tiagolobocastro: Should this be:
```suggestion
        if replica.as_ref().size > volume.as_ref().size {
```
Otherwise it could potentially grow rather than shrink?

dsharma-dc: > Otherwise it could potentially grow rather than shrink?

Changed to `>` . PTAL. Theoretically yes it would mistakenly grow, but I don't see a case that can put us in such situation.

tiagolobocastro: hmm what if an "extra" replica was offline so was not resized up as part of the volume resize?
although this is interesting, if this can happen I wonder if we need to expand a replica when we want to add it to the nexus but it's not the correct size?

tiagolobocastro (approved):

control-plane/agents/src/bin/core/controller/reconciler/volume/garbage_collector.rs
dsharma-dc: > hmm what if an "extra" replica was offline so was not resized up as part of the volume resize?

If a replica that's part of a volume is offline, then the pre-check fails and volume resize doesn't initiate at all.

dsharma-dc:
I'll add some enhancements to ReplicaOperations to test this reclaim, in an upcoming PR. Adding resize to ReplicaOperations will allow us to have one of the replica bigger than the volume, and hence picked up this reconciler.

dsharma-dc:
bors try

Abhinandan-Purkait (approved):

dsharma-dc:
bors merge

control-plane/agents/src/bin/core/controller/reconciler/volume/garbage_collector.rs
tiagolobocastro: So if a node goes offline volume resizes become blocked?

dsharma-dc: > So if a node goes offline volume resizes become blocked?

Not blocked I meant, it is rejected with failed precondition(from the scheduling pre-check that we do) saying 
> Replicas  can't be resized due to required capacity ({}) or state(Online) not met.

tiagolobocastro: I mean if a volume replica node is offline then you can't resize until you bring the node online? Even if the replica is not part of the nexus?

PR #737, opened 2024-02-09
refactor: use pagination to load values from pstor

Abhinandan-Purkait:
### Changes in the PR.

- Changes the spec population from etcd from non paged call to a paged call.
- Also incorporates the pagination to legacy migration flows.

### DIFF
--- a/control-plane/agents/src/bin/core/controller/resources/migration.rs
+++ b/control-plane/agents/src/bin/core/controller/resources/migration.rs
@@ -16,10 +16,13 @@ pub const MSP_OPERATOR: &str = "msp-operator";
 pub(crate) async fn migrate_product_v1_to_v2<S: Store>(
     store: &mut S,
     spec_type: StorableObjectType,
+    etcd_max_page_size: i64,
 ) -> Result<(), StoreError> {
     info!("Migrating {spec_type:?} from v1 to v2 key space");
     let prefix = &product_v1_key_prefix_obj(spec_type);
-    let store_entries = store.get_values_prefix(prefix).await?;
+    let store_entries = store
+        .get_values_paged_all(prefix, etcd_max_page_size)
+        .await?;
     for (k, v) in store_entries {
         let id = k
             .split('/')
--- a/control-plane/agents/src/bin/core/tests/controller/mod.rs
+++ b/control-plane/agents/src/bin/core/tests/controller/mod.rs
@@ -1,12 +1,16 @@
 use deployer_cluster::{etcd_client::Client, *};
 use stor_port::{
-    pstor::{etcd::Etcd, StoreObj},
+    pstor::{etcd::Etcd, key_prefix_obj, ApiVersion, StorableObjectType, StoreKv, StoreObj},
     types::v0::{
         openapi::models,
         store::registry::{ControlPlaneService, StoreLeaseOwner, StoreLeaseOwnerKey},
         transport,
     },
 };
+
+use serde_json::Value;
+use std::str::FromStr;
+use uuid::Uuid;
 
 /// Test that the content of the registry is correctly loaded from the persistent store on start up.
 #[tokio::test]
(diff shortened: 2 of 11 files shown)


utils/pstor/src/etcd.rs
```diff
@@ -233,7 +233,6 @@ impl StoreKv for Etcd {
                 Some(
                     GetOptions::new()
                         .with_prefix()
-                        .with_from_key()
```
Abhinandan-Purkait, line 236: Had to remove this otherwise it returns all keys.

tiagolobocastro: hmm right but then also without it it won't work right? it's not paginating?

tiagolobocastro: Yeah I don't think this is right either.
I think what we need to do is add the range end, so probably we need to pass the original prefix on every call, or compute the range end ahead of time.
Please add a test with small limits to ensure we get it right

Abhinandan-Purkait: For a specific prefix, for example, `/a/b/c/NodeSpec/id` all the prefix for these kvs will always be `/a/b/c/NodeSpec` how do we calculate range end?

Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors try

control-plane/agents/src/bin/core/controller/resources/migration.rs
```diff
@@ -16,31 +19,58 @@ pub const MSP_OPERATOR: &str = "msp-operator";
...
+                    &new_value,
+                )
+                .await?;
+        }
+
+        if let Some((key, _)) = store_entries.last() {
+            prefix = key;
+        } else {
+            break;
+        }
```
tiagolobocastro:
```suggestion
        last = store_entries.last().cloned();
        for (k, v) in store_entries {
            let id = k
                .split('/')
                .last()
                .ok_or_else(|| StoreError::InvalidKey { key: k.to_string() })?;

            let new_value = match spec_type {
                StorableObjectType::VolumeSpec => migrate_volume_labels(v)?,
                StorableObjectType::PoolSpec => migrate_pool_labels(v)?,
                _ => v,
            };

            store
                .put_kv(
                    &format!("{}/{}", key_prefix_obj(spec_type, API_VERSION), id),
                    &new_value,
                )
                .await?;
        }

        if let Some((key, _)) = last {
            prefix = key;
        } else {
            break;
        }
```

utils/pstor/src/etcd.rs
```diff
@@ -256,6 +261,42 @@ impl StoreKv for Etcd {
...
+
+            first = false;
+
+            all_values.extend(values.clone());
+
+            if let Some((key, _)) = values.last() {
+                prefix = key;
+            } else {
+                break;
+            }
```
tiagolobocastro:
```suggestion
            last = values.last().cloned();
            all_values.extend(values);

            if let Some((key, _)) = last() {
                prefix = key;
            } else {
                break;
            }
```

control-plane/agents/src/bin/core/tests/controller/mod.rs
```diff
@@ -212,3 +215,76 @@ async fn core_agent_lease_lock() {
...
+    }
+
+    // Persist some volumes in old key space in etcd.
+    for _i in 1 .. 6 {
+        let uuid = Uuid::new_v4();
+        let key = format!("{}/{}", OLD_VOLUME_PREFIX, uuid);
+        let json_str = r#"{"uuid":"456122b1-7e19-4148-a890-579ca785a119","size":2147483648,"labels":{"local":"true"},"num_replicas":3,"status":{"Created":"Online"},"target":{"node":"mayastor-node4","nexus":"d6ccbb97-d13e-4ffb-91a0-c7607bb01f8f","protocol":"nvmf"},"policy":{"self_heal":true},"topology":{"node":{"Explicit":{"allowed_nodes":["mayastor-node2","mayastor-master","mayastor-node3","mayastor-node1","mayastor-node4"],"preferred_nodes":["mayastor-node2","mayastor-node3","mayastor-node4","mayastor-master","mayastor-node1"]}},"pool":{"Labelled":{"exclusion":{},"inclusion":{"openebs.io/created-by":"msp-operator"}}}},"last_nexus_id":"d6ccbb97-d13e-4ffb-91a0-c7607bb01f8f","operation":null}"#;
+        let value = Value::from_str(json_str).unwrap();
+        etcd.put_kv(&key, &value).await.unwrap();
+    }
```
tiagolobocastro, line 266: maybe also add the old nexus style entries? $uuid

utils/pstor/src/etcd.rs
```diff
@@ -221,23 +221,28 @@ impl StoreKv for Etcd {
...
         key_prefix: &str,
         limit: i64,
+        range_end: &str,
     ) -> Result<Vec<(String, Value)>, Error> {
         if limit <= 2 {
             return Err(Error::PagedMinimum);
         }
 
+        let get_options = if range_end.is_empty() {
+            println!("range_end is empty");
```
tiagolobocastro:
```suggestion
```

utils/pstor/src/etcd.rs
```diff
@@ -221,23 +221,28 @@ impl StoreKv for Etcd {
...
+            GetOptions::new()
+                .with_from_key()
```
tiagolobocastro, line 232: what does this do in this case?

tiagolobocastro (approved):

utils/pstor/src/etcd.rs
Abhinandan-Purkait: Key becomes string with this, causes lifetime issue?

utils/pstor/src/etcd.rs
Abhinandan-Purkait: If the range end is empty then it `Gets all keys >= key.`

Abhinandan-Purkait: So for example if its `get "" ""` it will get us everything paginated

Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors merge

PR #744, opened 2024-02-15
feat(volume/resize): don't allow volume resize in presence of snapshots

dsharma-dc:
If a volume is expanded after taking a snapshot, then the IOs to the new io-units(block ranges) fails because the lba range falls out of bound for replica's parent device(snapshot). Till the issue is fixed, this check ensures we don't try to expand a volume if there is snapshot present. SPDK issue #3108.

### DIFF
--- a/control-plane/agents/src/bin/core/volume/operations.rs
+++ b/control-plane/agents/src/bin/core/volume/operations.rs
@@ -191,6 +191,13 @@ impl ResourceResize for OperationGuardArc<VolumeSpec> {
         let spec = self.as_ref().clone();
         let state = registry.volume_state(&request.uuid).await?;
 
+        // Pre-check - Don't allow resize if the volume has snapshots.
+        // NOTE: This is temporary till the fix for handling this in spdk is done.
+        if !registry.specs().snapshots_by_vol(self.uuid()).is_empty() {
+            return Err(SvcError::Internal {
+                details: "Volume can't be resized while it has snapshots".into(),
+            });
+        }
         // Pre-checks - volume state eligible for resize.
         vol_status_ok_for_resize(&spec)?;
         // Pre-check - Ensure pools that host replicas have enough space to resize the replicas,


control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -191,6 +191,13 @@ impl ResourceResize for OperationGuardArc<VolumeSpec> {
         let spec = self.as_ref().clone();
         let state = registry.volume_state(&request.uuid).await?;
 
+        // Pre-check - Don't allow resize if the volume has snapshots.
+        // NOTE: This is temporary till the fix for handling this in spdk is done.
+        if !registry.specs().snapshots_by_vol(self.uuid()).is_empty() {
+            return Err(SvcError::Internal {
```
tiagolobocastro, line 197: @Abhinandan-Purkait will Internal "avoid" the storm of retries or we need a different error code?

tiagolobocastro (approved):

control-plane/agents/src/bin/core/volume/operations.rs
Abhinandan-Purkait:
```
codes.Canceled,          // gRPC: Client Application cancelled the request
codes.DeadlineExceeded,  // gRPC: Timeout
codes.Unavailable,       // gRPC: Server shutting down, TCP connection broken - previous volume operation may be still in progress.
codes.ResourceExhausted, // gRPC: Server temporarily out of resources - previous volume operation may be still in progress.
codes.Aborted:           // CSI: Operation pending for volume
```

Only these error codes will do exponential backoff, all others including internal will be burst retries.

dsharma-dc: `Aborted` may be used here then I guess. Not a perfect fit, but will be ok.

dsharma-dc: Will map `Aborted` in the CSI controller, let's keep Internal here.

Abhinandan-Purkait (approved):

dsharma-dc:
bors merge

PR #746, opened 2024-02-16
feat(csi-driver): add volume expansion

niladrih:
Changes:
- Add offline and online expansion capabilities
- Add implementation for ControllerExpandVolume on the controller binary
- Add implmentation for the NodeExpandVolume on the node binary

### DIFF
--- a/control-plane/csi-driver/src/bin/controller/client.rs
+++ b/control-plane/csi-driver/src/bin/controller/client.rs
@@ -5,8 +5,8 @@ use stor_port::types::v0::openapi::{
     models,
     models::{
         AffinityGroup, AppNode, CreateVolumeBody, Node, NodeTopology, Pool, PoolTopology,
-        PublishVolumeBody, RestJsonError, Topology, Volume, VolumePolicy, VolumeShareProtocol,
-        Volumes,
+        PublishVolumeBody, ResizeVolumeBody, RestJsonError, Topology, Volume, VolumePolicy,
+        VolumeShareProtocol, Volumes,
     },
 };
 
@@ -41,6 +41,8 @@ pub enum ApiClientError {
     Unavailable(String),
     /// Precondition Failed.
     PreconditionFailed(String),
+    /// Not Acceptable (406)
+    NotAcceptable(String),
 }
 
 /// Placeholder for volume topology for volume creation operation.
@@ -80,6 +82,7 @@ impl From<clients::tower::Error<RestJsonError>> for ApiClientError {
                         StatusCode::SERVICE_UNAVAILABLE => Self::Unavailable(detailed),
                         StatusCode::PRECONDITION_FAILED => Self::PreconditionFailed(detailed),
                         StatusCode::BAD_REQUEST => Self::InvalidArgument(detailed),
+                        StatusCode::NOT_ACCEPTABLE => Self::NotAcceptable(detailed),
                         status => Self::GenericOperation(status, detailed),
                     }
                 }
@@ -497,5 +500,41 @@ impl RestApiClient {
             .get_app_node(app_node_id)
             .await?;
         Ok(response.into_body())
+    }
+
+    /// Expand volume.
+    #[instrument(fields(volume.uuid = %volume_id, volume.size = %required_volume_size), skip(self, volume_id, required_volume_size))]
+    pub(crate) async fn expand_volume(
+        &self,
+        volume_id: &uuid::Uuid,
+        required_volume_size: u64,
+    ) -> Result<Volume, ApiClientError> {
+        use clients::tower::{Error::Response, ResponseError::Expected};
+
+        let vol_client = self.rest_client.volumes_api();
+
+        // This call should be idempotent and should return success if the volume
+        // size is already greater than or equal to the target_volume_size.
+        let resize_result = vol_client
+            .put_volume_size(
+                volume_id,
+                ResizeVolumeBody::new(required_volume_size as usize),
+            )
+            .await
+            .map(|response| response.into_body());
+
+        match resize_result {
+            // Success case.
+            Ok(vol) => Ok(vol),
+            // The volume capacity is already greater than or equal to required capacity.
+            Err(Response(Expected(err))) if err.status() == StatusCode::NOT_ACCEPTABLE => {
+                Ok(vol_client
+                    .get_volume(volume_id)
+                    .await
+                    .map(|response| response.into_body())?)
+            }
+            // Something went wrong.
+            Err(err) => Err(err.into()),
+        }
     }
 }
--- a/control-plane/csi-driver/src/bin/controller/controller.rs
+++ b/control-plane/csi-driver/src/bin/controller/controller.rs
@@ -21,6 +21,7 @@ use std::{collections::HashMap, str::FromStr};
 use tonic::{Request, Response, Status};
 use tracing::{debug, error, instrument, trace, warn};
 use uuid::Uuid;
+use volume_capability::AccessType;
 
 const OPENEBS_TOPOLOGY_KEY: &str = "openebs.io/nodename";
 const VOLUME_NAME_PATTERN: &str =
@@ -790,6 +791,7 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
             controller_service_capability::rpc::Type::GetCapacity,
             controller_service_capability::rpc::Type::CreateDeleteSnapshot,
             controller_service_capability::rpc::Type::ListSnapshots,
+            controller_service_capability::rpc::Type::ExpandVolume,
         ];
 
         Ok(Response::new(ControllerGetCapabilitiesResponse {
@@ -976,9 +978,52 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
     #[instrument(err, skip(self))]
     async fn controller_expand_volume(
         &self,
-        _request: tonic::Request<ControllerExpandVolumeRequest>,
+        request: tonic::Request<ControllerExpandVolumeRequest>,
     ) -> Result<tonic::Response<ControllerExpandVolumeResponse>, tonic::Status> {
-        Err(Status::unimplemented("Not implemented"))
+        let request = request.into_inner();
+
+        let vol_uuid = Uuid::parse_str(&request.volume_id).map_err(|error| {
+            Status::invalid_argument(format!(
+                "Malformed volume UUID '{}': {error}",
+                request.volume_id
+            ))
+        })?;
+
+        let requested_size = request
+            .capacity_range
+            .as_ref()
+            .ok_or(Status::invalid_argument(format!(
+                "Cannot expand volume '{}': invalid request {request:?}: missing CapacityRange",
+                request.volume_id
+            )))?
+            .required_bytes;
+
+        // NodeExpandVolume should be avoided only if we know without a shred of doubt that the
+        // volume is that of a Block type. The volume_capability field is optional and we should
+        // assume NodeExpandVolume is required if it is not clearly defined that the volume is
+        // Block type volume.
+        let node_expansion_required = !matches!(
+            request.volume_capability.as_ref(),
+            Some(vc) if matches!(vc.access_type, Some(AccessType::Block(_)))
+        );
+
+        let _guard = csi_driver::limiter::VolumeOpGuard::new(vol_uuid)?;
+
+        let vol = RestApiClient::get_client()
+            // This call is idempotent and should return success without expansion,
+            // if the volume size is larger than or equal to the requested size.
+            .expand_volume(&vol_uuid, requested_size as u64)
+            .await
+            .map_err(|error| match error {
+                ApiClientError::PreconditionFailed(msg) => Status::failed_precondition(msg),
+                ApiClientError::ResourceExhausted(msg) => Status::out_of_range(msg),
+                error => Status::from(error),
+            })?;
+
+        Ok(tonic::Response::new(ControllerExpandVolumeResponse {
+            capacity_bytes: vol.spec.size as i64,
+            node_expansion_required,
+        }))
     }
 
     #[instrument(err, skip(self))]
--- a/control-plane/csi-driver/src/bin/node/filesystem_ops.rs
+++ b/control-plane/csi-driver/src/bin/node/filesystem_ops.rs
@@ -3,10 +3,12 @@
 //! specific filesystem, repairing of the filesystem, retrieving specific properties of the
 //! filesystem.
 
-use crate::mount;
+use crate::{findmnt::get_devicepath, mount};
 use csi_driver::filesystem::FileSystem as Fs;
-use devinfo::{blkid::probe::Probe, DevInfoError};
-use std::process::Output;
+
+use anyhow::anyhow;
+use devinfo::{blkid::probe::Probe, mountinfo::MountInfo, DevInfoError};
+use std::{process::Output, str, str::FromStr};
 use tokio::process::Command;
 use tonic::async_trait;
 use tracing::{debug, trace};
@@ -32,6 +34,23 @@ pub(crate) struct FileSystem(Fs);
 impl From<Fs> for FileSystem {
     fn from(value: Fs) -> Self {
         Self(value)
+    }
+}
+
+impl TryFrom<&MountInfo> for FileSystem {
+    type Error = anyhow::Error;
+
+    fn try_from(mnt: &MountInfo) -> Result<Self, Self::Error> {
+        if mnt.fstype.is_empty() {
+            return Err(anyhow!("fstype is empty"));
+        }
+
+        Fs::from_str(mnt.fstype.to_lowercase().as_str())
+            .map_err(|err| anyhow!("failed to parse FileSystem: {err}"))
+            .and_then(|fs_type| match fs_type {
+                Fs::Ext4 | Fs::Xfs | Fs::Btrfs => Ok(FileSystem::from(fs_type)),
+                _ => Err(anyhow!("Unsupported filesystem")),
+            })
     }
 }
 
@@ -107,6 +126,8 @@ pub(crate) trait FileSystemOps: Send + Sync {
         }
         Ok(())
     }
+    /// Write the existing filesystem on to new unused blocks on the block device.
+    async fn expand(&self, mount_path: &str) -> Result<(), Error>;
 }
 
 #[async_trait]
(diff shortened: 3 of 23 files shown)


niladrih:
bors try

niladrih:
~Pytests are in WIP state.~
EDIT: Pytests are in #759

dsharma-dc:
The commit need to be reworded as:
`test(volume/expansion): add volume expansion to csi-driver tests`

tiagolobocastro:
> The commit need to be reworded as: `test(volume/expansion): add volume expansion to csi-driver tests`

Good comment, I would make this myself :)
Also I think you can also probably remove "control-plane/" from the commits

control-plane/csi-driver/src/bin/controller/client.rs
```diff
@@ -13,6 +13,7 @@ use stor_port::types::v0::openapi::{
 use anyhow::{anyhow, Result};
 use once_cell::sync::OnceCell;
 use std::collections::HashMap;
+use stor_port::types::v0::openapi::models::ResizeVolumeBody;
```
tiagolobocastro: move this above together with other stor_ports

control-plane/csi-driver/src/bin/controller/client.rs
```diff
@@ -498,4 +499,25 @@ impl RestApiClient {
...
+    #[instrument(fields(volume.uuid = %volume_id, volume.size = %target_volume_size), skip(self, volume_id, target_volume_size))]
+    pub(crate) async fn expand_volume(
+        &self,
+        volume_id: &uuid::Uuid,
+        target_volume_size: u64,
+    ) -> Result<Volume, ApiClientError> {
+        let vol = self
+            .rest_client
+            .volumes_api()
+            // This call should be idempotent and should return success if the volume
```
tiagolobocastro: is this in relation to csi or the rest api?
For the rest idempotent does not mean success if something is already done, as that is not the strict definition of idempotency.
So this means we may need to catch the error here, though not sure how that bit was implemented for resize in control-plane

control-plane/csi-driver/src/bin/controller/main.rs
```diff
@@ -1,5 +1,7 @@
+use clap::{Arg, ArgMatches};
```
tiagolobocastro: keep external crates separate as they were pls

control-plane/csi-driver/src/bin/node/filesystem_ops.rs
```diff
@@ -35,6 +36,25 @@ impl From<Fs> for FileSystem {
...
 
+impl TryFrom<&MountInfo> for FileSystem {
+    type Error = anyhow::Error;
+
+    fn try_from(mnt: &MountInfo) -> Result<Self, Self::Error> {
+        if mnt.fstype.is_empty() {
+            return Err(anyhow!("fstype is empty"));
+        }
+
+        let fs_type = match mnt.fstype.to_lowercase().as_str() {
```
tiagolobocastro: Can't we use `FileSystem::from_str` ?

control-plane/csi-driver/src/bin/node/filesystem_expansion.rs
tiagolobocastro, line 1: Couldn't this exist in `filesystem_ops.rs`?

control-plane/csi-driver/src/bin/node/node.rs
```diff
@@ -1,10 +1,16 @@
...
     filesystem_vol::{publish_fs_volume, stage_fs_volume, unpublish_fs_volume, unstage_fs_volume},
+    mount::find_mount,
 };
-use csi_driver::csi::volume_capability::{access_mode::Mode, AccessType};
+use csi_driver::{
+    csi::volume_capability::{access_mode::Mode, AccessType},
+    filesystem::FileSystem as Fs,
+    limiter::VolumeOpGuard,
+};
+use nix::{errno::Errno, sys};
```
tiagolobocastro: keep the external imports separate

control-plane/csi-driver/src/bin/node/node.rs
```diff
@@ -432,9 +437,117 @@ impl node_server::Node for Node {
...
+        if request.volume_path.is_empty() {
+            return Err(Status::invalid_argument("'volume_path' is empty"));
+        }
+
+        let _guard = VolumeOpGuard::new(vol_uuid)?;
+
+        let dev_name = Device::lookup(&vol_uuid)
+            .await
+            .map_err(|error| {
+                Status::internal(format!(
```
tiagolobocastro: can use failure macro here I guess

control-plane/csi-driver/src/bin/node/node.rs
```diff
@@ -432,9 +437,117 @@ impl node_server::Node for Node {
...
+                ))
+            })?
+            .ok_or(Status::invalid_argument(format!(
+                "failed to find a device for volume {vol_uuid}"
+            )))?
+            .devname();
+        // Get device size.
+        // The underlying block device should already have been expanded to the
+        // required size as a part of the ControllerExpandVolume call.
+        let capacity_bytes = get_size_from_dev_name(dev_name.as_str()).map_err(|error| {
```
tiagolobocastro: Shouldn't we check that capacity_bytes is what we expect?

control-plane/csi-driver/src/bin/controller/main.rs
niladrih: Oh, is that why those blank lines exist.. sure, I'll put them back.

tiagolobocastro: Yes :)

control-plane/csi-driver/src/bin/controller/client.rs
dsharma-dc: w.r.t csi it must be idempotent. I don't think REST is gonna get an error today as well if the size is already what we request.
@niladrih Can we confirm if **we get the CSI RPC at all** in case PVC is modified to a size that's same or lesser than current size?

dsharma-dc: @tiagolobocastro Shall we throw back error to REST in such case?

niladrih: @dsharma-dc -- We should get the CSI call for changes to the PVC capacity (https://github.com/kubernetes-csi/external-resizer/blob/v1.10.0/pkg/controller/expand_and_recover.go#L160)
The spec requires that we be idempotent, so we ought to have idempotency here irrespective of what kubernetes descides to do here: https://github.com/container-storage-interface/spec/blob/v1.5.0/spec.md#controllerexpandvolume

control-plane/csi-driver/src/bin/node/node.rs
niladrih: This is a good idea. I think to perform this check correctly, we'd have to get the usable blockdevice capacity from the total blockdevice capacity. Otherwise a comparision against this size value would be hacky. Wdyt @tiagolobocastro @dsharma-dc ?

dsharma-dc: Yes, the usable capacity exported by nexus is typically bit lesser than spec after accounting various things, but `get_size_from_dev_name` will return the usable amount only I guess so that should be ok. However, since we don't specify size in `resize2fs` or `xfs_grow` commands, it'll figure out the usable size automatically. Comparing against what we expect may not be very fruitful here.

control-plane/csi-driver/src/bin/node/filesystem_expansion.rs
niladrih: Sure, it would fit in just fine there.

control-plane/csi-driver/src/bin/node/node.rs
niladrih: yep

control-plane/csi-driver/src/bin/node/node.rs
niladrih: Oh, I didn't know about that one. Yep, works!

control-plane/csi-driver/src/bin/node/filesystem_ops.rs
niladrih: Using `FromStr` 👍

control-plane/csi-driver/src/bin/controller/client.rs
niladrih: Oh, right!

control-plane/csi-driver/src/bin/node/node.rs
niladrih: > Comparing against what we expect may not be very fruitful here

While this plugin method doesn't handle the expansion for the block layer, the CSI spec requires that we only return OK iff the allocated capacity for the device is equal to or more than the required capacity. This implies that the less-than case is an error case. 
In principle, the controller plugin should return that kind of an error, but that would be quite difficult to reach out to linux API for for nodes from the controller plugin. However, the node plugin can do that.

niladrih: Oh, I think I get why the /sys/block size is completely usable now.

niladrih:
> The commit need to be reworded as: `test(volume/expansion): add volume expansion to csi-driver tests`

> Also I think you can also probably remove "control-plane/" from the commits

Made these changes. Also ref: https://github.com/openebs/mayastor-control-plane/pull/759.

control-plane/csi-driver/src/bin/controller/client.rs
niladrih: I've added a check. I'm GET-ing the volume and comparing the size against the requested size.

niladrih: WDYT @tiagolobocastro @dsharma-dc ?

control-plane/csi-driver/src/bin/node/node.rs
tiagolobocastro: @dsharma-dc  Why is the capacity exported by the nexus less than spec?
@niladrih can't we just test capacity_bytes against the request?
EDIT: oh I see you've indeed done that below :+1:

control-plane/csi-driver/src/bin/node/node.rs
```diff
@@ -432,9 +438,149 @@ impl node_server::Node for Node {
...
+                dev_name,
+                error
+            )
+        })? as i64;
+
+        // Ensure capacity_bytes is greater than or equal ot required_bytes.
+        if capacity_bytes.lt(&required_bytes) {
+            return Err(failure!(
+                Code::FailedPrecondition,
+                "block device capacity is invalid"
```
tiagolobocastro: nit: something more like these lines
```suggestion
                "block device capacity is lower than the required"
```

control-plane/csi-driver/src/bin/controller/controller.rs
```diff
@@ -976,9 +978,52 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
...
     ) -> Result<tonic::Response<ControllerExpandVolumeResponse>, tonic::Status> {
-        Err(Status::unimplemented("Not implemented"))
+        let request = request.into_inner();
+
+        let vol_uuid = Uuid::parse_str(&request.volume_id).map_err(|e| {
+            Status::invalid_argument(format!(
+                "Malformed volume UUID '{}': {}",
+                request.volume_id, e
+            ))
+        })?;
```
```suggestion
        let vol_uuid = Uuid::parse_str(&request.volume_id).map_err(|error| {
            Status::invalid_argument(format!(
                "Malformed volume UUID '{}': {error}",
                request.volume_id
            ))
        })?;
```

control-plane/csi-driver/src/bin/controller/controller.rs
```diff
@@ -976,9 +978,52 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
...
+
+        let requested_size = request
+            .capacity_range
+            .as_ref()
+            .ok_or(Status::invalid_argument(format!(
+                "Cannot expand volume '{}': invalid request {:?}: missing CapacityRange",
+                request.volume_id, request
```
```suggestion
                "Cannot expand volume '{}': invalid request {request:?}: missing CapacityRange",
                request.volume_id
```

control-plane/csi-driver/src/bin/controller/controller.rs
```diff
@@ -976,9 +978,52 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
...
+        );
+
+        let _guard = csi_driver::limiter::VolumeOpGuard::new(vol_uuid)?;
+
+        let vol = RestApiClient::get_client()
+            // This call is idempotent and should return success without expansion,
+            // if the volume size is larger than or equal to the requested size.
+            .expand_volume(&vol_uuid, requested_size as u64)
+            .await
+            .map_err(|e| match e {
```
```suggestion
            .map_err(|error| match error {
```

control-plane/csi-driver/src/bin/controller/controller.rs
```diff
@@ -976,9 +978,52 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
...
+                ApiClientError::PreconditionFailed(msg) => Status::failed_precondition(msg),
+                ApiClientError::ResourceExhausted(msg) => Status::out_of_range(msg),
+                err => Status::from(err),
```
```suggestion
                error => Status::from(error),
```

control-plane/csi-driver/src/bin/controller/client.rs
```diff
@@ -498,4 +498,35 @@ impl RestApiClient {
...
+        let vol_client = self.rest_client.volumes_api();
+
+        let vol = vol_client
+            .get_volume(volume_id)
+            .await
+            .map(|response| response.into_body())?;
+
+        // This call should be idempotent and should return success if the volume
+        // size is already greater than or equal to the target_volume_size.
+        if vol.spec.size.ge(&required_volume_size) {
```
tiagolobocastro: IIRC this can be done using the error code from put itself?

tiagolobocastro (approved):

control-plane/csi-driver/src/bin/node/node.rs
dsharma-dc: > @dsharma-dc Why is the capacity exported by the nexus less than spec?

That's because nexus bdev capacity is set as `end` - `start` block numbers, and these block numbers are adjusted after accounting for GPT and metadata details.

tiagolobocastro: Oh that's right, we probably should create replicas with extra capacity to account for this...
But then it seems like we need to check here against size - 5MiB @niladrih ?

dsharma-dc: A bit confusing part here is that the -5MiB is from the actual replica size, and not requested size. Example below:

- Create 105MiB replica and a nexus with it -> spdk replica bdev size is actually created 108MiB, and nexus bdev is 105MiB. Nexus size visible to host is 103MiB(108 - 5)
- Resize replica and nexus to 150MiB -> spdk replica bdev becomes 152MiB, nexus bdev is 150MiB. Nexus size visible to host is 147MiB (152 - 5)

so a check here like below seems ok perhaps?
`(capacity_bytes > requested_size - 5MiB) => OK`

niladrih: Is this in a const somewhere I can use, or do you want me to use the 5 MiB raw?
@tiagolobocastro @dsharma-dc ?

control-plane/csi-driver/src/bin/controller/client.rs
niladrih: Yes, I've refactored the idempotency implementation. WDYT?

control-plane/csi-driver/src/bin/controller/client.rs
```diff
@@ -498,4 +501,40 @@ impl RestApiClient {
...
+                ResizeVolumeBody::new(required_volume_size as usize),
+            )
+            .await
+            .map(|response| response.into_body());
+
+        match resize_result {
+            // Success case.
+            Ok(vol) => Ok(vol),
+            // The volume capacity is already greater than or equal to required capacity.
+            Err(Response(Expected(err))) if matches!(err.status(), StatusCode::NOT_ACCEPTABLE) => {
```
tiagolobocastro:
```suggestion
            Err(Response(Expected(err))) if err.status() == StatusCode::NOT_ACCEPTABLE => {
```

tiagolobocastro (approved):

control-plane/csi-driver/src/bin/node/node.rs
niladrih: Also, it seems like the nexus size is 3 MiB away from REST Volume resource size? Is the Volume resource size the same as the required_bytes from CreateVolumeRequest?

dsharma-dc: > Also, it seems like the nexus size is 3 MiB away from REST Volume resource size? Is the Volume resource size the same as the required_bytes from CreateVolumeRequest?

There is no const to use as such, you can define one and use perhaps. The volume resource size will be same as REST volume resource size, which will same in resource spec. The extras that we we see in bdev are done at spdk for book-keeping.

dsharma-dc (approved):

tests/bdd/features/volume/resize/test_resize_online.py
```diff
@@ -484,8 +484,8 @@ def the_new_capacity_should_be_available_for_the_application(test_volume_factory
 def the_failure_reason_should_be_invalid_arguments():
     """the failure reason should be invalid arguments."""
```
dsharma-dc, line 474: Unrelated to this change - there is a commented function in this file `a published volume with more than one replicas` L290. Could you just delete that please?

control-plane/csi-driver/src/bin/node/node.rs
niladrih: Adjusted the required_bytes by 5MiB.

tests/bdd/features/volume/resize/test_resize_online.py
niladrih: Removed this.

niladrih:
bors try

niladrih:
bors merge

niladrih:
bors merge

niladrih:
bors merge

niladrih:
bors merge

tiagolobocastro:
`test(volume/expansion): use HTTP 406 for VolumeResizeSize error` this is not a "test" ?
Should be refactor

niladrih:
bors try

niladrih:
bors try

dsharma-dc (approved):

niladrih:
bors merge

PR #747, opened 2024-02-16
check cluster capacity limit during volume resize, and also allow volume resize via rest plugin

dsharma-dc:
- Enforce cluster capacity limit checks during volume resize.
- Allow volume resize operation via REST plugin.

### DIFF
--- a/control-plane/agents/src/bin/core/volume/service.rs
+++ b/control-plane/agents/src/bin/core/volume/service.rs
@@ -47,6 +47,7 @@ use stor_port::{
 pub(super) struct Service {
     registry: Registry,
     create_volume_limiter: std::sync::Arc<tokio::sync::Semaphore>,
+    capacity_limit_borrow: std::sync::Arc<parking_lot::RwLock<u64>>,
 }
 
 #[tonic::async_trait]
@@ -243,6 +244,7 @@ impl Service {
             create_volume_limiter: std::sync::Arc::new(tokio::sync::Semaphore::new(
                 registry.create_volume_limit(),
             )),
+            capacity_limit_borrow: std::sync::Arc::new(parking_lot::RwLock::new(0)),
             registry,
         }
     }
@@ -578,6 +580,26 @@ impl Service {
     #[tracing::instrument(level = "info", skip(self), err, fields(volume.uuid = %request.uuid))]
     pub(super) async fn resize_volume(&self, request: &ResizeVolume) -> Result<Volume, SvcError> {
         let mut volume = self.specs().volume(&request.uuid).await?;
-        volume.resize(&self.registry, request).await
+
+        let Some(limit) = request.cluster_capacity_limit() else {
+            return volume.resize(&self.registry, request).await;
+        };
+
+        let required = request.requested_size - volume.as_ref().size;
+        *self.capacity_limit_borrow.write() += required;
+        // If there is a defined system wide capacity limit, ensure we don't breach that.
+        let current = *self.capacity_limit_borrow.read();
+        self.specs()
+            .check_capacity_limit_for_resize(limit, current)
+            .map_err(|e| {
+                *self.capacity_limit_borrow.write() -= required;
+                e
+            })?;
+
+        let resize_ret = volume.resize(&self.registry, request).await;
+        // Reset the capacity limit that we consumed and will now be accounted in the system's
+        // current total.
+        *self.capacity_limit_borrow.write() -= required;
+        resize_ret
     }
 }
--- a/control-plane/agents/src/bin/core/volume/specs.rs
+++ b/control-plane/agents/src/bin/core/volume/specs.rs
@@ -816,6 +816,24 @@ impl ResourceSpecsLocked {
         }
     }
 
+    pub(crate) fn check_capacity_limit_for_resize(
+        &self,
+        cluster_capacity_limit: u64,
+        current_borrowed_limit: u64,
+    ) -> Result<(), SvcError> {
+        let specs = self.write();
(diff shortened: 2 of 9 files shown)


dsharma-dc:
bors try

control-plane/agents/src/bin/core/volume/service.rs
```diff
@@ -578,6 +580,29 @@ impl Service {
...
+
+        let resize_ret = volume.resize(&self.registry, request).await;
+        if request.cluster_capacity_limit().is_some() {
+            // Reset the capacity limit that we consumed and will now be accounted in the system's
+            // current total.
+            let mut current_borrowed = self.capacity_limit_borrow.write();
+            tracing::info!(volume=%volume.uuid(),"resize_volume: return borrowed capacity limit {required}B, current {}B ", *current_borrowed);
+            *current_borrowed -= required;
+        }
+        resize_ret
```
tiagolobocastro:
```suggestion
        let Some(limit) = request.cluster_capacity_limit() else {
            return volume.resize(&self.registry, request).await;
        }
        let required = request.requested_size - volume.as_ref().size;
        // If there is a defined system wide capacity limit, ensure we don't breach that.
        *self.capacity_limit_borrow.write() += required;
        self.specs()
                .check_capacity_limit_for_resize(limit, *self.capacity_limit_borrow.read()).map_err(|e| {
                    *self.capacity_limit_borrow.write() -= required;
                    e
                })?;
        let resize_ret = volume.resize(&self.registry, request).await;
        // Reset the capacity limit that we consumed and will now be accounted in the system's
        // current total.
        *self.capacity_limit_borrow.write() -= required;
        resize_ret
```

control-plane/agents/src/bin/core/volume/specs.rs
```diff
@@ -828,6 +829,31 @@ impl ResourceSpecsLocked {
...
     }
 
+    pub(crate) fn check_capacity_limit_for_resize(
+        &self,
+        request: &ResizeVolume,
+        current_borrowed_limit: u64,
+    ) -> Result<(), SvcError> {
+        // We anyway called this function after checking cluster_capacity_limit is not None,
+        // but let's keep this match for completion, should the caller end change at some point.
+        match request.cluster_capacity_limit {
```
tiagolobocastro: We don't need request and do match again, just use the limit, see my suggestion above.

control-plane/agents/src/bin/core/volume/specs.rs
```diff
@@ -828,6 +829,31 @@ impl ResourceSpecsLocked {
...
+            None => {} // no limit, no check needed
+            Some(limit) => {
+                let specs = self.write();
+                let total: u64 = specs.volumes.values().map(|v| v.lock().size).sum();
+                let forthcoming_total = current_borrowed_limit + total;
+                tracing::info!(current_borrowed_limit=%current_borrowed_limit, total=%total, forthcoming_total=%forthcoming_total, "Cluster capacity limit checks ");
```
tiagolobocastro: or make it trace
```suggestion
```

control-plane/agents/src/bin/core/volume/specs.rs
```diff
@@ -828,6 +829,31 @@ impl ResourceSpecsLocked {
         }
     }
 
+    pub(crate) fn check_capacity_limit_for_resize(
+        &self,
+        request: &ResizeVolume,
```
tiagolobocastro: we already have the limit from request, don't need to pass Option again

```suggestion
        cluster_limit: u64,
```

control-plane/agents/src/bin/core/volume/service.rs
dsharma-dc: Thanks, looks cleaner. Except that I had to read and get the value before calling checks, otherwise the guard doesn't get dropped, and the `write()` call hangs in `map_err` path.
```
        let current = *self.capacity_limit_borrow.read()
        self.specs()
                .check_capacity_limit_for_resize(limit, current)
```

control-plane/agents/src/bin/core/volume/specs.rs
```diff
@@ -828,6 +828,26 @@ impl ResourceSpecsLocked {
         }
     }
 
+    pub(crate) fn check_capacity_limit_for_resize(
+        &self,
+        cluster_capacity_limit: u64,
+        current_borrowed_limit: u64,
+    ) -> Result<(), SvcError> {
+        // We anyway called this function after checking cluster_capacity_limit is not None,
+        // but let's keep this match for completion, should the caller end change at some point.
```
tiagolobocastro: Comment out of date? Don't see match

```suggestion
```

control-plane/plugin/src/resources/volume.rs
```diff
@@ -190,6 +190,39 @@ impl Scale for Volume {
...
+                }
+                OutputFormat::None => {
+                    // In case the output format is not specified, show a success message.
+                    println!("Volume {id} resized successfully 🚀")
+                }
+            },
+            Err(e) => {
+                return Err(Error::ScaleVolumeError {
+                    id: id.to_string(),
+                    source: e,
```
tiagolobocastro:
```suggestion
            Err(source) => {
                return Err(Error::ScaleVolumeError {
                    id: id.to_string(),
                    source,
```

tiagolobocastro (approved):
Your PR needs renaming as you've also added resize support to the plugin

Abhinandan-Purkait (approved):

dsharma-dc:
> Your PR needs renaming as you've also added resize support to the plugin

Done.

dsharma-dc:
bors merge

PR #750, opened 2024-02-20
feat(snapshot): coreagent change for setvolumeattributes

hrudaya21:


control-plane/stor-port/src/types/v0/store/volume.rs
```diff
@@ -681,6 +707,9 @@ impl From<VolumeOperation> for models::volume_spec_operation::Operation {
             VolumeOperation::DestroySnapshot(_) => {
                 models::volume_spec_operation::Operation::DestroySnapshot
             }
+            VolumeOperation::SetVolumeProperties(_) => {
+                todo!()
```
tiagolobocastro, line 681: need to update models here, otherwise we panic...

control-plane/stor-port/src/types/v0/store/volume.rs
```diff
@@ -643,6 +652,23 @@ impl PublishOperation {
         self.publish_context.clone()
     }
 }
+/// SetVolumeProperty Operation parameters.
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+pub struct VolumeProperties {
+    properties: HashMap<String, String>,
```
tiagolobocastro: We can use actual types here rather than String?

control-plane/stor-port/src/types/v0/transport/volume.rs
```diff
@@ -703,6 +703,22 @@ impl SetVolumeReplica {
...
 }
 
+/// Set the volume property.
+#[derive(Serialize, Deserialize, Default, Debug, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct SetVolumeProperties {
+    /// The uuid of the volume.
+    pub uuid: VolumeId,
+    /// Map of property name to value.
+    pub properties: HashMap<String, String>,
```
tiagolobocastro: We can use an enum here rather than a map?

control-plane/stor-port/src/types/v0/store/volume.rs
hrudaya21: This will be filled once, openapi changes are in placed. Planned to put it in another PR.
Draft PR https://github.com/openebs/mayastor-control-plane/pull/752 is there for reference.

control-plane/stor-port/src/types/v0/store/volume.rs
hrudaya21: Currently idea is to keep each attribute with its datatype in open api spec and map it to generic string based map.
VolumeAttr enum will match the name for matching the attribute to set it in Volumespec.

hrudaya21: fixed.

control-plane/stor-port/src/types/v0/transport/volume.rs
hrudaya21: fixed.

control-plane/stor-port/src/types/v0/transport/volume.rs
```diff
@@ -72,6 +73,16 @@ pub struct VolumeState {
     pub usage: Option<VolumeUsage>,
 }
 
+/// Volume properties.
+#[derive(
+    Serialize, Deserialize, EnumString, Debug, EnumCountMacro, EnumIter, PartialEq, Clone, Copy,
```
tiagolobocastro: Let's not do implicit Copy here, new attr size could be larger as we add more things 
Also do we need all of the strums derives btw?

control-plane/stor-port/src/types/v0/store/volume.rs
```diff
@@ -643,6 +639,23 @@ impl PublishOperation {
...
+impl VolumeProperty {
+    /// Return new `Self` from the given parameters.
+    pub fn new(property: VolumeAttr) -> Self {
+        match property {
+            VolumeAttr::MaxSnapshots(_) => Self { property },
+        }
+    }
+    /// Get property enum.
+    pub fn property(&self) -> VolumeAttr {
+        self.property
```
```suggestion
    pub fn property(&self) -> &VolumeAttr {
        &self.property
```

control-plane/stor-port/src/types/v0/store/volume.rs
```diff
@@ -643,6 +639,23 @@ impl PublishOperation {
         self.publish_context.clone()
     }
 }
+/// SetVolumeProperty Operation parameters.
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+pub struct VolumeProperty {
```
tiagolobocastro: what's the different between VolumeProperty and VolumeAttr?

Abhinandan-Purkait: Can we just directly use VolumeAttr everywhere?

tiagolobocastro: That's what I was wondering, just use the same (either Attr or Property). Or perhaps Property can be a type alias for Attr

hrudaya21: Fixed it. Before I have not used enum with datatype. With enum with datatype, this extra struct layering is not needed.

control-plane/stor-port/src/types/v0/transport/volume.rs
hrudaya21: strum is removed. But I think `copy` is needed while passing this enum in `set_property` function.

tiagolobocastro: Let's remove Copy, just use clone whenever needed

control-plane/stor-port/src/types/v0/store/volume.rs
```diff
@@ -519,6 +509,11 @@ impl SpecTransaction<VolumeOperation> for VolumeSpec {
                 VolumeOperation::Resize(size) => {
                     self.size = size;
                 }
+                VolumeOperation::SetVolumeProperty(property) => match property {
+                    VolumeProperty::MaxSnapshots(max_snapshots) => {
+                        self.max_snapshots = Some(max_snapshots);
```
tiagolobocastro, line 514: Probably at some point it'll be time to refactor the volume spec, and tidy up things, such as putting all properties together etc

control-plane/stor-port/src/types/v0/transport/volume.rs
```diff
@@ -72,6 +72,13 @@ pub struct VolumeState {
     pub usage: Option<VolumeUsage>,
 }
 
+/// Volume properties.
+#[derive(Serialize, Deserialize, Debug, Copy, PartialEq, Clone)]
```
tiagolobocastro:
```suggestion
#[derive(Serialize, Deserialize, Debug, PartialEq, Clone)]
```

tiagolobocastro (approved):

control-plane/stor-port/src/types/v0/store/volume.rs
hrudaya21: Yes, all such properties can be put in an enum in volume spec.

Abhinandan-Purkait (approved):

hrudaya21:
bors try

hrudaya21:
bors merge

PR #752, opened 2024-02-20
feat(snapshot): plugin changes for setvolumeattributes

hrudaya21:


abhilashshetty04 (approved):

control-plane/plugin/src/resources/volume.rs
```diff
@@ -225,6 +227,63 @@ impl Scale for Volume {
...
+    ) -> PluginResult {
+        let parts: Vec<&str> = property.trim().split('=').collect();
+        if parts.len() != 2 {
+            return Err(Error::SetVolumePropertyParseError {
+                reason: format!("volume: {id}, invalid property format: {property}, enter in the format property=value"),
+            });
+        }
+        let left = parts[0].trim();
+        let right = parts[1].trim();
+        let property_body = match left {
```
Abhinandan-Purkait: Unrelated to this PR. @tiagolobocastro Just wondering if we should use strum on generated code to get these conversions?

Abhinandan-Purkait (approved):

control-plane/plugin/src/resources/mod.rs
```diff
@@ -69,6 +70,18 @@ pub enum ScaleResources {
...
 
+/// The types of resources that support the 'SetProperty' operation.
+#[derive(clap::Subcommand, Debug)]
+pub enum SetPropertyResources {
+    /// Set volume property.
+    Volume {
+        /// ID of the volume.
+        id: VolumeId,
+        /// Property of volume.
+        property: String,
```
tiagolobocastro: let's use explicit types here too

control-plane/plugin/src/resources/volume.rs
tiagolobocastro: I think we should use explicit types to begin with, perhaps even use the openapi properties in the CLI

control-plane/plugin/src/resources/mod.rs
hrudaya21: Use explicit field with clap trait now.

control-plane/plugin/src/resources/mod.rs
```diff
@@ -69,6 +70,19 @@ pub enum ScaleResources {
...
+/// The types of resources that support the 'SetProperty' operation.
+#[derive(clap::Subcommand, Debug)]
+pub enum SetPropertyResources {
+    /// Set volume property.
+    Volume {
+        /// ID of the volume.
+        id: VolumeId,
+        /// Max snapshot limit per volume.
+        #[clap(long)]
+        max_snapshots: u32,
```
tiagolobocastro: Shouldn't this be an enum?

Abhinandan-Purkait: Yeah, was thinking the same, should it be rather an enum of Vol Properties then we can add more?

tiagolobocastro: https://github.com/openebs/mayastor-control-plane/tree/set-properties

hrudaya21:
bors try

control-plane/plugin/src/resources/error.rs
```diff
@@ -81,8 +81,8 @@ pub enum Error {
         source: openapi::tower::client::Error<openapi::models::RestJsonError>,
     },
     /// Error when set volume property request fails.
-    #[snafu(display("Failed to set volume {id} property, Error {source}"))]
-    ScaleVolumePropertyError {
+    #[snafu(display("Failed to set volume {id} property, Error: {source}"))]
```
tiagolobocastro: Existing ones don't have ':'.
Let's keep it consistent

tiagolobocastro (approved):

hrudaya21:
bors merge

PR #760, opened 2024-02-26
feat(topology/plugin): add get/list label node/nodes command

sinhaashish:
Add commands

***kubectl-mayastor get label nodes***
```
 ID             GRPC ENDPOINT          STATUS  LABELS
 node-1-149660  135.181.202.199:10124  Online  W=X, L=M, P=Q, A=B
 node-2-149660  65.108.250.82:10124    Online  A=B, P=Q, C=D
 ```

```
kubectl-mayastor get label node  node-1-149660
 ID             GRPC ENDPOINT          STATUS  LABELS
 node-1-149660  135.181.202.199:10124  Online  P=Q, W=X, A=B, L=M
 ```

### DIFF
--- a/control-plane/plugin/src/resources/label.rs
+++ b/control-plane/plugin/src/resources/label.rs
@@ -1,0 +1,53 @@
+pub struct NodeLabel {}
+pub struct NodeLabels {}
+
+use async_trait::async_trait;
+
+use crate::{
+    operations::{Get, List, PluginResult},
+    resources::{
+        error::Error,
+        node::{node_display_print, node_display_print_one, NodeDisplayFormat},
+        utils::OutputFormat,
+        NodeId,
+    },
+    rest_wrapper::RestClient,
+};
+
+#[async_trait(?Send)]
+impl Get for NodeLabel {
+    type ID = NodeId;
+    async fn get(id: &Self::ID, output: &OutputFormat) -> PluginResult {
+        match RestClient::client().nodes_api().get_node(id).await {
+            Ok(node) => node_display_print_one(node.into_body(), output, NodeDisplayFormat::Labels),
+            Err(e) => {
+                return Err(Error::GetNodeError {
+                    id: id.to_string(),
+                    source: e,
+                });
+            }
+        }
+        Ok(())
+    }
+}
+
+#[async_trait(?Send)]
+impl List for NodeLabels {
+    async fn list(output: &OutputFormat) -> PluginResult {
+        match RestClient::client().nodes_api().get_nodes(None).await {
+            Ok(nodes) => {
+                // iterate through the nodes and filter for only those that have labels
+                let nodelist = nodes.into_body();
+                let mut filteredlist = nodelist;
+                // remove nodes with labels
+                filteredlist
+                    .retain(|i| i.spec.is_some() && i.spec.as_ref().unwrap().labels.is_some());
+                node_display_print(filteredlist, output, NodeDisplayFormat::Labels)
+            }
+            Err(e) => {
+                return Err(Error::ListNodesError { source: e });
+            }
+        }
+        Ok(())
+    }
+}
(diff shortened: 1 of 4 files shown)


control-plane/plugin/src/resources/mod.rs
```diff
@@ -94,6 +98,16 @@ pub enum GetCordonArgs {
     Nodes,
 }
 
+/// The types of resources that support the 'get label' operation.
+#[derive(clap::Subcommand, Debug)]
+pub enum GetLabelArgs {
+    /// Get the label for the node with the given ID.
```
tiagolobocastro:
```suggestion
pub enum GetLabelsArgs {
    /// Get the labels for the node with the given ID.
```

control-plane/plugin/src/resources/mod.rs
```diff
@@ -94,6 +98,16 @@ pub enum GetCordonArgs {
...
+    Node {
+        id: NodeId,
+    },
+    Nodes,
```
tiagolobocastro, line 107:
```suggestion
    /// Get labels for all nodes.
    Nodes,
```

control-plane/plugin/src/resources/node.rs
```diff
@@ -353,6 +369,17 @@ impl CreateRows for NodeDisplay {
                 }
                 rows
             }
+            NodeDisplayFormat::Labels => {
+                let mut rows = vec![];
+                for node in self.inner.iter() {
+                    let mut row = node.create_rows();
+                    let labelstring = NodeDisplay::get_label_list(node).join(", ");
+                    // Add the labels to each row.
+                    row[0].add_cell(Cell::new(&labelstring));
```
tiagolobocastro, line 378: I wasn't aware of this thing being done for cordon. Seems confusing that we add the header only if we have the label if the command is specifically get cordon or get labels
What do you think @Abhinandan-Purkait ?

control-plane/plugin/src/resources/label.rs
```diff
@@ -0,0 +1,53 @@
...
+#[async_trait(?Send)]
+impl List for NodeLabels {
+    async fn list(output: &OutputFormat) -> PluginResult {
+        match RestClient::client().nodes_api().get_nodes(None).await {
+            Ok(nodes) => {
+                // iterate through the nodes and filter for only those that have labels
+                let nodelist = nodes.into_body();
+                let mut filteredlist = nodelist;
+                // remove nodes with labels
+                filteredlist
```
tiagolobocastro, line 43: hmm so if you do plugin get node labels you only get nodes with labels?

sinhaashish: yes 
its on the very same lines as of `./target/debug/kubectl-mayastor get drain nodes` and `./target/debug/kubectl-mayastor get cordone nodes`

control-plane/plugin/src/resources/node.rs
Abhinandan-Purkait: I too noticed this now, the ideal way would have been using a `--show-labels` on the `get nodes/node` but maybe fine for now since we are already doing it for cordon and drain? Not sure.

control-plane/plugin/src/resources/label.rs
tiagolobocastro: I think that might be different because cordon/drain are actually separate states for a node, whereas here they are just regular labels.

control-plane/plugin/src/resources/node.rs
tiagolobocastro: what is the kubectl ux for this?

Abhinandan-Purkait: kubectl get nodes has `--show-labels` for displaying the labels when the flag is passed. It adds a new column called labels to show the the labels comma separated. The normal get nodes doesn't have the labels column for obvious reasons i.e. it will overwhelm the get nodes output

tiagolobocastro: Maybe we can do the same here then?

sinhaashish:
Pr https://github.com/openebs/mayastor-control-plane/pull/779 does the same

PR #761, opened 2024-02-26
fix(legacy prefix detection): calculate range end for legacy prefix instead of empty

Abhinandan-Purkait:
#### What is the issue?

For legacy prefix detection we used the paginated call with range end as `""`, which was incorrect as `""` range end fetches all the values, irrespective of the key prefix. Now we calculate the range end with legacy prefix thus we get only the specific KVs.


Abhinandan-Purkait:
bors try

tiagolobocastro (approved):

Abhinandan-Purkait:
bors merge

Abhinandan-Purkait:
bors merge

PR #762, opened 2024-02-26
fix(legacy prefix detection): calculate range end for legacy prefix instead of empty

Abhinandan-Purkait:
Cherry pick #761


tiagolobocastro (approved):

Abhinandan-Purkait:
bors merge

PR #763, opened 2024-02-28
feat(topology): add node spread topology code

sinhaashish:
NodelSpreadTopologyKey: "performance" - is a exclusion
1: "performance: Gold"
2: "performance: Silver"
3: "performance: Emerald"

### DIFF
--- a/control-plane/agents/src/bin/core/volume/operations.rs
+++ b/control-plane/agents/src/bin/core/volume/operations.rs
@@ -35,10 +35,11 @@ use stor_port::{
             volume::{PublishOperation, RepublishOperation, VolumeOperation, VolumeSpec},
         },
         transport::{
-            CreateVolume, DestroyNexus, DestroyReplica, DestroyShutdownTargets, DestroyVolume,
-            Protocol, PublishVolume, Replica, ReplicaId, ReplicaOwners, RepublishVolume,
-            ResizeVolume, SetVolumeProperty, SetVolumeReplica, ShareNexus, ShareVolume,
-            ShutdownNexus, UnpublishVolume, UnshareNexus, UnshareVolume, Volume,
+            CreateReplica, CreateVolume, DestroyNexus, DestroyReplica, DestroyShutdownTargets,
+            DestroyVolume, NodeTopology, Protocol, PublishVolume, Replica, ReplicaId,
+            ReplicaOwners, RepublishVolume, ResizeVolume, SetVolumeProperty, SetVolumeReplica,
+            ShareNexus, ShareVolume, ShutdownNexus, UnpublishVolume, UnshareNexus, UnshareVolume,
+            Volume,
         },
     },
 };
@@ -995,8 +996,18 @@ impl CreateVolumeExe for CreateVolume {
         for replica in candidates.candidates() {
             if replicas.len() >= self.replicas as usize {
                 break;
-            } else if replicas.iter().any(|r| r.node == replica.node) {
-                // don't reuse the same node
+            } else if replicas.iter().any(|r| {
+                r.node == replica.node
+                    || spread_label_is_same(r, replica, context).unwrap_or_else(|error| {
+                        context.volume.error(&format!(
+                            "Failed to create replica {:?} for volume, error: {}",
+                            replica,
+                            error.full_string()
+                        ));
+                        false
+                    })
+            }) {
+                // don't re-use the same node or same exclusion labels
                 continue;
             }
             let replica = if replicas.is_empty() {
@@ -1052,5 +1063,38 @@ impl CreateVolumeExeVal for CreateVolumeSource<'_> {
             CreateVolumeSource::None(params) => params.pre_flight_check(),
             CreateVolumeSource::Snapshot(params) => params.pre_flight_check(),
         }
+    }
+}
+
+// spread_label_is_same checks if the node labels is same as that of all existing replicas.
+fn spread_label_is_same(
+    replica: &Replica,
+    replica_candidate: &CreateReplica,
+    context: &Context<'_>,
+) -> Result<bool, SvcError> {
+    let topology = context.volume.as_ref().topology.as_ref();
+    let node_topology = topology.and_then(|topology| topology.node.as_ref());
+
+    match node_topology {
+        Some(NodeTopology::Explicit(_)) => Ok(false),
+        Some(NodeTopology::Labelled(labelled_topology)) => {
+            if !labelled_topology.exclusion.is_empty() {
+                let candidate_node = context.registry.specs().node(&replica_candidate.node)?;
+                let replica_node = context.registry.specs().node(&replica.node)?;
+
+                for key in labelled_topology.exclusion.keys() {
+                    if let (Some(candidate_value), Some(replica_value)) = (
+                        candidate_node.labels().get(key),
+                        replica_node.labels().get(key),
+                    ) {
+                        if candidate_value == replica_value {
+                            return Ok(true);
+                        }
+                    }
+                }
+            }
+            Ok(false)
+        }
+        _ => Ok(false),
     }
 }
--- a/control-plane/agents/src/bin/core/controller/scheduling/volume_policy/mod.rs
+++ b/control-plane/agents/src/bin/core/controller/scheduling/volume_policy/mod.rs
@@ -70,7 +70,7 @@ impl DefaultBasePolicy {
 
 /// Return true if all the keys present in volume's pool/node inclusion matches with the pool/node
 /// labels otherwise returns false.
-pub(crate) fn qualifies_inclusion_labels(
+pub(crate) fn qualifies_label_criteria(
     vol_pool_inc_labels: HashMap<String, String>,
     pool_labels: &HashMap<String, String>,
 ) -> bool {
--- a/control-plane/agents/src/bin/core/controller/scheduling/volume_policy/node.rs
+++ b/control-plane/agents/src/bin/core/controller/scheduling/volume_policy/node.rs
@@ -2,7 +2,7 @@ use crate::controller::scheduling::{
     nexus::GetSuitableNodesContext,
     resources::{NodeItem, PoolItem},
     volume::GetSuitablePoolsContext,
-    volume_policy::qualifies_inclusion_labels,
+    volume_policy::qualifies_label_criteria,
 };
 use std::collections::HashMap;
 use stor_port::types::v0::transport::NodeTopology;
@@ -73,6 +73,7 @@ impl NodeFilters {
     /// Should only attempt to use nodes having specific creation label if topology has it.
     pub(crate) fn topology(request: &GetSuitablePoolsContext, item: &PoolItem) -> bool {
         let volume_node_topology_inclusion_labels: HashMap<String, String>;
+        let volume_node_topology_exclusion_labels: HashMap<String, String>;
         match &request.topology {
             None => return true,
             Some(topology) => match &topology.node {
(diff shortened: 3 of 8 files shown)


control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -1048,3 +1059,52 @@ impl CreateVolumeExeVal for CreateVolumeSource<'_> {
...
+    let node_exclusion_label_match = match node_topology {
+        NodeTopology::Explicit(_) => true,
+        NodeTopology::Labelled(labelled_topology) => {
+            // If the NodeSpreadTopologyKey is present in the Volume Node Topology, then
+            // get the node details from replica candidate and
+            // check if the node labels is same as that of all existing replicas.
+            // If its same then don't consider the pool for replica creation
+            let mut exclusion_label_match = false;
+            if !labelled_topology.exclusion.is_empty() {
+                let candidate_node = context
```
tiagolobocastro: Can't we use the node from the candidate itself?

sinhaashish: No, the node in candidate replica is just the `NodeId` . 
```
pub struct CreateReplica {
    /// Id of the io-engine instance.
    pub node: NodeId,

Comment (tiagolobocastro) [reply control-plane/agents/src/bin/core/volume/operations.rs]: I mean, could we build a ReplicaCandidate newtype with the NodeWrapper which we have in the Candidates list

File: control-plane/agents/src/bin/core/volume/operations.rs [hunk 2]
```diff
@@ -1054,3 +1065,52 @@ impl CreateVolumeExeVal for CreateVolumeSource<'_> {
...
+                    .registry
+                    .specs()
+                    .node(&replica_candidate.node.clone())?
```
tiagolobocastro: Below as well
```suggestion
                    .node(&replica_candidate.node)?
```

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -1054,3 +1065,52 @@ impl CreateVolumeExeVal for CreateVolumeSource<'_> {
...
+                    .labels()
+                    .clone();
+
+                let replica_node = context
```
tiagolobocastro: Naming if these variables is a bit confusing, we seem to be getting the labels and not the node?

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -1054,3 +1065,52 @@ impl CreateVolumeExeVal for CreateVolumeSource<'_> {
...
+                    .registry
+                    .specs()
+                    .node(&replica.node.clone())?
+                    .labels()
+                    .clone();
+
+                for (key, _) in labelled_topology.exclusion.iter() {
+                    if candidate_node[key] == replica_node[key] {
```
tiagolobocastro: this will cause a panic is keys are missing, need to cater for this

control-plane/agents/src/bin/core/volume/operations.rs
sinhaashish: renamed

control-plane/agents/src/bin/core/volume/operations.rs
sinhaashish: fixed

control-plane/agents/src/bin/core/volume/operations.rs
sinhaashish: Why do we need a new type. IMHO the present approach is good enough, Its just that we need the node id to get the labels. 

If you are not convinced , I can work on that on another PR

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -995,8 +996,18 @@ impl CreateVolumeExe for CreateVolume {
         for replica in candidates.candidates() {
             if replicas.len() >= self.replicas as usize {
                 break;
-            } else if replicas.iter().any(|r| r.node == replica.node) {
-                // don't reuse the same node
+            } else if replicas.iter().any(|r| {
+                r.node == replica.node
+                    || spread_label_is_same(r, replica, context).unwrap_or_else(|error| {
```
tiagolobocastro, line 1001: remind me what are we doing here again?
We've got the candidates list in step "1".
And on step "2" we check if the spread is the same? Is that because the list of candidates includes all replicas on a single list, which would not be "spread"?

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -1054,3 +1065,56 @@ impl CreateVolumeExeVal for CreateVolumeSource<'_> {
...
+            // check if the node labels is same as that of all existing replicas.
+            // If its same then don't consider the pool for replica creation
+            let mut exclusion_label_match = false;
+            if !labelled_topology.exclusion.is_empty() {
+                let candidate_node_label = context
+                    .registry
+                    .specs()
+                    .node(&replica_candidate.node)?
+                    .labels()
+                    .clone();
```
tiagolobocastro: Let's avoid the clone since we only need to compare the labels anyway

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -1054,3 +1065,56 @@ impl CreateVolumeExeVal for CreateVolumeSource<'_> {
...
+                    .node(&replica.node)?
+                    .labels()
+                    .clone();
+
+                for key in labelled_topology.exclusion.keys() {
+                    if let (Some(candidate_value), Some(replica_value)) =
+                        (candidate_node_label.get(key), replica_node_label.get(key))
+                    {
+                        if candidate_value == replica_value {
+                            exclusion_label_match = true;
```
tiagolobocastro: Why do we break out after a single match? What if the other keys don't match?

control-plane/agents/src/bin/core/volume/operations.rs
sinhaashish: Here the following thing are done
1. Get the candidate list
2. We check if the if the node on which the replica is already placed is again in the candidate list. If yes then we skip the node, to ensure that replica are present on different node. 
3. Along with step 2 we also check if spread is the same. 
Like if node with label `performance: Gold` is selected for 1st replica , the second replica doesn't goes on node with label `performance: Gold` . It can go on `performance: Silver` or `performance: Emerald`. 

The candidates contains all the replica candidate , it doesn't knows about anything spread till it reaches this part of code.
```
SpreadTopologyKey: "performance" - is a exclusion 
1: "performance: Gold"
2: "performance: Silver"
3: "performance: Emerald"
```

control-plane/agents/src/bin/core/volume/operations.rs
sinhaashish: In spread if any on the key matches , we must skip that node, thus we break even if a single key matches.

tiagolobocastro: Ah got it, btw we can just break out with return Ok(false), we don't need this variable

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -1054,3 +1065,56 @@ impl CreateVolumeExeVal for CreateVolumeSource<'_> {
         }
     }
 }
+
+// spread_label_is_same checks if the node labels is same as that of all existing replicas.
+fn spread_label_is_same(
```
tiagolobocastro, line 1070: I think this should "live" in the scheduling modules and not here.

tiagolobocastro (approved):

Abhinandan-Purkait (approved):

sinhaashish:
bors merge

sinhaashish:
bors merge

control-plane/agents/src/bin/core/volume/operations.rs
sinhaashish: tried putting it scheduling but it errored out 
```
error[E0446]: restricted type `volume::operations::Context<'_>` in public interface
   --> control-plane/agents/src/bin/core/volume/scheduling.rs:149:1
    |
149 | / pub(crate) fn spread_label_is_same(
150 | |     replica: &Replica,
151 | |     replica_candidate: &CreateReplica,
152 | |     context: &Context,
153 | | ) -> Result<bool, SvcError> {
    | |___________________________^ can't leak restricted type
    ```

**Comment by sinhaashish** [comment]
bors try

**Comment by pchandra19** [comment]
bors try

**Comment by pchandra19** [comment]
bors cancel

**Comment by pchandra19** [comment]
bors try

**Comment by sinhaashish** [comment]
bors cancel

**Comment by sinhaashish** [comment]
bors try

**Comment by sinhaashish** [comment]
bors merge

**Comment by sinhaashish** [comment]
bors try

**Comment by sinhaashish** [comment]
bors try

**Comment by sinhaashish** [comment]
bors merge

**Comment by sinhaashish** [comment]
bors merge

**Comment by sinhaashish** [comment]
bors merge

File: control-plane/agents/src/bin/core/volume/operations.rs (hunk 6 above)
Comment (tiagolobocastro) [reply control-plane/agents/src/bin/core/volume/operations.rs]: Marked as resolved but it's not...

Comment (sinhaashish) [reply control-plane/agents/src/bin/core/volume/operations.rs]: I tried removing clone().
but there as an errror 
```
error[E0716]: temporary value dropped while borrowed
    --> control-plane/agents/src/bin/core/volume/operations.rs:1091:44
     |
1091 |                   let candidate_node_label = context
     |  ____________________________________________^
1092 | |                     .registry
1093 | |                     .specs()
1094 | |                     .node(&replica_candidate.node)?
     | |___________________________________________________^ creates a temporary value which is freed while still in use
1095 |                       .labels();
     |                                - temporary value is freed at the end of this statement
...
1103 |                           (candidate_node_label.get(key), replica_node_label.get(key))
     |                            -------------------- borrow later used here
     |
     = note: consider using a `let` binding to create a longer lived value
```
since `get` uses this map , i had to use clone.

Comment (tiagolobocastro) [reply control-plane/agents/src/bin/core/volume/operations.rs]: Rather than simply ignore due to the error, let's have a look at what it's saying instead.
The problem is that labels returns a reference which goes out of scope once the temporary node is out of scope (at the end of the statement).
So I suspect you can simply do this:
```rust
let candidate_node = ....;
let candidate_node_labels = candidate_node.labels();
```
or just use `candidate_node.labels()` directly

sinhaashish: done

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -1054,3 +1065,50 @@ impl CreateVolumeExeVal for CreateVolumeSource<'_> {
...
+        NodeTopology::Labelled(labelled_topology) => {
+            // If the NodeSpreadTopologyKey is present in the Volume Node Topology, then
+            // get the node details from replica candidate and
+            // check if the node labels is same as that of all existing replicas.
+            // If its same then don't consider the pool for replica creation
+            let mut exclusion_label_match = false;
+            if !labelled_topology.exclusion.is_empty() {
+                let candidate_node = context.registry.specs().node(&replica_candidate.node)?;
+                // .labels()
+                // .clone();
```
```suggestion
```

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -1054,3 +1065,50 @@ impl CreateVolumeExeVal for CreateVolumeSource<'_> {
...
+
+                let replica_node = context.registry.specs().node(&replica.node)?;
+                // .labels()
```
```suggestion
```

sinhaashish:
bors try

sinhaashish:
bors try

sinhaashish:
bors merge

sinhaashish:
bors try

sinhaashish:
bors merge

sinhaashish:
bors merge

sinhaashish:
bors cancel

sinhaashish:
bors ping

sinhaashish:
bors merge

PR #764, opened 2024-02-29
add replica resize operation to ReplicaOperations 

dsharma-dc:
This will be mainly used for test purpose, and not in the normal prod workflow.

### DIFF
--- a/control-plane/grpc/proto/v1/replica/replica.proto
+++ b/control-plane/grpc/proto/v1/replica/replica.proto
@@ -157,6 +157,19 @@ message UnshareReplicaRequest {
   optional google.protobuf.StringValue pool_uuid = 5;
 }
 
+// Resize a replica to the requested size
+// For internal usage mostly.
+message ResizeReplicaRequest {
+  // id of the io-engine instance
+  string node_id = 1;
+  // uuid of the replica
+  string replica_id = 2;
+  // requested new size for the replica
+  uint64 requested_size = 3;
+  // id of the pool
+  string pool_id = 4;
+}
+
 // Reply type for a CreateReplica request
 message CreateReplicaReply {
   oneof reply {
@@ -189,6 +202,14 @@ message ShareReplicaReply {
 // Reply type for a UnshareReplica request
 message UnshareReplicaReply {
   optional common.ReplyError error = 1;
+}
+
+// Reply type for a ResizeReplica request
+message ResizeReplicaReply {
+  oneof reply {
+    Replica replica = 1;
+    common.ReplyError error = 2;
+  }
 }
 
 // Get all the replicas from specific node and pool
@@ -242,4 +263,5 @@ service ReplicaGrpc {
   rpc GetReplicas (GetReplicasRequest) returns (GetReplicasReply) {}
   rpc ShareReplica (ShareReplicaRequest) returns (ShareReplicaReply) {}
   rpc UnshareReplica (UnshareReplicaRequest) returns (UnshareReplicaReply) {}
+  rpc ResizeReplica (ResizeReplicaRequest) returns (ResizeReplicaReply) {}
 }
--- a/control-plane/grpc/src/operations/replica/traits.rs
+++ b/control-plane/grpc/src/operations/replica/traits.rs
@@ -2,10 +2,10 @@ use crate::{
     common,
     context::Context,
     misc::traits::{StringValue, ValidateRequestTypes},
-    replica,
+    operations::volume::traits::TryIntoId,
     replica::{
-        get_replicas_request, CreateReplicaRequest, DestroyReplicaRequest, ShareReplicaRequest,
-        UnshareReplicaRequest,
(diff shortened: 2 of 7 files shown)


dsharma-dc:
bors try

dsharma-dc:
bors try

control-plane/grpc/src/operations/replica/traits.rs
```diff
@@ -54,6 +53,12 @@ pub trait ReplicaOperations: Send + Sync {
         req: &dyn UnshareReplicaInfo,
         ctx: Option<Context>,
     ) -> Result<(), ReplyError>;
+    /// Resize a replica
```
tiagolobocastro:
```suggestion
    /// Resize a replica.
```

control-plane/grpc/proto/v1/replica/replica.proto
```diff
@@ -157,6 +157,19 @@ message UnshareReplicaRequest {
   optional google.protobuf.StringValue pool_uuid = 5;
 }
 
+// Resize a replica to the requested size
+// For internal usage mostly.
+message ResizeReplicaRequest {
+  // id of the io-engine instance
+  string node_id = 1;
+  // uuid of the replica
+  google.protobuf.StringValue replica_id = 2;
```
tiagolobocastro:
```suggestion
  string replica_id = 2;
```

control-plane/grpc/proto/v1/replica/replica.proto
```diff
@@ -157,6 +157,19 @@ message UnshareReplicaRequest {
...
+  // requested new size of the volume
```
tiagolobocastro:
```suggestion
  // requested new size for the replica
```

control-plane/grpc/proto/v1/replica/replica.proto
```diff
@@ -157,6 +157,19 @@ message UnshareReplicaRequest {
...
+  uint64 requested_size = 3;
+    // id of the pool
```
tiagolobocastro:
```suggestion
  // id of the pool
  string pool_id = 4;
  // uuid of the pool
  optional string pool_uuid = 5;
```

tiagolobocastro (commented):
What is the intended use case for this replica size?
Do we want to allow this for replicas which are part of a volume? What is the use case?

dsharma-dc:
> What is the intended use case for this replica size? Do we want to allow this for replicas which are part of a volume? What is the use case?

This is primarily added so we could test reconciler. It's not expected to be used for normal volume resize workflows.

control-plane/grpc/proto/v1/replica/replica.proto
dsharma-dc: @tiagolobocastro All the messages in this file have replica_id as StringValue. Any particular reason to make that `string` here?

control-plane/grpc/proto/v1/replica/replica.proto
dsharma-dc: There is no `pool_uuid` in the transport's `ResizeReplica` type as of today, hence didn't keep it here.

tiagolobocastro: interesting, there should be one there..

tiagolobocastro: ah I see we already have a mix of having uuid and not having it :( 
So seems like we can skip the pool altogether here as per existing requests

control-plane/grpc/proto/v1/replica/replica.proto
tiagolobocastro: they were added mistakenly... we've since switched to string

dsharma-dc: ok, will change. I guess having it `string` will help simplify code as we may not need `Validated*` type in between.

control-plane/grpc/src/operations/replica/traits.rs
tiagolobocastro: Did you forget to push? This is resolved but changes are not visible

dsharma-dc: > Did you forget to push? This is resolved but changes are not visible

Yeah it wasn't pushed yet. Done now.

control-plane/grpc/src/operations/replica/traits.rs
```diff
@@ -782,6 +788,86 @@ impl ValidateRequestTypes for UnshareReplicaRequest {
     }
 }
 
+/// ResizeReplicaInfo trait for the replica resize to be implemented by entities which want to
+/// avail this operation
```
tiagolobocastro: please terminate all doc comments with `.`

tiagolobocastro (approved):

control-plane/grpc/src/operations/replica/traits.rs
dsharma-dc: Done

hrudaya21 (approved):

dsharma-dc:
bors merge

PR #765, opened 2024-03-01
fix(csi-driver): support getting device size from devpath

niladrih:
Changes:
- the Device::lookup() API doesn't return a dev-name, but instead returns a dev-path. Changed the size function to accept either.
- the `failure!` macro prints errors event if they are never encountered, when used with the `ok_or()` combinator. Using the `ok_or_else()` combinator instead.
- Adjusting the expected capacity of the nexus during NodeExpandVolume call to be at most 10MiB less than the requested capacity.

### DIFF
--- a/control-plane/csi-driver/src/bin/controller/controller.rs
+++ b/control-plane/csi-driver/src/bin/controller/controller.rs
@@ -978,23 +978,24 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
     #[instrument(err, skip(self))]
     async fn controller_expand_volume(
         &self,
-        request: tonic::Request<ControllerExpandVolumeRequest>,
-    ) -> Result<tonic::Response<ControllerExpandVolumeResponse>, tonic::Status> {
-        let request = request.into_inner();
+        request: Request<ControllerExpandVolumeRequest>,
+    ) -> Result<Response<ControllerExpandVolumeResponse>, Status> {
+        let args = request.into_inner();
+        trace!(volume.uuid = %args.volume_id, request = ?args);
 
-        let vol_uuid = Uuid::parse_str(&request.volume_id).map_err(|error| {
+        let vol_uuid = Uuid::parse_str(&args.volume_id).map_err(|error| {
             Status::invalid_argument(format!(
                 "Malformed volume UUID '{}': {error}",
-                request.volume_id
+                args.volume_id
             ))
         })?;
 
-        let requested_size = request
+        let requested_size = args
             .capacity_range
             .as_ref()
             .ok_or(Status::invalid_argument(format!(
-                "Cannot expand volume '{}': invalid request {request:?}: missing CapacityRange",
-                request.volume_id
+                "Cannot expand volume '{}': invalid request {args:?}: missing CapacityRange",
+                args.volume_id
             )))?
             .required_bytes;
 
@@ -1003,7 +1004,7 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
         // assume NodeExpandVolume is required if it is not clearly defined that the volume is
         // Block type volume.
         let node_expansion_required = !matches!(
-            request.volume_capability.as_ref(),
+            args.volume_capability.as_ref(),
             Some(vc) if matches!(vc.access_type, Some(AccessType::Block(_)))
         );
 
@@ -1020,6 +1021,10 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
                 error => Status::from(error),
             })?;
 
+        debug!(
+            size_bytes = requested_size,
+            "Expansion succeeded for volume {vol_uuid}"
+        );
         Ok(tonic::Response::new(ControllerExpandVolumeResponse {
             capacity_bytes: vol.spec.size as i64,
             node_expansion_required,
--- a/control-plane/csi-driver/src/bin/node/dev.rs
+++ b/control-plane/csi-driver/src/bin/node/dev.rs
@@ -142,10 +142,20 @@ impl Device {
 }
 
 /// Get the block device capacity size from the sysfs block count.
-pub(crate) fn get_size_from_dev_name<N: AsRef<Path>>(dev_name: N) -> Result<usize, DeviceError> {
+/// Arg can be /dev/device-name style dev-path or just the device name itself.
+pub(crate) fn sysfs_dev_size<N: AsRef<Path>>(device: N) -> Result<usize, DeviceError> {
     // Linux sectors are 512 byte long. This is fixed.
     // Ref: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/linux/types.h?id=v5.15#n117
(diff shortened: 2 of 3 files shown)


control-plane/csi-driver/src/bin/node/dev.rs
```diff
@@ -142,11 +142,21 @@ impl Device {
 }
 
 /// Get the block device capacity size from the sysfs block count.
-pub(crate) fn get_size_from_dev_name<N: AsRef<Path>>(dev_name: N) -> Result<usize, DeviceError> {
+/// Arg can be /dev/device-name style dev-path or just the device name itself.
+pub(crate) fn get_sysfs_size_for_device<N: AsRef<Path>>(device: N) -> Result<usize, DeviceError> {
```
tiagolobocastro:
```suggestion
pub(crate) fn sysfs_dev_size<N: AsRef<Path>>(device: N) -> Result<usize, DeviceError> {
```

control-plane/csi-driver/src/bin/node/node.rs
```diff
@@ -512,21 +510,19 @@ impl node_server::Node for Node {
                     error
                 )
             })?
-            .ok_or(failure!(
-                Code::InvalidArgument,
-                "failed to find a device for volume {}",
-                vol_uuid
-            ))?
+            .ok_or(Status::invalid_argument(format!(
```
tiagolobocastro: ok_or_else would fix this too

dsharma-dc (approved):

control-plane/csi-driver/src/bin/node/node.rs
```diff
@@ -536,7 +540,7 @@ impl node_server::Node for Node {
         // down by 5 MiB so that the capacity of the device is verified to be greater than or
         // equal to this corrected capacity. This is required because we are not comparing the
         // required capacity against the REST API Volume resource.
-        const MAX_NEXUS_CAPACITY_DIFFERENCE: i64 = 5 * 1024 * 1024;
+        const MAX_NEXUS_CAPACITY_DIFFERENCE: i64 = 10 * 1024 * 1024;
```
tiagolobocastro, line 545: hmm might want to do the opposite and add 10 to device_capacity instead? Otherwise what if resize is to < 10MiB ?

control-plane/csi-driver/src/bin/node/node.rs
```diff
@@ -536,7 +540,7 @@ impl node_server::Node for Node {
         // down by 5 MiB so that the capacity of the device is verified to be greater than or
         // equal to this corrected capacity. This is required because we are not comparing the
         // required capacity against the REST API Volume resource.
-        const MAX_NEXUS_CAPACITY_DIFFERENCE: i64 = 5 * 1024 * 1024;
```
dsharma-dc, line 539: Comment needs update saying 10MiB includes additional leeway on top of 5MiB.

control-plane/csi-driver/src/bin/node/node.rs
tiagolobocastro: we could also add check on control-plane that resize must be >10MiB difference? Otherwise we could have a whole here where we'll simply not expand?

niladrih: > what if resize is to < 10MiB

It's an `i64`, should be able to work. Moving the 10MiB to the other side anyway.

control-plane/csi-driver/src/bin/controller/controller.rs
```diff
@@ -981,6 +981,7 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
         request: tonic::Request<ControllerExpandVolumeRequest>,
     ) -> Result<tonic::Response<ControllerExpandVolumeResponse>, tonic::Status> {
         let request = request.into_inner();
+        info!("Volume expansion request: {:?}", &request);
```
tiagolobocastro: in line with existing methods:
```suggestion
        tracing::trace!(volume.uuid = %args.volume_id, request = ?args);
```

control-plane/csi-driver/src/bin/controller/controller.rs
```diff
@@ -1020,6 +1021,10 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
                 error => Status::from(error),
             })?;
 
+        info!(
+            size_bytes = requested_size,
+            "Expansion succeeded for volume {}", &vol_uuid
+        );
```
```suggestion
        debug!(
            size_bytes = requested_size,
            "Expansion succeeded for volume {vol_uuid}"
        );
```

control-plane/csi-driver/src/bin/node/node.rs
```diff
@@ -475,6 +475,7 @@ impl node_server::Node for Node {
         //===============================CsiAccessType=============================================
 
         let request = request.into_inner();
+        info!("Volume expansion request: {:?}", &request);
```
tiagolobocastro: Again, keep with existing ones..
Also might want to move this up?

```suggestion
        trace!("node_expand_volume {request:?}");
```

control-plane/csi-driver/src/bin/node/node.rs
tiagolobocastro: it would underflow...

control-plane/csi-driver/src/bin/node/node.rs
```diff
@@ -590,6 +603,10 @@ impl node_server::Node for Node {
             .await
             .map_err(|err| failure!(Code::Internal, "{}", err))?;
 
+        info!(
+            size_bytes = required_bytes,
+            "Expansion succeeded for volume {}", &vol_uuid
```
tiagolobocastro:
```suggestion
            "Expansion succeeded for volume {vol_uuid}"
```

tiagolobocastro (approved):

control-plane/csi-driver/src/bin/node/node.rs
niladrih: Made this a debug.

control-plane/csi-driver/src/bin/node/node.rs
niladrih: Moved this up, keeping the request, instead of shadowing it.

niladrih:
bors merge

control-plane/csi-driver/src/bin/node/node.rs
```diff
@@ -555,20 +561,26 @@ impl node_server::Node for Node {
         // The CSI spec, as of v1.8.0, treats volume_capability as an optional field, so
         // in an attempt to be as spec-compliant as possible, we must have a plan-B for the
         // filesystem identification part.
-        let filesystem_handle = match CsiAccessType::from(&request.volume_capability) {
+        let filesystem_handle = match CsiAccessType::from(&args.volume_capability) {
             // volume_capability AccessType says this is not a 'Mount' type. Nothing to do.
-            CsiAccessType::NotAFilesystem => return success_result,
+            CsiAccessType::NotAFilesystem => {
+                info!(
```
tiagolobocastro:
```suggestion
                debug!(
```

niladrih:
bors merge

PR #768, opened 2024-03-01
fix(csi-controller/pv-watcher): race between creation and gc

tiagolobocastro:
Check for deletion candidates using PVC and when such candidate is found
also ensure there is no corresponding PV.
This is required to avoid a race during creation where the volume is created
but the PV is not yet.
Also, do this check twice, as otherwise we might find false positives when
a PVC is deleted before the PV is sucessfully created, leading to confusing
message about GC.

### DIFF
--- a/control-plane/csi-driver/src/bin/controller/pvwatcher.rs
+++ b/control-plane/csi-driver/src/bin/controller/pvwatcher.rs
@@ -1,17 +1,18 @@
 use crate::client::ListToken;
 use futures::TryStreamExt;
-use k8s_openapi::api::core::v1::PersistentVolume;
+use k8s_openapi::api::core::v1::{PersistentVolume, PersistentVolumeClaim};
 use kube::{
-    api::Api,
+    api::{Api, ListParams},
     runtime::{watcher, WatchStreamExt},
     Client, ResourceExt,
 };
 use tracing::{debug, error, info};
 
-/// Struct for PV Garbage collector
+/// Struct for PV Garbage collector.
 #[derive(Clone)]
-pub(crate) struct PvGarbageCollector {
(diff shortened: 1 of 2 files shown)


Abhinandan-Purkait (approved):

control-plane/csi-driver/src/bin/controller/pvwatcher.rs
```diff
@@ -101,8 +103,39 @@ impl PvGarbageCollector {
...
+            return;
+        };
+
+        let mut gc_uids = Vec::with_capacity(volumes.len());
+        for volume_uid in volumes {
+            if self.is_vol_orphan(volume_uid, &pvcs).await {
+                gc_uids.push(volume_uid);
+            }
+        }
+        tokio::time::sleep(std::time::Duration::from_secs(2)).await;
```
Abhinandan-Purkait: Do we need to sleep here?

tiagolobocastro: It's just to avoid interfering with the regular CSI deletion, by giving it some time.

niladrih (approved):

control-plane/csi-driver/src/bin/controller/pvwatcher.rs
niladrih: Should we wait a little longer? Maybe 15 or 20 seconds?

abhilashshetty04 (approved):

control-plane/csi-driver/src/bin/controller/pvwatcher.rs
tiagolobocastro: I've made it 60s

tiagolobocastro:
bors merge

PR #770, opened 2024-03-04
fix(volume/resize): more appropriate error mapping

dsharma-dc:
Quoting from gRPC documentation:
"There is a fair bit of overlap between FAILED_PRECONDITION and OUT_OF_RANGE. We recommend using OUT_OF_RANGE (the more specific error) when it applies so that callers who are iterating through a space can easily look for an OUT_OF_RANGE error to detect when they are done."

The Mayastor CSI driver maps ResourceExhausted to OutOfRange for VolumeExpansion.

### DIFF
--- a/control-plane/agents/src/common/errors.rs
+++ b/control-plane/agents/src/common/errors.rs
@@ -907,7 +907,7 @@ impl From<SvcError> for ReplyError {
                 extra,
             },
             SvcError::ResizeReplError { .. } => ReplyError {
-                kind: ReplyErrorKind::FailedPrecondition,
+                kind: ReplyErrorKind::ResourceExhausted,
                 resource: ResourceKind::Replica,
                 source,
                 extra,


niladrih (approved):

Abhinandan-Purkait:
When do we hit OutOfRange? Fyi this can cause burst retries.

dsharma-dc:
> When do we hit OutOfRange? Fyi this can cause burst retries.

We could come in this situation when core-agent throws `ReplicaResizeErr` which is either when replica(s)' pool don't have sufficient space, or a replica is offline. Since we can't differentiate between these today in scheduling logic, ResourceExhausted code maps closest.

Abhinandan-Purkait:
> > When do we hit OutOfRange? Fyi this can cause burst retries.
> 
> We could come in this situation when core-agent throws `ReplicaResizeErr` which is either when replica(s)' pool don't have sufficient space, or a replica is offline. Since we can't differentiate between these today in scheduling logic, ResourceExhausted code maps closest.

This makes sense, but why are mapping ResourceExhausted to OutOfRange?

niladrih:
> > > When do we hit OutOfRange? Fyi this can cause burst retries.
> > 
> > 
> > We could come in this situation when core-agent throws `ReplicaResizeErr` which is either when replica(s)' pool don't have sufficient space, or a replica is offline. Since we can't differentiate between these today in scheduling logic, ResourceExhausted code maps closest.
> 
> This makes sense, but why are mapping ResourceExhausted to OutOfRange?

@Abhinandan-Purkait -- the CSI spec recommends using the OutOfRange Status for invalid capacity_range values. This seems appropriate for a resize request which exceeds the capacity possible for Replicas. I am not sure if it fits the replica-offline case. The spec asks us to use a message to describe the issue, and with that it would see it fits the replica-offline case too.

Abhinandan-Purkait:
> > > > When do we hit OutOfRange? Fyi this can cause burst retries.
> > > 
> > > 
> > > We could come in this situation when core-agent throws `ReplicaResizeErr` which is either when replica(s)' pool don't have sufficient space, or a replica is offline. Since we can't differentiate between these today in scheduling logic, ResourceExhausted code maps closest.
> > 
> > 
> > This makes sense, but why are mapping ResourceExhausted to OutOfRange?
> 
> @Abhinandan-Purkait -- the CSI spec recommends using the OutOfRange Status for invalid capacity_range values. This seems appropriate for a resize request which exceeds the capacity possible for Replicas. I am not sure if it fits the replica-offline case. The spec asks us to use a message to describe the issue, and with that it would see it fits the replica-offline case too.

Okay, but my only concern was that if we return OutOfRange it will no do exponential backoff which would flood the csi logs. Whatever looks good to you guys is fine.

Abhinandan-Purkait (approved):

tiagolobocastro:
> Okay, but my only concern was that if we return OutOfRange it will no do exponential backoff which would flood the csi logs. Whatever looks good to you guys is fine.

That's a problem indeed. Also I don't think OutOfRange is the appropriate error code..

tiagolobocastro (approved):

dsharma-dc:
bors merge

dsharma-dc:
bors cancel

dsharma-dc:
bors merge

PR #771, opened 2024-03-04
fix(csi-controller): use 'internal' status with PreconditionFailed error

niladrih:
The CSI spec requires that FailedPrecondition GRPC status be returned iff
a CO asks for online resize when the plugin is only capable of offline
resize. To avoid generating error logs which suggest that this CSI driver
doesn't support online resize, I am mapping the HTTP 412 code to GRPC
Internal status code.
This alse maps REST ResourceExhaused to GRPC ResourceExhausted.

Ref:
- https://github.com/container-storage-interface/spec/blob/v1.5.0/spec.md#controllerexpandvolume-errors

### DIFF
--- a/control-plane/csi-driver/src/bin/controller/controller.rs
+++ b/control-plane/csi-driver/src/bin/controller/controller.rs
@@ -1016,8 +1016,16 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
             .expand_volume(&vol_uuid, requested_size as u64)
             .await
             .map_err(|error| match error {
-                ApiClientError::PreconditionFailed(msg) => Status::failed_precondition(msg),
-                ApiClientError::ResourceExhausted(msg) => Status::out_of_range(msg),
+                // This is mapped to the GRPC 'Internal' status because the failed_precondition
+                // holds special significance in the CSI external-resizer sidecar. If the
+                // FailedPrecondition status is returned to the external-resizer, then it fails
+                // and never retries, and assumes the cause to be 'online resize request to
+                // a plugin which only supports offline resize'. This is not the case for us, and
+                // we do want retries from the resizer for all possible errors, so we're making
+                // sure we never return it, even if ApiClientError::PreconditionFailed is mapped
+                // to Status::failed_precondition at some point, in some other context.
+                ApiClientError::PreconditionFailed(msg) => Status::internal(msg),
+                ApiClientError::ResourceExhausted(msg) => Status::resource_exhausted(msg),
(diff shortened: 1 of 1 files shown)


dsharma-dc (approved):

Abhinandan-Purkait:
Where does it say, `FailedPrecondition` can be used if and only if `CO asks for online resize when the plugin is only capable of offline resize`? Then does this mean we cannot return failed precondition for any other matching error?

niladrih:
> Where does it say, `FailedPrecondition` can be used if and only if `CO asks for online resize when the plugin is only capable of offline resize`? Then does this mean we cannot return failed precondition for any other matching error?

@Abhinandan-Purkait -- There's a link with the appropriate errors for a few well known cases in the PR description. With Kubernetes, we have seen PVC events which say that this plugin isn't capable of online resize, and the volume must be taken offline for resize to succeed. In the volume-expansion context, we cannot use the GRPC FailedPrecondition status for anything else, it would seem.

Abhinandan-Purkait:
@niladrih Thanks for explaining. Yeah in that case `iff` make sense.

Abhinandan-Purkait (approved):

control-plane/csi-driver/src/bin/controller/controller.rs
```diff
@@ -1016,7 +1016,6 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
             .expand_volume(&vol_uuid, requested_size as u64)
             .await
             .map_err(|error| match error {
-                ApiClientError::PreconditionFailed(msg) => Status::failed_precondition(msg),
                 ApiClientError::ResourceExhausted(msg) => Status::out_of_range(msg),
```
tiagolobocastro: Also I wonder, do we need to remap resource exhausted to out of range?
Why not return resource exhausted status?

tiagolobocastro (approved):

control-plane/csi-driver/src/bin/controller/controller.rs
niladrih: I've done it because the spec.md lists this as a special case for invalid capacity_range. ResourceExhausted fits here as well. I've just concluded testing the retry durations for error cases. And all, but FailedPrecondition, are exponential backoff errors. The errors don't have anything different about them in the retry context, so we might as well go ahead pick the one which best fits the case.

niladrih: I'll change this to ResourceExhausted.

niladrih:
bors merge

PR #776, opened 2024-03-07
Blocking nvme disconnect fixes

tiagolobocastro:
fix(agents/ha/cluster): increase path replacement timeout
    
    In a system test run we've seen the disconnect of an old path take 50s
    to complete.
    We're not sure why this is the case as we cannot reproduce it yet but
    for now we can increase the timeout.
    This is probably fine as if we can't talk to the node where the app is
    then we can't do much more anyway.
    
    We should also consider whether we need to wait for the disconnect at
    all. If we can find out if the disconnect has been started, then we
    can probably just complete that async and succeed the replace!
    
---

    fix(agents/ha/node): disconnect controller async
    
    Dependent nvmeadm crate is blocking and certain operations such as
    disconnecting the controller can take quite some time.
    Here we are just doing a WA by running the disconnect in a tokio
    blocking spawn.
    todo: Ideally we should either make nvmeadm async or somehow provide
    async as alternative, which won't be great as the code will
    likely get duplicated...
    
### DIFF
--- a/control-plane/agents/src/bin/ha/node/server.rs
+++ b/control-plane/agents/src/bin/ha/node/server.rs
@@ -97,52 +97,48 @@ impl NodeAgentSvc {
 }
 
 /// Disconnect cached NVMe controller.
-fn disconnect_controller(ctrlr: &NvmeController, new_path: String) -> Result<(), SvcError> {
-    let parsed_path = parse_uri(new_path.as_str())?;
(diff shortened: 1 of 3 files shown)


Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

PR #777, opened 2024-03-08
feat(volume/resize): don't allow volume resize for snapshot restore vols

dsharma-dc:
If a volume is expanded after taking a snapshot or a restore created from that snapshot is expanded, then the IOs to the new io-units(block ranges) fails because the lba range falls out of bound for replica's parent device(snapshot). Till the issue is fixed, this check ensures we don't try to expand a volume if there is snapshot present. SPDK issue 3108.


tiagolobocastro (approved):

Abhinandan-Purkait (approved):

dsharma-dc:
bors merge

PR #781, opened 2024-03-13
feat(topology): add affinity topology key to rest, grpc, stor-port & csi-driver

sinhaashish:
The code for core will be in follow up PR.

I have used string as affinity not as set as set is not supported in openapi v3.

### DIFF
--- a/control-plane/rest/openapi-specs/v0_api_spec.yaml
+++ b/control-plane/rest/openapi-specs/v0_api_spec.yaml
@@ -2475,6 +2475,8 @@ components:
           - ''
         inclusion:
           - ''
+        affinitykey:
+          - ''
       description: labelled topology
       type: object
       properties:
@@ -2504,9 +2506,22 @@ components:
           type: object
           additionalProperties:
             type: string
+        affinitykey:
+          example: ''
+          description: |-
+            This feature includes resources with identical $label keys. For example,
+             if the affinity key is set to "Zone":
+             Initially, a resource that matches the label is selected, example "Zone: A".
+             Subsequently, all other resources must match the given label "Zone: A",
+             effectively adding this requirement as an inclusion label.
+          type: array
+          items:
+            example: 'Zone'
+            type: string
       required:
         - exclusion
         - inclusion
+        - affinitykey
     Topology:
       description: node and pool topology for volumes
       type: object
--- a/control-plane/agents/src/bin/core/tests/deserializer.rs
+++ b/control-plane/agents/src/bin/core/tests/deserializer.rs
@@ -112,6 +112,7 @@ fn test_deserialization_v1_to_v2() {
                             );
                             labels
                         },
+                        affinity: Vec::new(),
                     })),
                 }),
                 sequencer: Default::default(),
--- a/control-plane/csi-driver/src/bin/controller/controller.rs
+++ b/control-plane/csi-driver/src/bin/controller/controller.rs
@@ -1068,6 +1068,8 @@ fn context_into_topology(context: &CreateParams) -> CreateVolumeTopology {
     let mut pool_inclusive_label_topology: HashMap<String, String> = HashMap::new();
     let mut node_inclusive_label_topology: HashMap<String, String> = HashMap::new();
     let mut node_exclusive_label_topology: HashMap<String, String> = HashMap::new();
+    let mut pool_affinity_label_topology: Vec<String> = Vec::<String>::new();
+    let mut node_affinity_label_topology: Vec<String> = Vec::<String>::new();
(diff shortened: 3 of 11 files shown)


control-plane/rest/openapi-specs/v0_api_spec.yaml
```diff
@@ -2497,9 +2499,22 @@ components:
           type: object
           additionalProperties:
             type: string
+        affinity  :
+          example: ''
+          description: |-
+            Includes resources with the same $label eg:
```
```suggestion
            Includes resources with the same $label key eg:
```

tiagolobocastro: Does this need to be a map? Can't it just be a set or a list?

Abhinandan-Purkait: I think a map is better then we would not need to parse to map again at some point?

Abhinandan-Purkait (approved):

control-plane/rest/openapi-specs/v0_api_spec.yaml
tiagolobocastro: hmm why do we need it as a map?
Once the key is set, don't we simply set equivalent key/val on the include map? CC @sinhaashish

sinhaashish: Map makes it look consistent , yes once the key is set we will set the equivalent key/value on include map

tiagolobocastro: It may be consistent, but include and exclude have to work with a map as they might have values. Here affinity will never have a value? So Set would be preferable IMHO as it clearly states the behaviour. 
But if you think map is better then at the very least lets ensure that value is always empty string when we process the request from rest.

tiagolobocastro (approved):

control-plane/rest/openapi-specs/v0_api_spec.yaml
sinhaashish: @tiagolobocastro  @Abhinandan-Purkait   I have modified the code to use list instead of map. PTAL

control-plane/rest/openapi-specs/v0_api_spec.yaml
```diff
@@ -2497,9 +2499,23 @@ components:
           type: object
           additionalProperties:
             type: string
+        affinity  :
```
```suggestion
        affinity:
```

tiagolobocastro: Should it be called affinityKeys ?

control-plane/rest/openapi-specs/v0_api_spec.yaml
```diff
@@ -2497,9 +2499,23 @@ components:
...
+          example: ''
+          description: |-
+            Includes resources with the same $label key eg:
+             if label is "Zone":
+             A resource matching the label is decided initially and
+             all other resources are paired with the same label
+             A resource with "Zone" would be choose the label as "Zone: A"
+             and all other resoures will be paired with label "Zone: A"
```
tiagolobocastro: I think this wording can be improved

sinhaashish: done

control-plane/rest/openapi-specs/v0_api_spec.yaml
sinhaashish: changed

control-plane/rest/openapi-specs/v0_api_spec.yaml
```diff
@@ -2497,9 +2499,24 @@ components:
           type: object
           additionalProperties:
             type: string
+        affinitykey  :
```
```suggestion
        affinitykey:
```

control-plane/rest/openapi-specs/v0_api_spec.yaml
```diff
@@ -2497,9 +2499,24 @@ components:
...
+          example: ''
+          description: |-
+            This feature includes resources with identical $label keys. For example,
+             if the affinity key is set to "Zone":
+             Initially, a resource that matches the label is selected.
+             Subsequently, all other resources are associated with the same label.
+             For instance, if a resource with the label "Zone" is chosen,
+             it will be assigned the label "Zone: A".
+             Consequently, all other resources will also be labeled as "Zone: A".
```
```suggestion
             Initially, a resource that matches the label is selected, example "Zone: A".
             Subsequently, all other resources must match the given label "Zone: A",
             effectively adding this requirement as an inclusion label.
```

tiagolobocastro (approved):

sinhaashish:
bors try

sinhaashish:
bors try

sinhaashish:
bors merge

sinhaashish:
bors merge

sinhaashish:
bors merge

sinhaashish:
bors merge

PR #782, opened 2024-03-14
Tracing and Platform fixes

tiagolobocastro:
feat(platform): support custom namespace
    
    When we run outside of the cluster, we can't just use the env to determine
    the correct namespace..
    For now we expose an additional method which allows overriding it.
    A better solution is perhaps to simply ensure we create clients with the
    correct namespace as default?
    
---

    fix(agents/core): use tracing api
    
    Incorrect log api was used for error logging.
    Also include error source in the service error.
    

Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

PR #784, opened 2024-03-18
fix(volume/resize): handle overflow during volume shrink attempt

dsharma-dc:
This fixes the case where an attempt to shrink the volume with a capacity limit set on cluster - panicked with below error:

> stderr:
thread 'tokio-runtime-worker' panicked at control-plane/agents/src/bin/core/volume/service.rs:611:24:
attempt to subtract with overflow
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

### DIFF
--- a/control-plane/agents/src/bin/core/volume/service.rs
+++ b/control-plane/agents/src/bin/core/volume/service.rs
@@ -608,7 +608,12 @@ impl Service {
             return volume.resize(&self.registry, request).await;
         };
 
-        let required = request.requested_size - volume.as_ref().size;
+        // If requested size is less than volume's current size(attempt to shrink volume),
+        // then required becomes zero because we won't need to borrow anything from capacity_limit.
+        let required = request
+            .requested_size
+            .checked_sub(volume.as_ref().size)
+            .unwrap_or_default();
         *self.capacity_limit_borrow.write() += required;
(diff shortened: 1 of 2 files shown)


dsharma-dc:
bors try

hrudaya21 (approved):
lgtm

Abhinandan-Purkait (approved):

control-plane/agents/src/bin/core/volume/service.rs
```diff
@@ -608,7 +608,10 @@ impl Service {
             return volume.resize(&self.registry, request).await;
         };
 
-        let required = request.requested_size - volume.as_ref().size;
+        let required = request
```
chriswldenyer, line 613: This could do with a comment explaining how this handles the shrink case. I suppose required becomes zero?

chriswldenyer (approved):
LGTM, one suggestion

control-plane/agents/src/bin/core/volume/service.rs
dsharma-dc: > This could do with a comment explaining how this handles the shrink case. I suppose required becomes zero?

Sure, added a comment. Yes it becomes zero. We could have added special handling to capacity handling for this case but this looks good with minimal change.

dsharma-dc:
bors merge

PR #786, opened 2024-03-19
fix(serde): fix serialization error for get/list nodes

sinhaashish:
The below error was seen while getting the nodes in json format 
```
[nix-shell:~/code/rust/mayastor-extensions]$ ./target/debug/kubectl-mayastor get node node-1-155678 -o json
thread 'main' panicked at dependencies/control-plane/control-plane/plugin/./src/resources/utils.rs:185:49:
called `Result::unwrap()` on an `Err` value: Error("can only flatten structs and maps (got a sequence)", line: 0, column: 0)
```

This  PR fixes it.

### DIFF
--- a/control-plane/plugin/src/resources/node.rs
+++ b/control-plane/plugin/src/resources/node.rs
@@ -118,7 +118,14 @@ impl ListWithArgs for Nodes {
             Ok(nodes) => {
                 let node_display =
                     NodeDisplayLabels::new_nodes(nodes.clone().into_body(), args.show_labels());
-                print_table(output, node_display);
+                match output {
+                    OutputFormat::Yaml | OutputFormat::Json => {
+                        print_table(output, node_display.inner);
+                    }
+                    OutputFormat::None => {
+                        print_table(output, node_display);
+                    }
+                }
             }
             Err(e) => {
                 return Err(Error::ListNodesError { source: e });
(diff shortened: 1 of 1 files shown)


control-plane/plugin/src/resources/node.rs
```diff
@@ -118,7 +118,14 @@ impl ListWithArgs for Nodes {
             Ok(nodes) => {
                 let node_display =
                     NodeDisplayLabels::new_nodes(nodes.clone().into_body(), args.show_labels());
-                print_table(output, node_display);
+                match output {
+                    OutputFormat::Yaml | OutputFormat::Json => {
+                        print_table(output, node_display.inner);
+                    }
+                    OutputFormat::None => {
+                        print_table(output, node_display);
```
Abhinandan-Purkait, line 126: Doesn't the print table do the print based on formats? Also why are we passing inner separately?

Abhinandan-Purkait (approved):

control-plane/plugin/src/resources/node.rs
Abhinandan-Purkait: Could we not have passed the inner to the print_table only?

sinhaashish: In case of json and yaml the serde can't  deserialize the struct. 
The similar changes are done in our code base earlier 
Refer [here](https://github.com/openebs/mayastor-control-plane/blob/1ee5592f1df0ce1d5e36c360315fa4c327361527/control-plane/plugin/src/resources/node.rs#L484)

Abhinandan-Purkait: Alright.

abhilashshetty04 (approved):

sinhaashish:
bors merge

control-plane/plugin/src/resources/node.rs
tiagolobocastro: I think the problem is actually in the NodeDisplay struct which should not have flatten on the vector directly.

sinhaashish: I tried the same way as its done in  [`pub struct NodeDisplay {`](https://github.com/openebs/mayastor-control-plane/blob/919e1a1afbfcbc76ebe08035889d59d2ff12ae61/control-plane/plugin/src/resources/node.rs#L380)

The node display format uses `node_display_print` which does the same as this PR

PR #791, opened 2024-03-25
feat(eventing): add target volumeid to nvme path events

datacore-vvarakantham:
Added target(volumeid) to name path events after obtaining it from nqn.

### DIFF
--- a/control-plane/agents/src/bin/ha/node/server.rs
+++ b/control-plane/agents/src/bin/ha/node/server.rs
@@ -4,10 +4,7 @@ use crate::{
     path_provider::get_nvme_path_entry,
     Cli,
 };
-use agents::{
-    errors::{SvcError, SvcError::SubsystemNotFound},
-    eventing::EventWithMeta,
-};
+use agents::errors::{SvcError, SvcError::SubsystemNotFound};
 use events_api::event::{EventAction, EventCategory, EventMessage, EventMeta, EventSource};
 use grpc::{
     common::MapWrapper,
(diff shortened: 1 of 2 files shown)


w3aman (approved):

control-plane/agents/src/bin/ha/node/server.rs
```diff
@@ -432,10 +432,23 @@ impl ParsedUri {
 
 impl EventWithMeta for NodeAgentSvc {
     fn event(&self, event_action: EventAction, meta: EventMeta) -> EventMessage {
+        let volume_id = meta
```
tiagolobocastro: Seems rather complicated, could we change the event fn to receive also optional nqn directly?

tiagolobocastro (approved):
Would be worth cherry-picking to 2.6 as well

control-plane/agents/src/bin/ha/node/server.rs
datacore-vvarakantham: We are using the Event trait here. We might have to add another event trait to pass the extra argument nqn.

tiagolobocastro: A custom function here would probably suffice?

datacore-vvarakantham: Added another private trait for this event.

tiagolobocastro (approved):

datacore-vvarakantham:
bors merge

PR #796, opened 2024-03-28
fix(upgrade): don't use reservations without ha

tiagolobocastro:
If HA is disabled (can happen during v1-v2 upgrade) then don't use reservations.
This is because there are some imcompats between v1 and v2


dsavitskiy (approved):

sinhaashish (approved):

tiagolobocastro:
bors merge

PR #806, opened 2024-04-11
fix(agents/ha/node): complete reconnect when path is deleting

tiagolobocastro:
During the reconnect of a new path, the old path might take some time to disconnect.
Rather than wait for this (and timeout, which leads to retries) simply complete call as soon as path is deleting.

### DIFF
--- a/control-plane/agents/src/bin/ha/cluster/switchover.rs
+++ b/control-plane/agents/src/bin/ha/cluster/switchover.rs
@@ -35,6 +35,8 @@ fn client() -> impl VolumeOperations {
     core_grpc().volume()
 }
 
+const REPLACE_PATH_DEADLINE_RETRIES: u64 = 3;
+
 /// Stage represents the steps for switchover request.
 #[derive(Serialize, Deserialize, Debug, PartialEq, Eq, Copy, Clone, Ord, PartialOrd)]
 pub(crate) enum Stage {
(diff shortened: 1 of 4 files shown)


tiagolobocastro:
bors try

control-plane/agents/src/bin/ha/node/detector.rs
```diff
@@ -147,6 +155,22 @@ impl PathRecord {
...
+            return;
+        }
+        tracing::info!(
+            from=%self.state,
+            %to,
+            target=self.nqn,
+            path=self.path,
+            "Target path is deleting",
+        );
+        //self.event(EventAction::NvmePathDeleting).generate();
```
Abhinandan-Purkait: Is this needed for eventing?

tiagolobocastro: Need a change on dependencies to add path deleting event

abhilashshetty04 (approved):
lgtm

Abhinandan-Purkait (approved):

control-plane/agents/src/bin/ha/node/detector.rs
```diff
@@ -147,6 +155,22 @@ impl PathRecord {
...
     }
+
+    fn report_deleting(&mut self) {
+        let to = PathState::Removing;
+        if self.state == to {
+            return;
+        }
+        tracing::info!(
+            from=%self.state,
+            %to,
```
dsharma-dc, line 166: `to=%to` may be?

control-plane/agents/src/bin/ha/cluster/switchover.rs
```diff
@@ -406,50 +408,57 @@ impl SwitchOverRequest {
...
+                        | ReplyErrorKind::Timeout
+                        | ReplyErrorKind::NotFound
+                        | ReplyErrorKind::DeadlineExceeded
+                ) =>
+            {
+                if !matches!(error.kind, ReplyErrorKind::NotFound) {
+                    warn!(path=%self.existing_nqn, "HA Node agent failed to replace path within timeout");
+                    if self.retry_count < REPLACE_PATH_DEADLINE_RETRIES {
+                        return Err(anyhow!("Nvme path deadline exceeded"));
+                    }
```
dsharma-dc, line 448: If I read this right, It seems we'd end up retrying when we get NotFound, is that intended?

dsharma-dc (approved):

control-plane/agents/src/bin/ha/node/detector.rs
tiagolobocastro: hmm that's the same as %to

control-plane/agents/src/bin/ha/cluster/switchover.rs
tiagolobocastro: `NotFound` is hit when the new path is not present for some odd reason (maybe crash after previous republish?).
In this case there's no point retrying to connect, better to go back to republish?

dsharma-dc: ok, so this logic is indeed doing republish(and not retry) for this(NotFound) case. Sounds alright.

tiagolobocastro: Yes though I guess maybe not unreasonable to retry again in case of sudden io-engine restart. Maybe something to look into as a separate PR.

tiagolobocastro:
bors merge

PR #809, opened 2024-04-18
fix(csi-node/mounts): update deps to get safeMountIter fix

tiagolobocastro:
Dep update fixes deadlock on the new safe mount which basically looped forever trying to read mounts.
This meant the csi-node got stuck on stage calls and volumes would not get staged therefore pods would stay stuck in ContainerCreating forever.


niladrih (approved):

tiagolobocastro:
bors merge

PR #812, opened 2024-04-29
feat(topology): add affinity key to core

sinhaashish:


control-plane/agents/src/bin/core/controller/scheduling/volume.rs
```diff
@@ -118,6 +122,103 @@ impl Deref for GetSuitablePools {
     }
 }
 
+/// Create replicas of a volume.
+/// Selects the best pool candidates to create lvol replicas on.
```
tiagolobocastro:
```suggestion
/// Selects the best pool candidates to create volume replicas on.
```

control-plane/agents/src/bin/core/controller/scheduling/volume.rs
```diff
@@ -118,6 +122,103 @@ impl Deref for GetSuitablePools {
...
+pub(crate) struct CreateVolumeReplica {
```
tiagolobocastro: Why do we need a new type?  Moreover it's quite similar to `AddVolumeReplicas`

control-plane/grpc/proto/v1/pool/pool.proto
```diff
@@ -64,6 +64,8 @@ message PoolState {
   uint64 used = 6;
   // total size of pool replicas
   optional uint64 committed = 7;
+  // optional labels of pool
```
tiagolobocastro: hmm this doesn't make sense? Dataplane pool state has no notion of labels...

control-plane/agents/src/bin/core/volume/scheduling.rs
```diff
@@ -15,9 +15,10 @@ use crate::{
     node::wrapper::NodeWrapper,
 };
 use agents::errors::{NotEnough, SvcError};
+use std::collections::HashMap;
```
Abhinandan-Purkait, line 18: Please move this down after a blank line.

sinhaashish:
PR https://github.com/openebs/mayastor-control-plane/pull/846 raised for the same.

PR #817, opened 2024-05-06
Filter pools based on labels

sinhaashish:
This will be used for the command 
`kubectl mayastor get pools --labels A=B,C=D`

### DIFF
--- a/control-plane/agents/src/bin/core/pool/registry.rs
+++ b/control-plane/agents/src/bin/core/pool/registry.rs
@@ -3,10 +3,58 @@ use crate::controller::{
     wrapper::{GetterOps, *},
 };
 use agents::errors::{SvcError, SvcError::PoolNotFound};
-use stor_port::types::v0::transport::{CtrlPoolState, NodeId, Pool, PoolId, Replica, ReplicaId};
+use std::collections::HashMap;
+use stor_port::types::v0::transport::{
+    CtrlPoolState, Labels, NodeId, Pool, PoolId, Replica, ReplicaId,
+};
 
 /// Pool helpers
 impl Registry {
+    pub(crate) async fn get_label_opts_pools(
+        &self,
+        labels: Option<Labels>,
+    ) -> Result<Vec<Pool>, SvcError> {
+        let pools = self.get_node_opt_pools(None).await?;
+        let pools_after_label_filtering = match labels {
+            Some(label) => {
+                let mut filtered_pool: Vec<Pool> = vec![];
+                for pool in pools.iter() {
+                    match pool.spec() {
+                        Some(spec) => match spec.labels {
+                            Some(pool_labels) => {
+                                let pool_label_match =
+                                    Self::labels_matched(pool_labels, label.clone())?;
+                                if pool_label_match {
+                                    filtered_pool.push(pool.clone());
+                                }
+                            }
+                            None => continue,
+                        },
+                        None => continue,
+                    }
+                }
+                return Ok(filtered_pool);
+            }
+            None => pools,
+        };
+        Ok(pools_after_label_filtering)
+    }
+
+    pub(crate) fn labels_matched(
+        pool_labels: HashMap<String, String>,
+        labels: Labels,
+    ) -> Result<bool, SvcError> {
+        for label in labels.split(',') {
+            let [key, value] = label.split('=').collect::<Vec<_>>()[..] else {
+                return Err(SvcError::LabelNodeFilter {});
+            };
+            if pool_labels.get(key) != Some(&value.to_string()) {
+                return Ok(false);
+            }
+        }
+        Ok(true)
+    }
+
     /// Get all pools from node `node_id` or from all nodes.
     pub(crate) async fn get_node_opt_pools(
         &self,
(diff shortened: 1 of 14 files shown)


control-plane/grpc/proto/v1/pool/pool.proto
```diff
@@ -84,6 +84,7 @@ message GetPoolsRequest {
     common.NodeFilter node = 1;
     common.PoolFilter pool = 2;
     common.NodePoolFilter node_pool = 3;
+    common.LabelFilter labels = 4;
```
tiagolobocastro: Should this also have NodePoolLabelFilter?

control-plane/agents/src/bin/core/pool/registry.rs
```diff
@@ -3,10 +3,75 @@ use crate::controller::{
...
+
+        let pools_from_spec = self
+            .specs()
+            .pools()
+            .into_iter()
+            .filter(|p| !pools.iter().any(|i| i.id() == &p.id))
+            .map(Pool::from_spec)
+            .collect::<Vec<_>>();
+
+        pools.extend(pools_from_spec);
```
tiagolobocastro: We can simply call:
```suggestion
        let pools = self.get_node_opt_pools(None).await?;
```

control-plane/agents/src/bin/core/pool/registry.rs
```diff
@@ -3,10 +3,75 @@ use crate::controller::{
...
+            None => pools,
+        };
+        Ok(pools_after)
+    }
+
+    pub(crate) fn labels_matched(
+        pool_labels: HashMap<String, String>,
+        labels: Labels,
+    ) -> Result<bool, SvcError> {
+        for label in labels.split(',') {
```
tiagolobocastro, line 47: I'm confused, doesn't the hashmap already have key and value? Why do we need further parsing?

control-plane/rest/openapi-specs/v0_api_spec.yaml
```diff
@@ -1146,6 +1146,32 @@ paths:
           $ref: '#/components/responses/ServerError'
       security:
         - JWT: []
+  '/label/pools':
```
tiagolobocastro: hmm this doesn't fit here, we're not getting pools from labels.
I think this should be another parameter on the /pools itself

control-plane/rest/openapi-specs/v0_api_spec.yaml
```diff
@@ -1146,6 +1146,32 @@ paths:
...
+    get:
+      tags:
+        - Pools
+      operationId: get_pools_topology
+      parameters:
+        - in: query
+          name: topology_label
+          required: true
+          schema:
+            $ref: '#/components/schemas/TopologyLabel'
```
tiagolobocastro, line 1139: Should be a map?

control-plane/stor-port/src/types/v0/transport/pool.rs
```diff
@@ -117,6 +117,7 @@ impl From<CtrlPoolState> for models::PoolState {
 
 rpc_impl_string_id!(PoolId, "ID of a pool");
 rpc_impl_string_uuid!(PoolUuid, "UUID of a pool");
+rpc_impl_string_id!(Labels, "Labels of a pool");
```
tiagolobocastro, line 120: Why are we adding labels as an identifier?

control-plane/rest/openapi-specs/v0_api_spec.yaml
sinhaashish: done

control-plane/rest/openapi-specs/v0_api_spec.yaml
sinhaashish: topology_label is a query param, and its passed as string  `A=B,C=D,E=F` . Cant pass map as a query param .

control-plane/agents/src/bin/core/pool/registry.rs
sinhaashish: no, the query params comes as a string lie `A=B,C=D` and here it is parsed. The labels of the pool are a hashmap

control-plane/stor-port/src/types/v0/transport/pool.rs
sinhaashish: It is used as a filter criteria [here ](https://github.com/sinhaashish/mayastor-control-plane/blob/d4608a04204c404ddd8733ee0c5eec667a3896c6/control-plane/stor-port/src/types/v0/transport/misc.rs#L51)

control-plane/rest/openapi-specs/v0_api_spec.yaml
tiagolobocastro: Alright, then let's label the type `TopologyLabels` and add comments to say how it's meant to be used

control-plane/agents/src/bin/core/pool/registry.rs
tiagolobocastro: Looks like that's an openapi limitation due to using query rather than body. 
Let's change it from string to hashmap on the grpc proto.

sinhaashish:
closing this as PR https://github.com/openebs/mayastor-control-plane/pull/823 contains the same functionality and is merged

PR #820, opened 2024-05-08
feat(snap-rebuild): add io-engine controller api

tiagolobocastro:
Adds simple unit test to ensure api is working.
More will follow when the api is either exported via gRPC or through component/integration level testing.


control-plane/agents/src/bin/core/controller/io_engine/mod.rs
```diff
@@ -220,3 +222,20 @@ pub(crate) trait HostApi {
     /// List blockdevices based on api versions.
     async fn list_blockdevices(&self, request: &GetBlockDevices) -> Result<BlockDevices, SvcError>;
 }
+
+/// Interface for snapshot rebuilding operations.
+#[async_trait]
+pub(crate) trait SnapshotRebuildApi {
```
dsharma-dc, line 228: Would it be simpler to make these a part of ReplicaSnapshotApi? We're going to deal with a replica's snapshots at a time during snap rebuild so looks logically near.

Abhinandan-Purkait (approved):

control-plane/agents/src/bin/core/controller/io_engine/mod.rs
tiagolobocastro: In previous dataplane api PR we had decided to keep snapshot rebuild itself separate as it has not much in common with snapshot api.

control-plane/agents/src/bin/core/controller/io_engine/types.rs
```diff
@@ -32,3 +32,51 @@ pub(crate) struct RebuildHistoryResp {
...
+/// List all or a specific snapshot rebuild.
+pub(crate) struct ListSnapRebuild {
+    /// Select a specific target replica.
+    pub(crate) uuid: Option<transport::ReplicaId>,
+}
+/// A list of snapshot rebuilds.
+pub(crate) type ListSnapRebuildRsp = Vec<SnapshotRebuild>;
+/// Destroy a snapshot rebuild.
+pub(crate) struct DestroySnapRebuild {
+    /// The uuid of the particular rebuild to destroy.
```
dsharma-dc:
```suggestion
    /// The uuid of the particular replica for which to destroy the rebuild.
```

dsharma-dc (approved):

abhilashshetty04 (approved):

abhilashshetty04:
@tiagolobocastro , should we call it TerminateSnapRebuild instead of DestroySnapRebuild ?

control-plane/agents/src/bin/core/controller/io_engine/types.rs
tiagolobocastro: Ah though this is also the uuid of  the rebuild itself, which inherits its uuid from the replica. I'll also add this info in the doc comment

tiagolobocastro:
> @tiagolobocastro , should we call it TerminateSnapRebuild instead of DestroySnapRebuild ?

On the dataplane api it's destroy, so let's keep the same terminology

tiagolobocastro:
bors merge

PR #821, opened 2024-05-10
Make image prefix configurable

niladrih:


niladrih:
bors try

avishnu (approved):
LGTM

default.nix
```diff
@@ -1,9 +1,9 @@
-{ system ? null, allInOne ? true, incremental ? false, img_tag ? "", tag ? "", img_org ? "" }:
+{ system ? null, allInOne ? true, incremental ? false, img_tag ? "", tag ? "", img_org ? "", product_prefix ? "" }:
```
tiagolobocastro, line 1: We already have img_tag and img_org, so let's use img_prefix here too

Jenkinsfile
```diff
@@ -85,6 +85,7 @@ pipeline {
   }
   environment {
     CARGO_INCREMENTAL = "0"
+    PRODUCT_PREFIX = 'mayastor'
```
tiagolobocastro: Maybe then here should also be IMAGE_PREFIX ?

niladrih: Eventually this value would we used with the kubectl-mayastor binary to decide on the suffix name.

niladrih: Does IMAGE_PREFIX still fit?

tiagolobocastro: hmm good point, maybe leave this one as is then

default.nix
niladrih: Yup, makes sense. Changed the name of the variable.

niladrih:
bors try

tiagolobocastro (approved):

niladrih:
bors merge

PR #823, opened 2024-05-13
feat(topology): add pool filter on node_id and labels

sinhaashish:
This will be used for the command
1. `kubectl mayastor get pools --labels A=B,C=D`
2. `kubectl mayastor get pools --labels A=B,C=D --node <node_id>`
3. `kubectl mayastor get pools --node <node_id>`

### DIFF
--- a/control-plane/plugin/src/resources/pool.rs
+++ b/control-plane/plugin/src/resources/pool.rs
@@ -1,15 +1,16 @@
 use crate::{
-    operations::{Get, List, PluginResult},
+    operations::{Get, ListWithArgs, PluginResult},
     resources::{
         error::Error,
         utils,
         utils::{CreateRow, GetHeaderRow},
-        PoolId,
+        NodeId, PoolId,
     },
     rest_wrapper::RestClient,
 };
 use async_trait::async_trait;
 use prettytable::Row;
+use std::collections::HashMap;
 
 /// Pools resource.
 #[derive(clap::Args, Debug)]
@@ -60,18 +61,62 @@ impl GetHeaderRow for openapi::models::Pool {
     }
 }
 
+/// Arguments used when getting pools.
+#[derive(Debug, Clone, clap::Args)]
+pub struct GetPoolsArgs {
+    /// Gets Pools from this node only.
+    #[clap(long)]
+    node_id: Option<NodeId>,
+
+    /// Selector (label query) to filter on, supports '=' only.
+    /// (e.g. -l key1=value1,key2=value2).
+    /// Pools must satisfy all of the specified label constraints.
+    #[clap(short = 'l', long)]
+    selector: Option<String>,
+}
+
+impl GetPoolsArgs {
+    /// Return the node ID.
+    pub fn node_id(&self) -> &Option<NodeId> {
+        &self.node_id
+    }
+    /// Select the pools based on labels.
+    pub fn selector(&self) -> &Option<String> {
+        &self.selector
+    }
+}
+
 #[async_trait(?Send)]
-impl List for Pools {
-    async fn list(output: &utils::OutputFormat) -> PluginResult {
-        match RestClient::client().pools_api().get_pools().await {
-            Ok(pools) => {
-                // Print table, json or yaml based on output format.
-                utils::print_table(output, pools.into_body());
-            }
-            Err(e) => {
-                return Err(Error::ListPoolsError { source: e });
-            }
-        }
+impl ListWithArgs for Pools {
+    type Args = GetPoolsArgs;
+    async fn list(args: &Self::Args, output: &utils::OutputFormat) -> PluginResult {
+        let mut pools = match args.node_id() {
+            Some(node_id) => RestClient::client()
+                .pools_api()
+                .get_node_pools(node_id)
+                .await
+                .map(|pools| pools.into_body())
+                .map_err(|e| Error::ListPoolsError { source: e }),
+            None => RestClient::client()
+                .pools_api()
+                .get_pools()
+                .await
+                .map(|pools| pools.into_body())
+                .map_err(|e| Error::ListPoolsError { source: e }),
+        }?;
+
+        pools.retain(|pool| match &pool.spec {
+            Some(spec) => match &spec.labels {
+                Some(pool_labels) => {
+                    let pool_label_match =
+                        labels_matched(pool_labels, args.selector()).unwrap_or(false);
+                    pool_label_match
+                }
+                None => true,
+            },
+            None => true,
+        });
+        utils::print_table(output, pools);
         Ok(())
     }
 }
@@ -98,4 +143,27 @@ impl Get for Pool {
         }
         Ok(())
     }
+}
+
+/// Check if the labels match the pool labels.
+pub(crate) fn labels_matched(
+    pool_labels: &HashMap<String, String>,
+    labels: &Option<String>,
+) -> Result<bool, Error> {
+    match labels {
+        Some(filter_labels) => {
+            for label in filter_labels.split(',') {
+                let [key, value] = label.split('=').collect::<Vec<_>>()[..] else {
+                    return Err(Error::LabelNodeFilter {
+                        labels: filter_labels.to_string(),
+                    });
+                };
+                if pool_labels.get(key) != Some(&value.to_string()) {
+                    return Ok(false);
+                }
+            }
+        }
+        None => return Ok(true),
+    }
+    Ok(true)
 }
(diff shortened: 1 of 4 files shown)


control-plane/plugin/src/resources/pool.rs
```diff
@@ -60,18 +61,77 @@ impl GetHeaderRow for openapi::models::Pool {
...
+#[derive(Debug, Clone, clap::Args)]
+/// Arguments used when getting nodes.
+pub struct GetPoolsArgs {
+    /// Id of the node
+    #[clap(long)]
+    node_id: Option<NodeId>,
+
+    /// Filter pools on labels
+    #[clap(long)]
+    labels: Option<String>,
```
tiagolobocastro: To match the same term kubectl uses?

```suggestion
    selector: Option<String>,
```

control-plane/plugin/src/resources/pool.rs
```diff
@@ -60,18 +61,77 @@ impl GetHeaderRow for openapi::models::Pool {
     }
 }
 
+#[derive(Debug, Clone, clap::Args)]
+/// Arguments used when getting nodes.
+pub struct GetPoolsArgs {
+    /// Id of the node
```
tiagolobocastro:
```suggestion
    /// Gets Pools from this node only.
```

control-plane/plugin/src/resources/pool.rs
```diff
@@ -60,18 +61,77 @@ impl GetHeaderRow for openapi::models::Pool {
...
+    #[clap(long)]
+    node_id: Option<NodeId>,
+
+    /// Filter pools on labels
```
tiagolobocastro: Would be nice if we could add support for all these, otherwise update accordingly.
```suggestion
    /// Selector (label query) to filter on, supports '=', '==', and '!='.(e.g. -l key1=value1,key2=value2).   
    /// Pools must satisfy all of the specified label constraints.
```

control-plane/plugin/src/resources/pool.rs
```diff
@@ -60,18 +61,77 @@ impl GetHeaderRow for openapi::models::Pool {
...
+        let pools = match args.node_id() {
+            Some(node_id) => {
+                match RestClient::client()
+                    .pools_api()
+                    .get_node_pools(&node_id)
+                    .await
+                {
+                    Ok(pools) => pools.into_body(),
+                    Err(e) => {
+                        return Err(Error::ListPoolsError { source: e });
```
tiagolobocastro: Add Ok above and use ? on 110.
```suggestion
                        Err(Error::ListPoolsError { source: e });
```

control-plane/plugin/src/resources/pool.rs
```diff
@@ -60,18 +61,77 @@ impl GetHeaderRow for openapi::models::Pool {
...
+        let pools_after_label_filtering = match args.filter_labels() {
+            Some(filter_labels) => {
+                for pool in pools.iter() {
+                    match &pool.spec {
+                        Some(spec) => match &spec.labels {
+                            Some(pool_labels) => {
+                                let pool_label_match =
+                                    labels_matched(pool_labels, filter_labels.clone())?;
+                                if pool_label_match {
+                                    filtered_pool.push(pool.clone());
```
tiagolobocastro: why not just filter pools directly? Don't need to create a new vector

control-plane/plugin/src/resources/pool.rs
```diff
@@ -60,18 +61,77 @@ impl GetHeaderRow for openapi::models::Pool {
...
+
+impl GetPoolsArgs {
+    /// Return the node ID.
+    pub fn node_id(&self) -> Option<NodeId> {
+        self.node_id.clone()
+    }
+    /// Filter on the labels provided
+    pub fn filter_labels(&self) -> Option<String> {
+        self.labels.clone()
+    }
```
tiagolobocastro:
```suggestion
    pub fn node_id(&self) -> &Option<NodeId> {
        &self.node_id.clone()
    }
    /// Filter on the labels provided
    pub fn filter_labels(&self) -> &Option<String> {
        &self.labels
    }
```

control-plane/plugin/src/resources/pool.rs
sinhaashish: done

control-plane/plugin/src/resources/pool.rs
```diff
@@ -60,18 +61,62 @@ impl GetHeaderRow for openapi::models::Pool {
...
+                .map_err(|e| Error::ListPoolsError { source: e }),
+            None => RestClient::client()
+                .pools_api()
+                .get_pools()
+                .await
+                .map(|pools| pools.into_body())
+                .map_err(|e| Error::ListPoolsError { source: e }),
+        }?;
+
+        pools.retain(|pool| match &pool.spec {
```
tiagolobocastro, line 108: much better :+1:

control-plane/plugin/src/resources/pool.rs
```diff
@@ -60,18 +61,62 @@ impl GetHeaderRow for openapi::models::Pool {
     }
 }
 
+#[derive(Debug, Clone, clap::Args)]
+/// Arguments used when getting nodes.
```
tiagolobocastro:
```suggestion
/// Arguments used when getting nodes.
#[derive(Debug, Clone, clap::Args)]
```

control-plane/plugin/src/resources/pool.rs
```diff
@@ -99,3 +144,26 @@ impl Get for Pool {
...
+
+/// Check if the labels match the pool labels.
+pub(crate) fn labels_matched(
+    pool_labels: &HashMap<String, String>,
+    labels: &Option<String>,
+) -> Result<bool, Error> {
+    match labels {
+        Some(filter_labels) => {
+            for label in filter_labels.split(',') {
+                let [key, value] = label.split('=').collect::<Vec<_>>()[..] else {
```
tiagolobocastro, line 156: Seems we're only supporting '='
Can we update this to include '==' and '!=' ?

sinhaashish: This parsing will require regex. IMHO we can keep it simple in this PR and later take it as an enhancement if needed.

tiagolobocastro: Sounds good, but please then update doc comment to reflect this (ie remove the '==' and '!=' from the doc comment)

tiagolobocastro (approved):

control-plane/plugin/src/resources/pool.rs
```diff
@@ -60,18 +61,62 @@ impl GetHeaderRow for openapi::models::Pool {
     }
 }
 
+/// Arguments used when getting nodes.
```
Abhinandan-Purkait: Should be `Arguments used when getting pools`?

control-plane/plugin/src/resources/pool.rs
```diff
@@ -60,18 +61,62 @@ impl GetHeaderRow for openapi::models::Pool {
...
+#[derive(Debug, Clone, clap::Args)]
+pub struct GetPoolsArgs {
+    /// Gets Pools from this node only.
+    #[clap(long)]
+    node_id: Option<NodeId>,
+
+    /// Selector (label query) to filter on, supports '=', '==',
+    /// and '!='.(e.g. -l key1=value1,key2=value2).
+    /// Pools must satisfy all of the specified label constraints.
```
Abhinandan-Purkait: Please update this with whatever is supported.

control-plane/plugin/src/resources/pool.rs
Abhinandan-Purkait: Why do we need the `==`? What would that query look like? 
In k8s its equality and set based. I think if we want an enhancement we should go for the k8s way.
ref: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#list-and-watch-filtering

Abhinandan-Purkait (approved):

sinhaashish:
bors merge

control-plane/plugin/src/resources/pool.rs
sinhaashish: kubectl get pods --help will show you the below option 
```
  -l, --selector='':
	Selector (label query) to filter on, supports '=', '==', and '!='.(e.g. -l key1=value1,key2=value2). Matching
	objects must satisfy all of the specified label constraints.
``` 

Refer https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#equality-based-requirement

PR #825, opened 2024-05-15
feat(topology): add validation for node topology

sinhaashish:
Error out if  `nodeSpreadTopologyKey` has same values as of `nodeHasTopologyKey` or `nodeAffinityTopologyLabel`

### DIFF
--- a/control-plane/csi-driver/src/context.rs
+++ b/control-plane/csi-driver/src/context.rs
@@ -392,6 +392,10 @@ impl TryFrom<&HashMap<String, String>> for PublishParams {
             args.get(Parameters::NodeSpreadTopologyKey.as_ref()),
         )
         .map_err(|_| tonic::Status::invalid_argument("Invalid node_spread_topology_key"))?;
+
+        validate_topology_params(&node_affinity_topology_label, &node_spread_topology_key)?;
+        validate_topology_params(&node_has_topology_key, &node_spread_topology_key)?;
+
         Ok(Self {
             io_timeout,
             nvme_io_timeout,
@@ -408,6 +412,25 @@ impl TryFrom<&HashMap<String, String>> for PublishParams {
             node_spread_topology_key,
         })
     }
+}
+
+/// Validate the topology parameters.
+/// Errors out of nodeSpreadTopologyKey has same values as of
+/// node_has_topology_key/nodeAffinityTopologyLabel
+pub(crate) fn validate_topology_params(
+    map1: &Option<HashMap<String, String>>,
+    map2: &Option<HashMap<String, String>>,
+) -> Result<(), tonic::Status> {
+    if let (Some(map1), Some(map2)) = (map1, map2) {
+        let has_common_keys = map1.len() == map2.len() && map1.keys().any(|k| map2.contains_key(k));
+
+        if has_common_keys {
+            return Err(tonic::Status::invalid_argument(
+                "Invalid node topology. `nodeSpreadTopologyKey` can't have same values as `nodeHasTopologyKey` or `nodeAffinityTopologyLabel`",
+            ));
+        }
+    }
+    Ok(())
 }
 
 /// Volume Creation parameters.


control-plane/csi-driver/src/context.rs
```diff
@@ -392,6 +392,16 @@ impl TryFrom<&HashMap<String, String>> for PublishParams {
             args.get(Parameters::NodeSpreadTopologyKey.as_ref()),
         )
         .map_err(|_| tonic::Status::invalid_argument("Invalid node_spread_topology_key"))?;
+
+        validate_topology_params(
+            node_affinity_topology_label.clone(),
```
tiagolobocastro: pass by reference rather than clone

control-plane/csi-driver/src/context.rs
```diff
@@ -410,6 +420,25 @@ impl TryFrom<&HashMap<String, String>> for PublishParams {
...
 
+/// Validate the topology parameters.
+/// Errors out of nodeSpreadTopologyKey has same values as of
+/// node_has_topology_key/nodeAffinityTopologyLabel
+pub(crate) fn validate_topology_params(
+    map1: Option<HashMap<String, String>>,
+    map2: Option<HashMap<String, String>>,
+) -> Result<(), tonic::Status> {
+    if let (Some(map1), Some(map2)) = (map1, map2) {
+        let has_common_keys = map1.len() == map2.len() && map1.keys().all(|k| map2.contains_key(k));
```
tiagolobocastro: Shouldn't it be any?

```suggestion
        let has_common_keys = map1.len() == map2.len() && map1.keys().any(|k| map2.contains_key(k));
```

control-plane/csi-driver/src/context.rs
```diff
@@ -410,6 +420,25 @@ impl TryFrom<&HashMap<String, String>> for PublishParams {
...
+
+        if has_common_keys {
+            return Err(tonic::Status::invalid_argument(
+                "Invalid node topology. `nodeSpreadTopologyKey` can't have same values as `nodeHasTopologyKey` or `nodeAffinityTopologyLabel`",
```
tiagolobocastro, line 429: What about nodeHasTopologyKey and nodeAffinityTopologyLabel. Do they need validation between them as well?

sinhaashish:
Ran this test . Please validate the step
The storage class has 
```
kind: StorageClass
apiVersion: storage.k8s.io/v1
metadata:
  name: mayastor-5
parameters:
  ioTimeout: "30"
  protocol: nvmf
  repl: "3"
  nodeTopologyAffinityLabel: |
    zone-us: us-west-1
  nodeSpreadTopologyKey: |
    zone-us
  nodeHasTopologyKey: |
    zone-us
provisioner: io.openebs.csi-mayastor
volumeBindingMode: Immediate
```

Then created a PVC using the above sc 
```
[nix-shell:~/code/rust/mayastor-control-plane]$ cat <<EOF | kubectl create -f -
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: ms-volume-claim
spec:
  accessModes:
  - ReadWriteOnce
  resources:
    requests:
      storage: 1Gi
  storageClassName: mayastor-5
EOF
persistentvolumeclaim/ms-volume-claim created
```

The pvc is in pending state 
upon checking the csi-controller  logs  it shows 
```
failed to provision volume with StorageClass "mayastor-5": rpc error: code = InvalidArgument desc = Invalid node topology. `nodeSpreadTopologyKey` can't have same values as `nodeHasTopologyKey` or `nodeAffinityTopologyLabel`
```

control-plane/csi-driver/src/context.rs
sinhaashish: IMO, we dont need that , `nodeAffinityTopologyLabel` is actually a subset of `nodeHasTopologyKey`
So of both are given the with the same key then `nodeAffinityTopologyLabel` will take the presedence as it has key and value both

Abhinandan-Purkait (approved):

tiagolobocastro (approved):

sinhaashish:
bors merge

PR #831, opened 2024-05-20
CI stability fixes

tiagolobocastro:
test(python/bdd): incorrect usage of retry decorator and full rebuild
    
    Retry decorator cannot be used on the fixture directly.
    Also check if a full rebuild happened using rebuild history in case child is
    reused for new nexus.
    
---

    build(openapi): don't depend on version.txt
    
    Not quite sure why the version.txt was added since we know that we should not
    be adding output files as an input to the build.rs of a crate.
    This is causing extra rebuilds during CI!
    I suspect it may have been added to handle git clean of the openapi dir!
    I think the correct way to handle that is to simply move the output of the
    build.rs to the target folder where it should have lived :-)
    Let's remove this and then pay attention to missing rebuilds..
    
---

    fix(deployer): increase timeout for io-engine connect
    
    On GCP testing sometimes we fail to connect to the io-engine within a second.
    Doubling this for now should be sufficient.
    
---

    test(python/bdd): enable core isolation for log rebuild
    
    Speeds up the io-engine bootstrap.
    Also to consider, should we just enable isolation by default on the tests?
    
---

    test(python/bdd): check error code of deployer start
    
    In case the deployer fails to start, let's error out early as it seems
    sometimes the tests still pass but when the tests fails it can lead into
    debugging the wrong path...
    
### DIFF
--- a/deployer/src/infra/io_engine.rs
+++ b/deployer/src/infra/io_engine.rs
@@ -107,6 +107,7 @@ impl ComponentAction for IoEngine {
                 options.latest_io_api_version(),
                 &name,
                 socket,
+                40,
                 tokio::time::sleep,
             )
             .await?;
--- a/openapi/build.rs
+++ b/openapi/build.rs
@@ -15,9 +15,9 @@ fn main() {
 
     println!("cargo:rerun-if-changed=../nix/pkgs/openapi-generator");
     println!("cargo:rerun-if-changed=../control-plane/rest/openapi-specs");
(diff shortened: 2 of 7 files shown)


Abhinandan-Purkait (approved):

chriswldenyer (approved):
Looks reasonable.

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

PR #841, opened 2024-06-11
Disallow NR rsnapshots and rebuilds without replica ancestry rebuild fix

tiagolobocastro:
feat: expose io-engine version via rest

---

    refactor: verify rebuild validity using replica nodes only

---

    fix: don't add replica to nexus which can't rebuild properly
    
    Check if all nodes have the rebuild fix for the following operations:
    - increase replica count on volume
    - move replica
    - add replica to nexus
    - take snapshot on nr volume
    
    Checking all nodes might lead to some false positives and also some issues if some nodes
    have been removed from power but not from the pstor, but it's a very
    simple way of ensuring all nodes are fixed.
    A potential improvement may be to check if the replica nodes have the fix.
    
---

    feat(deployer): add idle io-engines
    
    These are useful to sort of simulate upgrades.
    This is required because there's no easy way of updating a container image tag.
    A better alternative to add in the future would be to extend composer to add
    containers to existing cluster and extending cluster to allow for this via its
    components as well.
    
---

    feat: update deps and handle features and bugfixes info
    
    This is now parsed from the io-engine registration and can be used to determine
    if a certain function can be performed or not.
    Also persists these to the pstor.
    
### DIFF
--- a/control-plane/agents/src/bin/core/controller/io_engine/v0/host.rs
+++ b/control-plane/agents/src/bin/core/controller/io_engine/v0/host.rs
@@ -27,6 +27,9 @@ impl crate::controller::io_engine::HostApi for super::RpcClient {
             api_versions: Some(vec![ApiVersion::V0]),
             instance_uuid: None,
             node_nqn: None,
+            features: None,
+            bugfixes: None,
+            version: None,
         })
     }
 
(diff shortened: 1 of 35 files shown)


Abhinandan-Purkait (approved):
LGTM

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

PR #846, opened 2024-06-25
feat(topology): add pool affinity key

sinhaashish:


control-plane/agents/src/bin/core/controller/resources/mod.rs
```diff
@@ -147,7 +147,7 @@ impl<T: OperationSequencer + Sized, R> AsRef<R> for OperationGuard<T, R> {
     }
 }
 /// It unlocks the sequence lock on drop.
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
tiagolobocastro: The guard should not be clonable, this defeats the purpose of the guard itself.

control-plane/agents/src/bin/core/volume/clone_operations.rs
```diff
@@ -130,7 +134,8 @@ impl CreateVolumeExe for SnapshotCloneOp<'_> {
                 id: volume.uuid_str(),
             });
         }
-        Ok(clonable_snapshots)
+        vec_clonable_snapshot.push(clonable_snapshots);
```
tiagolobocastro: I'm confused as to what this is doing, perhaps what you want is to change the way the associated type is defined.

control-plane/grpc/src/operations/replica/traits.rs
```diff
@@ -943,6 +943,7 @@ impl From<&dyn CreateReplicaInfo> for CreateReplica {
             owners: data.owners(),
             allowed_hosts: data.allowed_hosts(),
             kind: None,
+            labels: None,
```
tiagolobocastro: Why are there labels on the replica?

tests/bdd/features/volume/topology/test_pool_topology.py
```diff
@@ -180,12 +201,80 @@
...
+    },
+
 ]
 
 
 @pytest.fixture(scope="module")
 def init():
     Deployer.start(NUM_IO_ENGINES, io_engine_coreisol=True)
+
+    time.sleep(10)
```
tiagolobocastro: Why is there a sleep here?

control-plane/agents/src/bin/core/controller/scheduling/volume_policy/mod.rs
```diff
@@ -68,26 +68,43 @@ impl DefaultBasePolicy {
...
     for (vol_inc_key, vol_inc_value) in vol_pool_inc_labels.iter() {
         match pool_labels.get(vol_inc_key) {
             Some(pool_val) => {
                 if vol_inc_value.is_empty() {
                     continue;
                 }
                 if pool_val != vol_inc_value {
-                    return false;
+                    satisfy_inclusion = false;
+                    break;
```
tiagolobocastro, line 88: why not just return?

control-plane/agents/src/bin/core/controller/scheduling/volume_policy/mod.rs
```diff
@@ -68,26 +68,43 @@ impl DefaultBasePolicy {
     }
 }
 
-/// Return true if all the keys present in volume's pool/node inclusion matches with the pool/node
-/// labels otherwise returns false.
+/// Return true if all the keys present in the volume's pool/node
+/// inclusion/affinity match the pool/node labels; otherwise, return false.
 pub(crate) fn qualifies_label_criteria(
     vol_pool_inc_labels: HashMap<String, String>,
+    vol_pool_affinity: HashMap<String, String>,
```
tiagolobocastro, line 75: can we take by reference instead?

control-plane/agents/src/bin/core/pool/wrapper.rs
```diff
@@ -116,6 +122,12 @@ impl PoolWrapper {
         self.state.status = PoolStatus::Unknown;
     }
 
+    /// Set labels in PoolWrapper.
+    #[allow(dead_code)]
+    pub(crate) fn set_labels(&mut self, labels: Option<String>) {
+        self.labels = labels;
+    }
+
```
tiagolobocastro: If unused, delete it.

```suggestion
```

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -762,97 +763,6 @@ impl ResourceShutdownOperations for OperationGuardArc<VolumeSpec> {
     }
 }
 
-#[async_trait::async_trait]
```
tiagolobocastro, line 765: Why is this removed?

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -924,28 +835,42 @@ pub(super) trait CreateVolumeExeVal: Sync + Send {
...
-        // we can't fulfil the required replication factor, so let the caller
-        // decide what to do next
-        if replicas.len() < context.volume.as_ref().num_replicas as usize {
-            self.undo(&mut context, replicas).await;
-            Err(SvcError::ReplicaCreateNumber {
-                id: context.volume.uid_str(),
-            })
-        } else {
-            Ok(replicas)
+        let mut count = 0;
```
tiagolobocastro: This is too much implied specific, it should be handled in the create itself.

control-plane/agents/src/bin/core/volume/operations.rs
sinhaashish: because it was not being used

control-plane/agents/src/bin/core/volume/operations.rs
sinhaashish: done

tests/bdd/features/volume/topology/test_pool_topology.py
sinhaashish: removed

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -888,6 +797,7 @@ impl ResourceLifecycleExt<CreateVolumeSource<'_>> for OperationGuardArc<VolumeSp
         volume
             .complete_create(result, registry, OnCreateFail::Delete)
             .await?;
+
```
```suggestion
```

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -945,6 +855,7 @@ pub(super) trait CreateVolumeExe: CreateVolumeExeVal {
             Ok(replicas)
         }
     }
+
```
```suggestion
```

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -966,12 +877,9 @@ impl CreateVolumeExeVal for CreateVolume {
 
 #[async_trait::async_trait]
 impl CreateVolumeExe for CreateVolume {
-    type Candidates = CreateReplicaCandidate;
+    type Candidates = Vec<CreateReplicaCandidate>;
```
tiagolobocastro, line 896: Yes that's it :+1:

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -980,49 +888,60 @@ impl CreateVolumeExe for CreateVolume {
...
     ) -> Vec<Replica> {
-        let mut replicas = Vec::<Replica>::with_capacity(candidates.candidates().len());
-        for replica in candidates.candidates() {
-            if replicas.len() >= self.replicas as usize {
-                break;
-            } else if replicas.iter().any(|r| {
-                r.node == replica.node
-                    || spread_label_is_same(r, replica, context).unwrap_or_else(|error| {
+        let mut fina_replicas = Vec::<Replica>::new();
+        for crc in candidates.iter() {
```
tiagolobocastro: Can consume here, rather than reference

```suggestion
        for crc in candidates {
```

tiagolobocastro: Also what does crc stand for?

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -980,49 +888,60 @@ impl CreateVolumeExe for CreateVolume {
...
+            let mut replicas = Vec::<Replica>::with_capacity(crc.candidates().len());
```
tiagolobocastro: We don't seem to be considering replicas which are already part of the volume, example from previous creation attempts?
How does this work and do they affect the creation of the candidates?

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -980,49 +888,60 @@ impl CreateVolumeExe for CreateVolume {
...
+                };
             }
-            let replica = if replicas.is_empty() {
-                let mut replica = replica.clone();
-                // the local replica needs to be connected via "bdev:///"
-                replica.share = Protocol::None;
-                replica
+
+            if replicas.len() < context.volume.as_ref().num_replicas as usize {
+                self.undo(context, replicas).await;
```
tiagolobocastro: Why don't we handle the errors here?

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -980,49 +888,60 @@ impl CreateVolumeExe for CreateVolume {
...
-                Err(error) => {
-                    context.volume.error(&format!(
-                        "Failed to create replica {:?} for volume, error: {}",
-                        replica,
-                        error.full_string()
-                    ));
-                    // continue trying...
-                }
-            };
+                fina_replicas = replicas;
```
tiagolobocastro: Why do we need this final_replicas?

control-plane/agents/src/bin/core/volume/scheduling.rs
```diff
@@ -14,12 +14,40 @@ use crate::{
...
+    let vec_of_pool_wrappers: Vec<PoolWrapper> =
+        volume::AddVolumeReplica::builder_with_defaults(request.clone(), registry)
+            .await
+            .collect()
+            .into_iter()
+            .map(|e| e.collect())
+            .collect();
+
+    if is_affinity {
+        result = second_pass_filter(vec_of_pool_wrappers, request, registry);
```
tiagolobocastro: We discussed this previously but I forgot, what is the second pass for? To build the list of lists for the specific case of the affinityKey?
I suggest renaming it to what it is.

control-plane/agents/src/bin/core/volume/scheduling.rs
```diff
@@ -143,3 +171,90 @@ pub(crate) async fn resizeable_replicas(
...
+    registry: &Registry,
+) -> Vec<Vec<PoolWrapper>> {
+    let mut map: HashMap<String, Vec<PoolWrapper>> = HashMap::new();
+    for pool in pools.iter_mut() {
+        let id = pool.state().id.clone();
+        let spec = registry.specs().pool(&id).unwrap_or_default();
+        let pool_labels = spec.labels.clone().unwrap_or_default();
+
+        // Get the affinity labels from the pool topology
+        let pool_affinity = match &suitable_pools.spec().topology.as_ref().unwrap().pool {
```
tiagolobocastro: probably use flat_map here

control-plane/agents/src/bin/core/volume/scheduling.rs
```diff
@@ -143,3 +171,90 @@ pub(crate) async fn resizeable_replicas(
...
+            Some(pool_topology) => match pool_topology {
+                PoolTopology::Labelled(label) => label.affinity.clone(),
+            },
+            None => HashMap::new(),
+        };
+
+        // performance: gold and zone: us-west, then the map key is "performance:gold;zone:us-west;"
```
tiagolobocastro, line 238: hmm I don't understand this, why is the map key `performance:gold;zone:us-west;`?
Shouldn't there be 2 separate keys?

control-plane/agents/src/bin/core/volume/operations.rs
sinhaashish: create_replica_candiate

control-plane/agents/src/bin/core/volume/operations.rs
sinhaashish: because we want to try out the set of other replica candiates. 
If all the elements in the set of replica candiates are over then we error out , which is done in the `run` method

control-plane/agents/src/bin/core/volume/operations.rs
sinhaashish: For every set of create replica candidates  , I am checking if the required number of replica is sufficed or not.

Let say  Set[[PoolA,PoolB, PoolC], [PoolD, PoolE]] 
If PoolA, PoolB , PoolC are good enough and replica can be created then , create the replica, dont look for PoolD, PoolE.
If the first element of set i.e. PoolA, PoolB, PoolC and accomodate the replicas, even if the first replica was present on PoolA, we dont check it. 

We just skip this set and move on to other i.e. PoolD and PoolE

control-plane/agents/src/bin/core/volume/operations.rs
sinhaashish: This is the final set of replicas on which volumes are provisioned.

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -964,86 +870,153 @@ impl CreateVolumeExeVal for CreateVolume {
     }
 }
 
+#[derive(Debug, Clone)]
+pub struct ReplicaWithLabels {
```
tiagolobocastro: Probably all these you added can just be pub(super)
Also missing doc comments

```suggestion
pub(super) struct ReplicaWithLabels {
```

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -964,86 +870,153 @@ impl CreateVolumeExeVal for CreateVolume {
...
-                    || spread_label_is_same(r, replica, context).unwrap_or_else(|error| {
+        candidates: Self::Candidates,
+    ) -> Result<Self::Replicas, SvcError> {
+        let mut final_replicas = Vec::<Replica>::new();
+        let mut hashmap: Option<HashMap<String, String>> = Some(HashMap::new());
+        // In case of affinty key, if some replica are already provisioned then delete them.
+        let request = GetSuitablePools::new(context.volume.as_ref(), None);
+        let is_affinity = is_affinity_requested(request.clone());
+
+        if context.volume.as_ref().num_replicas > 0 && is_affinity {
```
tiagolobocastro: What's >0 doing? Did you meant to do > 1 ?

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -964,86 +870,153 @@ impl CreateVolumeExeVal for CreateVolume {
...
+            match self.undo(context, &final_replicas).await {
```
tiagolobocastro: What are we undoing here?

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -964,86 +870,153 @@ impl CreateVolumeExeVal for CreateVolume {
...
+                continue;
+            } else {
+                final_replicas = replicas;
+                hashmap = create_replica_candidate.labels().clone();
+                break;
+            }
+        }
+
+        if final_replicas.len() < context.volume.as_ref().num_replicas as usize {
+            match self.undo(context, &final_replicas).await {
```
tiagolobocastro: Why is this repeated so many times?

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -964,86 +870,153 @@ impl CreateVolumeExeVal for CreateVolume {
...
+                Ok(_) => {}
+                Err(_) => {
+                    return Err(SvcError::ReplicaCreateNumber {
+                        id: context.volume.uid_str(),
+                    });
+                }
+            }
+        }
+        for create_replica_candidate in candidates.iter() {
+            let mut replicas =
```
tiagolobocastro, line 913: Why do we need a final_replicas, can't we just reset it on the for loop and just reuse it?
AFAICT you're never modifying final_replicas, you're only assigning it?

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -964,86 +870,153 @@ impl CreateVolumeExeVal for CreateVolume {
...
+                    Ok(_) => {}
+                    Err(_error) => {
+                        return Err(SvcError::ReplicaCreateNumber {
+                            id: context.volume.uid_str(),
+                        });
+                    }
                 }
+                continue;
+            } else {
+                final_replicas = replicas;
```
tiagolobocastro: can't we simply return here?

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -964,86 +870,153 @@ impl CreateVolumeExeVal for CreateVolume {
...
+            match context.registry.specs().replica(&replica_state.uuid).await {
                 Ok(mut replica) => {
                     let request = DestroyReplica::from(replica_state.clone());
                     replica
                         .destroy(context.registry, &request.with_disown_all())
-                        .await
+                        .await?;
                 }
-                Err(error) => Err(error),
+                Err(error) => return Err(error),
```
tiagolobocastro: Shouldn't we try to delete all before returning ?

control-plane/agents/src/bin/core/volume/scheduling.rs
tiagolobocastro: This is marked as resolved but not resolved, there's still unwrap here...

control-plane/agents/src/bin/core/volume/specs.rs
```diff
@@ -173,6 +181,54 @@ pub(crate) async fn nexus_attach_candidates(
     Ok(candidates)
 }
 
+/// Return a list of appropriate requests which can be used to create a replica on a pool.
+/// This can be used when the volume's current replica count is smaller than the desired volume's
+/// replica count.
+pub(crate) async fn volume_replica_candidates_new(
+    registry: &Registry,
+    volume_spec: &VolumeSpec,
+) -> Result<Vec<Vec<CreateReplicaNew>>, SvcError> {
```
tiagolobocastro: This doesn't seem right, again it's adding label per replica, rather than per vector of create_requests
Also use the `newtype` pattern rather than repeating all fields in the new definitions.
I think it should be more like this:
```rust
Result<Vec<CreateReplicas>, SvcError>....

struct CreateReplicas {
   replicas: Vec<CreateReplica>,
   affinity_labels: HashMap<String, String>,
}
```

control-plane/agents/src/bin/core/volume/operations.rs
sinhaashish: replaced with result

control-plane/agents/src/bin/core/volume/operations.rs
sinhaashish: done

control-plane/agents/src/bin/core/volume/operations.rs
sinhaashish: removed , added in todo

control-plane/agents/src/bin/core/volume/operations.rs
sinhaashish: added in todo

control-plane/agents/src/bin/core/volume/scheduling.rs
sinhaashish: It clubs all the keysand maps and create the key of the hash map inorder to club all pool wrappers having the same key value pair.

control-plane/agents/src/bin/core/volume/scheduling.rs
sinhaashish: done

control-plane/agents/src/bin/core/volume/scheduling.rs
tiagolobocastro: But why is this required, can't it simply put it as a map?

control-plane/agents/src/bin/core/volume/specs.rs
```diff
@@ -173,6 +181,84 @@ pub(crate) async fn nexus_attach_candidates(
...
+        volume_spec.trace(&format!(
+            "Creation pool candidates for volume: {:?}",
+            pools.iter().map(|p| p.state()).collect::<Vec<_>>()
+        ));
+        let mut create_replicas: Vec<CreateReplica> = Vec::new();
+        let mut labels = HashMap::new();
+        for pool_wrapper in pools.iter() {
+            let replica_uuid = ReplicaId::new();
+            let create_replica = CreateReplica {
+                node: pool_wrapper.node.clone(),
```
tiagolobocastro, line 238: hmm isn't there already code which setups the CreateReplica?

control-plane/agents/src/bin/core/volume/specs.rs
```diff
@@ -271,30 +357,41 @@ pub(crate) async fn create_volume_replicas(
...
 
-    if request.replicas > node_replicas.len() as u64 {
-        Err(SvcError::from(NotEnough::OfPools {
-            have: node_replicas.len() as u64,
+    let node_replicas_vec: Vec<CreateReplicas> =
+        volume_replica_candidates_new(registry, volume).await?;
+
+    // The first element of the node_replicas_vec contains the most suitable replicas.
+    // Error out if the first element node_replicas_vec doesn't have enough replicas.
+    if request.replicas > node_replicas_vec[0].replicas().len() as u64 {
```
tiagolobocastro, line 382: you're assuming index  0 always exists, is this true?

control-plane/agents/src/bin/core/volume/specs.rs
sinhaashish: its there but its in `volume_replica_candidates` which uses `volume_pool_candidates` . But in this case we are using `volume_pool_candidates_new`

control-plane/agents/src/bin/core/volume/specs.rs
sinhaashish: added more validation 
if node_replicas_vec.is_empty() {
        return Err(SvcError::from(NotEnough::OfPools {
            have: 0,
            need: request.replicas,
        }));
    }

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -920,38 +831,32 @@ pub(super) struct Context<'a> {
...
-        // decide what to do next
-        if replicas.len() < context.volume.as_ref().num_replicas as usize {
-            self.undo(&mut context, replicas).await;
-            Err(SvcError::ReplicaCreateNumber {
-                id: context.volume.uid_str(),
-            })
-        } else {
-            Ok(replicas)
-        }
+        let replicas = self.create(&mut context, candidates).await?;
```
tiagolobocastro, line 846: Why is this not calling undo anymore?

control-plane/agents/src/bin/core/volume/scheduling.rs
```diff
@@ -14,12 +14,40 @@ use crate::{
...
+    let vec_of_pool_wrappers: Vec<PoolWrapper> =
+        volume::AddVolumeReplica::builder_with_defaults(request.clone(), registry)
+            .await
+            .collect()
+            .into_iter()
+            .map(|e| e.collect())
+            .collect();
+
+    if is_affinity {
+        result = get_filtered_pools(vec_of_pool_wrappers, request, registry);
```
tiagolobocastro: Something like this..

```suggestion
        result = group_pools_by_affinity(vec_of_pool_wrappers, request, registry);
```

control-plane/agents/src/bin/core/volume/specs.rs
tiagolobocastro: Can we break down the existing function so we can reuse setting up the creation of the Vec<CreateReplica> ?

control-plane/agents/src/bin/core/volume/scheduling.rs
tiagolobocastro: Let's get a conclusion on this question so we can merge on this PR?

control-plane/agents/src/bin/core/volume/operations.rs
sinhaashish: it is taken care inside the create method itself

control-plane/agents/src/bin/core/volume/scheduling.rs
sinhaashish: done

sinhaashish:
>Can we break down the existing function so we can reuse setting up the creation of the Vec ? 

We cant as for every pool we are setting the labels as well .

control-plane/agents/src/bin/core/volume/specs.rs
sinhaashish: We cant as for every pool we are setting the labels as well .

control-plane/agents/src/bin/core/volume/scheduling.rs
```diff
@@ -143,3 +171,95 @@ pub(crate) async fn resizeable_replicas(
...
+            } else {
+                map.insert(result, vec![pool.clone()]);
+            }
+        }
+    }
+
+    let mut final_list_of_pools: Vec<Vec<PoolWrapper>> = map.into_values().collect();
+
+    // Sort vectors by their lengths in descending order
+    final_list_of_pools.sort_by_key(|b| std::cmp::Reverse(b.len()));
```
tiagolobocastro: Please add a rust unittest for this logic (in another pr is fine, but do it as the next)

control-plane/agents/src/bin/core/volume/specs.rs
tiagolobocastro: hmm I'm sure it can still be done somehow, but that's fine, I can take this up in refactoring later

tiagolobocastro (approved):

control-plane/agents/src/bin/core/pool/wrapper.rs
```diff
@@ -49,11 +54,16 @@ impl PoolWrapper {
         Self {
             state: pool,
             replicas,
+            labels,
             committed,
             free_space,
         }
     }
 
+    /// Get the labels
```
Abhinandan-Purkait: missing `.` at the end

Abhinandan-Purkait (approved):

sinhaashish:
bors merge

sinhaashish:
bors merge

sinhaashish:
bors merge

sinhaashish:
bors merge

sinhaashish:
bors merge

sinhaashish:
bors merge

sinhaashish:
bors merge

tiagolobocastro:
Had a look at this and I found the problem but the fact that this has taken so long is making me wonder if we've got things a bit too complicated here.
Was kinda difficult to navigate the code and figure out what's going on.
Let's close this PR, raise api changes only and let's get those merged in first.
Then we can try the implementation again.

PR #848, opened 2024-06-27
fix(deployer): increasing the max number of allowed connection attempts to the io-engine

sinhaashish:
increasing the max number of allowed connection attempts to the io-engine

### DIFF
--- a/deployer/src/infra/io_engine.rs
+++ b/deployer/src/infra/io_engine.rs
@@ -117,7 +117,7 @@ impl ComponentAction for IoEngine {
                 options.latest_io_api_version(),
                 &name,
                 socket,
-                40,
+                100,
                 tokio::time::sleep,
             )
             .await?;


deployer/src/infra/io_engine.rs
```diff
@@ -117,7 +117,7 @@ impl ComponentAction for IoEngine {
                 options.latest_io_api_version(),
                 &name,
                 socket,
-                40,
+                100,
```
Abhinandan-Purkait, line 120: Why do we need this change?

sinhaashish: some test cases were failing, as the deployer start command took time initialising the IO-Engines. This reduces the scope for that.

tiagolobocastro: Commit should be renamed, this is not increasing the sleep time, it's increasing the max number of allowed connection attempts to the io-engine.
This is needed because with a high number of io-engines it seems they can take a very long time to get ready on ashish's system. One culprit was: "Accel framework software module"

sinhaashish: done

Abhinandan-Purkait (approved):

tiagolobocastro (approved):

sinhaashish:
bors merge

PR #849, opened 2024-07-05
fix(fsfreeze): improved error message when volume is not staged

Abhinandan-Purkait:
### Describe the issue
It its seen that if a node has supposedly recovered from an abrupt shutdown and the application is not running, in that case the the filesystem mayastor volume is published but it's not mounted on the node where the application is. Mayastor control-plane volume being published gives the csi-controller a wrong notion that the volume is under use i.e mounted thus it tries to issue fsfreeze to ensure fs consistency, but fsfreeze wouldn't succeed as it would search for the mount path to issue FIFREEZE ioctl. In such case the flow returns a very vague error of `Terminated with error error=Cannot find volume: volume ID: e0316b8c-e4f4-454c-9803-c238dc68457c`.

### Changes
This PR enhances the error message stating what is the actual error i.e `Failed to freeze volume as the filesystem volume is not mounted`

### DIFF
--- a/control-plane/csi-driver/src/bin/controller/controller.rs
+++ b/control-plane/csi-driver/src/bin/controller/controller.rs
@@ -18,7 +18,7 @@ use utils::{dsp_created_by_key, DSP_OPERATOR};
 
 use regex::Regex;
 use std::{collections::HashMap, str::FromStr};
-use tonic::{Request, Response, Status};
+use tonic::{Code, Request, Response, Status};
 use tracing::{debug, error, instrument, trace, warn};
 use uuid::Uuid;
 use volume_capability::AccessType;
@@ -838,7 +838,7 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
         })?;
         let create_params = CreateSnapshotParams::try_from(&request.parameters)?;
 
-        // Get the volume object. Extract the app node endpoint if the quiesce is requested.
+        // Get the volume object. Extract the app node endpoint if quiesce is requested.
         let volume = RestApiClient::get_client().get_volume(&volume_uuid).await?;
         let app_node_endpoint_info = match volume_app_node(&volume) {
             // Volume is not published, so no need to quiesce.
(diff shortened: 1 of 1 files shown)


control-plane/csi-driver/src/bin/controller/controller.rs
```diff
@@ -861,10 +861,19 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
...
+                        issue_fs_freeze(app_node_endpoint.clone(), volume_uuid.to_string()).await
+                    {
+                        return match error.code() {
+                            Code::NotFound => Err(Status::not_found(format!(
+                                "Failed to freeze volume {}, filesystem volume is not mounted",
+                                volume_uuid
+                            ))),
+                            _ => Err(error),
+                        };
+                    }
```
tiagolobocastro: I think only 1 match suffices

```suggestion
                    match issue_fs_freeze(app_node_endpoint.clone(), volume_uuid.to_string()).await
                    {
                        Err(error) if error.code() == Code::NotFound => Err(Status::not_found(format!(
                                "Failed to freeze volume {}, filesystem volume is not mounted",
                                volume_uuid
                            ))),
                        _else => _else
                    }?;
```
Also is the volume not mounted, or not staged/attached?

tiagolobocastro (approved):
This is a bit confusing:
`feat(fs const snap): return meaningful error if volume is not attached by fsfreeze flow is triggered`
It's also a fix, not a feature. Example:
`fix(fsfreeze): improved error message when volume is not staged`

control-plane/csi-driver/src/bin/controller/controller.rs
Abhinandan-Purkait: I wanted to use attached, was wondering if users would be more used to the term mounted than attached?

tiagolobocastro: hmm not sure. Volume attachments is using the term attached, but they would have to know about it.

Abhinandan-Purkait: Alright. I will use attached here.

Abhinandan-Purkait: Done.

Abhinandan-Purkait:
bors merge

PR #853, opened 2024-08-07
feat(topology): restrict users from deleting created-by label on pools

sinhaashish:


control-plane/agents/src/bin/core/pool/service.rs
```diff
@@ -98,6 +99,15 @@ impl PoolOperations for Service {
             }
             .into());
         }
+
+        if !req.label_key().is_empty() && req.label_key() == dsp_created_by_key() {
```
tiagolobocastro:
```suggestion
        if req.label_key() == dsp_created_by_key() {
```

control-plane/agents/src/bin/core/pool/service.rs
```diff
@@ -98,6 +99,15 @@ impl PoolOperations for Service {
...
+            return Err(SvcError::InvalidLabelCreatedBy {
+                labels: req.label_key(),
+                resource_kind: ResourceKind::Pool,
+            }
+            .into());
+        }
```
tiagolobocastro: Actually, these checks should probably be done on the unlabel, rather than at this handler?

tiagolobocastro (commented):
hmm, I wonder if we should have distinguished internal from user labels.
And don't show internal labels to the user, and don't let them delete them unless a flag is specified?
Maybe for now also don't return the created_by_dsp label for the gets

control-plane/agents/src/bin/core/pool/service.rs
Abhinandan-Purkait: Yeah, we should do this from operations trait impl.

sinhaashish: done

control-plane/agents/src/common/errors.rs
```diff
@@ -208,6 +208,11 @@ pub enum SvcError {
         labels: String,
         resource_kind: ResourceKind,
     },
+    #[snafu(display("Invalid {}, labels: {} . Cannot delete `created-by` label as it is used for identifying mayastor pools", resource_kind, labels))]
+    InvalidLabelCreatedBy {
```
Abhinandan-Purkait: I don't think we should have an error variant for a specific key. Rather just call it `ForbiddenUnlabelKey` or `RestrictedUnlabelKey` and make it a generic one, for cases we have other labels to restrict. Also replace created-by with {labels} so that it's generic.

Abhinandan-Purkait (changes requested):

control-plane/agents/src/common/errors.rs
sinhaashish: done

Abhinandan-Purkait (approved):

control-plane/agents/src/common/errors.rs
```diff
@@ -208,6 +208,15 @@ pub enum SvcError {
         labels: String,
         resource_kind: ResourceKind,
     },
+    #[snafu(display(
+        "Forbidden {}, cannot delete forbidden labels: {} ",
```
tiagolobocastro:
```suggestion
        "Forbidden {}, cannot delete internal labels: {} ",
```

tiagolobocastro (approved):

control-plane/agents/src/common/errors.rs
sinhaashish: done

sinhaashish:
bors merge

sinhaashish:
bors ping

sinhaashish:
bors merge

PR #858, opened 2024-09-09
feat: handle new nvmf uri scheme

dsharma-dc:
This change lets the control plane csi node plugin to consume the new scheme deviceUri received from volume publish response during the volume attach path. For HA, things should work fundamentally with this. Some additional HA handling will be part of next patch where we handle HA path replacement with transport specific matching.

### DIFF
--- a/control-plane/agents/src/bin/ha/node/server.rs
+++ b/control-plane/agents/src/bin/ha/node/server.rs
@@ -24,7 +24,7 @@ use stor_port::{
 };
 
 use http::Uri;
-use nvmeadm::nvmf_subsystem::Subsystem;
+use nvmeadm::{nvmf_discovery::TrType, nvmf_subsystem::Subsystem};
 use std::{collections::HashMap, net::SocketAddr, sync::Arc, time::Instant};
 use tokio::time::{sleep, Duration};
 use utils::nvme_target_nqn_prefix;
@@ -181,6 +181,7 @@ impl NodeAgentSvc {
         let preexisted_subsystem = Subsystem::get(
             parsed_uri.host(),
             &parsed_uri.port(),
+            parsed_uri.transport(),
             parsed_uri.nqn().as_str(),
         )
         .is_ok();
@@ -198,6 +199,7 @@ impl NodeAgentSvc {
             Ok(_) => match Subsystem::get(
                 parsed_uri.host(),
                 &parsed_uri.port(),
+                parsed_uri.transport(),
                 parsed_uri.nqn().as_str(),
             ) {
                 Ok(subsystem) => {
@@ -293,6 +295,7 @@ impl NodeAgentSvc {
                     let curr_subsystem = match Subsystem::get(
                         parsed_uri.host(),
                         &parsed_uri.port(),
+                        parsed_uri.transport(),
                         parsed_uri.nqn().as_str(),
                     ) {
                         Ok(s) => s,
@@ -454,6 +457,7 @@ fn parse_uri(new_path: &str) -> Result<ParsedUri, SvcError> {
 struct ParsedUri {
     host: String,
     port: u16,
+    transport: TrType,
     nqn: String,
 }
 
@@ -461,15 +465,37 @@ impl ParsedUri {
     fn new(uri: Uri) -> Result<ParsedUri, SvcError> {
         let host = uri.host().ok_or(SvcError::InvalidArguments {})?.to_string();
         let port = uri.port().ok_or(SvcError::InvalidArguments {})?.as_u16();
+        // split returning a None indicates we possibly received old style uri scheme nvmf:// for
+        // new path, which implies transport as tcp.
+        let trstring = uri
+            .scheme_str()
+            .ok_or(SvcError::InvalidArguments {})?
+            .split('+')
+            .nth(1);
+
+        let transport = if let Some(t) = trstring {
+            t.parse().map_err(|_| SvcError::InvalidArguments {})?
+        } else {
+            TrType::default()
+        };
+
         let nqn = uri.path()[1 ..].to_string();
 
-        Ok(Self { host, port, nqn })
+        Ok(Self {
+            host,
+            port,
+            transport,
+            nqn,
+        })
     }
     fn host(&self) -> &str {
         &self.host
     }
     fn port(&self) -> u16 {
         self.port
+    }
+    fn transport(&self) -> TrType {
+        self.transport
     }
     fn nqn(&self) -> String {
         self.nqn.clone()
--- a/control-plane/csi-driver/src/bin/node/dev.rs
+++ b/control-plane/csi-driver/src/bin/node/dev.rs
@@ -79,7 +79,9 @@ impl Device {
         match url.scheme() {
             "file" => Ok(Box::new(nbd::Nbd::try_from(&url)?)),
             "iscsi" => Ok(Box::new(iscsi::IscsiAttach::try_from(&url)?)),
-            "nvmf" => Ok(Box::new(nvmf::NvmfAttach::try_from(&url)?)),
+            "nvmf" | "nvmf+tcp" | "nvmf+rdma+tcp" => {
+                Ok(Box::new(nvmf::NvmfAttach::try_from(&url)?))
+            }
             "nbd" => Ok(Box::new(nbd::Nbd::try_from(&url)?)),
             scheme => Err(DeviceError::from(format!(
                 "unsupported device scheme: {scheme}"
--- a/control-plane/csi-driver/src/bin/node/dev/nvmf.rs
+++ b/control-plane/csi-driver/src/bin/node/dev/nvmf.rs
@@ -2,11 +2,12 @@ use std::{
     collections::HashMap,
     convert::{From, TryFrom},
     path::Path,
+    str::FromStr,
 };
 
 use nvmeadm::{
     error::NvmeError,
-    nvmf_discovery::{disconnect, ConnectArgsBuilder},
+    nvmf_discovery::{disconnect, ConnectArgsBuilder, TrType},
 };
 
 use csi_driver::PublishParams;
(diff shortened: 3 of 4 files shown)


dsharma-dc:
bors try

dsharma-dc:
bors cancel
bors try

control-plane/agents/src/bin/ha/node/server.rs
```diff
@@ -454,23 +458,41 @@ fn parse_uri(new_path: &str) -> Result<ParsedUri, SvcError> {
...
         let host = uri.host().ok_or(SvcError::InvalidArguments {})?.to_string();
         let port = uri.port().ok_or(SvcError::InvalidArguments {})?.as_u16();
+        // split returning a None indicates we possibly received old style uri scheme nvmf:// for
+        // new path, which implies transport as tcp.
+        let xprt = uri
+            .scheme_str()
+            .ok_or(SvcError::InvalidArguments {})?
+            .split('+')
+            .nth(1)
+            .unwrap_or("tcp")
```
Abhinandan-Purkait: Let's take this from some enum or constants?

control-plane/csi-driver/src/bin/node/dev.rs
```diff
@@ -79,7 +79,7 @@ impl Device {
         match url.scheme() {
             "file" => Ok(Box::new(nbd::Nbd::try_from(&url)?)),
             "iscsi" => Ok(Box::new(iscsi::IscsiAttach::try_from(&url)?)),
-            "nvmf" => Ok(Box::new(nvmf::NvmfAttach::try_from(&url)?)),
+            "nvmf" | "nvmf+tcp" | "nvmf+rdma" => Ok(Box::new(nvmf::NvmfAttach::try_from(&url)?)),
```
Abhinandan-Purkait: rdma+tcp is equal to rdma here right?

control-plane/csi-driver/src/bin/node/dev/nvmf.rs
```diff
@@ -32,6 +33,7 @@ lazy_static::lazy_static! {
 pub(super) struct NvmfAttach {
     host: String,
     port: u16,
+    xprt: String,
```
Abhinandan-Purkait: Can we not use the TrType?

control-plane/csi-driver/src/bin/node/dev/nvmf.rs
```diff
@@ -335,3 +347,12 @@ pub(crate) fn volume_uuid_from_url(url: &Url) -> Result<Uuid, DeviceError> {
 
     extract_uuid(components[1]).map_err(|error| DeviceError::from(format!("invalid UUID: {error}")))
 }
+
+/// Extract nvmf fabric transport from Url.
+pub(crate) fn xprt_from_url(url: &Url) -> Result<String, DeviceError> {
```
Abhinandan-Purkait: Let's try to use the TrType wherever possible? We can parse into the enum from the first caller?

Abhinandan-Purkait (approved):

control-plane/csi-driver/src/bin/node/dev/nvmf.rs
dsharma-dc: We could, however the existing structures e.g. `Subsystem` have transport as String which makes comparisons easier. TrType is only used on connect args builder side.

control-plane/csi-driver/src/bin/node/dev.rs
dsharma-dc: > rdma+tcp is equal to rdma here right?

yes.

control-plane/csi-driver/src/bin/node/dev/nvmf.rs
dsharma-dc: https://github.com/openebs/mayastor-control-plane/pull/858#discussion_r1753197929

control-plane/agents/src/bin/ha/node/server.rs
```diff
@@ -454,23 +457,41 @@ fn parse_uri(new_path: &str) -> Result<ParsedUri, SvcError> {
 struct ParsedUri {
     host: String,
     port: u16,
+    transport: String,
```
tiagolobocastro: Should be a vector of transports? Or a enum such as { TCP, RDMA, RDMA+TCP } ?

control-plane/agents/src/bin/ha/node/server.rs
tiagolobocastro: If we use an enum then could simply be enum::TCP

control-plane/csi-driver/src/bin/node/dev/nvmf.rs
```diff
@@ -170,6 +181,7 @@ impl Attach for NvmfAttach {
                 };
                 let ca = ConnectArgsBuilder::default()
                     .traddr(&self.host)
+                    .transport(TrType::from_str(self.transport.as_str()).unwrap_or_default())
```
tiagolobocastro: We can impl TryFrom from the enum to TrType

```suggestion
                    .transport(TrType::try_from(self.transport)?)
```

tiagolobocastro (commented):
Will retry fallback as tcp come in new PR?

control-plane/agents/src/bin/ha/node/server.rs
dsharma-dc: > Should be a vector of transports? Or a enum such as { TCP, RDMA, RDMA+TCP } ?

Yes I'll make it a vector of transports. It'll be done when I implement the behaviour change on csi-node to connect rdma and fallback to tcp if initiator is not capable. With this change, the intention is to maintain the current behaviour.

control-plane/csi-driver/src/bin/node/dev/nvmf.rs
dsharma-dc: > We can impl TryFrom from the enum to TrType

What'd be benefit as compared to `FromStr` ?

dsharma-dc:
bors try

control-plane/agents/src/bin/ha/node/server.rs
dsharma-dc: modified to use enum, not in a very straightforward way though.

control-plane/csi-driver/src/bin/node/dev/nvmf.rs
tiagolobocastro: If we make transport an enum as I suggested then you have to use the type generic `TryFrom`

control-plane/agents/src/bin/ha/node/server.rs
```diff
@@ -454,23 +457,45 @@ fn parse_uri(new_path: &str) -> Result<ParsedUri, SvcError> {
...
+        // new path, which implies transport as tcp.
+        let transport = uri
+            .scheme_str()
+            .ok_or(SvcError::InvalidArguments {})?
+            .split('+')
+            .nth(1)
+            .map_or(TrType::default(), |s| {
+                s.parse()
+                    .map_err(|_| format!("Couldn't parse TrType from {s:?}"))
+                    .expect("Failure parsing TrType")
```
tiagolobocastro: this will crash, please handle error by returning up the stack

control-plane/csi-driver/src/bin/node/dev/nvmf.rs
```diff
@@ -335,3 +347,12 @@ pub(crate) fn volume_uuid_from_url(url: &Url) -> Result<Uuid, DeviceError> {
 
     extract_uuid(components[1]).map_err(|error| DeviceError::from(format!("invalid UUID: {error}")))
 }
+
+/// Extract nvmf fabric transport from Url.
+pub(crate) fn transport_from_url(url: &Url) -> Result<TrType, DeviceError> {
+    // Shouldn't expect nvmf:// scheme here in reality. However, if control plane is
+    // interacting with an old io-engine then old style uri scheme will be received.
+    // Default to tcp for handling that case.
+    let xprt = url.scheme().split('+').nth(1).unwrap_or("tcp");
```
tiagolobocastro: nit: use TrType::TCP

control-plane/csi-driver/src/bin/node/dev.rs
```diff
@@ -79,7 +79,9 @@ impl Device {
         match url.scheme() {
             "file" => Ok(Box::new(nbd::Nbd::try_from(&url)?)),
             "iscsi" => Ok(Box::new(iscsi::IscsiAttach::try_from(&url)?)),
-            "nvmf" => Ok(Box::new(nvmf::NvmfAttach::try_from(&url)?)),
+            "nvmf" | "nvmf+tcp" | "nvmf+rdma+tcp" => {
```
tiagolobocastro, line 82: have we reached consensus here? CC @Abhinandan-Purkait 
Btw another thing we could do is piggy back on the query parameter, and say `?fallback=tcp`

dsharma-dc: I've modified dataplane PR to return `nvmf+rdma+tcp` (both transports) so let's stick to that, and not introduce query parameter for now as we have enough context to decide on fallback.

control-plane/agents/src/bin/ha/node/server.rs
dsharma-dc: done, with some dancing around.

control-plane/csi-driver/src/bin/node/dev/nvmf.rs
dsharma-dc: done.

sinhaashish (approved):

control-plane/csi-driver/src/bin/node/dev.rs
Abhinandan-Purkait: Sounds good to me.

tiagolobocastro (approved):

dsharma-dc:
bors merge

PR #862, opened 2024-09-16
Don't destroy replicas part of shutdown nexuses

tiagolobocastro:
feat: re-shutdown nexus when node is online
    
    When nexus shutdown fails, if the node comes back online let's
    attempt the shutdown again.
    
---

    fix: don't disown replica from unshutdown nexus
    
    In case nexus shutdown failed we used to disown the replica from
    both the volume and the nexus.
    This can be a problem if the nexus is still running as the io-engine
    does not handle it gracefully, leading into pool lock issues.
    Instead, simply disown the replica from the volume and not the nexus.
    
### DIFF
--- a/control-plane/agents/src/bin/core/controller/reconciler/nexus/re_shutdown.rs
+++ b/control-plane/agents/src/bin/core/controller/reconciler/nexus/re_shutdown.rs
@@ -1,0 +1,50 @@
+use crate::controller::{
+    reconciler::PollTriggerEvent,
+    resources::operations_helper::OperationSequenceGuard,
+    task_poller::{PollContext, PollEvent, PollResult, PollerState, TaskPoller},
+};
+
+/// ReShutdown nexuses if node comes back online with the Nexus intact.
+#[derive(Debug)]
+pub(super) struct ReShutdown {}
+impl ReShutdown {
+    /// Return a new `Self`.
+    pub(super) fn new() -> Self {
+        Self {}
+    }
+}
+
+#[async_trait::async_trait]
+impl TaskPoller for ReShutdown {
+    async fn poll(&mut self, context: &PollContext) -> PollResult {
+        // Fetch all nexuses that are not properly shutdown
+        for nexus in context.registry().specs().failed_shutdown_nexuses().await {
+            let Some(volume_id) = &nexus.immutable_ref().owner else {
+                continue;
+            };
+            let Ok(_volume) = context.specs().volume(volume_id).await else {
+                continue;
+            };
+
+            let Ok(mut nexus) = nexus.operation_guard() else {
+                continue;
+            };
+
+            nexus.re_shutdown_nexus(context.registry()).await;
+        }
+        PollResult::Ok(PollerState::Idle)
+    }
+
+    async fn poll_timer(&mut self, _context: &PollContext) -> bool {
+        false
+    }
+
+    async fn poll_event(&mut self, context: &PollContext) -> bool {
+        matches!(
+            context.event(),
+            PollEvent::Triggered(PollTriggerEvent::Start)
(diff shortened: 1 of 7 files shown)


control-plane/agents/src/bin/core/controller/reconciler/volume/garbage_collector.rs
```diff
@@ -345,62 +338,42 @@ async fn disown_non_reservable_replicas(
...
         .registry()
         .specs()
         .volume_failed_shutdown_nexuses(volume.uuid())
         .await;
 
-    // Remove the local child from the shutdown pending nexuses as they are
-    // non reservable.
+    // Remove the local children from the volume as they are non-reservable.
     for nexus in shutdown_failed_nexuses {
+        if target.uuid() == nexus.uuid() {
```
dsharma-dc, line 349: IIUC `target` represents current nexus here? If that has failed shutdown, why don't we want to go down the same code path for such target?

control-plane/agents/src/bin/core/controller/reconciler/volume/garbage_collector.rs
```diff
@@ -345,62 +338,42 @@ async fn disown_non_reservable_replicas(
...
+            };
+            if !replica.as_ref().dirty()
+                && replica.as_ref().owners.owned_by(volume.uuid())
+                && !replica
+                    .as_ref()
+                    .owners
+                    .nexuses()
+                    .iter()
+                    .any(|p| p != nexus.uuid())
+                && !is_replica_healthy(context, &mut nexus_info, replica.as_ref(), volume.as_ref())
```
dsharma-dc, line 372: I wonder do we really need to have health check of replica here? Irrespective of it's healthy or not, we can probably disown it from volume but not nexus, because eventually it is bound to be destroyed when nexus shutdown/cleanup happens.

dsharma-dc (approved):

control-plane/agents/src/bin/core/controller/reconciler/volume/garbage_collector.rs
tiagolobocastro: In this case we probably don't want to be disowning children outside of the existing logic for managing the volume target, as it may interfere with existing logic

control-plane/agents/src/bin/core/controller/reconciler/volume/garbage_collector.rs
tiagolobocastro: I guess it's just a safety gate, because is shouldn't really be healthy

control-plane/agents/src/bin/core/controller/reconciler/nexus/re_shutdown.rs
```diff
@@ -0,0 +1,50 @@
...
+}
+
+#[async_trait::async_trait]
+impl TaskPoller for ReShutdown {
+    async fn poll(&mut self, context: &PollContext) -> PollResult {
+        // Fetch all nexuses that are not properly shutdown
+        for nexus in context.registry().specs().failed_shutdown_nexuses().await {
+            let Some(volume_id) = &nexus.immutable_ref().owner else {
+                continue;
+            };
```
Abhinandan-Purkait, line 24: Can we not use the `volume_failed_shutdown_nexuses` method here?

Abhinandan-Purkait (approved):

control-plane/agents/src/bin/core/controller/reconciler/nexus/re_shutdown.rs
tiagolobocastro: hmm no because we don't have a volume here

tiagolobocastro:
bors merge

PR #863, opened 2024-09-18
feat: handle non-matching transport path removal

dsharma-dc:
- If the transport changes when nexus is recreated on same node, we connect the new path with new transport and disconnect the old transport path. IO isn't disrupted as initiator switches path.
- When nexus is recreated on a different node, then old path is disconnected as today, irrespective of transport.
- If the transport doesn't change and nexus is recreated on same node, but tr_addr changes, then old path(with old tr_addr) is disconnected as today.

### DIFF
--- a/control-plane/agents/src/bin/ha/node/server.rs
+++ b/control-plane/agents/src/bin/ha/node/server.rs
@@ -93,6 +93,20 @@ impl NodeAgentSvc {
     }
 }
 
+fn is_same_controller(new_path_uri: &ParsedUri, subsystem: &Subsystem) -> Result<bool, SvcError> {
+    let ctrlr_transport = subsystem
+        .transport
+        .parse()
+        .map_err(|_| SvcError::InvalidArguments {})?;
+
+    let same_transport = new_path_uri.transport().eq(&ctrlr_transport);
+    let same_host_port = subsystem
+        .address
+        .match_host_port(new_path_uri.host(), &new_path_uri.port().to_string());
+
+    Ok(same_transport && same_host_port)
+}
+
 /// Disconnect cached NVMe controller.
 async fn disconnect_controller(
     ctrlr: &NvmeController,
@@ -116,10 +130,11 @@ async fn disconnect_controller(
                 });
             }
 
-            if subsystem
-                .address
-                .match_host_port(new_path_uri.host(), &new_path_uri.port().to_string())
-            {
+            // It is same controller only if transport also match. Otherwise we disconnect it.
+            // This also takes care to disconnect old path if transport hasn't changed but host addr
+            // changed, for example due to interface name modified in io-engine args.
+            let same_controller = is_same_controller(&new_path_uri, &subsystem)?;
+            if same_controller {
                 tracing::info!(path, "Not disconnecting same NVMe controller");
                 Ok(None)
             } else {


control-plane/agents/src/bin/ha/node/server.rs
```diff
@@ -116,9 +116,19 @@ async fn disconnect_controller(
...
+                .map_err(|_| SvcError::InvalidArguments {})?;
+            let ctrlr_xprt_same_as_new_path = new_path_uri.transport().eq(&ctrlr_transport);
+
+            // It is same controller only if transport also match. Otherwise we disconnect it.
+            // This also takes care to disconnect old path if transport hasn't changed but host addr
+            // changed, for example due to interface name modified in io-engine args.
+            if ctrlr_xprt_same_as_new_path
+                && subsystem
+                    .address
+                    .match_host_port(new_path_uri.host(), &new_path_uri.port().to_string())
```
tiagolobocastro:
```suggestion
            let transport = subsystem
                .transport
                .parse()
                .map_err(|_| SvcError::InvalidArguments {})?;
            let same_transport = new_path_uri.transport().eq(&ctrlr_transport);
            let same_host_port = subsystem.address.match_host_port(new_path_uri.host(), &new_path_uri.port().to_string());

            // It is same controller only if the transport is also a match. Otherwise we disconnect it.
            // This also takes care to disconnect old path if the transport hasn't changed but host addr
            // changed, for example due to interface name modified in io-engine args.
            if same_transport && same_host_port
```

tiagolobocastro: Also consider putting this in a helper function, taking both the new_path_uri and the subsystem.

tiagolobocastro (approved):
I wonder if the subsystem itself should have a match

dsharma-dc:
> I wonder if the subsystem itself should have a match

Actually I think our abstraction of `Subsystem` is not ideal. We are using it as a path abstraction. A nexus(nqn) is a single subsystem which can be connected via multiple paths, and also the way initiator actually organises it is that a subsystem contains multiple paths e.g.
```
nvme-subsys0 - NQN=nqn.2019-05.io.openebs:8fb3b048-8a87-45a4-9dc4-9c396821156d
\
 +- nvme0 tcp traddr=65.21.186.165 trsvcid=8420 src_addr=65.21.186.165 live 
 +- nvme1 rdma traddr=65.21.186.165 trsvcid=8420 live 
```

Abhinandan-Purkait (approved):

Abhinandan-Purkait:
> > I wonder if the subsystem itself should have a match
> 
> Actually I think our abstraction of `Subsystem` is not ideal. We are using it as a path abstraction. A nexus(nqn) is a single subsystem which can be connected via multiple paths, and also the way initiator actually organises it is that a subsystem contains multiple paths e.g.
> 
> ```
> nvme-subsys0 - NQN=nqn.2019-05.io.openebs:8fb3b048-8a87-45a4-9dc4-9c396821156d
> \
>  +- nvme0 tcp traddr=65.21.186.165 trsvcid=8420 src_addr=65.21.186.165 live 
>  +- nvme1 rdma traddr=65.21.186.165 trsvcid=8420 live 
> ```

Yeah, I agree. We should call it Subsystem + have List of SubsystemPaths in it or something probably.

control-plane/agents/src/bin/ha/node/server.rs
dsharma-dc: > Also consider putting this in a helper function, taking both the new_path_uri and the subsystem.

done.

dsharma-dc:
bors try

dsharma-dc:
bors try

dsharma-dc:
bors try

dsharma-dc:
bors try

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

PR #864, opened 2024-09-18
Cherry Pick 862

tiagolobocastro:
feat: re-shutdown nexus when node is online
    
    When nexus shutdown fails, if the node comes back online let's
    attempt the shutdown again.
    
---

    fix: don't disown replica from unshutdown nexus
    
    In case nexus shutdown failed we used to disown the replica from
    both the volume and the nexus.
    This can be a problem if the nexus is still running as the io-engine
    does not handle it gracefully, leading into pool lock issues.
    Instead, simply disown the replica from the volume and not the nexus.
    

Abhinandan-Purkait (approved):

sinhaashish (approved):

tiagolobocastro:
bors merge

PR #867, opened 2024-09-24
build(openapi): don't regenerate on same md5sum and delete files

tiagolobocastro:
Running the generate script twice on a nix-build was failing because we don't make use of the .git during this build.
Fix this by manually cleaning up files as required. This should not happen anyway as md5sum would have been the same so: Ensure md5sum is correctly calculated


Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

PR #869, opened 2024-09-27
Allow custom logging for the bdd tests

tiagolobocastro:
test(bdd): use new rust-log args
    
    Now you can setup RUST_LOG which will be forwarded to the containers.
    todo: we might not want to see this in deployer itself, so perhaps
    we can also add some knob to ensure it's only used for the containers.
    
---

    test: run cleanup only once
    
    Ensure the cleanup trap is disabled when exiting.
    Also add more robust checks for FAST and CLEAN modes.
    
---

    test: allow modifying rust-log for deployer containers
    
    This will aid debugging as allows us to increase verbosity of all
    or a subset of the rust modules.
    todo: we may want to do this with more fine grained control.
    
### DIFF
--- a/deployer/src/lib.rs
+++ b/deployer/src/lib.rs
@@ -332,6 +332,13 @@ pub struct StartOptions {
 
     #[clap(long, default_value = "false")]
     pub enable_app_node_registration: bool,
+
+    /// Add custom RUST_LOG env to all containers.
+    #[clap(long)]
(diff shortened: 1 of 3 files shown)


sinhaashish (approved):

Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

PR #875, opened 2024-10-14
feat: connect over tcp if initiator isn't rdma capable

dsharma-dc:
- Added a cli option to control whether fallback is required or not. default kept `false`. We can make this default true once tested thoroughly. have some setup limitations currently.
- todo: adding a check(if required) to see if nvme_rdma module is available or not.

### DIFF
--- a/control-plane/agents/src/bin/ha/node/server.rs
+++ b/control-plane/agents/src/bin/ha/node/server.rs
@@ -135,7 +135,7 @@ async fn disconnect_controller(
             // changed, for example due to interface name modified in io-engine args.
             let same_controller = is_same_controller(&new_path_uri, &subsystem)?;
             if same_controller {
-                tracing::info!(path, "Not disconnecting same NVMe controller");
+                tracing::info!(path, "Not disconnecting same NVMe controller. old subsys {subsystem:?}, new path {new_path_uri:?}");
                 Ok(None)
             } else {
                 tracing::info!(path, "Disconnecting NVMe controller");
@@ -476,6 +476,16 @@ struct ParsedUri {
     nqn: String,
 }
 
+impl std::fmt::Debug for ParsedUri {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("ParsedUri")
+            .field("host", &self.host())
+            .field("port", &self.port())
+            .field("transport", &self.transport())
+            .field("nqn", &self.nqn())
+            .finish()
+    }
+}
 impl ParsedUri {
     fn new(uri: Uri) -> Result<ParsedUri, SvcError> {
         let host = uri.host().ok_or(SvcError::InvalidArguments {})?.to_string();
--- a/control-plane/csi-driver/src/bin/node/dev/nvmf.rs
+++ b/control-plane/csi-driver/src/bin/node/dev/nvmf.rs
@@ -14,6 +14,7 @@ use csi_driver::PublishParams;
 use glob::glob;
 use nvmeadm::nvmf_subsystem::Subsystem;
 use regex::Regex;
+use tracing::warn;
 use udev::{Device, Enumerator};
 use url::Url;
 use uuid::Uuid;
@@ -22,6 +23,7 @@ use crate::{
     config::{config, NvmeConfig, NvmeParseParams},
     dev::util::extract_uuid,
     match_dev::match_nvmf_device,
+    node::RDMA_CONNECT_CHECK,
 };
 
 use super::{Attach, Detach, DeviceError, DeviceName};
@@ -307,6 +309,16 @@ pub(crate) fn check_nvme_tcp_module() -> Result<(), std::io::Error> {
     Ok(())
 }
 
+/// Check for the presence of nvme tcp kernel module.
+/// TODO: Handle the case where this(and for that matter nvme_tcp too)
+/// could be builtin module.
+#[allow(unused)]
+pub(crate) fn check_nvme_rdma_module() -> Result<(), std::io::Error> {
+    let path = "/sys/module/nvme_rdma";
+    std::fs::metadata(path)?;
+    Ok(())
+}
+
 /// Set the nvme_core module IO timeout
 /// (note, this is a system-wide parameter)
 pub(crate) fn set_nvmecore_iotimeout(io_timeout_secs: u32) -> Result<(), std::io::Error> {
@@ -359,5 +371,19 @@ pub(crate) fn transport_from_url(url: &Url) -> Result<TrType, DeviceError> {
         .split('+')
         .nth(1)
         .unwrap_or(default_xprt.as_str());
-    TrType::from_str(xprt).map_err(|e| DeviceError::new(format!("{e:?}").as_str()))
+
+    let ret_xprt = TrType::from_str(xprt).map_err(|e| DeviceError::new(format!("{e:?}").as_str()));
+    let connect_cap_check = RDMA_CONNECT_CHECK.get().unwrap_or(&(false, false));
+
+    if !connect_cap_check.0 {
+        ret_xprt
+    } else {
+        match ret_xprt {
+            Ok(t) if t == TrType::rdma && !connect_cap_check.1 => {
+                warn!("rdma incapable node, connecting over tcp");
+                Ok(TrType::tcp)
+            }
+            _else => _else,
+        }
+    }
 }
--- a/control-plane/csi-driver/src/bin/node/main_.rs
+++ b/control-plane/csi-driver/src/bin/node/main_.rs
@@ -9,7 +9,7 @@ use crate::{
     identity::Identity,
     k8s::patch_k8s_node,
     mount::probe_filesystems,
-    node::Node,
+    node::{Node, RDMA_CONNECT_CHECK},
     nodeplugin_grpc::NodePluginGrpcServer,
     nodeplugin_nvme::NvmeOperationsSvc,
     registration::run_registration_loop,
@@ -221,6 +221,16 @@ pub(super) async fn main() -> anyhow::Result<()> {
                 .value_parser(clap::value_parser!(bool))
                 .help("Enable ansi color for logs")
         )
+        .arg(
+            Arg::new("nvme-connect-fallback")
+                .long("nvme-connect-fallback")
+                .default_value("false")
+                .value_parser(clap::value_parser!(bool))
+                .help(
+                    "Enable falling back to nvme connect over tcp if initiator node is not rdma capable, \n\
+                    even though volume target is rdma capable."
+            )
+        )
         .subcommand(
             clap::Command::new("fs-freeze")
                 .arg(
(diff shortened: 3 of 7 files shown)


Abhinandan-Purkait (approved):

control-plane/csi-driver/src/bin/node/main_.rs
```diff
@@ -309,6 +319,31 @@ pub(super) async fn main() -> anyhow::Result<()> {
...
+
+    if ibv_output.as_ref().is_ok_and(|s| {
+        !String::from_utf8(s.stdout.clone())
+            .unwrap_or_default()
+            .starts_with("0 HCA")
+    }) {
+        info!(
+            "host node rdma capable. conn_fallback({conn_fallback:?}), {:?}",
+            String::from_utf8(ibv_output.unwrap().stdout.clone())
+        );
```
Abhinandan-Purkait: Let's also check the output.status is success or not as well?

abhilashshetty04 (approved):

control-plane/csi-driver/src/bin/node/main_.rs
```diff
@@ -203,6 +203,16 @@ pub(super) async fn main() -> anyhow::Result<()> {
                 .value_parser(clap::value_parser!(bool))
                 .help("Enable ansi color for logs")
         )
+        .arg(
+            Arg::new("nvme-connect-fallback")
```
tiagolobocastro, line 225: naming seems a bit non-specific?
nvme-tcp-connect-fallback? not sure I like this one better though

control-plane/csi-driver/src/bin/node/main_.rs
```diff
@@ -309,6 +319,31 @@ pub(super) async fn main() -> anyhow::Result<()> {
         check_ana_and_label_node(&kube_client, node_name, nvme_enabled).await?;
     }
 
+    let conn_fallback = matches.get_flag("nvme-connect-fallback");
+    // Check and store if this node has an RDMA device so that nvme connect over RDMA
+    // can be done. If this node is also an io-engine node with rdma enabled and working,
+    // then this check will always set capability true.
+    let ibv_output = tokio::process::Command::new("ibv_devinfo")
```
tiagolobocastro, line 344: Can this be done with a library call instead?

control-plane/csi-driver/src/bin/node/main_.rs
```diff
@@ -309,6 +319,31 @@ pub(super) async fn main() -> anyhow::Result<()> {
...
+        .await;
+
+    if ibv_output.as_ref().is_ok_and(|s| {
+        !String::from_utf8(s.stdout.clone())
+            .unwrap_or_default()
+            .starts_with("0 HCA")
+    }) {
+        info!(
+            "host node rdma capable. conn_fallback({conn_fallback:?}), {:?}",
+            String::from_utf8(ibv_output.unwrap().stdout.clone())
```
tiagolobocastro: reuse the output from above rather than rebuild it here

control-plane/csi-driver/src/bin/node/main_.rs
dsharma-dc: > Can this be done with a library call instead?

We could, there are some crates available not sure how stable. Until we start managing devices on our own and have the crates tested, I don't intend to integrate those in our code.

control-plane/csi-driver/src/bin/node/main_.rs
dsharma-dc: fair enough, will rename to `nvme-tcp-connect-fallback`

tiagolobocastro: As I said, I don't prefer `nvme-tcp-connect-fallback`, so fine with existing one. If you can find a better one great, otherwise it's fine.

control-plane/csi-driver/src/bin/node/main_.rs
tiagolobocastro: There's a crate from a very well known peep in the rust community, but it seems to build the rdma library itself during the build, which might complicate things :(
Let's use the binary then.

control-plane/csi-driver/src/bin/node/main_.rs
dsharma-dc: > reuse the output from above rather than rebuild it here

Directly using may be tricky (unless we rearrange using some nested if else?) as that's inside closure along with `status` check. I've removed the redundant `clone()` now, and `String::from_utf8` doesn't do any cloning/copying of vector internally.

tiagolobocastro: I would say split it into multiple statements.
Also I think there's a bug, if string parsing fails, then it's being taken as the host is rdma capable?

Maybe something like this:
```suggestion
    let output = ibv.output.as_ref().flat_map(|s| String::from_utf8(s.stdout.clone().ok());
    match output {
      Some(output) if ibv.status.success() && !output.starts_with("0 HCA") => {
        info!(
            "host node rdma capable. conn_fallback({conn_fallback:?}), {output}",
        );
        // todo: check for the presence of nvme_rdma too. But don't bail out as we do for nvme_tcp.
        let _ = RDMA_CONNECT_CHECK.set((conn_fallback, true));
      _ => {
        error!("Error executing ibv_devinfo, or no output received. conn_fallback({conn_fallback:?}), {ibv_output:?}");
        let _ = RDMA_CONNECT_CHECK.set((conn_fallback, false));
      }
    }
```

dsharma-dc: ok so I've rearranged to keep it simple now, no fancy combinators. 
> Also I think there's a bug, if string parsing fails, then it's being taken as the host is rdma capable?

Handled this by erring towards incapability.

nix/overlay.nix
```diff
@@ -9,4 +9,5 @@ self: super: {
   control-plane = super.callPackage ./pkgs/control-plane { inherit allInOne incremental tag; };
   openapi-generator = super.callPackage ./pkgs/openapi-generator { };
   xfsprogs_5_16 = (import (super.sources).nixpkgs-22_05 { }).xfsprogs;
+  rdmacore = (import (super.sources).nixpkgs-22_05 { }).rdma-core;
```
tiagolobocastro: why do we need to pull a special rdma-core?

tiagolobocastro: I mean rather than just taking the default one from nixpkgs

control-plane/csi-driver/src/bin/node/dev/nvmf.rs
```diff
@@ -359,5 +371,19 @@ pub(crate) fn transport_from_url(url: &Url) -> Result<TrType, DeviceError> {
         .split('+')
         .nth(1)
         .unwrap_or(default_xprt.as_str());
-    TrType::from_str(xprt).map_err(|e| DeviceError::new(format!("{e:?}").as_str()))
+
+    let ret_xprt = TrType::from_str(xprt).map_err(|e| DeviceError::new(format!("{e:?}").as_str()));
+    let connect_cap_check = RDMA_CONNECT_CHECK.get().unwrap_or(&(false, false));
+
+    if !connect_cap_check.0 {
```
tiagolobocastro, line 378: I'm a bit confused with the bools, if we don't check, shouldn't we check if RDMA is supported? otherwise  might as well bail out?

control-plane/agents/src/bin/ha/node/server.rs
```diff
@@ -99,6 +99,11 @@ fn is_same_controller(new_path_uri: &ParsedUri, subsystem: &Subsystem) -> Result
         .parse()
         .map_err(|_| SvcError::InvalidArguments {})?;
 
+    tracing::info!(
```
tiagolobocastro: Do we need this trace?

control-plane/csi-driver/src/bin/node/dev/nvmf.rs
```diff
@@ -307,6 +309,16 @@ pub(crate) fn check_nvme_tcp_module() -> Result<(), std::io::Error> {
     Ok(())
 }
 
+/// Check for the presence of nvme tcp kernel module.
+/// TODO: Handle the case where this(and for that matter nvme_tcp too)
```
tiagolobocastro, line 313: we need to fix the init-containers as well, which only check for module and not builtin

control-plane/csi-driver/src/bin/node/dev/nvmf.rs
dsharma-dc: > I'm a bit confused with the bools, if we don't check, shouldn't we check if RDMA is supported? otherwise might as well bail out?

The logic is: If fallback is not enabled via cli, we return `ret_xprt` as-is. It could be tcp or rdma depending on nvmf url. If connect fails for rdma, let it fail.

nix/overlay.nix
dsharma-dc: There is no special requirement. I think I just followed example from `xfsprogs` and added here. Are you suggesting adding package name to shell.nix and cargo-package.nix, and not pulling from a specific nix channel? If that also works, should be ok.

control-plane/agents/src/bin/ha/node/server.rs
dsharma-dc: I've added this to debug a particular scenario if that ever happens, we'll have a trace. Should be ok here as it's a non-frequent codepath.

control-plane/csi-driver/src/bin/node/dev/nvmf.rs
dsharma-dc: yeah, will do that as separate task. We have ticket for that AFAIK.

control-plane/agents/src/bin/ha/node/server.rs
tiagolobocastro: There's already a log on the caller side, I suggest enhancing it with more tracing fields rather than adding this one

nix/overlay.nix
tiagolobocastro: Yeah just remove it from here and use it as any other package

control-plane/csi-driver/src/bin/node/dev/nvmf.rs
tiagolobocastro: Do we need a good user message when we fail later on? Otherwise we can fail here with our own error message.

control-plane/csi-driver/src/bin/node/main_.rs
```diff
@@ -327,6 +337,35 @@ pub(super) async fn main() -> anyhow::Result<()> {
...
+            } else {
+                false
+            }
+        }
+        Ok(_) | Err(_) => {
+            error!("Error executing ibv_devinfo, or invalid command output. conn_fallback({conn_fallback:?}), {ibv_output:?}");
+            false
+        }
+    };
+    let _ = RDMA_CONNECT_CHECK.set((conn_fallback, cap));
```
tiagolobocastro, line 370: nit: I suggest you not use once cell, and simply store this in the service

tiagolobocastro (approved):

control-plane/csi-driver/src/bin/node/dev/nvmf.rs
dsharma-dc: let's not fail here.
If it fails due to incapability later on, the trace that we've in main_.rs:363 will be relevant.

tiagolobocastro: Can you please check which error message we log in this case?

dsharma-dc: ok, needed to add one in rearranged code. Will add this at L359

```
            } else {
                info!("host node is not rdma capable. conn_fallback({conn_fallback:?})");
                false
            }
```

control-plane/csi-driver/src/bin/node/main_.rs
dsharma-dc: As discussed, keeping it as-is.

control-plane/agents/src/bin/ha/node/server.rs
dsharma-dc: done

nix/overlay.nix
dsharma-dc: modified nix scripts.

dsharma-dc:
bors try

dsharma-dc:
> ## try
> Build failed:
> 
> * [continuous-integration/jenkins/branch](https://mayastor-ci.mayadata.io/job/Mayastor-Control-Plane/job/trying/130/display/redirect)

CI failed with `Operation not permitted` errors on tmpfs disk files. Perhaps some stale things left behind from earlier runs. Retrying once again.

bors try

dsharma-dc:
bors try

dsharma-dc:
> * continuous-integration/jenkins/branch

ok, not seeing the infra issue anymore, but this run failed at an unrelated(volume resize) test which is flaky it seems.

bors try

dsharma-dc:
bors merge

PR #881, opened 2024-10-28
build(openapi): take into account templates folder

tiagolobocastro:
Ensures the md5sum skips the template and that we don't delete it.


niladrih (approved):

tiagolobocastro:
bors merge

Abhinandan-Purkait (approved):

PR #887, opened 2024-11-20
Fix regression for pool creation timeout retry

tiagolobocastro:
test: use tmp in project workspace
    
    Use a tmp folder from the workspace allowing us to cleanup up things like
    LVM volumes a lot easier as we can just purge it.
    
---

    test(pool): create on very large or very slow disks
    
    Uses LVM Lvols as backend devices for the pool.
    We suspend these before pool creation, allowing us to simulate slow
    pool creation.
    This test ensures that the pool creation is completed by itself and also
    that a client can also complete it by calling create again.
    
---

    fix: allow pool creation to complete asynchronously
    
    When the initial create gRPC times out, the data-plane may still be creating
    the pool in the background, which can happen for very large pools.
    Rather than assume failure, we allow this to complete in the background up to
    a large arbitrary amount of time. If the pool creation completes before, then
    we retry the creation flow.
    The reason why we don't simply use very large timeouts is because the gRPC
    operations are currently sequential, mostly due to historical reasons.
    Now that the data-plane is allowing concurrent calls, we should also allow
    this on the control-plane.
    TODO: allow concurrent node operations
    
---

    fix: check for correct not found error code
    
    A previous fix ended up not working correctly because it was merged
    incorrectly, somehow!
    
---

    chore: update terraform node prep
    
    Pull the Release key from a recent k8s version since the old keys are no
    longer valid.
    This will have to be updated from time to time.

### DIFF
--- a/control-plane/agents/src/bin/core/controller/resources/operations_helper.rs
+++ b/control-plane/agents/src/bin/core/controller/resources/operations_helper.rs
@@ -51,9 +51,9 @@ enum SpecError {
 }
 
 /// What to do when creation fails.
+#[derive(Debug)]
 pub(crate) enum OnCreateFail {
     /// Leave object as `Creating`, could allow for frontend retries.
-    #[allow(unused)]
     LeaveAsIs,
     /// When frontend retries don't make sense, set it to deleting so we can clean-up.
     SetDeleting,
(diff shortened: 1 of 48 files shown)


tiagolobocastro:
Resolves https://github.com/openebs/mayastor/issues/1772

control-plane/agents/src/bin/core/controller/resources/operations_helper.rs
```diff
@@ -68,7 +68,28 @@ impl OnCreateFail {
     pub(crate) fn eeinval_delete<O>(result: &Result<O, SvcError>) -> Self {
         match result {
             Err(error) if error.tonic_code() == tonic::Code::InvalidArgument => Self::Delete,
-            Err(error) if error.tonic_code() == tonic::Code::NotFound => Self::Delete,
+            _ => Self::SetDeleting,
+        }
+    }
+    /// Map errors into `Self` for pool creation requests, specifically.
+    pub(crate) fn pool_create<O>(result: &Result<O, SvcError>) -> Self {
```
dsharma-dc: `on_pool_create_err()` may be?

dsharma-dc (approved):

control-plane/agents/src/bin/core/controller/resources/operations_helper.rs
tiagolobocastro: Done

tiagolobocastro:
bors try

tiagolobocastro:
bors try

Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

PR #886, opened 2024-11-07
feat(csi-driver): make csi-driver operations async

Abhinandan-Purkait:
- Change the mount, unmounts and nvme operations over to use spawn blocking.
- Remove async_stream for grpc over uds for controller and node.

PR #888, opened 2024-11-23
feat(csi driver): make mount and unmounts non-blocking

Abhinandan-Purkait:
- Adds the mount utils to frontend and backend sys mount calls via a child process
- Makes the currently mount and umount calls async

### DIFF
--- a/control-plane/csi-driver/src/bin/controller/server.rs
+++ b/control-plane/csi-driver/src/bin/controller/server.rs
@@ -1,71 +1,10 @@
 use crate::{controller::CsiControllerSvc, identity::CsiIdentitySvc};
 use rpc::csi::{controller_server::ControllerServer, identity_server::IdentityServer};
 
-use futures::TryFutureExt;
-use std::{
-    fs,
-    io::ErrorKind,
-    ops::Add,
-    pin::Pin,
-    sync::Arc,
-    task::{Context, Poll},
-};
-use tokio::{
-    io::{AsyncRead, AsyncWrite, ReadBuf},
-    net::UnixListener,
-};
-use tonic::transport::{server::Connected, Server};
+use std::{fs, io::ErrorKind, ops::Add};
+use tokio::net::UnixListener;
+use tonic::{codegen::tokio_stream::wrappers::UnixListenerStream, transport::Server};
 use tracing::{debug, error, info};
-
-#[derive(Debug)]
-struct UnixStream(tokio::net::UnixStream);
-
-impl Connected for UnixStream {
-    type ConnectInfo = UdsConnectInfo;
-
-    fn connect_info(&self) -> Self::ConnectInfo {
-        UdsConnectInfo {
-            peer_addr: self.0.peer_addr().ok().map(Arc::new),
-            peer_cred: self.0.peer_cred().ok(),
-        }
-    }
-}
-
-// Not sure why we need the inner fields, probably worth checking if we can remove them.
-#[derive(Clone, Debug)]
-#[allow(unused)]
-struct UdsConnectInfo {
-    peer_addr: Option<Arc<tokio::net::unix::SocketAddr>>,
-    peer_cred: Option<tokio::net::unix::UCred>,
-}
-
-impl AsyncRead for UnixStream {
-    fn poll_read(
-        mut self: Pin<&mut Self>,
-        cx: &mut Context<'_>,
-        buf: &mut ReadBuf<'_>,
-    ) -> Poll<std::io::Result<()>> {
-        Pin::new(&mut self.0).poll_read(cx, buf)
-    }
-}
-
-impl AsyncWrite for UnixStream {
-    fn poll_write(
-        mut self: Pin<&mut Self>,
-        cx: &mut Context<'_>,
-        buf: &[u8],
-    ) -> Poll<std::io::Result<usize>> {
-        Pin::new(&mut self.0).poll_write(cx, buf)
-    }
-
-    fn poll_flush(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<std::io::Result<()>> {
-        Pin::new(&mut self.0).poll_flush(cx)
-    }
-
-    fn poll_shutdown(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<std::io::Result<()>> {
-        Pin::new(&mut self.0).poll_shutdown(cx)
-    }
-}
 
 pub(super) struct CsiServer {}
 impl CsiServer {
@@ -91,6 +30,7 @@ impl CsiServer {
 
         let incoming = {
             let uds = UnixListener::bind(csi_socket)?;
+            info!("CSI plugin bound to {}", csi_socket);
 
             // Change permissions on CSI socket to allow non-privileged clients to access it
             // to simplify testing.
@@ -103,12 +43,7 @@ impl CsiServer {
                 debug!("Successfully changed file permissions for CSI socket");
             }
 
-            async_stream::stream! {
-                loop {
-                    let item = uds.accept().map_ok(|(st, _)| UnixStream(st)).await;
-                    yield item;
-                }
-            }
+            UnixListenerStream::new(uds)
         };
 
         let cfg = crate::CsiControllerConfig::get_config();
--- a/control-plane/csi-driver/src/bin/node/dev/nvmf.rs
+++ b/control-plane/csi-driver/src/bin/node/dev/nvmf.rs
@@ -13,6 +13,7 @@ use csi_driver::PublishParams;
 use glob::glob;
 use nvmeadm::nvmf_subsystem::Subsystem;
 use regex::Regex;
+use tokio::task::spawn_blocking;
 use tracing::warn;
 use udev::{Device, Enumerator};
 use url::Url;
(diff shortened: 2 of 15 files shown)


Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
Testing:

1. 150 Volumes creation followed deletion all at once.
2. Randomly killed(force) the csi node pods, while operations were in flight.

Observations:
1. All operations completed eventually.
2. No stale entries found eventually.
4. Did not see zombie processes :-;

control-plane/csi-driver/src/bin/node/mount_utils/bin/mount_utils.rs
```diff
@@ -0,0 +1,50 @@
...
+    fstype: Option<String>,
+) -> Result<(), MountError> {
+    let blocking_task = tokio::task::spawn_blocking(move || {
+        let mut mount = Mount::builder();
+
+        if let Some(data_str) = data.as_deref().filter(|d| !d.is_empty()) {
+            mount = mount.data(data_str);
+        }
+
+        mount
```
tiagolobocastro, line 20: wonder if we need this at all anymore, if it's not safe to call ioctl directly, maybe we can simply call the mount binary? Though it might be safer to do this first until we figure out how to migrate from this to mount/unmount arguments properly.

control-plane/csi-driver/src/bin/node/mount_utils/bin/mount_utils.rs
```diff
@@ -0,0 +1,50 @@
+use crate::error::MountError;
+
+use sys_mount::{FilesystemType, Mount, MountFlags, UnmountFlags};
+
+/// Calls sys mount's mount in a spawn blocking.
+pub(crate) async fn mount(
```
tiagolobocastro, line 6: If the mount/unmount calls get stuck due to the device being in a bad state, how are you ensuring you don't keep stacking up new processes trying to the same operation?

control-plane/csi-driver/src/bin/node/mount_utils/bin/mount_utils.rs
Abhinandan-Purkait: With using binaries directly, the errors become a bit vague, but anyhow yeah we can do that. For now kept it as it was to keep the same behaviour.

control-plane/csi-driver/src/bin/node/mount_utils/bin/mount_utils.rs
Abhinandan-Purkait: In my testing I did not see any process stacking up.

Abhinandan-Purkait:
bors try

control-plane/csi-driver/src/bin/node/mount_utils/bin/mount_utils.rs
tiagolobocastro: Don't the ioctl get stuck? Which hangs the server. I thought this was the issue causing the the lockups.

control-plane/csi-driver/src/bin/node/mount_utils/bin/flag.rs
dsharma-dc, line 1: Why don't we use sys_mount crate instead directly?

Abhinandan-Purkait: We are creating a process rather than just doing the spawn blocking directly.

control-plane/csi-driver/src/bin/node/mount_utils/bin/mount_utils.rs
dsharma-dc: When the problem happens(repro), have you checked the host's dmesg if it shows any softlockup or oops message from kernel? In that case the host is already unreliable.

Abhinandan-Purkait: No, I haven't seen that, also seems like the process is not blocking the CPU core in which case the softlockup shows up probably, I can see other tasks getting scheduled on the core where the ioctl process is running.

Abhinandan-Purkait:
TODO: Split into meaningful commits

control-plane/csi-driver/src/bin/node/main_.rs
```diff
@@ -440,17 +474,13 @@ impl CsiServer {
...
-                    yield item;
-                }
-            }
+            UnixListenerStream::new(uds)
         };
 
         let node = Node::new(node_name.into(), node_selector, probe_filesystems());
         Ok(async move {
             Server::builder()
+                .timeout(std::time::Duration::from_secs(30))
```
Abhinandan-Purkait: Should this be configurable?

control-plane/agents/src/bin/core/tests/controller/mod.rs
```diff
@@ -290,7 +291,7 @@ async fn etcd_pagination() {
 
     cluster.restart_core().await;
     cluster
-        .volume_service_liveness(None)
+        .volume_service_liveness(Some(TimeoutOptions::new().with_max_retries(10)))
```
tiagolobocastro: you need to rebase, I already fixed this I think

control-plane/csi-driver/src/bin/controller/server.rs
```diff
@@ -103,12 +43,7 @@ impl CsiServer {
                 debug!("Successfully changed file permissions for CSI socket");
             }
 
-            async_stream::stream! {
-                loop {
-                    let item = uds.accept().map_ok(|(st, _)| UnixStream(st)).await;
-                    yield item;
-                }
-            }
+            UnixListenerStream::new(uds)
```
tiagolobocastro, line 46: :+1:

control-plane/csi-driver/src/bin/node/dev/nvmf.rs
```diff
@@ -192,12 +193,17 @@ impl Attach for NvmfAttach {
...
                     .build()?;
-                match ca.connect() {
-                    // Should we remove this arm?
-                    Err(NvmeError::ConnectInProgress) => Ok(()),
-                    Err(err) => Err(err.into()),
-                    Ok(_) => Ok(()),
-                }
+
+                spawn_blocking(move || {
+                    match ca.connect() {
```
tiagolobocastro, line 212: Now that this is not blocking, what happens if the future is dropped but the connect/disconnect ioctls are still in progress.
Maybe we should issue these by non-cancellable futures (worker-threads)? Otherwise each retry will stack them up (probably can be another PR)

control-plane/csi-driver/src/bin/node/main_.rs
```diff
@@ -255,6 +202,61 @@ pub(super) async fn main() -> anyhow::Result<()> {
                         .help("Uuid of the volume to unfreeze")
                 )
         )
+        .subcommand(
```
tiagolobocastro, line 205: Do we still need the subcommands? Or can we just use the code directly with spawn_blocking?

control-plane/csi-driver/src/bin/node/main_.rs
tiagolobocastro: As discussed, we can even not have this at all if not required

control-plane/csi-driver/src/bin/node/nodeplugin_grpc.rs
```diff
@@ -76,6 +76,7 @@ impl NodePluginGrpcServer {
             endpoint
         );
         Server::builder()
+            .timeout(std::time::Duration::from_secs(30))
```
tiagolobocastro: Be mindful that the gRPC themselves have timeouts, setting this value needs to be done carefully.
I suggest not doing this until we figure out the correct value.

```suggestion
```

control-plane/csi-driver/src/bin/node/dev/nvmf.rs
dsharma-dc: As I understood, tasks spawned by `spawn_blocking` are already non-cancellable if they have started. We can choose to wait on `JoinHandle` as we are doing here.

control-plane/csi-driver/src/bin/node/main_.rs
Abhinandan-Purkait: Yeah yeah we can now since we have decided to live with the fact the manual kill of the pod is the way out. I already have the changes on the other PR. Both of them work. Please review the other one if you think we should do that one. cc @dsharma-dc

control-plane/csi-driver/src/bin/node/nodeplugin_grpc.rs
Abhinandan-Purkait: You mean the client side timeout?

control-plane/csi-driver/src/bin/node/dev/nvmf.rs
Abhinandan-Purkait: What happens if the join handle is dropped?

control-plane/agents/src/bin/core/tests/controller/mod.rs
Abhinandan-Purkait: Ahh, I see the fix is different from what I did, I missed this, will remove.

control-plane/csi-driver/src/bin/node/main_.rs
Abhinandan-Purkait: Yeah we can, although kubelet cancels it in less than 2mins either way as we have seen, so that should be good enough?

Abhinandan-Purkait: Removed it.

control-plane/agents/src/bin/core/tests/controller/mod.rs
Abhinandan-Purkait: Done.

control-plane/csi-driver/src/bin/node/nodeplugin_grpc.rs

Abhinandan-Purkait:
bors try

control-plane/csi-driver/src/bin/node/nodeplugin_grpc.rs
tiagolobocastro: > You mean the client side timeout?

Yes

control-plane/csi-driver/src/bin/node/dev/nvmf.rs
tiagolobocastro: > As I understood, tasks spawned by `spawn_blocking` are already non-cancellable if they have started. We can choose to wait on `JoinHandle` as we are doing here.

The ca.connect() will be non-cancellable because it's blocking and stuck in ioctl anyway but the code which is calling this is cancellable and will potentially re-run again, and schedule another cannot, or potentially a disconnect.

Ah, we could ref-count the volume guard in the sawn_blocking?

Abhinandan-Purkait:
Closing this for now. Thanks Overruled by #886

PR #889, opened 2024-11-25
fix(volume): don't allow unpublish from non-frontend node

dsharma-dc:
In case application and volume target have already moved to a different node and there is a lagging ControllerUnpublish request that comes later, then this unpublish call can delete the newly created target, which is undesirable and leads to volume staging failures.

### DIFF
--- a/control-plane/csi-driver/src/bin/controller/controller.rs
+++ b/control-plane/csi-driver/src/bin/controller/controller.rs
@@ -605,15 +605,18 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
             return Ok(Response::new(ControllerUnpublishVolumeResponse {}));
         }
 
-        // Do forced volume upublish as Kubernetes already detached the volume.
+        // Do forced volume unpublish as Kubernetes already detached the volume.
         RestApiClient::get_client()
-            .unpublish_volume(&volume_uuid, true)
+            .unpublish_volume(&volume_uuid, true, Some(args.node_id.as_str()))
             .await
-            .map_err(|e| {
-                Status::not_found(format!(
+            .map_err(|e| match e {
+                ApiClientError::NotAcceptable(_) => {
+                    Status::ok("Ignoring failure on unpublish due to mismatched host")
+                }
+                _ => Status::not_found(format!(
                     "Failed to unpublish volume {}, error = {:?}",
                     &args.volume_id, e
-                ))
+                )),
             })?;
 
         debug!("Volume {} successfully unpublished", args.volume_id);
--- a/control-plane/grpc/proto/v1/volume/volume.proto
+++ b/control-plane/grpc/proto/v1/volume/volume.proto
@@ -323,6 +323,8 @@ message UnpublishVolumeRequest {
   // the nexus. Note: this option should be used only when we know the node will not become
   // accessible again and it is safe to do so.
   bool force = 2;
+  // frontend host requesting for volume unpublish.
+  repeated string frontend_nodes = 3;
 }
 
 // Share Volume request
--- a/control-plane/grpc/src/operations/volume/traits.rs
+++ b/control-plane/grpc/src/operations/volume/traits.rs
@@ -1618,6 +1618,8 @@ pub trait UnpublishVolumeInfo: Send + Sync + std::fmt::Debug {
     fn uuid(&self) -> VolumeId;
     /// Force unpublish
     fn force(&self) -> bool;
+    /// Frontend node requesting unpublish.
+    fn frontend_nodes(&self) -> Vec<String>;
 }
 
 impl UnpublishVolumeInfo for UnpublishVolume {
@@ -1627,6 +1629,10 @@ impl UnpublishVolumeInfo for UnpublishVolume {
 
     fn force(&self) -> bool {
         self.force()
+    }
+
+    fn frontend_nodes(&self) -> Vec<String> {
+        self.frontend_nodes.clone()
     }
 }
 
@@ -1644,6 +1650,10 @@ impl UnpublishVolumeInfo for ValidatedUnpublishVolumeRequest {
     fn force(&self) -> bool {
         self.inner.force
     }
+
+    fn frontend_nodes(&self) -> Vec<String> {
+        self.inner.frontend_nodes.clone()
+    }
 }
 
 impl ValidateRequestTypes for UnpublishVolumeRequest {
@@ -1658,7 +1668,7 @@ impl ValidateRequestTypes for UnpublishVolumeRequest {
 
 impl From<&dyn UnpublishVolumeInfo> for UnpublishVolume {
     fn from(data: &dyn UnpublishVolumeInfo) -> Self {
-        UnpublishVolume::new(&data.uuid(), data.force())
+        UnpublishVolume::new(&data.uuid(), data.force(), data.frontend_nodes())
     }
 }
 
@@ -1667,6 +1677,7 @@ impl From<&dyn UnpublishVolumeInfo> for UnpublishVolumeRequest {
         Self {
             uuid: Some(data.uuid().to_string()),
             force: data.force(),
+            frontend_nodes: data.frontend_nodes().iter().map(|n| n.into()).collect(),
         }
     }
 }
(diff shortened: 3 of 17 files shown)


control-plane/csi-driver/src/bin/controller/controller.rs
```diff
@@ -605,7 +605,17 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
             return Ok(Response::new(ControllerUnpublishVolumeResponse {}));
         }
 
-        // Do forced volume upublish as Kubernetes already detached the volume.
+        // no-op in case the node where unpublish is coming from isn't the current frontend node.
+        // This might happen in case of out of order/delayed unpublish call after application is
+        // is already published elsewhere.
+        if let Some(tgt) = volume.spec.target {
```
tiagolobocastro: I think this needs to be done at the core-agent unpublish, otherwise it's racy.
We already do similar thing for the republish: https://github.com/openebs/mayastor-control-plane/blob/dd21e4e276d2d8701d3935c2299219a234ccd955/control-plane/agents/src/bin/core/volume/operations.rs#L467

dsharma-dc: Because we have resource locks in core-agent? Then I'd say yes that'd be better.

Abhinandan-Purkait: I think both the places?

tiagolobocastro: > Because we have resource locks in core-agent? Then I'd say yes that'd be better.

Yes

> I think both the places?

Just core-agent, otherwise here it's still racy

dsharma-dc: Made the changes now.

control-plane/grpc/proto/v1/volume/volume.proto
```diff
@@ -323,6 +323,8 @@ message UnpublishVolumeRequest {
   // the nexus. Note: this option should be used only when we know the node will not become
   // accessible again and it is safe to do so.
   bool force = 2;
+  // frontend host requesting for volume unpublish.
+  optional string frontend_host = 3;
```
tiagolobocastro: In-keeping with publish
```suggestion
  repeated string frontend_nodes = 3;
```

control-plane/grpc/src/operations/volume/traits.rs
```diff
@@ -1618,6 +1618,8 @@ pub trait UnpublishVolumeInfo: Send + Sync + std::fmt::Debug {
     fn uuid(&self) -> VolumeId;
     /// Force unpublish
     fn force(&self) -> bool;
+    /// Frontend node requesting unpublish.
+    fn frontend_host(&self) -> Option<NodeId>;
```
tiagolobocastro:
```suggestion
    fn frontend_nodes(&self) -> Vec<NodeId>;
```

tests/bdd/features/csi/controller/controller.feature
```diff
@@ -112,4 +112,10 @@ Scenario: list existing volumes with pagination max entries set to 0
     Given 2 existing volumes
     When a ListVolumesRequest is sent to CSI controller with max_entries set to 0
     Then all volumes should be returned
-    And the next token should be empty
\ No newline at end of file
+    And the next token should be empty
+
+Scenario: unpublish volume from non-frontend node
```
tiagolobocastro, line 117: We should cover the returned errors when the nodes don't match

control-plane/csi-driver/src/bin/controller/controller.rs
```diff
@@ -605,15 +605,18 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
...
-        // Do forced volume upublish as Kubernetes already detached the volume.
+        // Do forced volume unpublish as Kubernetes already detached the volume.
         RestApiClient::get_client()
-            .unpublish_volume(&volume_uuid, true)
+            .unpublish_volume(&volume_uuid, true, Some(args.node_id.as_str()))
             .await
-            .map_err(|e| {
-                Status::not_found(format!(
+            .map_err(|e| match e {
+                ApiClientError::PreconditionFailed(_) => {
```
tiagolobocastro: this feels somewhat wrong, what if target was not deleted because of some other precondition error.
Maybe we need a different error here that is clearer? Something more similar to NOT_FOUND perhaps? Given that the target is not available for this node.

dsharma-dc: How about `NotAcceptable`? I intended to do that earlier

tiagolobocastro: Yeah that seems good. wdyt @Abhinandan-Purkait ?

Abhinandan-Purkait: Yep. Although there is one more mapping of not acceptable but looking at other options this sounds reasonable.

tiagolobocastro: We can also use the error body for additional info if we need to do additional mapping

tiagolobocastro: And I guess this is another option for PreConditionFailed too, keep this error code but check the body for the internal error code to be this target mismatch!

dsharma-dc: I'll make it `NotAcceptable` and we can extend that if required using error msg body if any such more scenario comes. I guess likelihood of future NotAcceptable is less compared to precondition error so we should be ok with `NotAcceptable`

control-plane/grpc/proto/v1/volume/volume.proto
dsharma-dc: > In-keeping with publish

should it be `optional repeated string frontend_nodes` then ?
Unpublish logically can't have a vector(multiple host) as per the semantics.

tests/bdd/features/csi/controller/controller.feature
dsharma-dc: Leaving this one as-is, since here we do csi calls which will see success only. Will add a new test in rest/tests to validate error codes.

dsharma-dc:
bors try

control-plane/grpc/proto/v1/volume/volume.proto
tiagolobocastro: yeah repeated optional string I guess

tiagolobocastro:
@dsharma-dc did you push? I still see the repeated nodes changes missing

dsharma-dc:
> @dsharma-dc did you push? I still see the repeated nodes changes missing

Not yet. I had a question posted for that which is now answered. Will make the changes.

control-plane/grpc/proto/v1/volume/volume.proto
tiagolobocastro: > Unpublish logically can't have a vector(multiple host) as per the semantics.

On the openapi we can leave only one, it's just the gRPC which allows multiple nodes on the publish

dsharma-dc: ok, `optional repeated` isn't allowed. kept `repeated string `. Done.

dsharma-dc:
bors try

tests/bdd/features/csi/controller/controller.feature
```diff
@@ -112,4 +112,10 @@ Scenario: list existing volumes with pagination max entries set to 0
...
+    Given a 2 replica volume published on a node
+    When the ControllerUnpublishVolume request arrives from a non-frontend node
+    Then nvmf target which exposes the volume should not be destroyed
+    And volume should remain as published
```
Abhinandan-Purkait: Blank line

Abhinandan-Purkait (approved):

tests/bdd/features/csi/controller/test_controller.py
```diff
@@ -999,6 +1031,12 @@ def check_unpublish_not_existing_volume(unpublish_not_existing_volume):
     ), "Volume unpuplishing succeeded with unexpected response"
 
 
+@then("volume should remain as published")
+def volume_remains_published():
+    """volume should remain as published."""
+    pass
```
tiagolobocastro, line 1037: need to check it's published here?

tiagolobocastro (approved):

tests/bdd/features/csi/controller/test_controller.py
dsharma-dc: The previous step is checking that the volume target isn't deleted and remains available on the original node , so simply `pass`ing here.

dsharma-dc:
bors merge

PR #892, opened 2024-11-29
fix(csi-node): handle devices for existing subsystems

tiagolobocastro:
When a device is broken but the volume is not unstaged properly, the subsytem ends up lingering in the system.
If the device is staged again, the device is assined a different namespace id. Change the regex which is used in the device parameter fo account for this. Also extend the existing csi-node BDD testing.

### DIFF
--- a/control-plane/csi-driver/src/bin/node/dev/nvmf.rs
+++ b/control-plane/csi-driver/src/bin/node/dev/nvmf.rs
@@ -1,13 +1,12 @@
+use nvmeadm::{
+    error::NvmeError,
+    nvmf_discovery::{disconnect, ConnectArgsBuilder, TrType},
+};
 use std::{
     collections::HashMap,
     convert::{From, TryFrom},
     path::Path,
     str::FromStr,
-};
-
-use nvmeadm::{
-    error::NvmeError,
-    nvmf_discovery::{disconnect, ConnectArgsBuilder, TrType},
 };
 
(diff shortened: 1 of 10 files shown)


Abhinandan-Purkait (approved):

sinhaashish (approved):

tiagolobocastro:
bors merge

tiagolobocastro:
bors try

control-plane/csi-driver/src/bin/node/dev/nvmf.rs
```diff
@@ -302,6 +314,48 @@ impl Detach for NvmfDetach {
...
+    let Some(devname) = match_nvmf_device(device, key) else {
+        return Ok(None);
+    };
+
+    let glob = glob(&block_dev_q(device)?).unwrap();
+    if !glob.into_iter().any(|glob_result| glob_result.is_ok()) {
+        if warn_bad.load(std::sync::atomic::Ordering::Relaxed) {
+            let name = device.sysname().to_string_lossy();
+            warn!("Block device {name} for volume {key} has no controller!");
+            // todo: shoot-down the stale mounts?
```
tiagolobocastro, line 350: @Abhinandan-Purkait we could potentially shoot down the stale mounts here, though this is probably not needed as the controller publish incoming changes would be doing this right?

tiagolobocastro:
bors merge

PR #894, opened 2024-12-03
feat(rest): add health API for readiness and liveness probes

niladrih:


niladrih:
Should I also add a readiness endpoint?

control-plane/rest/service/src/main.rs
```diff
@@ -30,6 +36,10 @@ pub(crate) struct CliArgs {
     #[clap(long)]
     http: Option<String>,
 
+    /// The bind address for the health REST server.
+    #[clap(long, default_value = "[::]:9091")]
+    health_endpoint: SocketAddr,
```
tiagolobocastro: might want to consider separating into ip and port

control-plane/rest/service/src/main.rs
```diff
@@ -241,18 +255,30 @@ async fn main() -> anyhow::Result<()> {
...
     } else {
-        server
+        main_server
     }
-    .workers(workers(&CliArgs::args()))
-    .run()
-    .await;
+    .workers(workers(&CliArgs::args()));
+
+    let health_server = HttpServer::new(move || {
```
tiagolobocastro: can't we just use the same server?

niladrih: I found this: https://github.com/linkerd/linkerd2/issues/7560
WDYT?

control-plane/rest/service/src/main.rs
niladrih: Why not accept it together?

control-plane/rest/service/src/main.rs
niladrih: Also kubernetes does it with a different API server.

control-plane/rest/service/src/main.rs
tiagolobocastro: ipv6 and ipv4 have different format for ip:port
If we need to expose this via helm chart it becomes more difficult. If this is just for local debugging then like this is fine.

control-plane/rest/service/src/main.rs
tiagolobocastro: A separate server would again consume more memory, which is already a problem.
Also, what happens if the other server is stuck, but the health server is fine. Then we'd be returning potentially incorrect health information?
I'm not sure there's any clear-cut benefit as of now.

control-plane/rest/service/src/main.rs
niladrih: This is no longer needed as the health service is not a separate endpoint anymore.

control-plane/rest/service/src/main.rs
niladrih: I've merged the health service into the main server. I've left it out of the openapi spec, as it logically is separate from the Mayastor APIs, and are more of meta-APIs.

control-plane/rest/service/src/health/handlers.rs
```diff
@@ -0,0 +1,28 @@
...
+/// Liveness probe check. Failure will result in Pod restart. 200 on success.
+#[get("/live")]
+async fn liveness(_cached_core_state: Data<CachedCoreState>) -> impl Responder {
+    HttpResponse::Ok()
+        .content_type("text/plain; charset=utf-8")
+        .insert_header(("X-Content-Type-Options", "nosniff"))
+        .body("live")
+}
+
+/// Readiness probe check. Failure will result in removal of Container from Kubernetes service
```
tiagolobocastro, line 13: Isn't the container removal done by the liveness rather than readiness?

control-plane/rest/service/src/health/handlers.rs
```diff
@@ -0,0 +1,28 @@
...
+/// target pool. 200 on success, 503 on failure.
+#[get("/ready")]
+async fn readiness(cached_core_state: Data<CachedCoreState>) -> HttpResponse {
+    if cached_core_state.get_or_assume_unavailable().await {
```
tiagolobocastro, line 17: Is this needed? What if there's an issue with the core agent, would this then result in removal of the rest?

control-plane/rest/service/src/main.rs
```diff
@@ -34,6 +43,10 @@ pub(crate) struct CliArgs {
     #[clap(long, short = 'z', default_value = DEFAULT_GRPC_CLIENT_ADDR)]
     core_grpc: Uri,
 
+    /// Set the maximum frequency of probing the agent-core for a liveness check.
+    #[arg(value_parser = humantime::parse_duration, default_value = "2m")]
```
tiagolobocastro:
```suggestion
    #[arg(long, value_parser = humantime::parse_duration, default_value = "2m")]
```

control-plane/rest/service/src/health/handlers.rs
niladrih: Removal is not a delete/termination in this case. It'll not have Endpoint resources, so it won't participate it the service pool. It'll not be a part of the hypothetical round robin amonst other rest instances.
https://kubernetes.io/docs/concepts/configuration/liveness-readiness-startup-probes/#readiness-probe
>If the readiness probe returns a failed state, Kubernetes removes the pod from all matching service endpoints.

control-plane/rest/service/src/health/handlers.rs
niladrih: It's a readiness failure, that would just mean that REST won't serve requests sent to the kubernetes service endpoint for mayastor-rest. Other instances (if any) of the rest would serve it. The container itself doesn't get deleted, etc. just sits around doing nothing.

control-plane/rest/service/src/main.rs
niladrih: Oops!

tiagolobocastro (approved):

control-plane/rest/service/src/health/handlers.rs
Abhinandan-Purkait: So the deletion part is done only on liveness failure?

niladrih:
@tiagolobocastro I'll work on a pytest and try to add it into this PR itself.

Abhinandan-Purkait (approved):
LGTM

control-plane/rest/service/src/health/handlers.rs
niladrih: Yes, the pod is restarted on a liveness failure.

Abhinandan-Purkait: What error do we get if let's say core-agent is down, i.e readiness failure?

niladrih: The `1/1` on the `kubectl get po -n mayastor` turns to `0/1`. That's all. We might see some logs on the kubelet. I'm going to have to check for them. I'll post here if I manage to collect them.

niladrih:
@tiagolobocastro @Abhinandan-Purkait I've added a pytest for the readiness probe. The liveness probe is trivial and IMO doesn't need tests.

niladrih:
bors try

niladrih:
bors try

niladrih:
bors try

niladrih:
bors try

abhilashshetty04 (approved):

tests/bdd/features/health_probes/readiness_probe.feature
```diff
@@ -0,0 +1,12 @@
+Feature: Readiness Probe
+
+  Background:
+    Given a running agent-core service
+    And a running REST service with "--core-health-freq" set to "10s"
```
tiagolobocastro: Please use milliseconds here, no need to wait so long for this test

niladrih: Okay.

tests/bdd/features/health_probes/test_readiness_probe.py
```diff
@@ -0,0 +1,98 @@
...
+    """after a delay of 5 seconds the REST service returns 503 for /ready endpoint for the following 12 seconds."""
+    time.sleep(5)
+
+    start_time = time.time()
+    while time.time() - start_time < 12:
+        response = http_get(READINESS_API_ENDPOINT)
+        if response.status_code != 503:
+            raise ValueError(
+                "Expected Readiness probe to return 503 for this duration of 12 seconds"
+            )
```
Abhinandan-Purkait: NIT: Should we rather use the retry with a gap of 5 seconds let's say instead of waiting for exact time, just in case things are little slow and it becomes flaky.

Abhinandan-Purkait (approved):

tests/bdd/features/health_probes/test_readiness_probe.py
niladrih: On the final check, if it returns something other than 503, it should fail.
We are waiting for the entire frequency duration before checking, so it should most definitely have updated itself, else it's unexpected behaviour.

niladrih:
bors merge

tests/bdd/features/health_probes/test_readiness_probe.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+@then(
+    "after a delay of 5 seconds the REST service returns 503 for /ready endpoint for the following 12 milliseconds"
+)
+def after_a_delay_of_5_seconds_the_rest_service_returns_503_for_ready_endpoint_for_the_following_12_milliseconds(
+    setup,
+):
+    """after a delay of 5 seconds the REST service returns 503 for /ready endpoint for the following 12 milliseconds."""
+    time.sleep(5)
```
tiagolobocastro: This is still in seconds?

niladrih:
bors cancel

niladrih:
bors merge

niladrih:
@tiagolobocastro -- I've trimmed down on the wait times on this test, however, any lower than this and the test becomes flaky. WDYT?

tiagolobocastro:
> @tiagolobocastro -- I've trimmed down on the wait times on this test, however, any lower than this and the test becomes flaky. WDYT?

I would say try 100's of milliseconds.
Most tests use that sort of magnitude.

If it's still flaky then fine keep as is.

niladrih:
> > @tiagolobocastro -- I've trimmed down on the wait times on this test, however, any lower than this and the test becomes flaky. WDYT?
> 
> I would say try 100's of milliseconds.
> Most tests use that sort of magnitude.
> 
> If it's still flaky then fine keep as is.

I've tried 10s, 100s, quite a few... this is lowest it goes and stays fail free.

niladrih:
bors merge

PR #901, opened 2024-12-09
Cherry pick csi-trace and csi-node uri fixes to develop

tiagolobocastro:
chore(bors): merge pull request #903

    903: fix(csi/stage): fetch uri via REST r=tiagolobocastro a=tiagolobocastro

    On a NodeStage call, it's possible that the publish_context URI is out of date. This can happen when the volume has been moved to another node, and the app pod is pinned to a node and the node restarts.
    See Mayastor Issue 1781 for more details.

    For this fix, we add a new option to enable rest client for the csi-node. Ideally we'd like to strictly adhere to CSI spec, and avoid doing any mayastor specific operations (effectively being a mostly generic nvme-connect csi-driver but the immutability of the publish_context makes this a bit difficult.

    Anyways, we add a new flag --enable-rest which is optional, thus still allowing us to run without this layer.
    This will be enabled by default on the helm chart.

    We also further check for the Nexus Online/Degraded state, which should help avoid a bunch of nvme connect errors in the kernel.

    With this, we can also improve a few things down the line, such as ensuring resize before publish, etc... but we should take these 1 at a time and not suddendly do everything via rest...

    Co-authored-by: Tiago Castro <tiagolobocastro@gmail.com>

---

    fix(csi-driver): trace was mistakenly added as info

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -20,14 +20,14 @@ repos:
     -   id: rust-lint
         name: Rust lint
         description: Run cargo clippy on files included in the commit. clippy should be installed before-hand.
-        entry: nix-shell --pure --run 'cargo-clippy --all --all-targets -- -D warnings'
+        entry: nix-shell --pure --run './scripts/rust/linter.sh clippy'
         pass_filenames: false
         types: [file, rust]
         language: system
     -   id: rust-style
         name: Rust style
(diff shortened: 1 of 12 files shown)


sinhaashish (approved):

niladrih (approved):

tiagolobocastro:
bors merge

PR #903, opened 2024-12-10
fix(csi/stage): fetch uri via REST

tiagolobocastro:
On a NodeStage call, it's possible that the publish_context URI is out of date. This can happen when the volume has been moved to another node, and the app pod is pinned to a node and the node restarts.
See Mayastor Issue 1781 for more details.

For this fix, we add a new option to enable rest client for the csi-node. Ideally we'd like to strictly adhere to CSI spec, and avoid doing any mayastor specific operations (effectively being a mostly generic nvme-connect csi-driver but the immutability of the publish_context makes this a bit difficult.

Anyways, we add a new flag --enable-rest which is optional, thus still allowing us to run without this layer.
This will be enabled by default on the helm chart.

We also further check for the Nexus Online/Degraded state, which should help avoid a bunch of nvme connect errors in the kernel.

With this, we can also improve a few things down the line, such as ensuring resize before publish, etc... but we should take these 1 at a time and not suddendly do everything via rest...

### DIFF
--- a/control-plane/csi-driver/src/bin/node/main_.rs
+++ b/control-plane/csi-driver/src/bin/node/main_.rs
@@ -20,7 +20,7 @@ use grpc::csi_node_nvme::nvme_operations_server::NvmeOperationsServer;
 use stor_port::platform;
 use utils::tracing_telemetry::{FmtLayer, FmtStyle};
 
-use crate::client::AppNodesClientWrapper;
+use crate::client::{AppNodesClientWrapper, VolumesClientWrapper};
 use clap::Arg;
 use serde_json::json;
 use std::{
@@ -32,8 +32,7 @@ use std::{
     str::FromStr,
 };
 use tokio::net::UnixListener;
-use tonic::codegen::tokio_stream::wrappers::UnixListenerStream;
-use tonic::transport::Server;
+use tonic::{codegen::tokio_stream::wrappers::UnixListenerStream, transport::Server};
 use tracing::{debug, error, info};
 
 const GRPC_PORT: u16 = 50051;
@@ -57,6 +56,14 @@ pub(super) async fn main() -> anyhow::Result<()> {
                 .value_name("BOOLEAN")
                 .requires("rest-endpoint")
                 .help("Enable registration of the csi node with the control plane")
+        )
+        .arg(
+            Arg::new("enable-rest")
+                .long("enable-rest")
+                .action(clap::ArgAction::SetTrue)
+                .value_name("BOOLEAN")
+                .requires("rest-endpoint")
+                .help("Enhance csi with added rest functionality")
         )
         .arg(
             Arg::new("csi-socket")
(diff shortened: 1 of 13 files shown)


control-plane/csi-driver/src/bin/node/main_.rs
```diff
@@ -58,6 +57,14 @@ pub(super) async fn main() -> anyhow::Result<()> {
...
                 .help("Enable registration of the csi node with the control plane")
         )
+        .arg(
+            Arg::new("enable-rest")
+                .long("enable-rest")
+                .action(clap::ArgAction::SetTrue)
+                .value_name("BOOLEAN")
+                .requires("rest-endpoint")
+                .help("Enhance csi with added rest functionality")
+        )
```
Abhinandan-Purkait, line 67: If someone disables this then they will hit the issue right, should it be an option then?

Abhinandan-Purkait (approved):

control-plane/csi-driver/src/bin/node/main_.rs
tiagolobocastro: Yeah, not sure.
This was more in case there's some issue with using rest and we wanted to disable it.

tiagolobocastro: Wdyt?

control-plane/csi-driver/src/bin/node/node.rs
```diff
@@ -73,6 +77,32 @@ impl Node {
...
+
+        let Some(client) = &self.vol_client else {
+            return Ok(ctx_uri.to_string());
+        };
+
+        let uri = client.volume_uri(volume_id).await?;
+        if &uri != ctx_uri {
+            tracing::warn!(volume.uri=%uri, ctx.uri=%ctx_uri, "Overriding URI volume context with URI from REST");
+        }
+        Ok(uri)
```
dsharma-dc, line 104: After we override, the staging will succeed but volumeattachment will still keep holding old uri?

tiagolobocastro: Yes it'll still be there

dsharma-dc: > Yes it'll still be there

If the target moves back to old node again at some point in future, with a different uri(say due to transport change or ip/interface change) does the volumeattachment get updated with new publish_context/uri? or the volume attachment is deleted and recreated so it picks up correct new uri?

tiagolobocastro: The VA attach data is immutable, it never gets updated :(
If the application itself moves, then we go controller unpublish/publish and then we get a new VA which might have correct uri for some time again

niladrih (approved):
Semantics and choice of syntax looks good to me :)

control-plane/csi-driver/src/bin/node/node.rs
dsharma-dc: ok. so IIUC, for scenario I described - when app never moves and if target comes back again to original node, the original VA will still be there with stale uri, which we'll override with this fix. But without this new option enabled, the staging will still fail in this scenario?

tiagolobocastro: Yes correct. User can manually get around it by deleting the VA.

dsharma-dc (approved):

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

PR #905, opened 2024-12-11
feat(csi-driver): add pre-publish hook to cleanup stale entries

Abhinandan-Purkait:
Scenario:
If an node undergoes a network partition and the application pods on it are moved by out-of-service taint then it can happen that the stale mount paths and the nvme controllers get left around, because CSI NodeUnstage was never called.
Once the node comes back and joins the cluster and the application pod comes back to the same node, it can happen that the stale nvme path gets live again causing the existing buffer data to be flushed potentially causing data corruption.

Changes:
This adds a pre-publish hook that can intercept the ControllerPublish and issue cleanup on the node if stale entries are detected.

### DIFF
--- a/control-plane/csi-driver/proto/node-service.proto
+++ b/control-plane/csi-driver/proto/node-service.proto
@@ -20,6 +20,8 @@ service NodePlugin {
   rpc UnfreezeFS (UnfreezeFSRequest) returns (UnfreezeFSReply) {}
   // Find the volume identified by the volume ID, and return volume information.
   rpc FindVolume (FindVolumeRequest) returns (FindVolumeReply) {}
+  // Force unstage the volume.
+  rpc ForceUnstageVolume (ForceUnstageVolumeRequest) returns (ForceUnstageVolumeReply) {}
 }
 
 enum VolumeType {
@@ -53,4 +55,12 @@ message FindVolumeRequest {
 message FindVolumeReply {
   optional VolumeType volume_type = 1;
   string device_path = 2; // the device path for the volume
+}
+
+// The request message containing ID of the volume to be force unstaged.
+message ForceUnstageVolumeRequest {
+  string volume_id = 1;
+}
+// Message for response on volume force unstage.
+message ForceUnstageVolumeReply {
 }
--- a/control-plane/csi-driver/src/bin/controller/controller.rs
+++ b/control-plane/csi-driver/src/bin/controller/controller.rs
@@ -4,21 +4,24 @@ use crate::{
 };
 use csi_driver::{
     context::{CreateParams, CreateSnapshotParams, PublishParams, QuiesceFsCandidate},
-    node::internal::{node_plugin_client::NodePluginClient, FreezeFsRequest, UnfreezeFsRequest},
+    node::internal::{
+        node_plugin_client::NodePluginClient, ForceUnstageVolumeRequest, FreezeFsRequest,
+        UnfreezeFsRequest,
+    },
 };
 use rpc::csi::{volume_content_source::Type, Topology as CsiTopology, *};
 use stor_port::types::v0::openapi::{
     models,
     models::{
-        AffinityGroup, LabelledTopology, NodeSpec, NodeStatus, Pool, PoolStatus, PoolTopology,
-        SpecStatus, Volume, VolumeShareProtocol,
+        AffinityGroup, AppNode, LabelledTopology, NodeSpec, NodeStatus, Pool, PoolStatus,
+        PoolTopology, SpecStatus, Volume, VolumeShareProtocol,
     },
 };
-use utils::{dsp_created_by_key, DSP_OPERATOR};
+use utils::{dsp_created_by_key, DEFAULT_REQ_TIMEOUT, DSP_OPERATOR};
 
 use regex::Regex;
-use std::{collections::HashMap, str::FromStr};
-use tonic::{Code, Request, Response, Status};
+use std::{collections::HashMap, str::FromStr, time::Duration};
+use tonic::{transport::Uri, Code, Request, Response, Status};
 use tracing::{debug, error, instrument, trace, warn};
 use uuid::Uuid;
 use volume_capability::AccessType;
@@ -31,6 +34,7 @@ const SNAPSHOT_NAME_PATTERN: &str =
 #[derive(Debug)]
 pub(crate) struct CsiControllerSvc {
     create_volume_limiter: std::sync::Arc<tokio::sync::Semaphore>,
+    force_unstage_volume: bool,
 }
 impl CsiControllerSvc {
     pub(crate) fn new(cfg: &CsiControllerConfig) -> Self {
@@ -38,6 +42,7 @@ impl CsiControllerSvc {
             create_volume_limiter: std::sync::Arc::new(tokio::sync::Semaphore::new(
                 cfg.create_volume_limit(),
             )),
+            force_unstage_volume: cfg.force_unstage_volume(),
         }
     }
     async fn create_volume_permit(&self) -> Result<tokio::sync::SemaphorePermit, tonic::Status> {
@@ -91,12 +96,30 @@ fn volume_app_node(volume: &Volume) -> Option<String> {
     }
 }
 
-#[tracing::instrument]
+/// Create a new endpoint that connects to the provided Uri.
+/// This endpoint has default connect and request timeouts.
+fn tonic_endpoint(endpoint: String) -> Result<tonic::transport::Endpoint, Status> {
+    let uri =
+        Uri::try_from(endpoint).map_err(|error| Status::invalid_argument(error.to_string()))?;
+
+    let timeout = humantime::parse_duration(DEFAULT_REQ_TIMEOUT).unwrap();
+    Ok(tonic::transport::Endpoint::from(uri)
+        .connect_timeout(timeout)
+        .timeout(std::time::Duration::from_secs(30))
+        .http2_keep_alive_interval(Duration::from_secs(5))
+        .keep_alive_timeout(Duration::from_secs(10))
+        .concurrency_limit(utils::DEFAULT_GRPC_CLIENT_CONCURRENCY))
+}
+
+#[tracing::instrument(err, skip_all)]
 async fn issue_fs_freeze(endpoint: String, volume_id: String) -> Result<(), Status> {
     trace!("Issuing fs freeze");
-    let mut client = NodePluginClient::connect(format!("http://{endpoint}"))
+    let channel = tonic_endpoint(format!("http://{endpoint}"))?
+        .connect()
         .await
-        .map_err(|error| Status::failed_precondition(error.to_string()))?;
+        .map_err(|error| Status::unavailable(error.to_string()))?;
+    let mut client = NodePluginClient::new(channel);
+
     match client
         .freeze_fs(Request::new(FreezeFsRequest {
             volume_id: volume_id.clone(),
@@ -112,12 +135,15 @@ async fn issue_fs_freeze(endpoint: String, volume_id: String) -> Result<(), Stat
     }
 }
 
-#[tracing::instrument]
+#[tracing::instrument(err, skip_all)]
 async fn issue_fs_unfreeze(endpoint: String, volume_id: String) -> Result<(), Status> {
     trace!("Issuing fs unfreeze");
-    let mut client = NodePluginClient::connect(format!("http://{endpoint}"))
+    let channel = tonic_endpoint(format!("http://{endpoint}"))?
+        .connect()
         .await
-        .map_err(|error| Status::failed_precondition(error.to_string()))?;
+        .map_err(|error| Status::unavailable(error.to_string()))?;
+    let mut client = NodePluginClient::new(channel);
+
     match client
         .unfreeze_fs(Request::new(UnfreezeFsRequest {
             volume_id: volume_id.clone(),
@@ -131,6 +157,28 @@ async fn issue_fs_unfreeze(endpoint: String, volume_id: String) -> Result<(), St
         }
         Err(error) => Err(error),
     }
+}
+
+#[tracing::instrument(err, skip_all)]
+async fn force_unstage(app_node: AppNode, volume_id: String) -> Result<(), Status> {
+    tracing::info!(
+        "Issuing cleanup for volume: {} to node {}, to endpoint {}",
+        volume_id,
+        app_node.id,
+        app_node.spec.endpoint
+    );
+    let channel = tonic_endpoint(format!("http://{}", app_node.spec.endpoint))?
+        .connect()
+        .await
+        .map_err(|error| Status::unavailable(error.to_string()))?;
+    let mut client = NodePluginClient::new(channel);
+
+    client
+        .force_unstage_volume(Request::new(ForceUnstageVolumeRequest {
+            volume_id: volume_id.clone(),
+        }))
+        .await
+        .map(|_| ())
 }
 
 /// Get share URI for existing volume object and the node where the volume is published.
@@ -518,13 +566,12 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
                         error!("{}", m);
                         return Err(Status::internal(m));
                     }
-                },
+                }
                 _ => {
-
                     // Check for node being cordoned.
                     fn cordon_check(spec: Option<&NodeSpec>) -> bool {
                         if let Some(spec) = spec {
-                            return spec.cordondrainstate.is_some()
+                            return spec.cordondrainstate.is_some();
                         }
                         false
                     }
@@ -538,12 +585,18 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
                         // then let the control-plane decide where to place the target. Node should not be cordoned.
                         Ok(node) if node.state.as_ref().map(|n| n.status).unwrap_or(NodeStatus::Unknown) != NodeStatus::Online || cordon_check(node.spec.as_ref()) => {
                             Ok(None)
-                        },
+                        }
                         // For 1-replica volumes, don't pre-select the target node. This will allow the
                         // control-plane to pin the target to the replica node.
                         Ok(_) if volume.spec.num_replicas == 1 => Ok(None),
                         Ok(_) => Ok(Some(node_id.as_str())),
                     }?;
+
+                    // Issue a cleanup rpc to csi node to ensure the subsystem doesn't have any path present before publishing
+                    if self.force_unstage_volume {
+                        let app_node = RestApiClient::get_client().get_app_node(&args.node_id).await?;
+                        force_unstage(app_node, volume_id.to_string()).await?;
+                    }
 
                     // Volume is not published.
                     let v = RestApiClient::get_client()
--- a/control-plane/csi-driver/src/bin/controller/main.rs
+++ b/control-plane/csi-driver/src/bin/controller/main.rs
@@ -111,12 +111,19 @@ async fn main() -> anyhow::Result<()> {
                 .help("Formatting style to be used while logging")
         )
         .arg(
-        Arg::new("ansi-colors")
-            .long("ansi-colors")
-            .default_value("true")
-            .value_parser(clap::value_parser!(bool))
-            .help("Enable ansi color for logs")
-    )
+            Arg::new("ansi-colors")
+                .long("ansi-colors")
+                .default_value("true")
+                .value_parser(clap::value_parser!(bool))
+                .help("Enable ansi color for logs")
+        )
+        .arg(
+            Arg::new("force-unstage-volume")
+                .long("force-unstage-volume")
+                .default_value("true")
+                .value_parser(clap::value_parser!(bool))
+                .help("Enable force unstage volume feature")
+        )
         .get_matches();
 
     utils::print_package_info!();
--- a/control-plane/csi-driver/src/bin/node/findmnt.rs
+++ b/control-plane/csi-driver/src/bin/node/findmnt.rs
@@ -4,6 +4,7 @@ use csi_driver::filesystem::FileSystem as Fs;
 use serde_json::Value;
 use std::{collections::HashMap, process::Command, str::FromStr, string::String, vec::Vec};
 use tracing::{error, warn};
+use utils::csi_plugin_name;
 
 // Keys of interest we expect to find in the JSON output generated
 // by findmnt.
@@ -13,9 +14,18 @@ const FSTYPE_KEY: &str = "fstype";
 
 #[derive(Debug)]
 pub(crate) struct DeviceMount {
-    #[allow(dead_code)]
     mount_path: String,
     fstype: FileSystem,
+}
+
+impl std::fmt::Display for DeviceMount {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "MountPath: {}, FileSystem: {}",
+            self.mount_path, self.fstype
+        )
+    }
 }
 
 impl DeviceMount {
@@ -26,6 +36,10 @@ impl DeviceMount {
     /// File system type
     pub(crate) fn fstype(&self) -> FileSystem {
         self.fstype.clone()
+    }
+    /// Get the mount path.
+    pub(crate) fn mount_path(&self) -> String {
+        self.mount_path.clone()
     }
 }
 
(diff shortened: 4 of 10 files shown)


control-plane/csi-driver/proto/node-service.proto
```diff
@@ -20,6 +20,8 @@ service NodePlugin {
   rpc UnfreezeFS (UnfreezeFSRequest) returns (UnfreezeFSReply) {}
   // Find the volume identified by the volume ID, and return volume information.
   rpc FindVolume (FindVolumeRequest) returns (FindVolumeReply) {}
+  // Issue Cleanup for the volume, unmount and disconnect.
+  rpc Cleanup (CleanupRequest) returns (CleanupReply) {}
```
tiagolobocastro: how about simply `NodeUnstage` ?

control-plane/csi-driver/src/bin/node/findmnt.rs
```diff
@@ -213,3 +226,133 @@ pub(crate) fn get_mountpaths(device_path: &str) -> Result<Vec<DeviceMount>, Devi
...
+    Ok(results)
+}
+
+/// Filters JSON output from `findmnt` for matching CSI mount points as per the filter.
+fn filter_csi_findmnt(
+    json_val: &Value,
+    filter: &FilterCsiMounts,
+    results: &mut Vec<HashMap<String, String>>,
+) {
+    match json_val {
```
tiagolobocastro, line 274: why do we need to parse manually? Can't we just use serde?

control-plane/csi-driver/src/bin/node/main_.rs
```diff
@@ -144,6 +144,13 @@ pub(super) async fn main() -> anyhow::Result<()> {
                 .value_parser(clap::value_parser!(bool))
                 .help("Enable ansi color for logs")
         )
+        .arg(
+            Arg::new("stale-entry-cleanup")
+                .long("stale-entry-cleanup")
+                .action(clap::ArgAction::SetTrue)
+                .value_name("BOOLEAN")
+                .help("Enable cleanup of the stale entries before controller publish")
```
tiagolobocastro: why do we need to enable this on the node? Shouldn't it be a controller feature?
Should it be a flag considering it could lead to corruption if not used?

control-plane/csi-driver/src/bin/node/nodeplugin_grpc.rs
```diff
@@ -54,6 +65,73 @@ impl NodePlugin for NodePluginSvc {
...
+        if !self.stale_entry_cleanup {
+            return Ok(Response::new(CleanupReply {}));
+        }
+        let volume_id = request.into_inner().volume_id;
+        debug!("Starting cleanup for volume: {volume_id}");
+
+        let nqn = format!("{}:{volume_id}", nvme_target_nqn_prefix());
+        match Subsystem::try_from_nqn(&nqn) {
+            Ok(subsystem_paths) => {
+                for subsystem_path in subsystem_paths {
```
tiagolobocastro, line 86: Should we not also unmount the paths? Otherwise it might cause stale mounts?

control-plane/csi-driver/proto/node-service.proto
Abhinandan-Purkait: We can but it's a little bit different compared to current node unstage.

control-plane/csi-driver/src/bin/node/main_.rs
Abhinandan-Purkait: Okay yeah maybe we should do on controller. Okay maybe a flag with SetFalse as default?

control-plane/csi-driver/src/bin/node/nodeplugin_grpc.rs
Abhinandan-Purkait: We are unmounting, once the subsystem paths are disconnected, if you see below line 116 and 128.

control-plane/csi-driver/src/bin/node/findmnt.rs
Abhinandan-Purkait: You mean using iterative method rather than recursion?

control-plane/csi-driver/proto/node-service.proto
dsharma-dc:
```suggestion
  rpc CleanupVolume (CleanupVolumeRequest) returns (CleanupVolumeReply) {}
```

control-plane/csi-driver/src/bin/controller/controller.rs
```diff
@@ -133,6 +135,20 @@ async fn issue_fs_unfreeze(endpoint: String, volume_id: String) -> Result<(), St
     }
 }
 
+#[tracing::instrument]
+async fn issue_cleanup(endpoint: String, volume_id: String) -> Result<(), Status> {
+    tracing::info!("Issuing cleanup of stale entries before publish for volume: {volume_id}");
```
dsharma-dc: How about removing caller side tracing and only have this one to print node and endpoint also, by taking app node as parameter in this function.

control-plane/csi-driver/src/bin/node/nodeplugin_grpc.rs
```diff
@@ -54,6 +65,73 @@ impl NodePlugin for NodePluginSvc {
...
+                    debug!(
+                        "Processing subsystem path: addr: {:?}, transport: {}, state: {}",
+                        subsystem_path.address, subsystem_path.transport, subsystem_path.state
```
dsharma-dc:
```suggestion
                        "Processing subsystem: {subsystem_path:?}, 
```

control-plane/csi-driver/src/bin/node/nodeplugin_grpc.rs
dsharma-dc: If I am reading correctly, we are unmounting only in case of NqnNotFound, and not after subsystems were found with the nqn and paths disconnected.

Abhinandan-Purkait:
> If I am reading correctly, we are unmounting only in case of NqnNotFound, and not after subsystems were found with the nqn and paths disconnected.

Yes, that's true, the reason is we cannot unmount if the subsystem path is present as that will get stuck, also we cannot just disconnect and do the umount because the disconnect takes quite a while to remove the path and the path stays in deleting (no IO) in the process of getting removed. For that we issue disconnect and return error so that csi retries and the check happens again, eventually in one of the retries it will be removed and we get into the section of NqnNotFound where we do the unmounts.

control-plane/csi-driver/src/bin/node/nodeplugin_grpc.rs
Abhinandan-Purkait: > If I am reading correctly, we are unmounting only in case of NqnNotFound, and not after subsystems were found with the nqn and paths disconnected.

Yes, that's true, the reason is we cannot unmount if the subsystem path is present as that will get stuck, also we cannot just disconnect and do the umount because the disconnect takes quite a while to remove the path and the path stays in deleting (no IO) in the process of getting removed. For that we issue disconnect and return error so that csi retries and the check happens again, eventually in one of the retries it will be removed and we get into the section of NqnNotFound where we do the unmounts.

control-plane/csi-driver/src/bin/node/nodeplugin_grpc.rs
Abhinandan-Purkait: Isn't the addr and path important here ? nqn is kind of same for everyone

control-plane/csi-driver/src/bin/node/main_.rs
Abhinandan-Purkait: Done.

control-plane/csi-driver/src/bin/controller/controller.rs
Abhinandan-Purkait: Done.

control-plane/csi-driver/src/bin/node/nodeplugin_grpc.rs
dsharma-dc: > Isn't the addr and path important here ? nqn is kind of same for everyone

It is. I thought the Debug trait might take care of that, but may be it won't as it's only derived?
You can retain the print as-is then.

dsharma-dc (approved):

control-plane/csi-driver/src/bin/node/nodeplugin_grpc.rs
dsharma-dc: hmm, Debug is derived for `SubsystemAddr` too so might just work the way suggested earlier. You can try and take a call as you deem fit.

Abhinandan-Purkait: Done, as you suggested.

control-plane/csi-driver/proto/node-service.proto
Abhinandan-Purkait: @tiagolobocastro Does this sound good to you?

Abhinandan-Purkait: @tiagolobocastro Does this sound good to you as well?

control-plane/csi-driver/src/bin/node/findmnt.rs
Abhinandan-Purkait: Yeah possibly we can do that. The rest of the code above also followed the same way so kept it similar. Although we can change all of them to serde maybe at some point.

control-plane/csi-driver/proto/node-service.proto
tiagolobocastro: Maybe `ForceUnstageVolume` or `ReleaseVolume` ?

tiagolobocastro: Added above two options as well (for the name itself). But yes should have Volume key in the naming

control-plane/csi-driver/src/bin/node/main_.rs
tiagolobocastro: But do we need it at all? You'll potentially get corruption without this flag

control-plane/csi-driver/src/bin/controller/controller.rs
```diff
@@ -133,6 +138,25 @@ async fn issue_fs_unfreeze(endpoint: String, volume_id: String) -> Result<(), St
     }
 }
 
+#[tracing::instrument]
```
tiagolobocastro: Though do we even need it?

```suggestion
#[tracing::instrument(err, skip_all)]
```

control-plane/csi-driver/src/bin/controller/controller.rs
```diff
@@ -799,7 +828,8 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
         }))
     }
 
-    #[instrument(err, fields(volume.uuid = request.get_ref().source_volume_id, snapshot.source_uuid = request.get_ref().source_volume_id, snapshot.uuid), skip(self))]
+    #[instrument(err, fields(volume.uuid = request.get_ref().source_volume_id, snapshot.source_uuid = request.get_ref().source_volume_id, snapshot.uuid
```
tiagolobocastro: what's modified here?

control-plane/csi-driver/src/bin/controller/controller.rs
```diff
@@ -939,7 +969,8 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
         Ok(Response::new(DeleteSnapshotResponse {}))
     }
 
-    #[instrument(err, fields(volume.uuid = request.get_ref().source_volume_id, snapshot.source_uuid = request.get_ref().source_volume_id, snapshot.uuid), skip(self))]
+    #[instrument(err, fields(volume.uuid = request.get_ref().source_volume_id, snapshot.source_uuid = request.get_ref().source_volume_id, snapshot.uuid
```
tiagolobocastro: and here

control-plane/csi-driver/src/bin/node/findmnt.rs
```diff
@@ -13,11 +13,20 @@ const FSTYPE_KEY: &str = "fstype";
...
-    #[allow(dead_code)]
     mount_path: String,
     fstype: FileSystem,
 }
 
+impl std::fmt::Display for DeviceMount {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "[ Mount Path: {}, File System Type: {} ]",
```
tiagolobocastro:
```suggestion
            "MountPath: {}, FileSystem: {}",
```

control-plane/csi-driver/src/bin/node/findmnt.rs
```diff
@@ -27,6 +36,10 @@ impl DeviceMount {
     pub(crate) fn fstype(&self) -> FileSystem {
         self.fstype.clone()
     }
+    /// Mount path
```
tiagolobocastro:
```suggestion
    /// Get the mount path.
```

control-plane/csi-driver/src/bin/node/findmnt.rs
```diff
@@ -213,3 +226,133 @@ pub(crate) fn get_mountpaths(device_path: &str) -> Result<Vec<DeviceMount>, Devi
         Err(e) => Err(e),
     }
 }
+
+const DRIVER_NAME: &str = "driverName";
+const VOLUME_HANDLE: &str = "volumeHandle";
+const MAYASTOR_DRIVER: &str = "io.openebs.csi-mayastor";
```
tiagolobocastro: don't we already have this in a constant somewhere?

control-plane/csi-driver/src/bin/node/findmnt.rs
```diff
@@ -213,3 +226,133 @@ pub(crate) fn get_mountpaths(device_path: &str) -> Result<Vec<DeviceMount>, Devi
...
+#[derive(Debug)]
+struct FilterCsiMounts<'a> {
+    driver: &'a str,
+    volume_id: &'a str,
+    file_name: &'a str,
+    device_pattern: &'a str,
+}
+
+/// Finds CSI mount points using `findmnt` and filters based on the given criteria.
+fn find_csi_mount(filter: FilterCsiMounts) -> Result<Vec<HashMap<String, String>>, DeviceError> {
```
tiagolobocastro, line 252: async?

control-plane/csi-driver/src/bin/node/findmnt.rs
tiagolobocastro: Oh I see, didn't know we were already using findmnt!

control-plane/csi-driver/src/bin/node/findmnt.rs
```diff
@@ -213,3 +226,133 @@ pub(crate) fn get_mountpaths(device_path: &str) -> Result<Vec<DeviceMount>, Devi
...
+    let output = Command::new(FIND_MNT).args(FIND_MNT_ARGS).output()?;
```
tiagolobocastro, line 253: Can we reduce things here by just checking mounts on the base mount path ?
`-R $dir`

control-plane/csi-driver/src/bin/node/findmnt.rs
```diff
@@ -213,3 +226,133 @@ pub(crate) fn get_mountpaths(device_path: &str) -> Result<Vec<DeviceMount>, Devi
...
+                if let Some(mountpath) = entry.get(TARGET_KEY) {
+                    if let Some(fstype) = entry.get(FSTYPE_KEY) {
+                        mountpaths.push(DeviceMount::new(
+                            mountpath.to_string(),
+                            Fs::from_str(fstype)
+                                .unwrap_or(Fs::Unsupported(fstype.to_string()))
+                                .into(),
+                        ))
+                    } else {
+                        error!("Missing fstype for {}", mountpath);
```
tiagolobocastro:
```suggestion
                        tracing::warn!(volume.id=%volume_id, "Missing fstype for {mountpath}");
```

control-plane/csi-driver/src/bin/node/findmnt.rs
```diff
@@ -213,3 +226,133 @@ pub(crate) fn get_mountpaths(device_path: &str) -> Result<Vec<DeviceMount>, Devi
...
+                        mountpaths.push(DeviceMount::new(
+                            mountpath.to_string(),
+                            Fs::Unsupported("".to_string()).into(),
+                        ))
+                    }
+                } else {
+                    warn!("missing target field {:?}", entry);
```
tiagolobocastro:
```suggestion
                    tracing::warn!(volume.id=%volume_id, ?entry, "Missing target field");
```

control-plane/csi-driver/src/bin/node/mount.rs
```diff
@@ -425,3 +426,48 @@ pub(crate) async fn unmount_on_fs_id_diff(
...
+                let mut unmount_flags = UnmountFlags::empty();
+                unmount_flags.insert(UnmountFlags::DETACH);
+                sys_mount::unmount(target_path, unmount_flags)
+                    .map_err(|error| Status::aborted(error.to_string()))
+            }
+        })
+        .await
+        .map_err(|error| Status::aborted(error.to_string()))??;
+
+        if target_path.ends_with("globalmount") {
```
tiagolobocastro: Since this didn't work, we can let it be?
Btw did you test cleaning up the entire path?

control-plane/csi-driver/src/bin/node/nodeplugin_grpc.rs
```diff
@@ -54,6 +63,67 @@ impl NodePlugin for NodePluginSvc {
             device_path: device.devname(),
         }))
     }
+
+    async fn cleanup(
```
tiagolobocastro: Should have instrument

control-plane/csi-driver/src/bin/node/nodeplugin_grpc.rs
tiagolobocastro: But we still should unmount the paths nevertheless
You can send unmount lazy after you disconnect?

control-plane/csi-driver/src/bin/node/nodeplugin_grpc.rs
```diff
@@ -54,6 +63,67 @@ impl NodePlugin for NodePluginSvc {
...
+                            subsystem_path
+                                .disconnect()
+                                .map_err(|error| Status::aborted(error.to_string()))
+                        })
+                        .await
+                        .map_err(|error| Status::aborted(error.to_string()))??;
+                    }
+                }
+                Err(Status::internal(format!(
+                    "Cleanup initiated for all stale entries of: {volume_id}. Returning error for validation on retry."
```
tiagolobocastro, line 99: Why do we return error here?

control-plane/csi-driver/src/bin/controller/main.rs
```diff
@@ -111,12 +111,19 @@ async fn main() -> anyhow::Result<()> {
...
+                .default_value("true")
+                .value_parser(clap::value_parser!(bool))
+                .help("Enable ansi color for logs")
+        )
+        .arg(
+            Arg::new("disable-stale-cleanup")
+                .long("disable-stale-cleanup")
+                .action(clap::ArgAction::SetFalse)
+                .value_name("BOOLEAN")
+                .help("Enable cleanup of the stale entries before controller publish")
```
dsharma-dc: help is saying `Enable` while the arg name says disable.

control-plane/csi-driver/src/bin/controller/controller.rs
```diff
@@ -538,13 +561,19 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
...
                             Ok(None)
-                        },
+                        }
                         // For 1-replica volumes, don't pre-select the target node. This will allow the
                         // control-plane to pin the target to the replica node.
                         Ok(_) if volume.spec.num_replicas == 1 => Ok(None),
                         Ok(_) => Ok(Some(node_id.as_str())),
                     }?;
 
+                    if self.disable_stale_cleanup {
```
dsharma-dc: Should become `!self.disable_stale_cleanup` ?

control-plane/csi-driver/src/bin/node/main_.rs
Abhinandan-Purkait: That's true was just thinking in terms of if it fails for someone for some reason should there be some way to disable this.

control-plane/csi-driver/src/bin/controller/controller.rs
Abhinandan-Purkait: Ohh nothing, the formatter did this, it seems.

control-plane/csi-driver/proto/node-service.proto
Abhinandan-Purkait: Okay `ForceUnstageVolume` sounds good.

control-plane/csi-driver/src/bin/node/main_.rs
Abhinandan-Purkait: Removing for now.

control-plane/csi-driver/proto/node-service.proto
Abhinandan-Purkait: Went with `ForceUnstageVolume`

control-plane/csi-driver/src/bin/node/findmnt.rs
Abhinandan-Purkait: Nope. We don't have this anywhere in control-plane.

control-plane/csi-driver/src/bin/node/findmnt.rs
Abhinandan-Purkait: Doesn't need to be right?

control-plane/csi-driver/src/bin/node/findmnt.rs
Abhinandan-Purkait: Not sure if there are other mounts?

control-plane/csi-driver/src/bin/node/mount.rs
Abhinandan-Purkait: Yeah tried that as well. I will remove this for now.

control-plane/csi-driver/src/bin/controller/main.rs
Abhinandan-Purkait: Removed this now as per above comments.

control-plane/csi-driver/src/bin/controller/controller.rs
Abhinandan-Purkait: Removed this.

control-plane/csi-driver/src/bin/node/nodeplugin_grpc.rs
Abhinandan-Purkait: As discussed kept it same for now, would refactor to retry in the method itself.

control-plane/csi-driver/src/bin/node/findmnt.rs
Abhinandan-Purkait: Correction: Took it from the constants

control-plane/csi-driver/src/bin/node/findmnt.rs
Abhinandan-Purkait: Yeah we need to take this as refactor of the findmnt.

control-plane/csi-driver/src/bin/node/findmnt.rs
Abhinandan-Purkait: Would add it as a refactor.

control-plane/csi-driver/src/bin/node/nodeplugin_grpc.rs
Abhinandan-Purkait: To utilise the retry without timing out, but would refactor this.

control-plane/csi-driver/src/bin/controller/controller.rs
```diff
@@ -91,12 +95,30 @@ fn volume_app_node(volume: &Volume) -> Option<String> {
...
+fn tonic_endpoint(endpoint: String) -> Result<tonic::transport::Endpoint, Status> {
+    let uri =
+        Uri::try_from(endpoint).map_err(|error| Status::invalid_argument(error.to_string()))?;
+
+    let timeout = humantime::parse_duration(DEFAULT_REQ_TIMEOUT).unwrap();
+    Ok(tonic::transport::Endpoint::from(uri)
+        .connect_timeout(timeout)
+        .timeout(std::time::Duration::from_secs(30))
+        .http2_keep_alive_interval(Duration::from_secs(10))
+        .keep_alive_timeout(Duration::from_secs(20))
```
```suggestion
        .http2_keep_alive_interval(Duration::from_secs(5))
        .keep_alive_timeout(Duration::from_secs(10))
```

control-plane/csi-driver/src/bin/controller/controller.rs
```diff
@@ -91,12 +95,30 @@ fn volume_app_node(volume: &Volume) -> Option<String> {
...
+
+#[tracing::instrument(err, skip_all)]
 async fn issue_fs_freeze(endpoint: String, volume_id: String) -> Result<(), Status> {
     trace!("Issuing fs freeze");
-    let mut client = NodePluginClient::connect(format!("http://{endpoint}"))
+    let channel = tonic_endpoint(format!("http://{endpoint}"))?
+        .connect()
         .await
-        .map_err(|error| Status::failed_precondition(error.to_string()))?;
+        .map_err(|error| Status::internal(error.to_string()))?;
```
tiagolobocastro: unavailable?

control-plane/csi-driver/src/bin/controller/controller.rs
```diff
@@ -112,12 +134,15 @@ async fn issue_fs_freeze(endpoint: String, volume_id: String) -> Result<(), Stat
...
-#[tracing::instrument]
+#[tracing::instrument(err, skip_all)]
 async fn issue_fs_unfreeze(endpoint: String, volume_id: String) -> Result<(), Status> {
     trace!("Issuing fs unfreeze");
-    let mut client = NodePluginClient::connect(format!("http://{endpoint}"))
+    let channel = tonic_endpoint(format!("http://{endpoint}"))?
+        .connect()
         .await
-        .map_err(|error| Status::failed_precondition(error.to_string()))?;
+        .map_err(|error| Status::internal(error.to_string()))?;
```
tiagolobocastro: same comment here

control-plane/csi-driver/src/bin/controller/controller.rs
```diff
@@ -133,6 +158,28 @@ async fn issue_fs_unfreeze(endpoint: String, volume_id: String) -> Result<(), St
...
+    tracing::info!(
+        "Issuing cleanup for volume: {} to node {}, to endpoint {}",
+        volume_id,
+        app_node.id,
+        app_node.spec.endpoint
+    );
+    let channel = tonic_endpoint(format!("http://{}", app_node.spec.endpoint))?
+        .connect()
+        .await
+        .map_err(|error| Status::internal(error.to_string()))?;
```
tiagolobocastro: and here

control-plane/csi-driver/src/bin/controller/controller.rs
```diff
@@ -133,6 +158,28 @@ async fn issue_fs_unfreeze(endpoint: String, volume_id: String) -> Result<(), St
     }
 }
 
+#[tracing::instrument(err, skip_all)]
+async fn issue_cleanup(app_node: AppNode, volume_id: String) -> Result<(), Status> {
```
```suggestion
async fn force_unstage(app_node: AppNode, volume_id: String) -> Result<(), Status> {
```

control-plane/csi-driver/src/bin/node/nodeplugin_grpc.rs
```diff
@@ -54,6 +68,71 @@ impl NodePlugin for NodePluginSvc {
             device_path: device.devname(),
         }))
     }
+
+    #[tracing::instrument(err, skip_all)]
```
tiagolobocastro: Add volume id here, similar to csi node.rs?

control-plane/csi-driver/src/bin/node/nodeplugin_grpc.rs
```diff
@@ -54,6 +68,71 @@ impl NodePlugin for NodePluginSvc {
...
+
+        let _guard = VolumeOpGuard::new_str(&volume_id)?;
+
+        debug!("Starting cleanup for volume: {volume_id}");
+
+        let nqn = format!("{}:{volume_id}", nvme_target_nqn_prefix());
+        match Subsystem::try_from_nqn(&nqn) {
+            Ok(subsystem_paths) => {
+                for subsystem_path in subsystem_paths {
+                    debug!("Processing subsystem: {subsystem_path:?}");
```
tiagolobocastro: Info?
Will not every be logged when we do find subsystems, so might be useful to see on info?

tiagolobocastro (approved):

control-plane/csi-driver/src/bin/node/nodeplugin_grpc.rs
Abhinandan-Purkait: Right could be.

control-plane/csi-driver/src/bin/controller/controller.rs
Abhinandan-Purkait: Done

control-plane/csi-driver/src/bin/controller/controller.rs
Abhinandan-Purkait: Done.

control-plane/csi-driver/src/bin/controller/controller.rs
Abhinandan-Purkait: Done.

control-plane/csi-driver/src/bin/controller/controller.rs
Abhinandan-Purkait: Done.

control-plane/csi-driver/src/bin/controller/controller.rs
Abhinandan-Purkait: Done.

control-plane/csi-driver/src/bin/node/nodeplugin_grpc.rs

Abhinandan-Purkait: `#[tracing::instrument(err, fields(volume.uuid = request.get_ref().volume_id), skip(self))]`

Abhinandan-Purkait:
bors merge

Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors merge

Abhinandan-Purkait:
bors cancel

Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors merge

PR #906, opened 2024-12-17
fix(topology): use topology when decreasing replicas

tiagolobocastro:
When removing volume replicas, take the pool topology into account and prefer to remove replicas which don't line up with the volume topology. This can happen when a volume's topology has been modified.

### DIFF
--- a/control-plane/agents/src/bin/core/controller/scheduling/volume.rs
+++ b/control-plane/agents/src/bin/core/controller/scheduling/volume.rs
@@ -5,13 +5,12 @@ use crate::controller::{
         affinity_group::{get_pool_ag_replica_count, get_restricted_nodes},
         pool::replica_rebuildable,
         resources::{ChildItem, PoolItem, PoolItemLister, ReplicaItem},
-        volume_policy::{SimplePolicy, ThickPolicy},
+        volume_policy::{node::NodeFilters, SimplePolicy, ThickPolicy},
         AddReplicaFilters, AddReplicaSorters, ChildSorters, ResourceData, ResourceFilter,
     },
     wrapper::PoolWrapper,
 };
 use agents::errors::SvcError;
-use std::{collections::HashMap, ops::Deref};
 use stor_port::types::v0::{
     store::{
         nexus::NexusSpec, nexus_persistence::NexusInfo, snapshots::replica::ReplicaSnapshot,
@@ -19,6 +18,8 @@ use stor_port::types::v0::{
     },
     transport::{NodeId, PoolId, VolumeState},
 };
+
+use std::{collections::HashMap, ops::Deref};
 
 /// Move replica to another pool.
 #[derive(Default, Clone)]
@@ -343,6 +344,21 @@ impl GetChildForRemovalContext {
                     }),
                     ag_rep_count,
                 )
+                .with_node_topology({
+                    Some(NodeFilters::topology_replica(
+                        &self.spec,
+                        &self.registry,
+                        replica_spec.pool_name(),
+                    ))
+                })
+                .with_pool_topology({
+                    use crate::controller::scheduling::volume_policy::pool::PoolBaseFilters;
+                    Some(PoolBaseFilters::topology_(
+                        &self.spec,
+                        &self.registry,
+                        replica_spec.pool_name(),
+                    ))
+                })
             })
             .collect::<Vec<_>>()
     }
--- a/control-plane/agents/src/bin/core/controller/scheduling/volume_policy/node.rs
+++ b/control-plane/agents/src/bin/core/controller/scheduling/volume_policy/node.rs
@@ -1,11 +1,18 @@
-use crate::controller::scheduling::{
-    nexus::GetSuitableNodesContext,
-    resources::{NodeItem, PoolItem},
-    volume::GetSuitablePoolsContext,
-    volume_policy::qualifies_label_criteria,
+use crate::controller::{
+    registry::Registry,
+    scheduling::{
+        nexus::GetSuitableNodesContext,
+        resources::{NodeItem, PoolItem},
+        volume::GetSuitablePoolsContext,
+        volume_policy::qualifies_label_criteria,
+    },
 };
+use stor_port::types::v0::{
+    store::volume::VolumeSpec,
+    transport::{NodeId, NodeTopology, PoolId},
+};
+
 use std::collections::HashMap;
-use stor_port::types::v0::transport::NodeTopology;
 
 /// Filter nodes used for replica creation.
 pub(crate) struct NodeFilters {}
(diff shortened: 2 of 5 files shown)


control-plane/agents/src/bin/core/controller/scheduling/volume_policy/pool.rs
```diff
@@ -77,27 +83,32 @@ impl PoolBaseFilters {
...
-                Some(pool_topology) => match pool_topology {
+                Some(pool_topology) => match &pool_topology {
                     PoolTopology::Labelled(labelled_topology) => {
                         // The labels in Volume Pool Topology should match the pool labels if
                         // present, otherwise selection of any pool is allowed.
-                        if !labelled_topology.inclusion.is_empty() {
-                            volume_pool_topology_inclusion_labels = labelled_topology.inclusion
-                        } else {
+                        if labelled_topology.inclusion.is_empty() {
+                            // todo: missing exclusion check?
```
tiagolobocastro, line 102: @sinhaashish do we need to take into account exclusion labels on filter?
I guess we're not making use of it atm?

control-plane/agents/src/bin/core/controller/scheduling/volume.rs
```diff
@@ -343,6 +343,22 @@ impl GetChildForRemovalContext {
                     }),
                     ag_rep_count,
                 )
+                .with_node_topology({
+                    use crate::controller::scheduling::volume_policy::node::NodeFilters;
```
Abhinandan-Purkait: Nit: Should we put it on top?

control-plane/agents/src/bin/core/controller/scheduling/volume_policy/node.rs
```diff
@@ -1,11 +1,17 @@
...
+    scheduling::{
+        nexus::GetSuitableNodesContext,
+        resources::{NodeItem, PoolItem},
+        volume::GetSuitablePoolsContext,
+        volume_policy::qualifies_label_criteria,
+    },
 };
 use std::collections::HashMap;
-use stor_port::types::v0::transport::NodeTopology;
+use stor_port::types::v0::{
```
Abhinandan-Purkait, line 10: Move this up? And keep the std after a blank line?

control-plane/agents/src/bin/core/controller/scheduling/volume_policy/pool.rs
```diff
@@ -1,10 +1,16 @@
...
+    registry::Registry,
+    scheduling::{
+        resources::{ChildItem, PoolItem},
+        volume::{GetSuitablePoolsContext, ReplicaResizePoolsContext},
+        volume_policy::qualifies_label_criteria,
+    },
 };
 use std::collections::HashMap;
-use stor_port::types::v0::transport::{PoolStatus, PoolTopology};
+use stor_port::types::v0::{
```
Abhinandan-Purkait, line 9: Same here?

Abhinandan-Purkait (approved):

control-plane/agents/src/bin/core/controller/scheduling/volume_policy/node.rs
```diff
@@ -72,9 +78,24 @@ impl NodeFilters {
     }
     /// Should only attempt to use nodes having specific creation label if topology has it.
     pub(crate) fn topology(request: &GetSuitablePoolsContext, item: &PoolItem) -> bool {
+        Self::topology_(request, request.registry(), &item.pool.node)
```
dsharma-dc, line 82: Shouldn't this be `request.spec` in first argument?

control-plane/agents/src/bin/core/controller/scheduling/volume_policy/pool.rs
```diff
@@ -77,27 +83,32 @@ impl PoolBaseFilters {
 
     /// Should only attempt to use pools having specific creation label if topology has it.
     pub(crate) fn topology(request: &GetSuitablePoolsContext, item: &PoolItem) -> bool {
+        Self::topology_(request, request.registry(), &item.pool.id)
```
dsharma-dc, line 87: Similarly here.

control-plane/agents/src/bin/core/controller/scheduling/volume.rs
tiagolobocastro: It could, but it's quite specific to this closure only.
Alright I'll move it.

control-plane/agents/src/bin/core/controller/scheduling/volume_policy/node.rs
tiagolobocastro: It's derefing into it

control-plane/agents/src/bin/core/controller/scheduling/volume_policy/pool.rs
tiagolobocastro: Same as above

control-plane/agents/src/bin/core/controller/scheduling/volume_policy/node.rs
tiagolobocastro: Done

control-plane/agents/src/bin/core/controller/scheduling/volume_policy/pool.rs
tiagolobocastro: Done

control-plane/agents/src/bin/core/controller/scheduling/volume_policy/pool.rs
sinhaashish: Its been used for `nodeSpreadTopologyKey` . Refer [here ](https://github.com/openebs/mayastor-control-plane/blob/27ab9dc45774584760579616e3d328f507d4e835/control-plane/agents/src/bin/core/volume/operations.rs#L1080C4-L1080C24)

tiagolobocastro: Ah, because the spread is done outside of the scheduling logic!
How do you suggest we take the spread into account here? Some further checks in `volume_replica_remove_candidate` ?
Also seems spread is also broken? It's only being used when creating the volume but not when new volume replicas are added?

sinhaashish: yes some checks needs to be added here  `volume_replica_remove_candidate`. 
The code for `spread_label_is_same` is present in create only and not in `increase_volume_replica`. 

But why is this required after all.  IIRC these all requirement only arises if the volume replica topology is modified and it was agreed that the replica topology if modified will leave all the replicas as is and wont reshuffle to other nodes or pools.

tiagolobocastro: >  The code for spread_label_is_same is present in create only and not in increase_volume_replica.

What do you mean by this, is it not required in increase, does it just work or is this spread broken when adding replicas?

>  wont reshuffle to other nodes or pools.

Yeah we don't do that automatically, but without these changes you can't even do it manually as a work-around.

sinhaashish: I mean spread is broken when adding replica.

tiagolobocastro: Could you please create an issue on mayastor for that?
And so in this case I think it's probably fine to not handle spread atm, since it's broken anyway @Abhinandan-Purkait ?

tiagolobocastro:
Merging since <https://github.com/openebs/mayastor/issues/1799> means spread is already broken.
bors merge

tiagolobocastro:
bors merge

PR #910, opened 2024-12-30
ci: switch from jenkins to github actions

tiagolobocastro:
Jenkins has become troublesome to maintain and makes it harder for users to see and test output without adding new users.
We could probably spend some time improving that, but the current Jenkins is also setup on systems managed by the sponsoring company so it makes sense to move away from that at least.
In order to run the CI tests we need a relatively powerful CI system so we are making use of the ubuntu-latest-16-cores runners.

The tests seem to run slower than on Jenkins. Part of it may be explained because the dependencies are not pre-cache and also the pre-requisites which take time to install.


control-plane/agents/src/bin/core/tests/controller/mod.rs
```diff
@@ -81,6 +81,7 @@ async fn store_lease_lock() {
         .await
         .unwrap();
 
+    tokio::time::sleep(std::time::Duration::from_secs(1)).await;
```
Abhinandan-Purkait, line 86: Do we need this sleep?

docker-logs.txt
```diff
@@ -0,0 +1,1478 @@
+2025-01-02 16:15:20.389430: Logs for Test tests/bdd/features/volume/topology/test_node_topology.py::test_suitable_nodes_which_contain_volume_spread_topology_key[True-zone-us-1-<=-succeed-must be-zone-us=us-west-1,zone-us=different-value] (call):
```
Abhinandan-Purkait: This file is not needed right?

tests/bdd/features/volume/topology/test_node-topology.py
```diff
@@ -82,7 +82,7 @@
         "node_name": NODE_1_NAME,
         "pool_uuid": NODE_1_POOL_1_UUID,
         "pool_body": CreatePoolBody(
-            ["malloc:///node1pool1?size_mb=50"],
+            ["malloc:///node1pool1?size_mb=32"],
```
Abhinandan-Purkait, line 85: Since you are already here? Would it be a good idea to make it file?

Abhinandan-Purkait (approved):

control-plane/agents/src/bin/core/tests/controller/mod.rs
tiagolobocastro: Seems on gha etcd is not ready without it, not sure why atm...
I'll add a comment for the effect

tests/bdd/features/volume/topology/test_node-topology.py
tiagolobocastro: Probably yes, but would rather get this in first as I've already spent quite some time on it

docker-logs.txt
tiagolobocastro: Thanks, seems like the gitignore was added after the  file had been commited so has to be removed. Done

tiagolobocastro:
bors try

tiagolobocastro:
bors merge

PR #911, opened 2025-01-03
refactor(plugin): expand scope of ExecuteOperations trait

niladrih:
The Trait ExecuteOptions requires a specific type from the this repo to work. This poses as a limitation for repos which use this codebase as a git module. This change adheres to the patterns where a clap::Parser struct instance is passed to the function, while removing dependency on the exact type of clap::Parser. Also, adds an output argument to support the existing users of this trait who depend on the CliArgs.output member from the cli_args arg.


niladrih:
This PR is not required.

PR #912, opened 2025-01-03
fix(ha-node): use transport info from nvme connect response

dsharma-dc:
With tcpFallback option enabled, the csi-node can make a decision to connect an rdma capable target over tcp if the initiator is not rdma capable. However, after the NvmeConnectResponse is received at ha-node, the ha-node doesn't know that fallback could've happened. The ha-node blindly goes ahead validating the new path presence by matching using rdma transport parsed from target uri, which it doesn't find because new path got connected over tcp. This cascades a SubsystemNotFound error up to ha-cluster agent that reinitiates republish and keeps looping in same problem.
This fix passes the actual transport used for connect info from csi-node to ha-node in NvmeConnectResponse, letting ha-node validate using accurate transport.

### DIFF
--- a/control-plane/csi-driver/src/bin/node/dev.rs
+++ b/control-plane/csi-driver/src/bin/node/dev.rs
@@ -27,6 +27,7 @@
 //!     }
 //! ```
 
+use nvmeadm::nvmf_discovery::TrType;
 use std::{
     collections::HashMap,
     convert::TryFrom,
@@ -50,6 +51,21 @@ use crate::match_dev;
 
 pub(crate) type DeviceName = String;
 
+// Could've directly used NvmfAttach object but it contains an AtomicBool that
+// doesn't have Clone implemented.
+#[derive(Default)]
+pub struct NvmfAttachParameters {
+    _host: String,
+    _port: u16,
+    pub transport: TrType,
+}
+
+pub enum AttachParameters {
+    Nbd,
+    Iscsi,
+    Nvmf(NvmfAttachParameters),
+}
+
 #[tonic::async_trait]
 pub(crate) trait Attach: Sync + Send {
     async fn parse_parameters(
@@ -60,6 +76,9 @@ pub(crate) trait Attach: Sync + Send {
     async fn find(&self) -> Result<Option<DeviceName>, DeviceError>;
     /// Fixup parameters which cannot be set during attach, eg IO timeout
     async fn fixup(&self) -> Result<(), DeviceError>;
+    /// Get basic connections parameters like target address, transport and port,
+    /// for this attach object.
+    fn attach_parameters(&self) -> AttachParameters;
 }
 
 #[tonic::async_trait]
--- a/control-plane/csi-driver/src/bin/node/nodeplugin_nvme.rs
+++ b/control-plane/csi-driver/src/bin/node/nodeplugin_nvme.rs
@@ -1,4 +1,4 @@
-use crate::dev::{nvmf::volume_uuid_from_url_str, Device};
+use crate::dev::{nvmf::volume_uuid_from_url_str, AttachParameters, Device};
 use csi_driver::limiter::VolumeOpGuard;
 use grpc::csi_node_nvme::{
     nvme_operations_server::NvmeOperations, NvmeConnectRequest, NvmeConnectResponse,
@@ -42,6 +42,16 @@ impl NvmeOperations for NvmeOperationsSvc {
             .await
             .map_err(|error| warn!(error=%error, "Failed to do fixup after connect"));
 
-        Ok(tonic::Response::new(NvmeConnectResponse {}))
+        let conn_params = match device.attach_parameters() {
+            AttachParameters::Nvmf(p) => p,
+            // Must not reach here.
+            _ => {
+                unreachable!("Params must be of variant AttachParameters::Nvmf here.");
+            }
+        };
+
+        Ok(tonic::Response::new(NvmeConnectResponse {
+            transport_used: Some(conn_params.transport as i32),
+        }))
     }
 }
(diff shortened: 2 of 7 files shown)


dsharma-dc:
This change depends on https://github.com/openebs/mayastor-dependencies/pull/107

control-plane/csi-driver/src/bin/node/dev.rs
```diff
@@ -60,6 +60,9 @@ pub(crate) trait Attach: Sync + Send {
     async fn find(&self) -> Result<Option<DeviceName>, DeviceError>;
     /// Fixup parameters which cannot be set during attach, eg IO timeout
     async fn fixup(&self) -> Result<(), DeviceError>;
+    /// Get basic connections parameters like target address, transport and port,
+    /// for this attach object.
+    async fn get_conn_params(&self) -> Result<HashMap<String, String>, DeviceError>;
```
tiagolobocastro: Should this be an output of `attach`? Image if for example the attach ever tries multiple protocols, there'd be no way of knowing which one was the "winner".
Also we can use an associated type here to return exact information.

control-plane/csi-driver/src/bin/node/nodeplugin_nvme.rs
```diff
@@ -42,6 +42,20 @@ impl NvmeOperations for NvmeOperationsSvc {
...
 
-        Ok(tonic::Response::new(NvmeConnectResponse {}))
+        let conn_params = device.get_conn_params().await?;
+        let transport = match conn_params
+            .get("transport")
+            .cloned()
+            .unwrap_or_default()
+            .as_str()
+        {
+            "rdma" => Some(1),
```
tiagolobocastro: Let's make use of the type system here rather than literals

control-plane/grpc/proto/v1/ha/csi_node_nvme.proto
```diff
@@ -10,8 +10,16 @@ message NvmeConnectRequest {
   common.MapWrapper publish_context = 2;
 }
 
-message NvmeConnectResponse {
+// chose enum variant values to match these values as TrType in nvmeadm.
```
tiagolobocastro:
```suggestion
// choose enum variant values to match these values as TrType in nvmeadm.
```

control-plane/grpc/proto/v1/ha/csi_node_nvme.proto
```diff
@@ -10,8 +10,16 @@ message NvmeConnectRequest {
...
+enum TrType {
+  invalid = 0;
```
tiagolobocastro: Should these be `Capitalised ? I think other enums are but can't quite remember

control-plane/grpc/proto/v1/ha/csi_node_nvme.proto
dsharma-dc: `chose` because comment is informing about a decision that was made :)

control-plane/csi-driver/src/bin/node/dev.rs
dsharma-dc: Ideally `attach` should be doing what it is told to, after all decisions are already done w.r.t transport, so we shouldn't be changing that part of interface. I can make this info return as part of attach, though I prefer to have a separate interface as it has clearly defined purpose e.g. like `parse_parameters`.

control-plane/grpc/proto/v1/ha/csi_node_nvme.proto
dsharma-dc: > Should these be `Capitalised ? I think other enums are but can't quite remember

We have mix of everything today, but mostly PascalCase. I'll change to PascalCase.

control-plane/grpc/proto/v1/ha/csi_node_nvme.proto
tiagolobocastro: Ah then it'd be `chosen`

control-plane/csi-driver/src/bin/node/dev.rs
tiagolobocastro: Alright, then in that case let's keep it as a separate function then. Maybe can simplify into something like this:

```suggestion
    fn attach_parameters(&self) -> Self::Parameters;
```

dsharma-dc: hmm, it's getting troublesome to use associated type in Attach trait because there are places e.g in `impl Device::parse()` where we return the trait as `Box<dyn Attach>` , which complains about associated type ambiguity then. To handle that will need much changes by using generics. I can rather define a concrete `struct AttachParameters` and use that. It compiles fine.

```
#[derive(Default)]
pub struct AttachParameters {
    pub _host: String,
    pub _port: u16,
    pub transport: TrType,
}

#[tonic::async_trait]
pub(crate) trait Attach: Sync + Send {
[...]
    fn attach_parameters(&self) -> AttachParameters;
}
```

tiagolobocastro: Ah that's right.
Then I'd say perhaps make it an enum:
```rust
enum AttachParameters {
  Nbd,
  Iscsi,
  Nvmf(NvmfAttachParameters),
}
```

dsharma-dc: done now.

control-plane/csi-driver/src/bin/node/nodeplugin_nvme.rs
dsharma-dc: > Let's make use of the type system here rather than literals

made changes.

dsharma-dc:
bors try

Abhinandan-Purkait (approved):

tiagolobocastro (approved):

dsharma-dc:
bors merge

dsharma-dc:
bors cancel

will update dependency commit

dsharma-dc:
bors merge

PR #914, opened 2025-01-09
fix(plugin/block-devices): don't panic on unmanaged pools 

tiagolobocastro:
fix(plugin/block-devices): don't panic on unmanaged pools
    
    When fetching the block-devices, don't panic if the pool spec is missing,
    that is, if the pool is unmanaged.
    (this can happen when troubleshooting and we remove pool
    from control plane).
    
---

    test: add pool create test rpc
    
In addition, the current plugin output tests are not running it seems as they otherwise fail when ran manually (at least the pool).
I fixed the pool test, but I don't think these are maintainable the way they are written. I'd rather smart parse the human output into json and then compare json with expected output.
Also we are calling rest directly, and thus we're not testing the actual calling of rest itself, we're just testing the human output formatting.


control-plane/plugin/src/resources/tests.rs
```diff
@@ -340,25 +398,35 @@ where
...
-        pool_state.capacity,
-        pool_state.used,
-        disks,
-        pool_state.node,
-        pool_state.status.to_string(),
-        true
+        " {id}  {disks}  {managed}    {node}  {status}  {capacity}     {allocated}        {available}      {committed} \n",
+        id = pool_state.id,
+        node = pool_state.node,
+        status = pool_state.status.to_string(),
```
Abhinandan-Purkait, line 429: Yeah agreed this is not a good way anymore now that the plugin has become more complex. The earlier intention was just to validate the stdout, tabular output, when the plugin was in initial stages.

Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

PR #915, opened 2025-01-10
add tracing method taking jaeger arg

abhilashshetty04:


control-plane/plugin/src/lib.rs
```diff
@@ -64,28 +64,33 @@ pub struct CliArgs {
...
+/// Init tracing with jaeger.
+pub fn init_tracing_with_jaeger(jaeger: Option<String>) -> TracingFlusher {
+    let git_version = option_env!("GIT_VERSION").unwrap_or_else(utils::raw_version_str);
+    let tags =
+        utils::tracing_telemetry::default_tracing_tags(git_version, env!("CARGO_PKG_VERSION"));
 
-        let jaeger = self.jaeger.as_ref();
+    let fmt_layer = match std::env::var("RUST_LOG") {
+        Ok(_) => FmtLayer::Stderr,
+        Err(_) => FmtLayer::None,
```
niladrih, line 89: This goes with Ok(), if RUST_LOG is empty, i.e. RUST_LOG="". Is this expected behaviour?

niladrih (approved):

Abhinandan-Purkait (approved):

control-plane/plugin/src/lib.rs
abhilashshetty04: Do you mean if someone explictly sets "" or doesnt set this env at all?

abhilashshetty04: @tiagolobocastro , As this binary doesnt have logging like our other services. Does FmtLayer variant (None, Json, Pretty) really matter? Does it also define what trace format we see in jaeger?

tiagolobocastro: It does have logging from its dependencies. We could also add logging in the future for anything we like I guess.
So yes, we could potentially configure fmtStyle via argument as well, up to you

control-plane/plugin/src/lib.rs
```diff
@@ -64,28 +64,33 @@ pub struct CliArgs {
...
+        init_tracing_with_jaeger(self.jaeger.clone())
+    }
+}
 
-        let fmt_layer = match std::env::var("RUST_LOG") {
-            Ok(_) => FmtLayer::Stderr,
-            Err(_) => FmtLayer::None,
-        };
+/// Init tracing with jaeger.
+pub fn init_tracing_with_jaeger(jaeger: Option<String>) -> TracingFlusher {
```
tiagolobocastro:
```suggestion
pub fn init_tracing_with_jaeger(jaeger: Option<&String>) -> TracingFlusher {
```

control-plane/plugin/src/lib.rs
```diff
@@ -64,28 +64,33 @@ pub struct CliArgs {
 impl CliArgs {
     /// Initialize tracing (including opentelemetry).
     pub fn init_tracing(&self) -> TracingFlusher {
-        let git_version = option_env!("GIT_VERSION").unwrap_or_else(utils::raw_version_str);
-        let tags =
-            utils::tracing_telemetry::default_tracing_tags(git_version, env!("CARGO_PKG_VERSION"));
+        init_tracing_with_jaeger(self.jaeger.clone())
```
tiagolobocastro:
```suggestion
        init_tracing_with_jaeger(self.jaeger.as_ref())
```

tiagolobocastro (approved):

control-plane/plugin/src/lib.rs
abhilashshetty04: Lets keep the behavior as is for now. We will set Fmt to Stderr if RUST_LOG is set.

abhilashshetty04:
bors merge

abhilashshetty04:
bors merge

PR #916, opened 2025-01-15
Cherry Pick 914: use topology when decreasing replicas

tiagolobocastro:
When removing volume replicas, take the topology into account and prefer to remove replicas which don't line up with the volume topology. This can happen when a volume's topology has been modified.


Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

PR #919, opened 2025-01-22
fix(csi/controller): disable orphan vol garbage collector

tiagolobocastro:
When a retain PV is deleted, we used to delete the mayastor volume. Whilst this is usually acceptable behaviour by our users it doesn't line up with expected K8s behaviour and has lead to data loss by a user.
In this change we add a flag to configure it and disable it by default.
In a future change we'll configure this via helm and also provide another way of deleting these orphan volumes.

### DIFF
--- a/control-plane/csi-driver/src/bin/controller/main.rs
+++ b/control-plane/csi-driver/src/bin/controller/main.rs
@@ -104,6 +104,13 @@ async fn main() -> anyhow::Result<()> {
                 )
         )
         .arg(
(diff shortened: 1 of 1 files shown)


tiagolobocastro:
Resolves https://github.com/openebs/mayastor/issues/1787

avishnu (approved):
lgtm. Thanks.

dsharma-dc (approved):

niladrih (approved):

tiagolobocastro:
bors merge

PR #921, opened 2025-01-22
Cherry-pick PR 917

tiagolobocastro:
fix(csi/controller): disable orphan vol garbage collector

When a retain PV is deleted, we used to delete the mayastor volume. Whilst this is usually acceptable behaviour by our users it doesn't line up with expected K8s behaviour and has lead to data loss by a user.
In this change we add a flag to configure it and disable it by default.
In a future change we'll configure this via helm and also provide another way of deleting these orphan volumes.


dsharma-dc (approved):

tiagolobocastro:
bors merge

PR #922, opened 2025-01-22
fix(reconciler/volume): disown replica from disowned volumes nexus

abhilashshetty04:
While we disown a replica from volume we used to validate if the replica is owned by any nexus. In some cases involving Outage on volume target node we observed that the new nexus do get created after the volumeattachments are deleted from the node. However, we werent able to disown the Unknown replica from volume since it was still owned by old target.

Now we check if replica is owned by the current target of the volume rather then if it is owned by any target.

### DIFF
--- a/control-plane/agents/src/bin/core/controller/reconciler/volume/garbage_collector.rs
+++ b/control-plane/agents/src/bin/core/controller/reconciler/volume/garbage_collector.rs
@@ -10,7 +10,7 @@ use crate::controller::{
 use agents::errors::SvcError;
 use stor_port::types::v0::{
     store::{nexus_persistence::NexusInfo, replica::ReplicaSpec, volume::VolumeSpec},
-    transport::{DestroyVolume, NexusOwners, ReplicaOwners, ResizeReplica, VolumeStatus},
+    transport::{DestroyVolume, NexusId, NexusOwners, ReplicaOwners, ResizeReplica, VolumeStatus},
 };
 use tracing::Instrument;
 
@@ -150,7 +150,6 @@ async fn disown_unused_nexuses(
     context: &PollContext,
 ) -> PollResult {
     let mut results = vec![];
-
     for nexus in context.specs().volume_nexuses(volume.uuid()) {
         if let Ok(mut nexus) = nexus.operation_guard() {
             if let Some(target) = volume.as_ref().target() {
@@ -158,8 +157,9 @@ async fn disown_unused_nexuses(
                     continue;
                 }
             }
-
-            nexus.warn_span(|| tracing::warn!("Attempting to disown unused nexus"));
+            nexus.info_span(|| {
+                tracing::info!("Disowning unused nexus from its volume");
+            });
             // the nexus garbage collector will destroy the disowned nexuses
             let owner = NexusOwners::Volume(volume.uuid().clone());
             match nexus.remove_owners(context.registry(), &owner, false).await {
@@ -206,7 +206,8 @@ async fn disown_unused_replicas(
         // don't attempt to disown the replicas if the volume state is faulted or unknown
         return PollResult::Ok(PollerState::Busy);
     }
-
+    let vol_nexus = context.specs().volume_nexuses(volume.uuid());
+    let vol_nexus_ids: Vec<&NexusId> = vol_nexus.iter().map(|n| n.uuid()).collect();
     let mut nexus_info = None; // defer reading from the persistent store unless we find a candidate
     let mut results = vec![];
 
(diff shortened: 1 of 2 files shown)


control-plane/agents/src/bin/core/controller/reconciler/volume/garbage_collector.rs
```diff
@@ -150,16 +150,16 @@ async fn disown_unused_nexuses(
...
         if let Ok(mut nexus) = nexus.operation_guard() {
             if let Some(target) = volume.as_ref().target() {
                 if target.nexus() == nexus.uid() || nexus.as_ref().is_shutdown() {
                     continue;
                 }
             }
-
-            nexus.warn_span(|| tracing::warn!("Attempting to disown unused nexus"));
+            nexus.info_span(|| {
+                tracing::info!("Disowning volume from the nexus");
```
tiagolobocastro:
```suggestion
                tracing::info!("Disowning unused nexus from its volume");
```

control-plane/stor-port/src/types/v0/store/replica.rs
```diff
@@ -109,6 +110,17 @@ impl ReplicaSpec {
     pub fn owned_by(&self, id: &VolumeId) -> bool {
         self.owners.owned_by(id)
     }
+    /// Check if this replica is owned by any of the nexus in volume spec.
+    pub fn owned_by_nexus(&self, nexus_id: Vec<NexusId>) -> bool {
```
tiagolobocastro: Pass by reference

control-plane/stor-port/src/types/v0/store/replica.rs
```diff
@@ -109,6 +110,17 @@ impl ReplicaSpec {
...
+        let mut result = false;
+        for nexus in self.owners.nexuses() {
+            if nexus_id.contains(nexus) {
+                result = true;
+                break;
+            }
+        }
+        result
```
tiagolobocastro:
```suggestion
        self.owners.iter().any(|owner| nexus_id.contains(owner))
```

control-plane/agents/src/bin/core/controller/reconciler/volume/garbage_collector.rs
```diff
@@ -206,7 +206,8 @@ async fn disown_unused_replicas(
         // don't attempt to disown the replicas if the volume state is faulted or unknown
         return PollResult::Ok(PollerState::Busy);
     }
-
+    let vol_nexus = context.specs().volume_nexuses(volume.uuid());
+    let n_id_list: Vec<NexusId> = vol_nexus.into_iter().map(|n| n.uuid().clone()).collect();
```
tiagolobocastro:
```suggestion
    let vol_nexus_ids: Vec<NexusId> = vol_nexus.into_iter().map(|n| n.uuid().clone()).collect();
```

tiagolobocastro (approved):

control-plane/agents/src/bin/core/controller/reconciler/volume/garbage_collector.rs
```diff
@@ -206,7 +206,8 @@ async fn disown_unused_replicas(
         // don't attempt to disown the replicas if the volume state is faulted or unknown
         return PollResult::Ok(PollerState::Busy);
     }
-
+    let vol_nexus = context.specs().volume_nexuses(volume.uuid());
```
dsharma-dc, line 209: Are we certain that this list won't contain the currently active nexus ever?

abhilashshetty04: I had doubts about it. I checked it was having only new nexus.

dsharma-dc: `volume_nexuses` functions says:
```
    /// Get a list of resourced NexusSpecs's which are owned by the given volume `id`
    /// but may not be active anymore.
```
So this list **may** contain more than one old nexus possibly, but **must no**t contain the currently active one IIUC, otherwise `owned_by_nexus` will return true.

dsharma-dc (approved):

abhilashshetty04:
@dsharma-dc , Before the `disown_unused_replicas`  we run `disown_unused_nexuses` which will clear old nexus's owner id as the new nexus is already in the `volume.state.target`. 

So we will receive only new nexus when we call from `disown_unused_replicas`

We can improve on docs i suppose. Its kind of race condition. Suppose we create new nexus and its published to the volume and reconciler already ran `disown_unused_nexuses`. So we will receive both nexuses in that scenario and the `Unknown` replica will not be removed. But in the next iteration we can guarantee that `disown_unused_nexuses` will be run before `disown_unused_replicas` and we get only active nexus of volume.

We have PR on develop. I will see how we can change docs in that.

abhilashshetty04:
bors merge

control-plane/agents/src/bin/core/controller/reconciler/volume/garbage_collector.rs
tiagolobocastro: This function will return all volume nexuses

PR #927, opened 2025-02-05
Feature: Configuring rest components for TLS

Johnaius:
This PR configures components using rest protocols to speak tls.  It is dependent on [this PR in extensions repo](https://github.com/openebs/mayastor-extensions/pull/622). 

### Rest-api Clients 
- Modified certificate loading to use pkcs8_private_keys instead of rsa_private_keys.  This was a change I made to get the generated certificates to work, they were apparently created in pkcs8 format, not rsa - I attempted to convert them with no luck, and found the pkcs8_private_keys in the [rustl-pemfile crate](https://docs.rs/rustls-pemfile/latest/rustls_pemfile/)...
### Csi-controller and Diskpool operator:
- Added support for TLS configuration based on CA certificate path.
- if certs are provided, use https, if not use http.
-  error handling for HTTPS connections without a certificate etc

### DIFF
--- a/control-plane/csi-driver/src/bin/controller/client.rs
+++ b/control-plane/csi-driver/src/bin/controller/client.rs
@@ -123,16 +123,49 @@ impl RestApiClient {
         let url = clients::tower::Url::parse(endpoint)
             .map_err(|error| anyhow!("Invalid API endpoint URL {}: {:?}", endpoint, error))?;
         let concurrency_limit = cfg.create_volume_limit() * 2;
-        let tower = clients::tower::Configuration::builder()
-            .with_timeout(cfg.io_timeout())
-            .with_concurrency_limit(Some(concurrency_limit))
-            .build_url(url)
-            .map_err(|error| {
-                anyhow::anyhow!(
-                    "Failed to create openapi configuration, Error: '{:?}'",
-                    error
-                )
-            })?;
+        let ca_certificate_path = cfg.ca_certificate_path();
+        let cert = match ca_certificate_path {
+            Some(path) => {
+                let cert = std::fs::read(path).map_err(|error| {
+                    anyhow::anyhow!(
+                        "Failed to create openapi configuration at path {}, Error: '{:?}'",
+                        path.display(),
+                        error
+                    )
+                })?;
+                Some(cert)
+            }
+            None => None,
+        };
+        let tower = match (url.scheme(), cert) {
+            ("https", Some(cert)) => clients::tower::Configuration::builder()
+                .with_timeout(Some(cfg.io_timeout()))
+                .with_concurrency_limit(Some(concurrency_limit))
+                .with_certificate(cert.as_slice())
+                .build_url(url)
+                .map_err(|error| {
+                    anyhow::anyhow!(
+                        "Failed to create openapi configuration, Error: '{:?}'",
+                        error
+                    )
+                })?,
+            ("https", None) => {
+                anyhow::bail!("HTTPS endpoint requires a CA certificate path");
+            }
+            (_, Some(_path)) => {
+                anyhow::bail!("CA certificate path is only supported for HTTPS endpoints");
+            }
+            _ => clients::tower::Configuration::builder()
+                .with_timeout(Some(cfg.io_timeout()))
+                .with_concurrency_limit(Some(concurrency_limit))
+                .build_url(url)
+                .map_err(|error| {
+                    anyhow::anyhow!(
+                        "Failed to create openapi configuration, Error: '{:?}'",
+                        error
+                    )
+                })?,
+        };
 
         REST_CLIENT.get_or_init(|| Self {
             rest_client: clients::tower::ApiClient::new(tower.clone()),
--- a/control-plane/csi-driver/src/bin/controller/config.rs
+++ b/control-plane/csi-driver/src/bin/controller/config.rs
@@ -1,7 +1,11 @@
 use anyhow::Context;
 use clap::ArgMatches;
 use once_cell::sync::OnceCell;
-use std::{collections::HashMap, time::Duration};
+use std::{
+    collections::HashMap,
+    path::{Path, PathBuf},
+    time::Duration,
+};
 
 static CONFIG: OnceCell<CsiControllerConfig> = OnceCell::new();
 
@@ -17,6 +21,8 @@ pub(crate) struct CsiControllerConfig {
     create_volume_limit: usize,
     /// Force unstage volume.
     force_unstage_volume: bool,
+    /// Path to the CA certificate file.
+    ca_certificate_path: Option<PathBuf>,
 }
 
 impl CsiControllerConfig {
@@ -50,7 +56,9 @@ impl CsiControllerConfig {
             tracing::warn!(
                 "Force unstage volume is disabled, can trigger potential data corruption!"
             );
-        }
+        };
+
+        let ca_certificate_path: Option<&PathBuf> = args.get_one::<PathBuf>("tls-client-ca-path");
 
         CONFIG.get_or_init(|| Self {
             rest_endpoint: rest_endpoint.into(),
@@ -58,6 +66,7 @@ impl CsiControllerConfig {
             node_selector,
             create_volume_limit,
             force_unstage_volume,
+            ca_certificate_path: ca_certificate_path.cloned(),
         });
         Ok(())
     }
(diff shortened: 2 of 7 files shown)


control-plane/csi-driver/src/bin/controller/client.rs
```diff
@@ -123,16 +123,50 @@ impl RestApiClient {
...
+                    .build_url(url)
+                    .map_err(|error| {
+                        anyhow::anyhow!(
+                            "Failed to create openapi configuration, Error: '{:?}'",
+                            error
+                        )
+                    })?
+            },
+            ("https", None) => {
+                anyhow::bail!("HTTPS endpoint requires a CA certificate path");
```
tiagolobocastro, line 153: Should we keep this as an option? This is what we were thinking of enabling as a first step.
Granted the existing certificate is a hardcoded dummy, but would already be better than plain HTTP as a default option?

Johnaius: If certs aren't provided, we shouldn't be using https as server can't be configured without certs.  It makes more sense to me to not rely on dummy certs at all and I would lobby for cert-manager to be installed and certs created if tls is enabled and remove any reliance on dummy certs.  That's something that needs to be decided- should cert-manager be used? -what are the benefits and drawbacks of dependency on cert-manager and if not used how will users generate certs?

tiagolobocastro: Yeah fair enough, let's keep HTTP by default so at least it's clear.
Let's start as you have it, using the self signed cert on install.
But would be great to see how the flow with cert-manager would look like, maybe on a OEP?

control-plane/csi-driver/src/bin/controller/client.rs
```diff
@@ -123,16 +123,50 @@ impl RestApiClient {
...
-            .map_err(|error| {
-                anyhow::anyhow!(
-                    "Failed to create openapi configuration, Error: '{:?}'",
-                    error
-                )
-            })?;
+        let ca_certificate_path = cfg.ca_certificate_path();
+        let cert = match ca_certificate_path {
+            Some(path) => {
+                let cert = std::fs::read(path).expect("Failed to read certificate file");
```
tiagolobocastro: use map_err to anyhow, like below?

control-plane/csi-driver/src/bin/controller/config.rs
```diff
@@ -17,6 +17,8 @@ pub(crate) struct CsiControllerConfig {
     create_volume_limit: usize,
     /// Force unstage volume.
     force_unstage_volume: bool,
+    /// Path to the CA certificate file.
+    ca_certificate_path: Option<String>,
```
tiagolobocastro:
```suggestion
    ca_certificate_path: Option<std::path::PathBuf>,
```

control-plane/csi-driver/src/bin/controller/main.rs
```diff
@@ -131,6 +131,11 @@ async fn main() -> anyhow::Result<()> {
                 .value_parser(clap::value_parser!(bool))
                 .help("Enable force unstage volume feature")
         )
+        .arg(
+            Arg::new("tls-client-ca-path")
+                .long("tls-client-ca-path")
+                .help("path to the CA certificate file"),
```
tiagolobocastro:
```suggestion
                .help("Path to the CA certificate file")
```

control-plane/csi-driver/src/bin/controller/client.rs
Johnaius: https://github.com/openebs/openebs/pull/3856

tiagolobocastro (commented):
What about [NoCertificateVerification](https://github.com/openebs/mayastor-control-plane/blob/cfc415908c3e797ec8bae36602a6adbfd9567948/openapi/templates/default/tower-hyper/client/configuration.mustache#L301) as well? 

Is that still required? Only for self-signed IIRC?

tiagolobocastro:
> What about [NoCertificateVerification](https://github.com/openebs/mayastor-control-plane/blob/cfc415908c3e797ec8bae36602a6adbfd9567948/openapi/templates/default/tower-hyper/client/configuration.mustache#L301) as well?
> 
> Is that still required? Only for self-signed IIRC?

Ah no, seems that's when not specifying the certificate on the client side

tiagolobocastro:
bors try

tiagolobocastro:
bors delegate

tiagolobocastro:
bors d+

tiagolobocastro:
You can now runs bors try @Johnaius

Johnaius:
bors try

Johnaius:
bors try

k8s/operators/src/pool/main.rs
```diff
@@ -129,14 +129,51 @@ async fn pool_controller(args: ArgMatches) -> anyhow::Result<()> {
...
+    let cert = match ca_certificate_path {
+        Some(path) => {
+            let cert = std::fs::read(path).expect("Failed to read certificate file");
+            Some(cert)
+        }
+        None => None,
+    };
+    let cfg = match (url.scheme(), cert) {
+        ("https", Some(cert)) => {
+            debug!("Attempting TLS connection to {}", url);
```
tiagolobocastro: Do we needs these? Otherwise could also make it more explicit like: 
```suggestion
            tracing::info!(url, "Connecting to the api-rest service with TLS");
```

k8s/operators/src/pool/main.rs
```diff
@@ -129,14 +129,51 @@ async fn pool_controller(args: ArgMatches) -> anyhow::Result<()> {
...
-            anyhow::anyhow!(
-                "Failed to create openapi configuration, Error: '{:?}'",
-                error
+    let ca_certificate_path: Option<&str> = args
+        .get_one::<String>("tls-client-ca-path")
+        .map(|x| x.as_str());
+    // take in cert path and make pem file
+    let cert = match ca_certificate_path {
+        Some(path) => {
+            let cert = std::fs::read(path).expect("Failed to read certificate file");
```
tiagolobocastro: return error up

control-plane/csi-driver/src/bin/controller/client.rs
```diff
@@ -123,16 +123,53 @@ impl RestApiClient {
...
+                        error
+                    )
+                })?;
+                Some(cert)
+            }
+            None => None,
+        };
+        let tower = match (url.scheme(), cert) {
+            ("https", Some(cert)) => {
+                debug!("Attempting TLS connection to {}", url);
```
tiagolobocastro: Same as for the pool

control-plane/csi-driver/src/bin/controller/client.rs
```diff
@@ -123,16 +123,53 @@ impl RestApiClient {
...
-                )
-            })?;
+        let ca_certificate_path = cfg.ca_certificate_path();
+        let cert = match ca_certificate_path {
+            Some(path) => {
+                let cert = std::fs::read(path).map_err(|error| {
+                    anyhow::anyhow!(
+                        "Failed to create openapi configuration, Error: '{:?}'",
+                        error
+                    )
```
tiagolobocastro: Is the `error` mentioning which file path encountered the error? Otherwise could do something like this instead:

```suggestion
                    anyhow::anyhow!("Failed to read certificate file at {}", path.display())
```

tiagolobocastro (commented):
Bors issues a merge with all commits, so could you please either remove debugging and code review commits, or even simply squash them all as as a single commit

Johnaius:
bors try

tiagolobocastro (approved):

Johnaius:
Added csi-node configurations

Johnaius:
bors try

dsharma-dc (approved):

Abhinandan-Purkait (approved):

tiagolobocastro:
Please go ahead and issue bors merge when you're ready @Johnaius 
Thank you!

Johnaius:
bors merge

Johnaius:
> bors merge

image

tiagolobocastro:
odd, this should work, any clues @niladrih ?
for now, I can issue it:
bors merge

Abhinandan-Purkait:
Seems like it did not trigger a merge @tiagolobocastro. I will try once.

Abhinandan-Purkait:
bors merge

Abhinandan-Purkait:
bors ping

Abhinandan-Purkait:
bors cancel

Abhinandan-Purkait:
bors try-

Abhinandan-Purkait (approved):

Abhinandan-Purkait:
bors merge

Abhinandan-Purkait:
How weird, it does work on the other PR but not here. @Johnaius Can you please rebase your PR once?

tiagolobocastro:
bors cancel

tiagolobocastro:
bors merge

PR #928, opened 2025-02-07
HA: ReShutdown the Target

tiagolobocastro:
fix(ha/shutdown): don't remove the current target
    
    Ensure we don't destroy the current target during a shutdown target removal
    call from the ha cluster agent.
    
---

    fix(ha/target/shutdown): reshutdown current target
    
    When the only remaining healthy replica is local to a failed shutdown target,
    the volume will fail to republish.
    In this case, ensure we confirm the shutdown of the target, which will allow
    the volume to be re-published.
    
    Adds a test which simulates this condition.

### DIFF
--- a/control-plane/agents/src/bin/core/volume/operations.rs
+++ b/control-plane/agents/src/bin/core/volume/operations.rs
@@ -498,7 +498,10 @@ impl ResourcePublishing for OperationGuardArc<VolumeSpec> {
         let mut older_nexus = specs.nexus(target_cfg.target().nexus()).await?;
         let mut move_nexus = true;
         let mut nexus_node = None;
-        match healthy_volume_replicas(&spec, &older_nexus.as_ref().node, registry).await {
+        let healthy_replicas_result =
+            healthy_volume_replicas(&spec, &older_nexus.as_ref().node, registry).await;
+        let healthy_replicas = healthy_replicas_result.is_ok();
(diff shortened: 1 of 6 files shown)


Abhinandan-Purkait (approved):
Nit: Typo on the PR description, curren --> current

abhilashshetty04 (approved):

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -742,6 +742,10 @@ impl ResourceShutdownOperations for OperationGuardArc<VolumeSpec> {
         for nexus_res in shutdown_nexuses {
             match nexus_res.operation_guard_wait().await {
                 Ok(mut guard) => {
+                    if self.as_ref().target_uuid() == Some(nexus_res.uuid()) {
```
dsharma-dc, line 745: Why is this check needed **here**? `volume_shutdown_nexuses()` should ensure it doesn't return the current target in the list. It already seems to check shutdown flag.

tiagolobocastro: The current target may be shutdown so we need to account for that. If we change the `volume_shutdown_nexuses` then we need some more changes.
So for now I'd like to keep as is to avoid having to refactor the other bits too.

tiagolobocastro:
bors merge

PR #930, opened 2025-02-07
Cherry-Pick: HA: ReShutdown the Target

tiagolobocastro:
fix(ha/shutdown): don't remove the current target
    
    Ensure we don't destroy the current target during a shutdown target removal
    call from the ha cluster agent.
    
---

    fix(ha/target/shutdown): reshutdown current target
    
    When the only remaining healthy replica is local to a failed shutdown target,
    the volume will fail to republish.
    In this case, ensure we confirm the shutdown of the target, which will allow
    the volume to be re-published.
    
    Adds a test which simulates this condition.


Abhinandan-Purkait (approved):

abhilashshetty04 (approved):

tiagolobocastro:
bors merge

PR #941, opened 2025-02-20
Volume Health Api

tiagolobocastro:
Add new volume health information to better showcase what the current state of the volume is.
The naming might still be a bit confusing and will need another discussion on this.

### DIFF
--- a/control-plane/grpc/proto/v1/volume/volume.proto
+++ b/control-plane/grpc/proto/v1/volume/volume.proto
@@ -58,7 +58,7 @@ message VolumeSpec {
   // Affinity Group related information.
   optional AffinityGroup affinity_group = 10;
   // Volume Content Source i.e the snapshot or a volume.
-  optional VolumeContentSource content_source  = 11;
+  optional VolumeContentSource content_source = 11;
   // Number of snapshots taken on this volume.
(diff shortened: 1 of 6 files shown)


control-plane/grpc/proto/v1/volume/volume.proto
```diff
@@ -179,6 +179,26 @@ message VolumeState {
...
   // volume usage information
   optional VolumeUsage usage = 6;
+  // volume health information
+  optional VolumeHealth health = 7;
+}
+
+// Extra volume information.
+message VolumeHealth {
+  // Indicates if the volume target was cleanly shutdown.
+  bool cleanShutdown = 1;
```
Abhinandan-Purkait, line 189: LGTM, but the word clean might not be very clear, what about consistent or insync?

Abhinandan-Purkait (approved):

control-plane/grpc/proto/v1/volume/volume.proto
tiagolobocastro: Will merge as is for now, and will re-discuss this later on

tiagolobocastro:
bors merge

PR #942, opened 2025-02-25
feat(pstor/watcher): add pstor watcher

tiagolobocastro:
Adds a pstor watcher api which registers watches with the pstor(etcd) and
also ensures the watches are live, reconnecting the stream if necessary.
A single stream per watch is used at the moment.
This might not scale well if we need to have a high watch update rate
though we don't intend it to be so.

### DIFF
--- a/utils/pstor/src/api.rs
+++ b/utils/pstor/src/api.rs
@@ -44,6 +44,11 @@ pub trait StoreKv: Sync + Send + Clone {
     ) -> Result<Vec<(String, Value)>, Error>;
     /// Deletes all key values from a given prefix.
     async fn delete_values_prefix(&mut self, key_prefix: &str) -> Result<(), Error>;
+    /// Get a StoreKv watcher.
+    fn kv_watcher<Ctx: WatcherCtx, W: WatchCallback<Ctx>>(
+        &self,
+        ctx_cb: W,
+    ) -> impl StoreKvWatcher<Ctx> + 'static;
 }
 
 /// Trait defining the operations that can be performed on a key-value store using object semantics.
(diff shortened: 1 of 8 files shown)


utils/pstor/src/api.rs
```diff
@@ -60,6 +65,74 @@ pub trait StoreObj: StoreKv + Sync + Send + Clone {
...
+}
+/// Watch callback function to be triggered on update events for `WatchKey`.
+pub trait WatchCallback<Ctx: WatcherCtx>:
+    Fn(WatchCbArg<Ctx>) -> WatchResult + Send + Sync + 'static
+{
+}
+impl<T, Ctx: WatcherCtx> WatchCallback<Ctx> for T where
+    T: Fn(WatchCbArg<Ctx>) -> WatchResult + Send + Sync + 'static
+{
+}
```
Abhinandan-Purkait: Do we need this trait?

utils/pstor/src/etcd_watcher.rs
```diff
@@ -0,0 +1,497 @@
...
+                tracing::info!("Successfully re-connected the watch stream");
+                WatchState::Watching(w)
+            }
+            state => state,
+        }
+    }
+
+    /// When the connection is lost and we're attempting to reconnect, ensure the requests keep
+    /// getting processed.
+    fn drip_pending_requests(
```
Abhinandan-Purkait, line 367: drip --> drop?

Abhinandan-Purkait (approved):

abhilashshetty04 (approved):

utils/pstor/src/api.rs
tiagolobocastro: We don't strictly need it. Though if we don't we have to basically put callback definition everywhere rather than use `WatchCallback`.
I can do that if you prefer?

utils/pstor/src/etcd_watcher.rs
tiagolobocastro: hmm it's not dropping them right? It's processing some of the pending request to avoid starving them out.
So I thought of drip as it's dripping them :D 
Any other suggestion?

utils/pstor/src/api.rs
Abhinandan-Purkait: No I think it's fine. I just wanted to understand.

utils/pstor/src/etcd_watcher.rs
Abhinandan-Purkait: Okay, yeah drip sounds right in that case.

utils/pstor/src/api.rs
tiagolobocastro: I've removed the custom traits in favor of the bare ones.
If trait aliases are stabilized then we can use aliases.

tiagolobocastro:
bors merge

PR #946, opened 2025-03-11
Add Volume Health Information

tiagolobocastro:
refactor(pstor/etcd-watcher): remove watch status update
    
    The keep alive is already useful for triggering watch restarts in case the
    connection is in a zombie state.
    Also, to address what happens when no progress status is received within
    the timeout, we can also issue an explicit request for progress and update
    the internal rev this way. This is also a much safer way of updating the
    rev as there might some race between getting the status and watches?
    
---

    feat(pstor/etcd): add http2 keep-alives
    
    This helps recover from half open connections automatically.
    
---

    test(etcd): allow for persistent data
    
    Allows testing etcd restarts, whilst persisting its data.
    
---

    refactor: add new volume health info keys
    
    Use new types for volume health info and info prefix, rather than
    modify the existing NexusInfoKey.
    We should also split the NexusInfoKey into non-volume info keys (which
    are only used for testing anyway) and volume info keys.
    
---

    security(GHSA-4p46-pwfr-66x6): update ring dependency
    
    advisory: https://github.com/advisories/GHSA-4p46-pwfr-66x6
    
---

    feat(health): add knob to disable volume health usage
    
---

    test(volume/watcher): add test for the various conditions
    
    Checks the volume health as replicas are faulted, rebuilt and
    targets shutdown cleanly or otherwise.

---

    fix(rest-plugin): show missing target if missing
    
    If the target is meant to be there, but it's missing, state that.
    
---

    feat(rest-plugin): add clean shutdown to volume output
    
---

    feat(rest-plugin): add replica topology health info
    
---

    feat(rest): add health information
    
    Enhance the volume with pstor health information which is retried from
    etcd using watches.
    
---

    feat(core/watcher): add pstor watcher for volume health info
    
    Add a volume health watcher which caches updates from etcd for every volume.
    When etcd entry is deleted, the equivalent entry in the cache is also deleted.
    
    When volume is republished sucessfully, the older target health information are
    also removed from the cache.
    
---

    fix: remove volume health entries on delete
    
    When a volume is deleted, destroy all related health entries.
    It has been observed that upon republished, the previous info
    is not deleted, so even if we fix that a lot of entries may
    already be leaked, so let's ensure we delete all entries for
    a given volume on volume deletion.
    
---

    refactor(pstor/watcher): add missing value on watch callback

### DIFF
--- a/control-plane/agents/src/bin/core/controller/pstor_cache.rs
+++ b/control-plane/agents/src/bin/core/controller/pstor_cache.rs
@@ -6,12 +6,15 @@ use stor_port::{
     pstor::{error::DeserialiseValue, Error, ObjectKey, StorableObject, StoreWatchReceiver},
 };
 
+/// A persistent store cache, which fetches all entries and them allows reading
+/// them via the `pstor::StoreObj` interface.
 #[derive(Clone)]
 pub(crate) struct PStorCache {
     entries: BTreeMap<String, serde_json::Value>,
 }
 
 impl PStorCache {
+    /// Create a new `PStorCache`.
     pub(crate) async fn new(
         pstor: &mut impl pstor::StoreKv,
         page_size: i64,
--- a/control-plane/agents/src/bin/core/controller/pstor_cache.rs
+++ b/control-plane/agents/src/bin/core/controller/pstor_cache.rs
@@ -1,0 +1,51 @@
+use agents::errors::SvcError;
+use snafu::ResultExt;
+use std::collections::BTreeMap;
+use stor_port::{
+    pstor,
+    pstor::{error::DeserialiseValue, Error, ObjectKey, StorableObject, StoreWatchReceiver},
+};
+
+#[derive(Clone)]
+pub(crate) struct PStorCache {
+    entries: BTreeMap<String, serde_json::Value>,
+}
+
+impl PStorCache {
+    pub(crate) async fn new(
+        pstor: &mut impl pstor::StoreKv,
+        page_size: i64,
+        prefix: &str,
+    ) -> Result<Self, SvcError> {
+        let entries = pstor
+            .get_values_paged_all(prefix, page_size)
+            .await?
+            .into_iter()
+            .collect::<BTreeMap<_, _>>();
+
+        Ok(Self { entries })
+    }
+}
+
+#[async_trait::async_trait]
+impl pstor::StoreObj for PStorCache {
+    async fn put_obj<O: StorableObject>(&mut self, _object: &O) -> Result<(), Error> {
+        unimplemented!()
+    }
+
+    async fn get_obj<O: StorableObject>(&mut self, key: &O::Key) -> Result<O, Error> {
+        let key = key.key();
+        match self.entries.get(&key) {
+            Some(kv) => Ok(
+                serde_json::from_value(kv.clone()).context(DeserialiseValue {
+                    value: kv.to_string(),
+                })?,
+            ),
+            None => Err(Error::MissingEntry { key }),
+        }
+    }
+
+    async fn watch_obj<K: ObjectKey>(&mut self, _key: &K) -> Result<StoreWatchReceiver, Error> {
+        unimplemented!()
+    }
+}
--- a/control-plane/agents/src/bin/core/controller/registry.rs
+++ b/control-plane/agents/src/bin/core/controller/registry.rs
@@ -386,6 +386,32 @@ impl Registry {
         }
     }
 
+    /// Serialized delete to the persistent store, with a prefix.
+    pub(crate) async fn delete_kv_prefix<K: StoreKey>(&self, key: &K) -> Result<(), SvcError> {
+        let store = self.store.clone();
+        match tokio::time::timeout(self.store_timeout, async move {
+            let mut store = store.lock().await;
+            let key = key.to_string();
+            Self::op_with_threshold(async move { store.delete_values_prefix(&key).await }).await
+        })
+        .await
+        {
+            Ok(result) => match result {
+                Ok(_) => Ok(()),
+                // already deleted, no problem
+                Err(StoreError::MissingEntry { .. }) => {
+                    tracing::warn!("Entry with key {} missing from store.", key.to_string());
+                    Ok(())
+                }
+                Err(error) => Err(SvcError::from(error)),
+            },
+            Err(_) => Err(SvcError::from(StoreError::Timeout {
+                operation: "Delete".to_string(),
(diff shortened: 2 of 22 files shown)


control-plane/agents/src/bin/core/controller/pstor_cache.rs
```diff
@@ -0,0 +1,51 @@
+use agents::errors::SvcError;
+use snafu::ResultExt;
+use std::collections::BTreeMap;
+use stor_port::{
+    pstor,
+    pstor::{error::DeserialiseValue, Error, ObjectKey, StorableObject, StoreWatchReceiver},
+};
+
+#[derive(Clone)]
+pub(crate) struct PStorCache {
```
Abhinandan-Purkait, line 12: Add a doc comment?

Abhinandan-Purkait (approved):

control-plane/agents/src/bin/core/controller/pstor_cache.rs
tiagolobocastro: Done

Abhinandan-Purkait (approved):

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -109,15 +109,7 @@ impl ResourceLifecycle for OperationGuardArc<VolumeSpec> {
...
-        if let Some(config) = self.as_ref().config() {
-            let nexus_id = config.target().nexus();
-            // Delete the NexusInfo entry persisted by the IoEngine.
-            ResourceSpecsLocked::delete_nexus_info(
-                &NexusInfoKey::new(&Some(self.uuid().clone()), nexus_id),
-                registry,
-            )
-            .await;
-        }
+        self.delete_all_info(registry).await;
```
dsharma-dc: Comment needs to be updated to reflect new behaviour.
Also, suggest renaming to `delete_all_nexusinfo`.

control-plane/stor-port/src/types/v0/store/nexus_persistence.rs
```diff
@@ -55,6 +55,7 @@ pub struct NexusInfoKey {
     volume_id: Option<VolumeId>,
     nexus_id: NexusId,
     mayastor_compat_v1: bool,
+    rm_all_nexuses: bool,
```
dsharma-dc: This seems bit out of place on (one) NexusInfoKey while it intends to operate on volume level(all nexusinfo keys).

tiagolobocastro: Yeah I wasn't sure where to place it, since the logic which builds the exact key is only contained in this type. Any suggestions?

dsharma-dc: How about not setting any nexus_id while creating NexusInfoKey from `delete_all_info`? And then during key creation assume that `volume_id + no nexus_id` means delete all nexus infos for volume. AFAIK, currently the code always expects a nexus_id while volume_id is optional.

control-plane/agents/src/bin/core/volume/operations.rs
tiagolobocastro: Done

control-plane/stor-port/src/types/v0/store/nexus_persistence.rs
tiagolobocastro: Making nexus optional here would require a lot of changes.
Instead I added new types, reused logic which builds key for the nexus info.
Please take a look.

control-plane/agents/src/bin/core/volume/registry.rs
```diff
@@ -106,8 +112,37 @@ impl Registry {
...
+                } else {
+                    info.nr_healthy_replicas().min(1)
+                },
+                online_healthy_replicas,
+                online_clean_replicas: if info.clean_shutdown {
+                    online_healthy_replicas
+                } else {
+                    online_healthy_replicas.min(1)
+                },
+                live_healthy_replicas: healthy
```
dsharma-dc, line 151: what's the difference b/w `online_healthy_replicas` and `live_healthy_replicas` ?

control-plane/agents/src/bin/core/controller/resources/volume/mod.rs
```diff
@@ -55,3 +63,93 @@ macro_rules! volume_span {
...
+        let health = Arc::new(Mutex::new(BTreeMap::new()));
+
+        let health_cln = health.clone();
+        let watcher = store.kv_watcher(move |arg| {
+            let cnt = pstor::WatchResult::Continue;
+
+            if arg.value.is_empty() {
+                match NexusInfoKey::parse_id(arg.updated_key) {
+                    Ok(id) => {
+                        //tracing::warn!(key=%arg.updated_key, "Removing key");
```
dsharma-dc: remove this line?

control-plane/agents/src/bin/core/controller/registry.rs
```diff
@@ -165,6 +168,7 @@ impl Registry {
             inner: Arc::new(RegistryInner {
                 nodes: Default::default(),
                 specs: ResourceSpecsLocked::new(),
+                health: VolumeHealthWatcher::new(&store),
```
dsharma-dc: IIUC, the first time health info is cached blank when registry initializes, i.e. in init_health. And afterwards, this watcher updates those `entries` upon any event on nexusinfo key?
What happens to updated health info upon next registry refresh(cache_period)?

control-plane/agents/src/bin/core/volume/registry.rs
tiagolobocastro: `live_healthy_replicas` are `online_healthy_replicas` which are in use by the volume target (nexus)

control-plane/agents/src/bin/core/controller/resources/volume/mod.rs
tiagolobocastro: doh, indeed

control-plane/agents/src/bin/core/controller/registry.rs
tiagolobocastro: On `new` it's blank but we register the watch, so any updates will start being inserted.
Then on `init_health` we read from etcd and we add to the cache only if the watch hasn't yet inserted them.

> What happens to updated health info upon next registry refresh(cache_period)?

The cache is not used by this health, health is driven by the watch only (ie no polling)

dsharma-dc (approved):

tiagolobocastro:
bors merge

PR #947, opened 2025-03-15
feat(rest-plugin): delete volume by uuid

tiagolobocastro:
Add delete volume subcommand to the rest-plugin utility. The K8s plugin utility will add additional checks, such as ensuring the equivalent PV is already deleted.

### DIFF
--- a/control-plane/plugin/src/resources/mod.rs
+++ b/control-plane/plugin/src/resources/mod.rs
@@ -176,6 +176,55 @@ pub enum GetDrainArgs {
     Nodes,
 }
 
+/// Delete resources.
+#[derive(Debug, clap::Args)]
+pub struct DeleteArgs {
+    /// Ignore error if resource is not found.
+    #[clap(long, short, global = true)]
+    pub ignore_not_found: bool,
+
+    /// Automatically confirm and assume yes for all prompts.
+    #[clap(long, short, global = true)]
+    pub yes: bool,
+
+    #[clap(subcommand)]
+    pub resource: DeleteResources,
+}
+
+/// The type of resources which support the delete operation.
+#[derive(clap::Subcommand, Debug)]
+pub enum DeleteResources {
+    /// Deletes the specified volume resource.
+    Volume {
+        /// The id of the volume to delete.
+        id: VolumeId,
+    },
+}
+
+/// Prompt the user with the given question, and collect a y/n answer.
+/// # NOTE: A newline must be added to ensure it's not a mistake.
+pub fn prompt(question: &str, help: &str) -> Result<bool, Error> {
+    inquire::Confirm::new(question)
+        .with_default(false)
+        .with_help_message(help)
+        .prompt()
+        .map_err(|source| Error::Dialogue { source })
+}
+
+/// Confirm the operation by prompting user with the question, and specify whether the answer has been given
+/// automatically as part of cli args.
+/// In case of negative answer, return an error.
+pub fn confirm(question: &str, help: &str, yes: bool) -> Result<(), Error> {
+    if yes {
+        return Ok(());
+    }
+    if prompt(question, help)? {
+        Ok(())
+    } else {
+        Err(Error::DialogueAbort {})
+    }
+}
+
 /// Tabular Output Tests.
 #[cfg(test)]
 mod tests;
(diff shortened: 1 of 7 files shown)


tiagolobocastro:
First step of https://github.com/openebs/mayastor/issues/1835

Abhinandan-Purkait:
Shouldn't the csi controller garbage collector extend this feature rather than a manual cli command, looks risky to me, how do you ensure we don't end up deleting a wrong volume. Maybe add a double confirmation?

tiagolobocastro:
> Shouldn't the csi controller garbage collector extend this feature rather than a manual cli command, looks risky to me, how do you ensure we don't end up deleting a wrong volume. Maybe add a double confirmation?

What do you mean by "extend this feature"? The whole point is that it's not safe to garbage collect these volumes, it should be a manual procedure.
The kubectl plugin will double check and ensure that the deletion volume does not have an equivalent PV.

By double confirmation, you mean like: "type Y or YES" to confirm?

Abhinandan-Purkait:
> By double confirmation, you mean like: "type Y or YES" to confirm?
 
Yes, just in case it was a wrong id. Also, what happens if someone issues this for an in use volume? Do we fail the delete? 
 
 The reason I pointed can extend csi-controller to do this is because we can do more things, like stale entry cleanup(mounts etc) for these volumes in case any left around.

tiagolobocastro:
> > By double confirmation, you mean like: "type Y or YES" to confirm?
> 
> Yes, just in case it was a wrong id. Also, what happens if someone issues this for an in use volume? Do we fail the delete?

Alright, we can add a prompt.
Yes, we fail to delete if volume is in use.

> The reason I pointed can extend csi-controller to do this is because we can do more things, like stale entry cleanup(mounts etc) for these volumes in case any left around.

Stale entry cleanups would be good to have, but IMHO that needs to come up in a separate task as an OEP.  This change is just to replace a crude portforward+curl, and have this documented.

If we change how we handle deletes, we could do other nifty things as well. There's a slack user which removed volume by mistake, and today that's potentially catastrophic. We could introduce another stage to deletes, where the volume could first be placed in a trash bin for X many days before its automatically deleted for real. This would allow users to recover data within the X days :)

Abhinandan-Purkait:
> If we change how we handle deletes, we could do other nifty things as well. There's a slack user which removed volume by mistake, and today that's potentially catastrophic. We could introduce another stage to deletes, where the volume could first be placed in a trash bin for X many days before its automatically deleted for real. This would allow users to recover data within the X days :)

Yeah, I think we can consider this, having a deletion timestamp on the volumespec and the reconcile to check elapsed from the timestamp and clean it up. Anyway that's a separate discussion. 

So are you adding the confirmation thing in this PR? Otherwise LGTM.

control-plane/plugin/src/resources/mod.rs
```diff
@@ -176,6 +176,27 @@ pub enum GetDrainArgs {
     Nodes,
 }
 
+/// Delete resources.
+#[derive(Debug, clap::Args)]
+pub struct DeleteArgs {
+    /// Ignore error if resource is not found.
+    #[clap(long, short, global = true)]
+    pub ignore_not_found: bool,
```
abhilashshetty04, line 184: Is this flag to ignore volume not present on mayastor and user executing the cmd to delete it? 

I see we check this flag if we get StatusCode::NOT_FOUND from mayastor delete_volume api

abhilashshetty04:
@tiagolobocastro , Where we will check if PV is present for the volume being getting deleted? In plugin or rest?

I feel if we do that validation then we are sure that there might not be any App/Pod using it.  We can add additional confirmation too but maybe not super important if we do PV validation..

control-plane/plugin/src/resources/mod.rs
tiagolobocastro: It's to not exit process with error

Abhinandan-Purkait (approved):

tiagolobocastro:
> @tiagolobocastro , Where we will check if PV is present for the volume being getting deleted? In plugin or rest?
> 
> I feel if we do that validation then we are sure that there might not be any App/Pod using it. We can add additional confirmation too but maybe not super important if we do PV validation..

kubectl-plugin will check if PV is deleted
I've also now added validation with prompt as suggested above
bors merge

PR #949, opened 2025-03-18
feat(encryption): allow storage class to accept encrypted parameter

Abhinandan-Purkait:
- Adds the `encrypted` parameter to be passed via the storage class, possible values are undefined, true, false.

### DIFF
--- a/control-plane/agents/src/bin/core/tests/pool/mod.rs
+++ b/control-plane/agents/src/bin/core/tests/pool/mod.rs
@@ -570,7 +570,7 @@ async fn reconciler_missing_pool_state() {
 
     // create volume to fill up some of the pool space
     for _ in 0..10 {
-        let body = CreateVolumeBody::new(VolumePolicy::default(), 1, 8388608u64, false);
+        let body = CreateVolumeBody::new(VolumePolicy::default(), 1, 8388608u64, false, false);
         let volume = VolumeId::new();
         volumes_api.put_volume(&volume, body).await.unwrap();
     }
@@ -867,7 +867,13 @@ async fn disown_unused_replicas() {
     let volume = volumes_api
         .put_volume(
             &"1e3cf927-80c2-47a8-adf0-95c481bdd7b7".parse().unwrap(),
-            models::CreateVolumeBody::new(models::VolumePolicy::default(), 1, 5242880u64, false),
+            models::CreateVolumeBody::new(
+                models::VolumePolicy::default(),
+                1,
+                5242880u64,
+                false,
+                false,
+            ),
         )
         .await
         .unwrap();
--- a/control-plane/csi-driver/src/context.rs
+++ b/control-plane/csi-driver/src/context.rs
@@ -68,6 +68,8 @@ pub enum Parameters {
     NodeHasTopologyKey,
     #[strum(serialize = "nodeSpreadTopologyKey")]
     NodeSpreadTopologyKey,
+    #[strum(serialize = "encrypted")]
+    Encrypted,
 }
 impl Parameters {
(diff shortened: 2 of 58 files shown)


control-plane/csi-driver/src/context.rs
```diff
@@ -466,6 +472,7 @@ pub struct CreateParams {
     sts_affinity_group: Option<String>,
     clone_fs_id_as_volume_id: Option<bool>,
     max_snapshots: Option<u32>,
+    encrypted: Option<bool>,
```
dsharma-dc, line 475: Can we rather have this only `bool` and ensure a concrete value of boolean on setter side? because we mandatorily need this value beyond this point.

Abhinandan-Purkait: This is the storage class bit, where users can choose to not specify the params they don't want to use. Otherwise they would need to set `encrypted: false` even if they don't care about encryption everytime they create a storage class. Making it an option relaxes that, meaning `unspecifed` or `encrypted: false` will be same eventually.

control-plane/agents/src/bin/core/tests/pool/mod.rs
```diff
@@ -867,7 +867,13 @@ async fn disown_unused_replicas() {
     let volume = volumes_api
         .put_volume(
             &"1e3cf927-80c2-47a8-adf0-95c481bdd7b7".parse().unwrap(),
-            models::CreateVolumeBody::new(models::VolumePolicy::default(), 1, 5242880u64, false),
+            models::CreateVolumeBody::new(
```
tiagolobocastro, line 870: not for this PR;
we should modify the openapi generated code to use builder pattern
I had intended to eventually modify it to match the paperclip v2 generated code but never got around to that

control-plane/rest/openapi-specs/v0_api_spec.yaml
```diff
@@ -2705,11 +2705,15 @@ components:
           type: integer
           format: int32
           minimum: 0
+        encrypted:
+          description: flag indicating whether or not the volume replicas are to be placed on encrypted pool
```
tiagolobocastro: Do we want to be so specific on the description?
Ex: what happens if we introduce encryption on replicas in the future?

```suggestion
        encrypted:
          description: flag indicating whether or not the volume replicas are to be placed on encrypted pool
```

tiagolobocastro (approved):

control-plane/agents/src/bin/core/tests/pool/mod.rs
Abhinandan-Purkait: Yeah, that would be really nice actually, we wouldn't need to change all the code everywhere in tests. Just set the needed ones.

control-plane/rest/openapi-specs/v0_api_spec.yaml
Abhinandan-Purkait: Right, how about `flag to enable data encryption` ?

dsharma-dc (approved):

control-plane/rest/openapi-specs/v0_api_spec.yaml
tiagolobocastro: Yeah I think that's enough

Abhinandan-Purkait: Done

Abhinandan-Purkait:
bors merge

Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors merge

PR #950, opened 2025-03-18
fix(csi/node): support non-multipath enabled hosts

tiagolobocastro:
The csi-node may run on hosts which have nvme enabled, but multipath may be disabled in the nvme_core module.
In this case, there's no nvmeXcYn1 controller available. A regression was introduced on v2.7.2 and v2.8.0 which whilst aiming to fix some issues around controller reconnect for multipath seems to have broken the non-multipath case as it's relying solely on nvmeXcYn1.

The fix is simply to check for the nvmeXn1 if multipath is disabled!

### DIFF
--- a/control-plane/csi-driver/src/bin/node/dev/nvmf.rs
+++ b/control-plane/csi-driver/src/bin/node/dev/nvmf.rs
@@ -82,9 +82,11 @@ impl NvmfAttach {
         enumerator.match_subsystem("block")?;
         enumerator.match_property("DEVTYPE", "disk")?;
 
+        let multipath = utils::check_nvme_core_ana()?;
+
         let mut first_error = Ok(None);
         for device in enumerator.scan_devices()? {
-            match match_device(&device, &key, &self.warn_bad) {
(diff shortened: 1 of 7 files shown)


control-plane/csi-driver/src/bin/node/dev/nvmf.rs
dsharma-dc, line 1: Does this gracefully also handle the case like:
1. multipath = N on host
2. controller namespace is connected
3. controller is lost
4. set multipath = Y on host
5. controller reconnects

tiagolobocastro: If controller is not removed, then you cannot modify multipath because there's no way to do so, you have to remove all nvme drivers and reload them with multipath=Y

Abhinandan-Purkait (approved):

abhilashshetty04 (approved):

dsharma-dc (approved):

tiagolobocastro:
Relates to https://github.com/openebs/mayastor/issues/1838

tiagolobocastro:
bors merge

PR #952, opened 2025-03-18
Cherry-pick PR 950

tiagolobocastro:
The csi-node may run on hosts which have nvme enabled, but multipath may be disabled in the nvme_core module.
In this case, there's no nvmeXcYn1 controller available. A regression was introduced on v2.7.2 and v2.8.0 which whilst aiming to fix some issues around controller reconnect for multipath seems to have broken the non-multipath case as it's relying solely on nvmeXcYn1.

The fix is simply to check for the nvmeXn1 if multipath is disabled!

### DIFF
--- a/control-plane/agents/src/bin/ha/node/main.rs
+++ b/control-plane/agents/src/bin/ha/node/main.rs
@@ -170,7 +170,7 @@ async fn main() -> anyhow::Result<()> {
     }
 
     // Check whether ana is enabled or not.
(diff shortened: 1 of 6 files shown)


Abhinandan-Purkait (approved):

tiagolobocastro:
Relates to https://github.com/openebs/mayastor/issues/1838

tiagolobocastro:
bors merge

PR #953, opened 2025-03-20
feat(encryption): replica scheduling on encrypted pools

Abhinandan-Purkait:
- Schedule volumes on encrypted pools if encryption is asked

### DIFF
--- a/control-plane/grpc/proto/v1/volume/volume.proto
+++ b/control-plane/grpc/proto/v1/volume/volume.proto
@@ -78,7 +78,7 @@ message VolumeSpec {
   }
 
   // Data encryption.
-  optional bool encrypted = 15;
+  optional bool encrypted = 14;
 }
 
 message Metadata {
--- a/control-plane/grpc/proto/v1/volume/volume.proto
+++ b/control-plane/grpc/proto/v1/volume/volume.proto
@@ -76,6 +76,9 @@ message VolumeSpec {
     google.protobuf.StringValue snapshot = 1;
     google.protobuf.StringValue snap_source_vol = 2;
   }
+
+  // Data encryption.
+  optional bool encrypted = 15;
 }
 
 message Metadata {
@@ -291,6 +294,8 @@ message CreateVolumeRequest {
   optional uint64 cluster_capacity_limit = 10;
   // Max snapshots limit per volume.
   optional uint32 max_snapshots = 11;
+  // Data encryption.
+  optional bool encrypted = 12;
 }
 
 // Publish a volume on a node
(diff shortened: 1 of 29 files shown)


Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors try

control-plane/grpc/proto/v1/volume/volume.proto
```diff
@@ -76,6 +76,9 @@ message VolumeSpec {
     google.protobuf.StringValue snapshot = 1;
     google.protobuf.StringValue snap_source_vol = 2;
   }
+
+  // Data encryption.
+  optional bool encrypted = 15;
```
```suggestion
  optional bool encrypted = 14;
```
?

control-plane/stor-port/src/types/v0/transport/replica.rs
```diff
@@ -441,6 +441,8 @@ pub struct CreateReplica {
     pub allowed_hosts: Vec<HostNqn>,
     /// The replica kind, eg: regular or clone.
     pub kind: Option<ReplicaKind>,
+    /// Data encryption.
+    pub encrypted: bool,
```
tiagolobocastro: Does this need option as well?

```suggestion
    pub encrypted: Option<bool>,
```

tiagolobocastro (approved):

control-plane/grpc/proto/v1/volume/volume.proto
Abhinandan-Purkait: Ohh or is it supposed to be 16? Because of the two in between?

control-plane/stor-port/src/types/v0/transport/replica.rs
Abhinandan-Purkait: Not sure, any advantage? It should just be true or false, None is indicative of false only.

dsharma-dc (approved):
for completion, this change needs corresponding io-engine change to set replica state as encrypted as well.

control-plane/grpc/proto/v1/volume/volume.proto
tiagolobocastro: No, I think should be the next number, which is 14?

control-plane/stor-port/src/types/v0/transport/replica.rs
tiagolobocastro: Option would let us know more information, ex: replica coming from node which doesn't know about encryption support.

Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors merge

Abhinandan-Purkait:
bors cancel

Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors merge

PR #954, opened 2025-03-20
add support for formatoptions

abhilashshetty04:
This PR adds 2 new storageClass parameter. `overrideGlobalFormatOpts` and `formatOptions`

formatOptions allows user to pass extra filesystem format flags depending on the `fsType`.

`overrideGlobalFormatOpts` will allows `xfs` users to override global format options set via helm values.

### DIFF
--- a/control-plane/agents/src/bin/core/tests/volume/format_options.rs
+++ b/control-plane/agents/src/bin/core/tests/volume/format_options.rs
@@ -1,0 +1,146 @@
+#![cfg(test)]
+
+use deployer_cluster::ClusterBuilder;
+use stor_port::types::v0::openapi::models;
+use stor_port::types::v0::transport::VolumeId;
+
+use std::{collections::HashMap, time::Duration};
+
+const VOLUME_SIZE: u64 = 32u64 * 1024 * 1024;
+const POOL_SIZE: u64 = 140u64 * 1024 * 1024;
+const BYTES_PER_INODE: u64 = 64u64 * 1024;
+
+struct DeviceDisconnect(nvmeadm::NvmeTarget);
+impl Drop for DeviceDisconnect {
+    fn drop(&mut self) {
+        if self.0.disconnect().is_err() {
+            std::process::Command::new(env!("SUDO"))
+                .args(["nvme", "disconnect-all"])
+                .status()
+                .unwrap();
+        }
+    }
+}
+
+/// Creates a volume and format using formatOption = bytes-per-inode.
+/// Mounts the fs onto the mount path inside csi-node container.
+/// Inspects tune2fs output for the device and ensure inode count is correct.
+#[tokio::test]
+async fn filesystem_volume_format_options() {
+    let cache_period = Duration::from_millis(3);
+    let reconcile_period = Duration::from_millis(3);
+    let cluster = ClusterBuilder::builder()
+        .with_rest(true)
+        .with_io_engines(1)
+        .with_tmpfs_pool_ix(0, POOL_SIZE)
+        .with_csi(true, true)
+        .with_cache_period(&humantime::Duration::from(cache_period).to_string())
+        .with_reconcile_period(reconcile_period, reconcile_period)
+        .build()
+        .await
+        .expect("Cluster to be built");
+
+    let api_client = cluster.rest_v00();
+    let volumes_api = api_client.volumes_api();
+    let vol_id = VolumeId::new();
+    let volume = volumes_api
+        .put_volume(
+            &vol_id,
+            models::CreateVolumeBody::new(
+                models::VolumePolicy::new(true),
+                1,
+                VOLUME_SIZE,
+                false,
+                false,
+            ),
+        )
+        .await
+        .expect("Volume to be created");
+
+    let mut node = cluster
+        .csi_node_client(0)
+        .await
+        .expect("To get node client");
+
+    let mut controller = cluster
+        .csi_controller_client()
+        .await
+        .expect("To get controller client");
+
+    let publish_result = controller
+        .controller_publish_volume(&volume, &cluster.csi_node(0))
+        .await
+        .expect("To publish volume");
+
+    let uri = publish_result
+        .publish_context
+        .get("uri")
+        .expect("To get publish uri")
+        .to_string();
+
+    let _nvme_io_subsys = DeviceDisconnect(nvmeadm::NvmeTarget::try_from(uri).unwrap());
+
+    let mut volume_context = HashMap::new();
+    let flags = format!("-i {}", BYTES_PER_INODE);
+    volume_context.insert("formatOptions".to_string(), flags);
+
+    let vol = volumes_api
+        .get_volume(&vol_id)
+        .await
+        .expect("To get the volume");
+
+    node.node_stage_volume_fs(
+        &vol,
+        "ext4",
+        publish_result.publish_context.clone(),
+        volume_context,
+    )
+    .await
+    .expect("To stage volume");
+
+    node.node_publish_volume_fs(&volume, "ext4", publish_result.publish_context)
+        .await
+        .expect("To publish volume");
+
+    let path = format!("/var/tmp/target/mount/{}", volume.spec.uuid);
+    let expected_inodes = VOLUME_SIZE / BYTES_PER_INODE;
+    let composer = cluster.composer().clone();
+
+    // Gets device backing the mountpath.
+    let findmnt_args: Vec<&str> = vec!["findmnt", "-n", "-o", "SOURCE", "--target", path.as_str()];
+    let (_, device) = composer.exec("csi-node-1", findmnt_args).await.unwrap();
+
+    // Gets tune2fs output of the device.
+    let tune2fs_args: Vec<&str> = vec!["tune2fs", "-l", device.trim()];
+    let (_, tune2fs) = composer.exec("csi-node-1", tune2fs_args).await.unwrap();
+
+    // We expect actual inode count to be lesser as we don't get the actual device size of the
+    // created volume. Its almost 4 to 5 Mb lesser. This affects the number of inode count
+    // when we use -i as it creates inode for each block (65k in this test).
+    let mut inode_count: u64 = 1000;
+    for line in tune2fs.trim().lines() {
+        if line.starts_with("Inode count:") {
+            let i_count = line.split_whitespace().nth(2).unwrap_or("Unknown");
+            inode_count = i_count
+                .parse::<u64>()
+                .expect("To parse inode count into u64");
+            break;
+        }
+    }
+
+    assert!(
+        inode_count < expected_inodes,
+        "inode count was greater then expected"
+    );
+
+    node.node_unpublish_volume(&volume)
+        .await
+        .expect("To unpublish volume");
+    node.node_unstage_volume(&volume)
+        .await
+        .expect("To unstage volume");
+    controller
+        .controller_unpublish_volume(&volume, &cluster.csi_node(0))
+        .await
+        .expect("To Controller Unpublish volume");
+}
--- a/control-plane/csi-driver/src/bin/node/filesystem_ops.rs
+++ b/control-plane/csi-driver/src/bin/node/filesystem_ops.rs
@@ -91,7 +91,7 @@ impl FileSystem {
 #[async_trait]
 pub(crate) trait FileSystemOps: Send + Sync {
     /// Create the filesystem using its fs util.
-    async fn create(&self, device: &str) -> Result<(), Error>;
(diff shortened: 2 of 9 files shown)


Abhinandan-Purkait (approved):

control-plane/csi-driver/src/bin/node/format.rs
```diff
@@ -27,6 +28,9 @@ pub(crate) async fn prepare_device(
         }
         return Ok(());
     }
-    debug!("Creating new filesystem ({}) on device {}", fstype, device);
-    fs_ops.create(device).await
+    info!(
+        "formatting device {} with {} using ({}) additional format options",
+        device, fstype, format_options
+    );
```
niladrih:
```suggestion
    info!(%format_options, "Formatting device {device} with filesystem {fstype}");
```

niladrih (approved):

control-plane/csi-driver/src/bin/node/filesystem_ops.rs
```diff
@@ -257,6 +258,7 @@ impl FileSystemOps for XFs {
         let output = Command::new(binary)
             .args(args)
             .arg(device)
+            .arg(format_options)
```
tiagolobocastro: What happens now if format_options clashes with `MKFS_XFS_ARGS` ?
1. If they set the same parameter to the same value, is this allowed?
2. If they set same parameter to different value, does the latter (format_options) win?
3. Wonder if we need another storage class parameter to dictate whether we keep the global options or not?

abhilashshetty04:
bors try

abhilashshetty04:
bors try

control-plane/agents/src/bin/core/tests/volume/format_options.rs
```diff
@@ -0,0 +1,164 @@
...
+        .await
+        .expect("Failed to unstage volume");
+    controller
+        .controller_unpublish_volume(&volume, &cluster.csi_node(0))
+        .await
+        .expect("Failed to publish volume");
+}
+
+/// Creates files enough to exhaust inode count of mount path present in csi-node-1 container.
+fn exhaust_inodes(path: String, inode_count: u64) -> Result<(), anyhow::Error> {
```
tiagolobocastro: we don't need to do this, just inspect output if tune2fs is what we expect
Also consider adding this test as a BDD test instead

control-plane/csi-driver/src/bin/node/filesystem_ops.rs
abhilashshetty04: 1. Same values are okay to pass for the same parameter.
2. Different value with of same parameter, Latter arg win.
3. For XFS, if user passes format_options then shall we just apply the format_options passed via sc. As its a latest intent. Helm format opts are more of a global/default which applies to all XFS volumes. 

Incase user wants more format options they can specify same default which is used in helm values +  additional args also in sc formatOptions?

control-plane/agents/src/bin/core/tests/volume/format_options.rs
abhilashshetty04: I tried libc statvfs module to get free inodes. However, Since path is mounted inside csi-node tune2fs/libc verification  was not working in test code. Same logic was giving correct inode count in csi-node stage/publish code. So thought of exhausting inodes.

Do you mean with more scenarios in BDD?

tiagolobocastro: You need to run the tune2fs command as a docker exec, like so: https://github.com/openebs/mayastor-control-plane/blob/51ba1c5491c67432db32da6b54d8e3dfc60a2811/tests/bdd/features/csi/node/test_node.py#L847

abhilashshetty04:
bors try

abhilashshetty04:
bors try

abhilashshetty04:
bors try

control-plane/csi-driver/src/bin/node/filesystem_ops.rs
abhilashshetty04: i tried:
 
"mkfs.xfs" "/dev/nvme3n1" "-m" "bigtime=0" "-m" "inobtcount=0" "-m" "bigtime=1" "-m" "inobtcount=1"
 
This fails with error `command failed: -m bigtime option respecified`
 
It seems like sending the same flag is not a problem. We can use -m again and again. I have tried -i (inode ratio 4096/65536) in same command it works.
 
However, here we set value of the flag (bigtime) it doesnt seem to work.

We should document this scenario for xfs users. They can use `overrideGlobalFormatOpts` along with `formatOptions` if they want to change global args.

control-plane/agents/src/bin/core/tests/volume/format_options.rs
```diff
@@ -0,0 +1,157 @@
...
+    )
+    .await
+    .expect("Failed to stage volume");
+    node.node_publish_volume_fs(&volume, "ext4", publish_result.publish_context)
+        .await
+        .expect("Failed to publish volume");
+    let path = format!("/var/tmp/target/mount/{}", volume.spec.uuid);
+    let expected_inodes = VOLUME_SIZE / BYTES_PER_INODE;
+
+    let device = Command::new("docker")
```
tiagolobocastro: can use composer exec

control-plane/csi-driver/src/bin/node/filesystem_vol.rs
```diff
@@ -78,6 +80,20 @@ pub(crate) async fn stage_fs_volume(
         }
     };
 
+    let xfs_global_args = std::env::var("MKFS_XFS_ARGS").unwrap_or_default();
+
+    let combined_format_opts = format!("{} {}", xfs_global_args, format_options);
+
+    if fstype.to_string() == "xfs" {
```
tiagolobocastro: use enum rather than string

control-plane/csi-driver/src/bin/node/filesystem_vol.rs
```diff
@@ -78,6 +80,20 @@ pub(crate) async fn stage_fs_volume(
...
+        let format_options_empty = format_options.is_empty();
+        format_options = match (override_global_format_opts, format_options_empty) {
+            (false, true) => xfs_global_args.as_ref(),
+            (false, false) => combined_format_opts.as_ref(),
+            (true, false) => format_options,
+            (true, true) => "",
+        };
+    }
```
tiagolobocastro: can this whole thing be simplified, ex:
```rust
let mut format_opts = args.format_opts;
if fstype == FsType::Xfs && !override {
format_opts = format!("{global_opts} {format_opts});
}
....
prepare_device(...., format_opts.trim_whitespace())
```

utils/deployer-cluster/src/lib.rs
```diff
@@ -1252,11 +1252,18 @@ impl CsiNodeClient {
...
     ) -> Result<NodeStageVolumeResponse, Error> {
         let mut context = std::collections::HashMap::new();
         context.insert(
             "uri".into(),
-            volume.state.target.as_ref().unwrap().device_uri.to_string(),
+            volume
+                .state
+                .target
+                .as_ref()
+                .expect("Cannot get volume target")
```
tiagolobocastro:
```suggestion
                .expect("volume target to exist")
```

control-plane/csi-driver/src/bin/node/filesystem_ops.rs
```diff
@@ -132,10 +132,14 @@ pub(crate) trait FileSystemOps: Send + Sync {
...
 impl FileSystemOps for Ext4Fs {
-    async fn create(&self, device: &str) -> Result<(), Error> {
+    async fn create(&self, device: &str, format_options: &str) -> Result<(), Error> {
         let binary = "mkfs.ext4";
-        let output = Command::new(binary)
-            .arg(device)
+        let mut cmd = Command::new(binary);
+        cmd.arg(device);
+        if !format_options.is_empty() {
+            cmd.arg(format_options);
```
tiagolobocastro: do we not need to split into a vector like for xfs?

control-plane/agents/src/bin/core/tests/volume/format_options.rs
```diff
@@ -0,0 +1,157 @@
...
+                false,
+            ),
+        )
+        .await
+        .expect("Failed to create volume");
+
+    let mut node = cluster
+        .csi_node_client(0)
+        .await
+        .expect("Failed to get node client");
```
tiagolobocastro:
```suggestion
        .expect("To get node client");
```
same throughout, though a simple unwrap also works for this test code

control-plane/csi-driver/src/bin/node/filesystem_ops.rs
abhilashshetty04: Are you asking should we split here also?

During xfs arg testing noticed that `"mkfs.xfs" "/dev/nvme0n1" "-m bigtime=0 -m inobtcount=0"` resulted in error. 

After splitting into vec it worked.

`"mkfs.xfs" "-m" "bigtime=0" "-m" "inobtcount=0" "/dev/nvme0n1"`

Ive done that for Ext4 and BtrFs also.

abhilashshetty04:
bors try

control-plane/agents/src/bin/core/tests/volume/format_options.rs
```diff
@@ -0,0 +1,146 @@
...
+}
+
+/// This test.
+/// Create a volume and format using formatOption = bytes-per-inode(determines no. of inodes).
+/// Mounts the fs onto the mount path inside csi-node container.
+/// Inspects tune2fs output for the device and ensure inode count is correct.
+#[tokio::test]
+async fn filesystem_volume_format_options() {
+    let cache_period = Duration::from_millis(250);
+    let reconcile_period = Duration::from_millis(3000);
```
tiagolobocastro: why is this set to 3s? If it makes a difference set it to ms as well

control-plane/agents/src/bin/core/tests/volume/format_options.rs
```diff
@@ -0,0 +1,146 @@
...
+        if self.0.disconnect().is_err() {
+            std::process::Command::new(env!("SUDO"))
+                .args(["nvme", "disconnect-all"])
+                .status()
+                .unwrap();
+        }
+    }
+}
+
+/// This test.
```
tiagolobocastro: Reads a bit funny

control-plane/agents/src/bin/core/tests/volume/format_options.rs
```diff
@@ -0,0 +1,146 @@
...
+    let (_, device) = composer.exec("csi-node-1", findmnt_args).await.unwrap();
+
+    // Gets tune2fs output of the device.
+    let tune2fs_args: Vec<&str> = vec!["tune2fs", "-l", device.trim()];
+    let (_, tune2fs) = composer.exec("csi-node-1", tune2fs_args).await.unwrap();
+
+    // We expect actual inode count to be lesser as we don't get the actual device size of the
+    // created volume. Its almost 4 to 5 Mb lesser. This affects the number of inode count
+    // when we use -i as it creates inode for each block (65k in this test).
+    let mut inode_count: u64 = 1000;
```
tiagolobocastro, line 120: ensure it fails if not found in the output of tune2fs

```suggestion
    let mut inode_count: u64 = 0;
```

control-plane/agents/src/bin/core/tests/volume/format_options.rs
```diff
@@ -0,0 +1,146 @@
...
+    for line in tune2fs.trim().lines() {
+        if line.starts_with("Inode count:") {
+            let i_count = line.split_whitespace().nth(2).unwrap_or("Unknown");
+            inode_count = i_count
+                .parse::<u64>()
+                .expect("To parse inode count into u64");
```
tiagolobocastro, line 126:
```suggestion
                .expect("To parse inode count into u64");
            break;
```

tiagolobocastro (approved):

control-plane/agents/src/bin/core/tests/volume/format_options.rs
abhilashshetty04: f tune2fs output is not retrieved for the device code might not reach parse code at line 126. Im expecting the inode_count to be not more then 512. Because for 32Mb device with 65536 blocks/inode ratio we get 512. Due to some reduced bytes on the target/device we get 416 consistently.

Hence kept `inode_count` 1000. If it doesnt get modified from tune2fs parse. The assert will fail.

tiagolobocastro: You can set it as 0 so it's more obvious that it's for that case, or perhaps even make it None, and then set it to Some

abhilashshetty04:
bors merge

PR #955, opened 2025-03-24
fix(ipv6): support nvmf uri host with ipv6

tiagolobocastro:
In rust, in case of ipv6 we must remove the [ ] from the url host. This is because with ipv6 the url is like so:
http://[$ip]:$port

The host returned by the url library is [$ip] which is not a valid ipv6!

For the testing, the python urllib is already removing the [] and so no changes are required there, though I haven't tested running CI on an ipv6 only linux system.

### DIFF
--- a/control-plane/csi-driver/src/bin/node/dev/nvmf.rs
+++ b/control-plane/csi-driver/src/bin/node/dev/nvmf.rs
@@ -128,7 +128,9 @@ impl TryFrom<&Url> for NvmfAttach {
         let hostnqn = hash_query.get("hostnqn").map(ToString::to_string);
 
         Ok(NvmfAttach::new(
-            host.to_string(),
+            host.trim_start_matches("[")
+                .trim_end_matches("]")
+                .to_string(),
             port,
             transport,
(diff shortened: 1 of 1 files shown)


michaelbeaumont:
Just jumping in here cause I got a notification :grimacing:, shouldn't we parse this to an actual IpAddr type like elsewhere in the code?

tiagolobocastro:
> Just jumping in here cause I got a notification 😬, shouldn't we parse this to an actual IpAddr type like elsewhere in the code?

Yeah probably should have been, but the `ConnectArgsBuilder` takes it as a string argument rather than an IpAddr.
Btw there's another PR [here](https://github.com/openebs/mayastor/pull/1840) for the data-plane.

tiagolobocastro:
bors merge

Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

tiagolobocastro:
bors cancel

tiagolobocastro:
bors merge

PR #956, opened 2025-03-26
feat(encryption): add encrypted fields to pool, volume and the replica topology

Abhinandan-Purkait:
- Adds encrypted column to pool.
- Adds encrypted column to volume.
- Adds encrypted column to replica-topology.

### DIFF
--- a/control-plane/agents/src/bin/core/volume/registry.rs
+++ b/control-plane/agents/src/bin/core/volume/registry.rs
@@ -250,6 +250,7 @@ impl Registry {
                     child.map(|c| c.state_reason.clone()),
                     child.and_then(|c| c.rebuild_progress),
                     healthy,
+                    state.encrypted,
                 )
             }
             Err(_) => {
@@ -268,6 +269,8 @@ impl Registry {
                             Some(child.state_reason.clone()),
                             child.rebuild_progress,
                             healthy,
+                            // If state is not available show from spec?
+                            Some(spec.encrypted),
                         );
                     }
                 }
@@ -280,6 +283,8 @@ impl Registry {
                     None,
                     None,
                     healthy,
+                    // If state is not available show from spec?
+                    Some(spec.encrypted),
                 )
             }
         }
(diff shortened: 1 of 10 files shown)


control-plane/agents/src/bin/core/volume/registry.rs
```diff
@@ -268,6 +269,8 @@ impl Registry {
                             Some(child.state_reason.clone()),
                             child.rebuild_progress,
                             healthy,
+                            // If state is not available show from spec?
```
dsharma-dc, line 272: I'd say if state is not available don't show anything.

dsharma-dc (approved):

control-plane/plugin/src/resources/volume.rs
```diff
@@ -71,7 +71,8 @@ impl CreateRow for openapi::models::Volume {
                     VolumeContentSource::snapshot(_) => "Snapshot",
                 }
             })),
-            optional_cell(state.health.as_ref().map(|h| h.clean_shutdown))
+            optional_cell(state.health.as_ref().map(|h| h.clean_shutdown)),
+            self.spec.encrypted
```
tiagolobocastro: Yeah see above the snapped thing,
so maybe here we could do something similar:
```suggestion
            match self.spec.encrypted {
                false => "false",
                true if all_healthy_replicas_encrypted => "true",
                true => "true (requires rebuild)",
},
```
or

```suggestion
            match self.spec.encrypted {
                false => "disabled",
                true if all_healthy_replicas_encrypted => "enabled (secured)",
                true => "enabled (rebuild required)",
},
```

tiagolobocastro (approved):

control-plane/plugin/src/resources/volume.rs
Abhinandan-Purkait: ~Sure. I will take this up on a follow up PR. For now keeping it the way it is.~
Done.

Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors merge

PR #957, opened 2025-04-02
test(encryption): add bdd tests for encryption

Abhinandan-Purkait:
- Adds the feature file for encryption.
- Adds the bdd test for the same.

### DIFF
--- a/tests/bdd/features/encryption/feature.feature
+++ b/tests/bdd/features/encryption/feature.feature
@@ -1,0 +1,27 @@
+Feature: Encryption-at-Rest on Mayastor DiskPool
+
+  Background:
+    Given the product is installed and running
+
+  Scenario Outline: Creating and Importing an encrypted DiskPool
+    Given a user created Secret file containing key parameters with <cipher> and <keysize>
+    When a diskpool is created with this Secret file
+    And the node hosting the pool reboots
+    Then the encrypted disk pool gets imported successfully eventually
+    Examples:
+      | cipher  | keysize |
+      | AesXts  | 128     |
+
+  Scenario Outline: Volume replica scheduling
+    Given a user created Secret file containing key parameters with <cipher> and <keysize>
+    When a diskpool is created with this Secret file
+    Then the encrypted disk pool gets created successfully
+    When a diskpool is created without encryption
+    Then the non encrypted disk pool gets created successfully
+    When a single replica volume is created with encryption
+    Then the replica for encrypted volume should be on encrypted pool
+    When a single replica volume is created without encryption
+    Then the replica for non encrypted volume should be on non encrypted pool
+    Examples:
+      | cipher  | keysize |
+      | AesXts  | 128     |
--- a/tests/bdd/features/encryption/test_encryption.py
+++ b/tests/bdd/features/encryption/test_encryption.py
@@ -1,0 +1,213 @@
+"""Encryption-at-Rest on Mayastor DiskPool feature tests."""
+
+import pytest
+import os
+import json
+
+from pytest_bdd import (
+    given,
+    scenario,
+    then,
+    when,
+    parsers,
+)
+from retrying import retry
+from common.deployer import Deployer, workspace_tmp
+from common.apiclient import ApiClient
+from common.docker import Docker
+from common.operations import Cluster
+from openapi.model.node_status import NodeStatus
+from openapi.model.pool_status import PoolStatus
+from openapi.model.create_pool_body import CreatePoolBody
+from openapi.model.encryption import Encryption
+from openapi.model.create_volume_body import CreateVolumeBody
+from openapi.model.encryption_secret import EncryptionSecret
+from openapi.model.volume_policy import VolumePolicy
+
+ENCR_VOLUME_UUID = "5cd5378e-3f05-47f1-a830-a0f5873a1449"
+NON_ENCR_VOLUME_UUID = "05451170-ac6a-43cb-a92a-e5e3581d1111"
+VOLUME_SIZE = 10485761
+NUM_VOLUME_REPLICAS = 1
+ENCR_POOL = "67cda0eb-1772-4e96-8463-ca686a76196d"
+NON_ENCR_POOL = "d12eaa63-2c4c-4256-a825-fdc426d1fa2d"
+NODE_NAME = "io-engine-1"
+
+SECRET_FILE_NAME = "pool-encr-secret"
+SECRET_FILE_PATH = os.path.join(workspace_tmp(), SECRET_FILE_NAME)
+
+
+@pytest.fixture(scope="module")
+def init():
+    Deployer.start(1)
+    yield
+    Deployer.stop()
+
+
+@pytest.fixture(scope="module")
+def disks():
+    pools = Deployer.create_disks(2, size=100 * 1024 * 1024)
+    yield pools
+    Deployer.cleanup_disks(len(pools))
+
+
+@pytest.fixture(autouse=True)
+@given("an io-engine cluster")
+def setup_io_engine_cluster(init, disks):
+    """an io-engine cluster."""
+    pytest.disks = disks
+    yield
+    Cluster.cleanup()
+
+
+@scenario("feature.feature", "Creating and Importing an encrypted DiskPool")
+def test_creating_and_importing_an_encrypted_diskpool():
+    """Creating and Importing an encrypted DiskPool."""
+
+
+@scenario("feature.feature", "Volume replica scheduling")
+def test_volume_replica_scheduling():
+    """Volume replica scheduling."""
+
+
+@given(
+    parsers.parse(
+        "a user created Secret file containing key parameters with {cipher} and {keysize:d}"
+    )
+)
+def _(cipher, keysize):
+    """a user created Secret file containing key parameters with <cipher> and <keysize>."""
+    if cipher == "AesXts":
+        key1 = generate_key(keysize)
+        key2 = generate_key(keysize)
+    elif cipher == "AesCbc":
+        key1 = generate_key(keysize)
+        key2 = None
+
+    SECRET_DATA = {
+        "cipher": cipher,
+        "key": key1,
+        "key_len": keysize,
+        "key2": key2,
+        "key_len2": keysize if key2 else None,
+    }
+
+    with open(SECRET_FILE_PATH, "w") as file:
+        json.dump(SECRET_DATA, file, indent=4)
+
+    yield
+    if os.path.exists(SECRET_FILE_PATH):
+        os.remove(SECRET_FILE_PATH)
+
+
+@given("the product is installed and running")
+def _():
+    """the product is installed and running."""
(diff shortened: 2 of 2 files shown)


Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors try

tests/bdd/features/encryption/feature.feature
```diff
@@ -0,0 +1,27 @@
...
+  Scenario Outline: Volume replica scheduling
+    Given a user created Secret file containing key parameters with <cipher> and <keysize>
+    When a diskpool is created with this Secret file
+    And another diskpool is created without encryption
+    Then the encrypted disk pool gets created successfully
+    And the non encrypted disk pool gets created successfully
+    When a single replica volume is created with encryption
+    And another single replica volume is created without encryption
+    Then the encryption enabled volume replicas should be on encrypted pool
+    And the non encryption not enabled volume replicas should be on non encrypted pool
```
tiagolobocastro: Maybe reorder in sequence:
When a 
Then
When the
Then the

tests/bdd/features/encryption/feature.feature
```diff
@@ -0,0 +1,27 @@
+Feature: Encryption-at-Rest on Mayastor DiskPool
+
+  Background:
+    Given io-engine is installed and running
```
tiagolobocastro:
```suggestion
    Given the product is installed and running
```

tests/bdd/features/encryption/test_encryption.py
```diff
@@ -0,0 +1,210 @@
...
+    """the encrypted disk pool gets imported successfully eventually."""
+    pool = ApiClient.pools_api().get_node_pool(NODE_NAME, ENCR_POOL)
+    assert pool.state.status == PoolStatus("Online")
+
+
+@then("the encryption enabled volume replicas should be on encrypted pool")
+def _():
+    """the encryption enabled volume replicas should be on encrypted pool."""
+    volume = ApiClient.volumes_api().get_volume(ENCR_VOLUME_UUID)
+    assert next(iter(volume.state.replica_topology.values()))["pool"] == ENCR_POOL
```
tiagolobocastro, line 187: nit: could check all rather than just assume 1 exists

tests/bdd/features/encryption/test_encryption.py
```diff
@@ -0,0 +1,210 @@
...
+    """the non encrypted disk pool gets created successfully."""
+    pool = ApiClient.pools_api().get_node_pool(NODE_NAME, NON_ENCR_POOL)
+    assert pool.state.status == PoolStatus("Online")
+
+
+@then("the non encryption not enabled volume replicas should be on non encrypted pool")
+def _():
+    """the non encryption not enabled volume replicas should be on non encrypted pool."""
+    volume = ApiClient.volumes_api().get_volume(NON_ENCR_VOLUME_UUID)
+    assert next(iter(volume.state.replica_topology.values()))["pool"] == NON_ENCR_POOL
```
tiagolobocastro, line 203: same here

tests/bdd/features/encryption/test_encryption.py
```diff
@@ -0,0 +1,210 @@
...
+
+ENCR_VOLUME_UUID = "5cd5378e-3f05-47f1-a830-a0f5873a1449"
+NON_ENCR_VOLUME_UUID = "05451170-ac6a-43cb-a92a-e5e3581d1111"
+VOLUME_SIZE = 10485761
+NUM_VOLUME_REPLICAS = 1
+ENCR_POOL = "67cda0eb-1772-4e96-8463-ca686a76196d"
+NON_ENCR_POOL = "d12eaa63-2c4c-4256-a825-fdc426d1fa2d"
+NODE_NAME = "io-engine-1"
+
+ROOT_DIR = os.getenv("ROOT_DIR")
```
tiagolobocastro: don't these already exist in helper methods?

tiagolobocastro (approved):

tests/bdd/features/encryption/feature.feature
dsharma-dc:
```suggestion
    And the replicas for non encrypted volume should be on non encrypted pool
```

tests/bdd/features/encryption/test_encryption.py
```diff
@@ -0,0 +1,210 @@
...
+    with open(SECRET_FILE_PATH, "w") as file:
+        json.dump(SECRET_DATA, file, indent=4)
+
+    yield
+    if os.path.exists(SECRET_FILE_PATH):
+        os.remove(SECRET_FILE_PATH)
+
+
+@given("io-engine is installed and running")
+def _():
```
dsharma-dc, line 103: Need to name these functions for clarity. Most of them are un-named currently.

tests/bdd/features/encryption/feature.feature
```diff
@@ -0,0 +1,27 @@
...
+      | AesXts  | 128     |
+
+  Scenario Outline: Volume replica scheduling
+    Given a user created Secret file containing key parameters with <cipher> and <keysize>
+    When a diskpool is created with this Secret file
+    And another diskpool is created without encryption
+    Then the encrypted disk pool gets created successfully
+    And the non encrypted disk pool gets created successfully
+    When a single replica volume is created with encryption
+    And another single replica volume is created without encryption
```
dsharma-dc:
```suggestion
    And a single replica volume is created without encryption
```

tests/bdd/features/encryption/feature.feature
```diff
@@ -0,0 +1,27 @@
...
+    Then the encryption enabled volume replicas should be on encrypted pool
```
dsharma-dc:
```suggestion
    Then the replicas for encrypted volume should be on encrypted pool
```

tests/bdd/features/encryption/test_encryption.py
```diff
@@ -0,0 +1,210 @@
...
+    """the node hosting the pool reboots."""
+    Docker.restart_container(NODE_NAME)
+
+
+@then("the encrypted disk pool gets created successfully")
+def _():
+    """the encrypted disk pool gets created successfully."""
+    pool = ApiClient.pools_api().get_node_pool(NODE_NAME, ENCR_POOL)
+    assert pool.state.status == PoolStatus("Online")
```
dsharma-dc, line 169: also check `pool.state.encrypted` perhaps.

tests/bdd/features/encryption/test_encryption.py
```diff
@@ -0,0 +1,210 @@
...
+
+
+@then("the encrypted disk pool gets imported successfully eventually")
+@retry(wait_fixed=10, stop_max_attempt_number=200)
+def _():
+    """the encrypted disk pool gets imported successfully eventually."""
+    pool = ApiClient.pools_api().get_node_pool(NODE_NAME, ENCR_POOL)
+    assert pool.state.status == PoolStatus("Online")
```
dsharma-dc, line 178: here also, `pool.state.encrypted`

tests/bdd/features/encryption/feature.feature
Abhinandan-Purkait: I think we use that for the system test not component. All the component tests have io-engine.

tests/bdd/features/encryption/test_encryption.py
Abhinandan-Purkait: Assume as in? There is only one pool, so even if I check for all in theory it would not be a valid check, as all replicas aren't going to be on same pool for a given volume.

tests/bdd/features/encryption/test_encryption.py
Abhinandan-Purkait: Couldn't find any, can you point me to it?

tests/bdd/features/encryption/test_encryption.py
Abhinandan-Purkait: Yeah, that's how the new pytest-bdd generates the definitions.

tests/bdd/features/encryption/feature.feature
tiagolobocastro: Is it, I see only one instance of it.

tests/bdd/features/encryption/test_encryption.py
tiagolobocastro: I mean, we're assuming here volume has only 1 replica right?

tests/bdd/features/encryption/test_encryption.py
tiagolobocastro: Yeah I also "complained" about this before, but it seems like the new way of doing it now so I'd say leave as is.

tests/bdd/features/encryption/test_encryption.py
tiagolobocastro: https://github.com/openebs/mayastor-control-plane/blob/d3c4a4e666b4d47de5b58b40182e4d9bb1cf6ddc/tests/bdd/common/deployer.py#L252

tests/bdd/features/encryption/feature.feature
Abhinandan-Purkait: I can change it should be fine.

tests/bdd/features/encryption/test_encryption.py
Abhinandan-Purkait: Well assuming in the sense, we created a single replica volume. So even if I loop over the replica_topologies and check there is only one pool to compare, which would not be the case ideally right if there was supposedly multiple replicas?

tests/bdd/features/encryption/test_encryption.py
Abhinandan-Purkait: Yeah.

tests/bdd/features/encryption/test_encryption.py
tiagolobocastro: Well that holds true today, but we may add more scenarios to the bdd, and currently this step reads as follows:
> the encryption enabled volume replicas should be on encrypted pool

Now, image what happens if someone reuses the same condition on another scenario, without knowing that even though it says "volume replicas", internally it's only checking the first one.

So we can either check all replicas, or rename the condition so it's obvious it can only be used on 1 volume replica, or perhaps even add an assert for spec replica count of 1 which would then fail and also make it obvious

Abhinandan-Purkait: Okay, changing the scenario statement to `the replica for non encrypted volume should be on non encrypted pool` after incorporating Diwakar's comment and adding the spec replica count assert. Thanks

Abhinandan-Purkait:
bors try

dsharma-dc (approved):

Abhinandan-Purkait:
bors merge

PR #958, opened 2025-04-02
Cherry-pick PR 955

tiagolobocastro:
In rust, in case of ipv6 we must remove the [ ] from the url host. This is because with ipv6 the url is like so:
http://[$ip]:$port

The host returned by the url library is [$ip] which is not a valid ipv6!

For the testing, the python urllib is already removing the [] and so no changes are required there, though I haven't tested running CI on an ipv6 only linux system.


Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

PR #961, opened 2025-04-08
make thin by default during snapshot restore

abhilashshetty04:
We do not support snapshot restore into thick volume as of now. With this change user does not need to recreate `thick` sc as `thin` during snapshot restore.  Unless user has specified `thin: false` in sc while resore. We will pass `thin: true`. 

This will fix backup usecase also as currently backup fails if ns has thick provisioned volumes.


tiagolobocastro (approved):
Let's cherry-pick this to 27 and 28?

Abhinandan-Purkait (approved):

abhilashshetty04:
bors merge

abhilashshetty04:
bors try

abhilashshetty04:
bors merge

PR #962, opened 2025-04-08
feat(encryption): restrict different devlink of same device for pool creation

Abhinandan-Purkait:
## Issue

Since we can't read the metadata of an encrypted pool, there isn't any definitive way in data-plane to stop a encrypted pool from getting recreated with a different devlink of a same device with a different pool name.

## Solution

Compare the devlinks across all pools on the specified node, if any pool shares a devlink with the requested disk's devlink/devpath/devname then reject the creation.

## Test

Adds a test to validate the scenario by creating an lvm lv on a loop device and using it's devlinks across multiple creations.

### DIFF
--- a/control-plane/agents/src/bin/core/node/wrapper.rs
+++ b/control-plane/agents/src/bin/core/node/wrapper.rs
@@ -1,3 +1,4 @@
+use crate::controller::io_engine::HostApi;
 use crate::{
     controller::{
         io_engine::{
@@ -15,8 +16,11 @@ use crate::{
     NumRebuilds,
 };
 use agents::{errors::SvcError, eventing::EventWithMeta};
+use async_trait::async_trait;
 use events_api::event::{EventAction, EventCategory, EventMessage, EventMeta, EventSource};
 use grpc::operations::snapshot::SnapshotInfo;
+use stor_port::transport_api::v0::BlockDevices;
+use stor_port::types::v0::transport::GetBlockDevices;
 use stor_port::{
     transport_api::{Message, MessageId, ResourceKind},
     types::v0::{
@@ -36,7 +40,6 @@ use stor_port::{
     },
 };
 
-use async_trait::async_trait;
 use parking_lot::RwLock;
 use std::{future::Future, ops::DerefMut, sync::Arc};
 use tracing::{debug, trace, warn};
@@ -1752,5 +1755,18 @@ impl ReplicaSnapshotApi for Arc<tokio::sync::RwLock<NodeWrapper>> {
 impl From<&NodeWrapper> for NodeState {
     fn from(node: &NodeWrapper) -> Self {
         node.node_state().clone()
+    }
+}
+
+#[async_trait]
+impl HostApi for Arc<tokio::sync::RwLock<NodeWrapper>> {
+    async fn liveness_probe(&self) -> Result<Register, SvcError> {
+        let dataplane = self.read().await.grpc_client().await?;
+        dataplane.liveness_probe().await
+    }
+
+    async fn list_blockdevices(&self, request: &GetBlockDevices) -> Result<BlockDevices, SvcError> {
+        let dataplane = self.read().await.grpc_client().await?;
+        dataplane.list_blockdevices(request).await
     }
 }
--- a/control-plane/agents/src/bin/core/pool/operations_helper.rs
+++ b/control-plane/agents/src/bin/core/pool/operations_helper.rs
@@ -1,3 +1,5 @@
+use crate::controller::io_engine::HostApi;
+use crate::node::wrapper::NodeWrapper;
 use crate::{
     controller::{
         io_engine::PoolApi,
@@ -11,8 +13,8 @@ use crate::{
     node::wrapper::GetterOps,
 };
 use agents::{errors, errors::SvcError};
-use snafu::OptionExt;
-use std::ops::Deref;
+use stor_port::transport_api::ResourceKind;
+use stor_port::types::v0::transport::{GetBlockDevices, PoolDeviceUri};
 use stor_port::types::v0::{
     store::{
         pool::PoolSpec,
@@ -20,6 +22,13 @@ use stor_port::types::v0::{
     },
     transport::{CreatePool, DestroyReplica, NodeId, ReplicaOwners},
 };
+
+use itertools::Itertools;
+use snafu::OptionExt;
+use std::collections::HashSet;
+use std::ops::Deref;
+use std::sync::Arc;
+use tokio::sync::RwLock;
 
 impl OperationGuardArc<PoolSpec> {
     /// Retries the creation of the pool which is being done in the background by the io-engine.
@@ -129,4 +138,87 @@ impl OperationGuardArc<ReplicaSpec> {
             disowners: by,
         }
     }
+}
+
+pub(crate) async fn devlink_preflight_checks(
+    request: &CreatePool,
+    node: Arc<RwLock<NodeWrapper>>,
+    registry: &Registry,
+) -> Result<(), SvcError> {
+    let request_disks: HashSet<String> = request
+        .disks
+        .iter()
+        .map(|disk| utils::disk::normalize_disk(disk.as_str()))
+        .collect();
+
+    let node_pools = registry
+        .get_node_opt_pools(Some(request.node.clone()))
+        .await?;
+
+    if !node_pools.is_empty() {
+        let node_pools_disks: Vec<PoolDeviceUri> = node_pools
+            .into_iter()
+            .filter_map(|pool| pool.spec().map(|spec| spec.disks))
+            .flatten()
+            .collect();
+
+        let node_pools_disks_normalized: HashSet<String> = node_pools_disks
+            .iter()
+            .map(|disk| utils::disk::normalize_disk(disk.as_str()))
+            .collect();
+
+        let common_disks: HashSet<_> = request_disks
+            .intersection(&node_pools_disks_normalized)
+            .cloned()
+            .collect();
+
+        // Same devpaths or devlinks should be rejected.
+        if !common_disks.is_empty() {
+            return Err(SvcError::InUse {
+                kind: ResourceKind::Block,
+                id: common_disks.iter().join(","),
+            });
+        }
+
+        let node_block_devices = node
+            .list_blockdevices(&GetBlockDevices {
+                node: request.node.clone(),
+                all: true,
+            })
+            .await?
+            .into_inner();
+
+        let matched_devices: Vec<_> = node_block_devices
+            .iter()
+            .filter(|device| {
+                request_disks.contains(&device.devname)
+                    || request_disks.contains(&device.devpath)
+                    || device
+                        .devlinks
+                        .iter()
+                        .any(|link| request_disks.contains(link))
+            })
+            .collect();
+
+        // If the requested disk was not found, that could be because it could be a malloc or a file,
+        // in that case ignore and move ahead to allow tests. If it is an actual disk that was not
+        // detected by blockdevice api then the disk might not be visible to io-engine, so let it fail from io-engine
+        // rather than bailing out from control-plane to keep the behaviour as before.
+        if !matched_devices.is_empty() {
+            if let Some(conflict) = matched_devices.iter().find(|bd| {
+                node_pools_disks_normalized.contains(&bd.devname)
+                    || node_pools_disks_normalized.contains(&bd.devpath)
+                    || bd
+                        .devlinks
+                        .iter()
+                        .any(|link| node_pools_disks_normalized.contains(link))
+            }) {
+                return Err(SvcError::InUse {
+                    kind: ResourceKind::Block,
+                    id: conflict.devname.clone(),
+                });
+            }
+        }
+    }
+    Ok(())
 }
--- a/control-plane/agents/src/bin/core/pool/operations_helper.rs
+++ b/control-plane/agents/src/bin/core/pool/operations_helper.rs
@@ -24,6 +24,7 @@ use stor_port::types::v0::{
 };
 
 use itertools::Itertools;
+use regex::Regex;
 use snafu::OptionExt;
 use std::collections::HashSet;
 use std::ops::Deref;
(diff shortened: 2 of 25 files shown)


Abhinandan-Purkait:
bors try

control-plane/agents/src/bin/core/node/wrapper.rs
```diff
@@ -1754,3 +1757,16 @@ impl From<&NodeWrapper> for NodeState {
...
 }
+
+#[async_trait]
+impl HostApi for Arc<tokio::sync::RwLock<NodeWrapper>> {
+    async fn liveness_probe(&self) -> Result<Register, SvcError> {
+        let dataplane = self.read().await.grpc_client().await?;
+        dataplane.liveness_probe().await
+    }
+
+    async fn list_blockdevices(&self, request: &GetBlockDevices) -> Result<BlockDevices, SvcError> {
```
tiagolobocastro, line 1768: Was this not already in place? How is the pool operator getting the block list?

control-plane/agents/src/bin/core/pool/operations_helper.rs
```diff
@@ -130,3 +138,73 @@ impl OperationGuardArc<ReplicaSpec> {
...
+
+pub(crate) async fn reject_same_devlink(
+    request: &CreatePool,
+    node: Arc<RwLock<NodeWrapper>>,
+    registry: &Registry,
+) -> Result<(), SvcError> {
+    let request_disks: HashSet<String> = request
+        .disks
+        .iter()
+        .map(|disk| utils::disk::normalize_disk(disk.as_str()))
```
tiagolobocastro, line 152: this won't work with pcie unbound devices?

control-plane/agents/src/bin/core/pool/operations_helper.rs
```diff
@@ -130,3 +138,73 @@ impl OperationGuardArc<ReplicaSpec> {
...
+    // rather than bailing out from control-plane to keep the behaviour as before.
+    if !matched_devices.is_empty() {
+        let node_pools = registry
+            .get_node_opt_pools(Some(request.node.clone()))
+            .await?;
+
+        if !node_pools.is_empty() {
+            let node_pools_disks: Vec<PoolDeviceUri> = node_pools
+                .iter()
+                .filter_map(|pool| pool.spec().map(|spec| spec.disks.clone()))
```
tiagolobocastro: if you reorder the `let` you can then consume iterator for spec.disks

control-plane/agents/src/bin/core/tests/pool/mod.rs
```diff
@@ -1130,3 +1132,197 @@ async fn slow_create() {
...
+            .expect("Destroying pool-2 should succeed");
+
+        client
+            .pool()
+            .destroy(&DestroyPool::from(create_pool_3_request.clone()), None)
+            .await
+            .expect_err("Destroying pool-3 should fail");
+    }
+
+    async fn write_secret_file() -> std::io::Result<()> {
```
tiagolobocastro: use Drop to delete the secret as well

deployer/src/infra/io_engine.rs
```diff
@@ -57,6 +57,12 @@ impl ComponentAction for IoEngine {
             .with_bind(&host_tmp, "/host/tmp")
             .with_bind("/var/run/dpdk", "/var/run/dpdk");
 
+            if options.mount_host_dev_udev {
+                spec = spec
+                    .with_bind("/dev", "/dev")
```
tiagolobocastro, line 62: can you not use `with_io_engine_devices` ? see test `slow_create`

deployer/src/lib.rs
```diff
@@ -351,6 +351,9 @@ pub struct StartOptions {
     /// Add custom RUST_LOG_SILENCE env to all containers.
     #[clap(long)]
     rust_log_silence: Option<String>,
+    /// Mount host dev and udev.
+    #[clap(long, default_value = "false")]
```
tiagolobocastro: default should be false already

```suggestion
    #[clap(long)]
```

utils/utils-lib/src/lib.rs
```diff
@@ -13,6 +13,7 @@ pub use version_info::{version_info as version_info_inner, VersionInfo};
 
 /// Byte conversion helpers.
 pub mod bytes;
+pub mod disk;
```
tiagolobocastro, line 17: missing doc comment

utils/utils-lib/src/disk.rs
```diff
@@ -0,0 +1,13 @@
+use url::Url;
+
+/// Normalize the disks if they have a schema, we dont want to change anything
+/// or do any error checking -- the loop will converge to the error state eventually
+pub fn normalize_disk(disk: &str) -> String {
```
dsharma-dc, line 5: We can do the `scheme()` check here to reject `pci` device uri. 
Also, do we always get a disk path that can be interpreted as file path? I think we do. Only then `to_file_path` will work for all cases.

control-plane/agents/src/bin/core/pool/operations_helper.rs
```diff
@@ -130,3 +138,73 @@ impl OperationGuardArc<ReplicaSpec> {
...
+    node: Arc<RwLock<NodeWrapper>>,
+    registry: &Registry,
+) -> Result<(), SvcError> {
+    let request_disks: HashSet<String> = request
+        .disks
+        .iter()
+        .map(|disk| utils::disk::normalize_disk(disk.as_str()))
+        .collect();
+
+    let bds = node
```
dsharma-dc:
```suggestion
    let node_block_devices = node
```

utils/utils-lib/src/disk.rs
dsharma-dc: As discussed, let's not block pci, rather let it go and create. Put an additional check before devlinks check - to check simply if the current request diskpath is present in any existing pool spec as-is. This will catch pci uri usage.

control-plane/agents/src/bin/core/node/wrapper.rs
Abhinandan-Purkait: It was getting the client and doing it manually. I have changed it now. PTAL.

control-plane/agents/src/bin/core/pool/operations_helper.rs
Abhinandan-Purkait: Have made the changes as discussed.

control-plane/agents/src/bin/core/tests/pool/mod.rs
Abhinandan-Purkait: Done

control-plane/agents/src/bin/core/pool/operations_helper.rs
Abhinandan-Purkait: Done

deployer/src/infra/io_engine.rs
Abhinandan-Purkait: No, doesn't work without that, the blockdevice api can't find the device, without this.

utils/utils-lib/src/disk.rs
Abhinandan-Purkait: Done.

Abhinandan-Purkait:
bors try

control-plane/agents/src/bin/core/pool/operations_helper.rs
```diff
@@ -130,3 +139,86 @@ impl OperationGuardArc<ReplicaSpec> {
         }
     }
 }
+
+pub(crate) async fn devlink_preflight_checks(
```
dsharma-dc, line 144: As discussed, we shall do the changes to block kernel devpath in the io-engine as that'll be easier/feasible to do there via `file` command equivalents from `std::fs` crate.

utils/utils-lib/src/disk.rs
```diff
@@ -0,0 +1,13 @@
...
+    Url::parse(disk).map_or(disk.to_string(), |u| {
+        u.to_file_path()
```
dsharma-dc, line 13: As discussed, this will cause pci address to be parsed as a filepath which will incorrect with a `/`. For pci scheme we may want to use Url.host() or similar function.

Abhinandan-Purkait:
bors try

control-plane/agents/src/bin/core/pool/operations_helper.rs
```diff
@@ -130,3 +142,100 @@ impl OperationGuardArc<ReplicaSpec> {
...
+    let request_disks: HashSet<String> = request
+        .disks
+        .iter()
+        .map(|disk| utils::disk::normalize_disk(disk.as_str()))
+        .collect();
+
+    if request_disks
+        .iter()
+        // Only attempt to validate if it starts with "/dev".
+        .filter(|disk| disk.starts_with("/dev"))
```
dsharma-dc: The regex is validating this anyway.

control-plane/agents/src/bin/core/pool/operations_helper.rs
Abhinandan-Purkait: Have added the regex and check here. Also have added a cli to disable the check, for those who want to use the same.

control-plane/agents/src/bin/core/pool/operations_helper.rs
Abhinandan-Purkait: This is to do the check only on devices starting with /dev, othewise something like `"malloc:///disk?size_mb=100"` or `/tmp/mydisk` or `pcie:///0000.1.2.3` will also get rejected which we don't want.

utils/utils-lib/src/disk.rs
Abhinandan-Purkait: I have made the method return canonical pcie now, so it will always be the same, because normalize runs on both the compare candidates.

utils/utils-lib/src/disk.rs
```diff
@@ -0,0 +1,35 @@
...
+    Url::parse(&disk).map_or(disk.to_string(), |u| {
+        u.to_file_path()
+            .unwrap_or_else(|_| disk.into())
+            .as_path()
+            .display()
+            .to_string()
+    })
+}
+
+fn canonical_pcie_address(addr: &str) -> Option<String> {
```
dsharma-dc, line 21: It must be `pcie://` if I'm not wrong. The scheme usually has two slashes. The third slash with other schemes like aio is part of the path following scheme.

Abhinandan-Purkait: Right. Will make that change.

Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors try

control-plane/agents/src/bin/core/tests/pool/mod.rs
```diff
@@ -1130,3 +1134,264 @@ async fn slow_create() {
...
+                std::thread::spawn(move || {
+                    let _ = std::fs::remove_file(path);
+                    println!("Removed secretfile");
+                });
+            }
+        }
+    }
+}
+
+#[tokio::test]
```
tiagolobocastro: Doesn't need to be async
Also since this is unittest, add it where `DEVLINK_REGEX`  lives

utils/utils-lib/src/disk.rs
```diff
@@ -0,0 +1,35 @@
+use url::Url;
+
+/// Normalize the disks if they have a schema, we dont want to change anything
+/// or do any error checking -- the loop will converge to the error state eventually
+pub fn normalize_disk(disk: &str) -> String {
+    let disk = if disk.starts_with("pcie://") {
+        canonical_pcie_address(disk).unwrap_or_else(|| disk.to_string())
```
tiagolobocastro, line 7: if it's pcie can't we just return the `canonical_pcie_address` if we change that to return just the "normalized" content?

utils/utils-lib/src/disk.rs
```diff
@@ -0,0 +1,35 @@
...
+        u.to_file_path()
+            .unwrap_or_else(|_| disk.into())
+            .as_path()
+            .display()
+            .to_string()
+    })
+}
+
+fn canonical_pcie_address(addr: &str) -> Option<String> {
+    let stripped = addr.strip_prefix("pcie://")?;
```
tiagolobocastro: Shouldn't we strip the 3 `///` here? otherwise the parts[0] would have a slash or no?

utils/utils-lib/src/disk.rs
```diff
@@ -0,0 +1,35 @@
...
+    Url::parse(&disk).map_or(disk.to_string(), |u| {
+        u.to_file_path()
+            .unwrap_or_else(|_| disk.into())
+            .as_path()
+            .display()
+            .to_string()
+    })
+}
+
+fn canonical_pcie_address(addr: &str) -> Option<String> {
```
tiagolobocastro, line 21: Should add a unittest for this function

control-plane/agents/src/bin/core/tests/pool/mod.rs
Abhinandan-Purkait: Done.

utils/utils-lib/src/disk.rs
Abhinandan-Purkait: I think it's better to keep the functionalities separate. Like `canonical_pcie_address` does only canonical conversion and `normalize_disk` does the normalize.

utils/utils-lib/src/disk.rs
Abhinandan-Purkait: Yes, done.

utils/utils-lib/src/disk.rs
Abhinandan-Purkait: Done.

Abhinandan-Purkait:
bors try

tiagolobocastro (approved):

dsharma-dc (approved):

Abhinandan-Purkait:
bors merge

Abhinandan-Purkait:
bors merge

Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
bors merge

PR #967, opened 2025-04-16
feat: allow setting encryption property on volume spec

dsharma-dc:
This way of setting encryption property should be used with caution when the user knows what they are doing, e.g. planned manual migration of pools from non-encrypted to encrypted.

### DIFF
--- a/control-plane/grpc/proto/v1/volume/volume.proto
+++ b/control-plane/grpc/proto/v1/volume/volume.proto
@@ -391,6 +391,7 @@ message SetVolumePropertyRequest {
 message VolumeProperty {
   oneof attr {
     MaxSnapshotValue max_snapshots = 1;
+    bool encrypted = 2;
   }
 }
 // Max snapshots limit per volume.
--- a/control-plane/plugin/src/resources/mod.rs
+++ b/control-plane/plugin/src/resources/mod.rs
@@ -91,6 +91,13 @@ pub enum SetPropertyResources {
 pub enum SetVolumeProperties {
     /// Max snapshot limit per volume.
     MaxSnapshots { max_snapshots: u32 },
+    #[clap(
+        about = "Encryption required for volume.\n\x1b[31m\x1b[1mCAUTION:\x1b[0m Use carefully during for example non-encrypted to encrypted pool migration. Refer to documentation for more details."
+    )]
+    Encryption {
+        #[clap(action = clap::ArgAction::Set)]
+        enabled: bool,
+    },
 }
 
 /// The types of resources that support cordoning.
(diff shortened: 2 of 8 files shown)


control-plane/grpc/proto/v1/volume/volume.proto
```diff
@@ -391,6 +391,7 @@ message SetVolumePropertyRequest {
 message VolumeProperty {
   oneof attr {
     MaxSnapshotValue max_snapshots = 1;
+    bool encrypted = 2;
```
tiagolobocastro, line 394: this should be a tri-state for the soft encryption during migrations

tiagolobocastro (commented):
This seems like more of a `feat` rather than a `chore`

control-plane/grpc/proto/v1/volume/volume.proto
dsharma-dc: > this should be a tri-state for the soft encryption during migrations

As discussed, we'll have this as boolean and introduce a helm variable to interpret `Prefer` mode in combination with this.

abhilashshetty04:
> e.g. planned manual migration of pools from non-encrypted to encrypted.

It is also applicable for just moving volume from non-encrypted to encrypted pool? 

User might not want to change pool to encrypted. Just move volume to encypted pool.

dsharma-dc:
> > e.g. planned manual migration of pools from non-encrypted to encrypted.
> 
> It is also applicable for just moving volume from non-encrypted to encrypted pool?
> 
> User might not want to change pool to encrypted. Just move volume to encypted pool.

Yes it's applicable for that scenario. Essentially when for example a few pools are migrated, user may pick and choose few volumes as candidates to become encrypted using those migrated pools.

Abhinandan-Purkait (approved):

control-plane/plugin/src/resources/mod.rs
```diff
@@ -91,6 +91,13 @@ pub enum SetPropertyResources {
 pub enum SetVolumeProperties {
     /// Max snapshot limit per volume.
     MaxSnapshots { max_snapshots: u32 },
+    #[clap(
+        about = "Encryption required for volume.\n\x1b[31m\x1b[1mCAUTION:\x1b[0m Use carefully during for example non-encrypted to encrypted pool migration. Refer documentation for more details."
```
tiagolobocastro:
```suggestion
        about = "Encryption required for volume.\n\x1b[31m\x1b[1mCAUTION:\x1b[0m Use carefully during for example non-encrypted to encrypted pool migration. Refer to documentation for more details."
```

tiagolobocastro (approved):

dsharma-dc:
bors merge

PR #968, opened 2025-04-20
feat(scaling volume): prefer cordoned nodes while removing replicas

Abhinandan-Purkait:
Currently on scaling down if a replica exists on a node which is cordoned we don't seem to consider that replica explicitly for removal. With this change we will prefer replicas on cordoned nodes first.

### DIFF
--- a/control-plane/agents/src/bin/core/controller/scheduling/mod.rs
+++ b/control-plane/agents/src/bin/core/controller/scheduling/mod.rs
@@ -168,6 +168,7 @@ impl ChildSorters {
         match Self::sort_by_health(a, b) {
             Ordering::Equal => match Self::sort_by_child(a, b) {
                 Ordering::Equal => {
+                    // Remove replicas from nodes which are cordoned with most priority.
                     // remove mismatched topology replicas first
                     if let (Some(a), Some(b)) = (a.valid_node_topology(), b.valid_node_topology()) {
                         match a.cmp(b) {
@@ -181,6 +182,15 @@ impl ChildSorters {
                             Ordering::Equal => {}
                             _else => return _else,
                         }
+                    }
+
+                    match if let (Some(a), Some(b)) = (a.node_spec(), b.node_spec()) {
+                        b.cordoned().cmp(&a.cordoned())
+                    } else {
+                        a.node_spec().is_some().cmp(&b.node_spec().is_some())
+                    } {
+                        Ordering::Equal => {}
+                        _else => return _else,
                     }
 
                     let childa_is_local = !a.spec().share.shared();
--- a/control-plane/agents/src/bin/core/controller/scheduling/resources/mod.rs
+++ b/control-plane/agents/src/bin/core/controller/scheduling/resources/mod.rs
@@ -7,6 +7,7 @@ use stor_port::types::v0::{
     store::{
         nexus_child::NexusChild,
         nexus_persistence::{ChildInfo, NexusInfo},
+        node::NodeSpec,
         replica::ReplicaSpec,
         snapshots::replica::ReplicaSnapshot,
         volume::VolumeSpec,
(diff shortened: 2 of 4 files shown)


Abhinandan-Purkait:
The priority is in this order:
1. Health
2. Child Status, rebuilding
3. Node Cordon
4. Node Topology
5. Pool Topology
6. Child Locality

Any thoughts on the order?

abhilashshetty04 (approved):

control-plane/agents/src/bin/core/controller/scheduling/mod.rs
```diff
@@ -168,6 +168,18 @@ impl ChildSorters {
...
+                    if let (Some(a), Some(b)) = (a.node_spec(), b.node_spec()) {
+                        match a.cordoned().cmp(&b.cordoned()) {
+                            Ordering::Less => {
+                                return Ordering::Greater;
+                            }
+                            Ordering::Equal => {}
+                            Ordering::Greater => {
+                                return Ordering::Less;
+                            }
+                        }
```
tiagolobocastro:
```suggestion
                        match b.cordoned().cmp(&a.cordoned()) {
                            Ordering::Equal => {}
                            _else => return _else,
                        }
```

control-plane/agents/src/bin/core/controller/scheduling/mod.rs
```diff
@@ -168,6 +168,18 @@ impl ChildSorters {
         match Self::sort_by_health(a, b) {
             Ordering::Equal => match Self::sort_by_child(a, b) {
                 Ordering::Equal => {
+                    // Remove replicas from nodes which are cordoned with most priority.
+                    if let (Some(a), Some(b)) = (a.node_spec(), b.node_spec()) {
```
tiagolobocastro: Looks this this spec is optional, meaning a replica may point to a node which does not exist anymore?
In such case should a replica without node spec always be preferred to be removed?

tiagolobocastro:
> The priority is in this order:
> 
>     1. Health
> 
>     2. Child Status, rebuilding
> 
>     3. Node Cordon
> 
>     4. Node Topology
> 
>     5. Pool Topology
> 
>     6. Child Locality
> 
> 
> Any thoughts on the order?

hmm should the topology take precedence over the cordon?

tiagolobocastro (approved):

Abhinandan-Purkait:
> > The priority is in this order:
> > ```
> > 1. Health
> > 
> > 2. Child Status, rebuilding
> > 
> > 3. Node Cordon
> > 
> > 4. Node Topology
> > 
> > 5. Pool Topology
> > 
> > 6. Child Locality
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Any thoughts on the order?
> 
> hmm should the topology take precedence over the cordon?

Probably, yes I think. Should I do that?

control-plane/agents/src/bin/core/controller/scheduling/mod.rs
Abhinandan-Purkait: Should it not be opposite, the way I did? false is Lesser than true, that means the one with true should be Ordering::Lower to be preffered more?

control-plane/agents/src/bin/core/controller/scheduling/mod.rs
Abhinandan-Purkait: Okay, we can do that.

control-plane/agents/src/bin/core/controller/scheduling/mod.rs
tiagolobocastro: See b swapped with a, so it's the same as what you had, with less lines.

tiagolobocastro:
> > > The priority is in this order:
> > > ```
> > > 1. Health
> > > 
> > > 2. Child Status, rebuilding
> > > 
> > > 3. Node Cordon
> > > 
> > > 4. Node Topology
> > > 
> > > 5. Pool Topology
> > > 
> > > 6. Child Locality
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Any thoughts on the order?
> > 
> > 
> > hmm should the topology take precedence over the cordon?
> 
> Probably, yes I think. Should I do that?

Yeah let's do that then

control-plane/agents/src/bin/core/controller/scheduling/mod.rs
Abhinandan-Purkait: Done.

Abhinandan-Purkait:
bors try

dsharma-dc (approved):

Abhinandan-Purkait:
bors merge

PR #971, opened 2025-04-30
[OEP 3916]: First pass at volume cloning implementation

adam-lithus:
**🚨 Huge disclaimer 🚨**

* I'm not a rust developer
* This was coded with Claude 3.7 Thinking
* I was careful with prompting, but it could still be a bunch of nonsense
* My main goal is to do the leg-work for someone who actually knows what they are doing (@tiagolobocastro )
* If this isn't useful then that is totally fine

### DIFF
--- a/control-plane/csi-driver/src/bin/controller/controller.rs
+++ b/control-plane/csi-driver/src/bin/controller/controller.rs
@@ -294,6 +294,11 @@ impl VolumeTopologyMapper {
     }
 }
 
+enum VolumeContentSourceInfo {
+    Snapshot(Uuid),
+    Volume(Uuid),
+}
+
 #[tonic::async_trait]
 impl rpc::csi::controller_server::Controller for CsiControllerSvc {
     #[instrument(err, fields(volume.uuid = tracing::field::Empty), skip(self))]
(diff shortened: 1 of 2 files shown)


control-plane/csi-driver/src/bin/controller/controller.rs
```diff
@@ -294,185 +294,466 @@ impl VolumeTopologyMapper {
...
     async fn create_volume(
         &self,
         request: tonic::Request<CreateVolumeRequest>,
     ) -> Result<tonic::Response<CreateVolumeResponse>, tonic::Status> {
-        let args = request.into_inner();
+        let request = request.into_inner();
+        debug!("create_volume parameters: {:#?}", request);
 
-        let _permit = self.create_volume_permit().await?;
+        let name = request.name.clone();
```
Abhinandan-Purkait, line 312: We seem to have removed the permit from here. We shouldn't do that. More over I think we should not change the order of any existing validations, that can change failure behaviour.

control-plane/csi-driver/src/bin/controller/controller.rs
```diff
@@ -294,185 +294,466 @@ impl VolumeTopologyMapper {
...
+                match RestApiClient::get_client().get_volume(&source_volume_uuid).await {
+                    Ok(_source_volume) => {
+                        // Create a temporary snapshot of the source volume
+                        let temp_snapshot_id = Uuid::new_v4();
+
+                        // Create the temporary snapshot
+                        let snapshot = match RestApiClient::get_client()
+                            .create_volume_snapshot(&source_volume_uuid, &temp_snapshot_id)
+                            .await 
+                        {
```
Abhinandan-Purkait, line 638: I would suggest to keep the other code as it is and just change this section in this order:
1. Get source volume, if exists --> next step, else bail.
2. If exists create a snapshot, if fails --> delete the snapshot and bail.
3. If succeeds get the snapshot object, if exists, create restore.
4. If fails then delete the snapshot and restore, and bail.
5. If succeeds, goes through usual success flow.

Let's hear from @tiagolobocastro and @dsharma-dc

Abhinandan-Purkait (changes requested):

Abhinandan-Purkait:
Hey @adam-lithus quick nudge, are you still working on this or intend to?

PR #973, opened 2025-05-02
fix: dsp migration was not considering topology

Abhinandan-Purkait:
- There was a bug where on v1beta2 to v1beta3 migration the topology were set as null, which is fixed now.


tiagolobocastro (approved):

abhilashshetty04 (approved):

Abhinandan-Purkait:
bors merge

PR #978, opened 2025-05-08
chore(ci): inject github token post nix installation

pchandra19:
This PR address the security advisory where the github token was being exposed.
As a solution: 
- We are first installing nix (without gh token) and then injecting the token in a seperate step.

Tested here : https://github.com/openebs/mayastor-control-plane/actions/runs/14901217891/job/41853358307

Please note : 
- Testing was carried out by disabling "if failure" flag so that the report could be generated 
- I have re-enabled the journalctl logs to present in the artifacts post testing.
- The if failure flag has been re-enabled.

### DIFF
--- a/scripts/ci-report.sh
+++ b/scripts/ci-report.sh
@@ -10,8 +10,7 @@ set -eu
 mkdir -p "$ROOT_DIR/ci-report"
 cd "$ROOT_DIR/ci-report"
 
-# Disable until we find how some CI envs are being included
-journalctl --since="$CI_REPORT_START_DATE" -o short-precise > journalctl.txt
+journalctl --since="$CI_REPORT_START_DATE" -o short-precise | sed -E 's/ghs_[^[:space:]]+/ghs_***/g' > journalctl.txt
 journalctl -k -b0 -o short-precise > dmesg.txt
 lsblk -tfa > lsblk.txt
 $SUDO nvme list -v > nvme.txt
--- a/scripts/ci-report.sh
+++ b/scripts/ci-report.sh
@@ -11,7 +11,7 @@ mkdir -p "$ROOT_DIR/ci-report"
 cd "$ROOT_DIR/ci-report"
 
 # Disable until we find how some CI envs are being included
-# journalctl --since="$CI_REPORT_START_DATE" -o short-precise > journalctl.txt
+journalctl --since="$CI_REPORT_START_DATE" -o short-precise > journalctl.txt
 journalctl -k -b0 -o short-precise > dmesg.txt
 lsblk -tfa > lsblk.txt
 $SUDO nvme list -v > nvme.txt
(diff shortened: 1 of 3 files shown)


scripts/ci-report.sh
```diff
@@ -11,7 +11,7 @@ mkdir -p "$ROOT_DIR/ci-report"
 cd "$ROOT_DIR/ci-report"
 
 # Disable until we find how some CI envs are being included
-# journalctl --since="$CI_REPORT_START_DATE" -o short-precise > journalctl.txt
+journalctl --since="$CI_REPORT_START_DATE" -o short-precise > journalctl.txt
```
tiagolobocastro: What if we just sed the token out here as discussed on the advisory?

pchandra19: We could sed it too as an extra measure. But I believe with the above approach we have already addressed the root cause for the problem and ensure the gh_token is not logged in journalctl.
I will modify the above with 
journalctl --since="$CI_REPORT_START_DATE" -o short-precise | \
sed -E 's/ghs_[^[:space:]]+/ghs_***/g' > journalctl.txt

tiagolobocastro: With the above you ensure it won't be logged with the nix-installer action, but as we modify the workflow, other actions may eventually do the same. Use sed would be a way of removing it completely.
With the sed I think we can also skip the changes to the installer, and just let it be as is.

pchandra19: Oh, I just pushed the commit with the change. So you want me to remove the previous changes as well or keep it as is since I am reducing the permissions on token as well with the previous change ?

tiagolobocastro: I think reducing the permissions we should, and as a rule of thumb we should always try to set as minimal as possible from now on.
The setting of token to "" and then resetting I think we can probably remove now, but we can let others review and comment as well. CC @Abhinandan-Purkait @niladrih

pchandra19: I'm looking at it this way: the fix I’ve implemented addresses the root cause in this specific case, and also serves as a reference for how we can avoid token leaks in the future. Any new action or step we develop should take these considerations into account.

So, this fix is largely precautionary and should be treated as a standard practice going forward. The journalctl masking is more of a "cure" — something applied post-event. I've included both preventive and corrective measures, and I’d prefer to keep it that way.

That said, I’m open to feedback — let others review it, and based on the consensus, I’ll apply any suggested changes.

Abhinandan-Purkait (approved):
LGTM

scripts/ci-report.sh
Abhinandan-Purkait: The current changes look good to me.  +1 on that.

avishnu (approved):

scripts/ci-report.sh
avishnu: The changes seem fine to me.

pchandra19: As discussed, we'll keep th entire change (root cause fix, permissions and post fix-journalctl). I am going ahead and merging the PR.

pchandra19:
bors merge

PR #983, opened 2025-05-18
fix: allow scaling up of volumes of affinity group

Abhinandan-Purkait:
### Issue
Volumes that are part of the affinity group consider restriction while scaling up for first time.

### Fix
The fix relaxes the restriction on first scale up.

### DIFF
--- a/control-plane/agents/src/bin/core/controller/scheduling/affinity_group.rs
+++ b/control-plane/agents/src/bin/core/controller/scheduling/affinity_group.rs
@@ -1,7 +1,7 @@
 use crate::controller::{registry::Registry, resources::ResourceUid};
 use std::collections::HashMap;
 use stor_port::types::v0::{
-    store::volume::{AffinityGroupSpec, VolumeSpec},
+    store::volume::{AffinityGroupSpec, VolumeOperation, VolumeSpec},
     transport::{NodeId, PoolId},
 };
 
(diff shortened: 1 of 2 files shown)


control-plane/agents/src/bin/core/controller/scheduling/affinity_group.rs
```diff
@@ -12,32 +12,36 @@ pub(crate) fn get_restricted_nodes(
...
-            // for that volume.
-            let node_ids = specs.volume_replica_nodes(volume);
-            // Add a new node in the list if it doesn't already exist.
-            let filtered_nodes: Vec<NodeId> = node_ids
-                .into_iter()
-                .filter(|id| !restricted_nodes.contains(id))
-                .collect();
-            restricted_nodes.extend(filtered_nodes);
-        }
+    // Since num_replicas in VolumeSpec is equal to 1 when scaling up from 1 replica to 2 replicas
```
tiagolobocastro, line 15: But SetReplica(2) can also happen when scaling down?

Abhinandan-Purkait: This code path is not a part of the decrease volume replica flow.

Abhinandan-Purkait: This only gets called from increase volume replicas

control-plane/agents/src/bin/core/controller/scheduling/affinity_group.rs
```diff
@@ -12,32 +12,36 @@ pub(crate) fn get_restricted_nodes(
...
+        .operation
+        .as_ref()
+        .map(|op| op.operation.clone())
+    {
+        return Vec::new();
+    }
+
+    // Only apply restriction when the volume currently has 1 replica
+    if volume_spec.num_replicas != 1 {
+        return Vec::new();
```
dsharma-dc: This will be concise perhaps?

```suggestion
if matches!(
        volume_spec.operation.as_ref().map(|op| op.operation.clone()),
        Some(VolumeOperation::SetReplica(2))
    ) || volume_spec.num_replicas != 1
{
    return Vec::new();
}
```

control-plane/agents/src/bin/core/controller/scheduling/affinity_group.rs
```diff
@@ -12,32 +12,36 @@ pub(crate) fn get_restricted_nodes(
...
     }
-    restricted_nodes
+
+    let specs = registry.specs();
+    // List of restricted nodes, which already host replicas from the volumes of the affinity group
+    affinity_group_spec
```
dsharma-dc, line 32: hmm and this block has become too concise :). Please ensure to test this logic block once for sanity.

dsharma-dc (approved):

control-plane/agents/src/bin/core/controller/scheduling/affinity_group.rs
Abhinandan-Purkait: I will extend my current to validate we are able to scale all the way up to 3 and then come back to 2, as we can't scale down beyond 2 at the moment as per initial design.

tiagolobocastro: Oh I see, fair enough then

tiagolobocastro (approved):
Pls cherry-pick to 2.8 and develop

control-plane/agents/src/bin/core/controller/scheduling/affinity_group.rs
Abhinandan-Purkait: Yeah, I have tested it. Let's see if the CI passes.

Abhinandan-Purkait:
bors merge

PR #985, opened 2025-05-19
feat(encryption): relaxed replica scheduling during pool migration

dsharma-dc:
This change handles the new option prefer-encrypted-pools enabled via helm chart. Once migration activity is finished, the option MUST be disabled from the helm chart.

### DIFF
--- a/control-plane/agents/src/bin/core/controller/registry.rs
+++ b/control-plane/agents/src/bin/core/controller/registry.rs
@@ -116,6 +116,8 @@ pub(crate) struct RegistryInner<S: Store> {
     etcd_max_page_size: i64,
     /// Allow pool creation using non-persistent devlinks.
     allow_non_persistent_devlinks: bool,
+    /// Prefer encrypted pools for volume replicas.
+    encrypted_pools_soft_scheduling: bool,
 }
 
 impl Registry {
@@ -143,6 +145,7 @@ impl Registry {
         etcd_max_page_size: i64,
         no_volume_health: bool,
         allow_non_persistent_devlinks: bool,
+        encrypted_pools_soft_scheduling: bool,
     ) -> Result<Self, SvcError> {
         let store_endpoint = Self::format_store_endpoint(&store_url);
         tracing::info!("Connecting to persistent store at {}", store_endpoint);
@@ -203,6 +206,7 @@ impl Registry {
                 ha_disabled: ha_enabled,
                 etcd_max_page_size,
                 allow_non_persistent_devlinks,
+                encrypted_pools_soft_scheduling,
             }),
         };
         registry.init().await?;
@@ -352,6 +356,11 @@ impl Registry {
         if let Some(h) = self.health() {
             h.retain(retain)
         }
+    }
+    /// If we should enforce encrypted pool requirement softly, i.e. if encrypted pool not
+    /// found, go ahead with non-encrypted one.
+    pub(crate) fn encryption_preference_soft(&self) -> bool {
+        self.encrypted_pools_soft_scheduling
     }
 
     /// Serialized write to the persistent store.
--- a/control-plane/agents/src/bin/core/controller/scheduling/volume_policy/pool.rs
+++ b/control-plane/agents/src/bin/core/controller/scheduling/volume_policy/pool.rs
@@ -56,7 +56,11 @@ impl PoolBaseFilters {
     }
     /// Should only attempt to use encrypted pools.
     pub(crate) fn encrypted(request: &GetSuitablePoolsContext, item: &PoolItem) -> bool {
-        request.encrypted() == item.pool.encrypted
+        let use_this = request.encrypted() == item.pool.encrypted;
+        if !use_this && request.encrypted() && request.registry().encryption_preference_soft() {
+            return !use_this;
+        }
+        use_this
     }
     /// Return true if the pool has enough capacity to resize the replica by the requested
     /// value.
--- a/control-plane/agents/src/bin/core/controller/scheduling/volume_policy/simple.rs
+++ b/control-plane/agents/src/bin/core/controller/scheduling/volume_policy/simple.rs
@@ -44,8 +44,7 @@ impl ResourcePolicy<AddVolumeReplica> for SimplePolicy {
             .filter(affinity_group::SingleReplicaPolicy::replica_anti_affinity)
             .filter_param(&self, SimplePolicy::min_free_space)
             .filter_param(&self, SimplePolicy::pool_overcommit)
-            // sort pools in order of total weight of certain field values.
-            .sort_ctx(SimplePolicy::sort_by_weights)
+            .sort_ctx(SimplePolicy::sort_pools)
     }
 }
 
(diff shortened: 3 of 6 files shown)


control-plane/agents/src/bin/core/controller/scheduling/volume_policy/pool.rs
```diff
@@ -56,7 +56,15 @@ impl PoolBaseFilters {
     }
     /// Should only attempt to use encrypted pools.
     pub(crate) fn encrypted(request: &GetSuitablePoolsContext, item: &PoolItem) -> bool {
-        request.encrypted() == item.pool.encrypted
+        let use_this = request.encrypted() == item.pool.encrypted;
+        if !use_this && request.registry().encryption_preference_soft() {
+            tracing::info!(
```
tiagolobocastro: let's no swarm the logs, on startup should be fine

control-plane/agents/src/bin/core/controller/scheduling/volume_policy/thick.rs
```diff
@@ -81,6 +80,31 @@ impl ThickPolicy {
             |item| item.len().into(),
         )
     }
+
+    /// A note about sort upon encryption status:
+    /// If prefer_encrypted_pools is set in registry, it means system is under some
+    /// kind of migration for volumes from non-encrypted to encrypted pools. In such case,
+    /// give preference to encrypted pool if available. During such migration, volume can
+    /// possibly have some replicas on encrypted pool and some on non-encrypted.
+    pub(crate) fn sort_pools(
```
tiagolobocastro, line 89: Is this function the same, can it be made generic?

dsharma-dc: > Is this function the same, can it be made generic?

It's following the existing semantics of other functions that are placed under both policies. We could refactor but we'll have to do all, so not on this change I'd say.

control-plane/agents/src/bin/core/controller/scheduling/volume_policy/thick.rs
```diff
@@ -81,6 +80,31 @@ impl ThickPolicy {
...
+        request: &GetSuitablePoolsContext,
+        a: &PoolItem,
+        b: &PoolItem,
+    ) -> std::cmp::Ordering {
+        match request.registry().encryption_preference_soft() {
```
dsharma-dc, line 94: hmm, apparently the logic isn't working correctly. While scaling up a volume that has spec set to `encrypted`, I'd expect an encrypted pool to picked up for replica, if there is one encrypted and one non-encrypted pool available. Somehow, I see non-encrypted is picked up. Example: pool-on-node-0 is non-encrypted, and pool-on-node-3 is encrypted. Perhaps comparator ordering issue.

```
    at control-plane/agents/src/bin/core/controller/scheduling/volume_policy/pool.rs:61
    in core::volume::service::set_volume_replica with request: SetVolumeReplica { uuid: VolumeId(576637c6-5ac9-4927-afdf-1e1f44092ff7, "576637c6-5ac9-4927-afdf-1e1f44092ff7"), replicas: 3 }, volume.uuid: 576637c6-5ac9-4927-afdf-1e1f44092ff7

  2025-05-22T09:39:06.086216Z  INFO core::volume::specs: volume_pool_candidates: [PoolWrapper { state: PoolState { node: NodeId("node-0-352937"), id: PoolId("pool-on-node-0"), disks: [PoolDeviceUri("aio:///dev/disk/by-id/scsi-SHC_Volume_102610434?uuid=125f24a0-8fc6-4a98-b411-c65a9ab17883")], status: Online, capacity: 10724835328, used: 25165824, committed: Some(3221225472), encrypted: false }, replicas: [], committed: 3221225472, free_space: 10699669504 }, PoolWrapper { state: PoolState { node: NodeId("node-3-352937"), id: PoolId("pool-on-node-3"), disks: [PoolDeviceUri("aio:///dev/disk/by-id/scsi-SHC_Volume_102610436?uuid=ac206e27-ef1e-45e8-83ea-52f794962728")], status: Online, capacity: 10724835328, used: 0, committed: Some(0), encrypted: true }, replicas: [], committed: 0, free_space: 10724835328 }]
```

dsharma-dc: This is fixed now. Needed to sort in reverse order.

control-plane/agents/src/bin/core/controller/scheduling/volume_policy/simple.rs
```diff
@@ -127,6 +126,31 @@ impl SimplePolicy {
...
+    /// possibly have some replicas on encrypted pool and some on non-encrypted.
+    pub(crate) fn sort_pools(
+        request: &GetSuitablePoolsContext,
+        a: &PoolItem,
+        b: &PoolItem,
+    ) -> std::cmp::Ordering {
+        match request.registry().encryption_preference_soft() {
+            true => match b.pool.encrypted().partial_cmp(&a.pool.encrypted()) {
+                Some(Ordering::Greater) => Ordering::Greater,
+                Some(Ordering::Less) => Ordering::Less,
```
Abhinandan-Purkait, line 143:
```suggestion
                Some(order) => order,
```
Will this work?

Abhinandan-Purkait (approved):

Abhinandan-Purkait:
It would be good if we can validate the stsAffinityGroup combo here.

control-plane/agents/src/bin/core/controller/scheduling/volume_policy/simple.rs
dsharma-dc: > Will this work?

It'll work most likely but all across agent-core code we have explicitly matched, so let's keep consistent.

tiagolobocastro: It wouldn't work because below we have the equal?

control-plane/agents/src/bin/core/main.rs
```diff
@@ -150,6 +150,12 @@ pub(crate) struct CliArgs {
     /// Allow pools to be created using non persistent devlinks.
     #[clap(long)]
     allow_non_persistent_devlink: bool,
+
+    /// Prefer to use encrypted pools for volume replicas if the this is set in addition to
+    /// encryption on volume spec. Applicable if the volume was initially provisioned via
+    /// a non-encryption storage class.
+    #[clap(long)]
+    prefer_encrypted_pools: bool,
```
tiagolobocastro: nit: parameter name itself could be misunderstood, maybe something more like this:
`encrypted_pools_soft_scheduling` ?

control-plane/agents/src/bin/core/controller/scheduling/volume_policy/pool.rs
```diff
@@ -56,7 +56,11 @@ impl PoolBaseFilters {
     }
     /// Should only attempt to use encrypted pools.
     pub(crate) fn encrypted(request: &GetSuitablePoolsContext, item: &PoolItem) -> bool {
-        request.encrypted() == item.pool.encrypted
+        let use_this = request.encrypted() == item.pool.encrypted;
+        if !use_this && request.registry().encryption_preference_soft() {
+            return !use_this;
+        }
+        use_this
```
tiagolobocastro: We should do this only for encrypted volumes?

```suggestion
        let use_this = request.encrypted() == item.pool.encrypted;
        if !use_this && request.encrypted() && request.registry().encryption_preference_soft() {
            return true;
        }
        use_this
```

dsharma-dc:
> It would be good if we can validate the stsAffinityGroup combo here.

Sure. I'm testing that. Regular testing has been successful with 4 nodes and 9 volumes. All volumes eventually migrated.

control-plane/agents/src/bin/core/controller/scheduling/volume_policy/pool.rs
dsharma-dc: > We should do this only for encrypted volumes?

The current logic will give benefit to non-encypted volumes too, in case non-encrypted pools are scarce. e.g stsAffinityGroup I was testing scales up fine from 1 -> 2 replica, which otherwise will fail. wdyt.

tiagolobocastro: possibly could be useful yes though may not be always desireable and we'd need to change the sorting to prefer the correct one, ie encrypted pools to encrypted volumes, with soft fallback to non-encrypted pools and non-encrypted pools to non-encrypted volumes with fallback to encrypted pools

And if so, maybe this should then become a multiple option, rather than a single bool

dsharma-dc: ok, let's not complicate with multiple option. Will restrict this to only encrypted.

control-plane/agents/src/bin/core/controller/scheduling/volume_policy/simple.rs
dsharma-dc: not changing here.

control-plane/agents/src/bin/core/controller/registry.rs
```diff
@@ -168,6 +171,7 @@ impl Registry {
                     description: "Product v1 prefix detection failed".to_string(),
                 })?;
 
+        tracing::info!("Flag - encrypted_pools_soft_scheduling: {encrypted_pools_soft_scheduling}");
```
tiagolobocastro: This may not be necessary, IIRC we already print all cli args at startup?

tiagolobocastro (approved):

dsharma-dc:
bors merge

Abhinandan-Purkait:
bors cancel

dsharma-dc:
bors merge

PR #991, opened 2025-05-29
feat(stsAffinityGroup): honor stsAffinity on backup restores via external tools

Abhinandan-Purkait:
### Problem
If a backup is created and restored via external data populator, like velero, the sts affinity group is not honored if while restore the intermediate volumes for data population is not a part of an sts.

### Solution
Annotate the source pvc upon succesfully assigning it to an affinity group and while restoring read the same annotation to figure out the same.

### DIFF
--- a/control-plane/csi-driver/src/bin/controller/controller.rs
+++ b/control-plane/csi-driver/src/bin/controller/controller.rs
@@ -394,12 +394,21 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
                 debug!(
                     "Volume {volume_uuid} already exists and is compatible with requested config"
                 );
+                let sts_affinity_group_name = context.sts_affinity_group().await?;
+                if let Some(info) = sts_affinity_group_name {
+                    info.set_sts_affinity_annotation().await?;
+                    debug!(
+                        volume.uuid = volume_uuid,
+                        "Volume successfully patched with affinity group annotation"
+                    );
+                }
             }
             // If the volume doesn't exist, create it.
             Err(ApiClientError::ResourceNotExists(_)) => {
                 let volume_topology = context_into_topology(&context);
 
-                let sts_affinity_group_name = context.sts_affinity_group();
+                let sts_affinity_group_info = context.sts_affinity_group().await?;
+                let affinity_group_name = sts_affinity_group_info.as_ref().map(|info| info.name());
                 let max_snapshots = context.max_snapshots();
                 let encrypted = context.encrypted().unwrap_or_default();
 
@@ -418,7 +427,7 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
                                 size,
                                 volume_topology,
                                 thin,
-                                sts_affinity_group_name.clone().map(AffinityGroup::new),
+                                affinity_group_name.map(AffinityGroup::new),
                                 max_snapshots,
                                 encrypted,
                             )
@@ -432,7 +441,7 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
                                 size,
                                 volume_topology,
                                 thin,
-                                sts_affinity_group_name.clone().map(AffinityGroup::new),
+                                affinity_group_name.map(AffinityGroup::new),
                                 max_snapshots,
                                 encrypted,
                             )
@@ -448,12 +457,13 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
                     volume_context.insert("fsId".to_string(), volume_uuid.clone());
                 }
 
-                if let Some(ag_name) = sts_affinity_group_name {
+                if let Some(info) = sts_affinity_group_info {
                     debug!(
                         volume.uuid = volume_uuid,
-                        volume.affinity_group = ag_name,
+                        volume.affinity_group = info.name(),
                         "Volume successfully created"
                     );
+                    info.set_sts_affinity_annotation().await?;
                 } else {
                     debug!(volume.uuid = volume_uuid, "Volume successfully created");
                 }
--- a/control-plane/csi-driver/src/context.rs
+++ b/control-plane/csi-driver/src/context.rs
@@ -1,4 +1,7 @@
 use crate::filesystem::FileSystem;
+use k8s_openapi::api::core::v1::PersistentVolumeClaim;
+use kube::api::{Patch, PatchParams};
+use kube::{Api, Client};
 use regex::Regex;
 use std::{
     collections::HashMap,
@@ -6,9 +9,11 @@ use std::{
     num::ParseIntError,
     str::{FromStr, ParseBoolError},
 };
+use stor_port::platform;
 use stor_port::types::v0::openapi::models::VolumeShareProtocol;
 use strum_macros::{AsRefStr, Display, EnumString};
 use tracing::log::warn;
+use tracing::{debug, trace};
 use utils::K8S_STS_PVC_NAMING_REGEX;
 
 use uuid::{Error as UuidError, Uuid};
@@ -472,6 +477,8 @@ pub(crate) fn validate_topology_params(
     Ok(())
 }
 
+const ANNOTATION_KEY: &str = "openebs.io/stsAffinityGroup";
+
 /// Volume Creation parameters.
 #[allow(dead_code)]
 #[derive(Debug)]
@@ -479,10 +486,12 @@ pub struct CreateParams {
     publish_params: PublishParams,
     share_protocol: VolumeShareProtocol,
     replica_count: u8,
-    sts_affinity_group: Option<String>,
+    sts_affinity_group: Option<bool>,
     clone_fs_id_as_volume_id: Option<bool>,
     max_snapshots: Option<u32>,
     encrypted: Option<bool>,
+    pvc_name: Option<String>,
+    pvc_namespace: Option<String>,
 }
 impl CreateParams {
     /// Get the `Parameters::PublishParams` value.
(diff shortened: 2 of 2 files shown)


control-plane/csi-driver/src/bin/controller/controller.rs
```diff
@@ -394,12 +394,21 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
                 debug!(
                     "Volume {volume_uuid} already exists and is compatible with requested config"
                 );
+                let sts_affinity_group_name = context.sts_affinity_group().await?;
+                if let Some(ag_name) = sts_affinity_group_name {
+                    debug!(
+                        volume.uuid = volume_uuid,
+                        volume.affinity_group = ag_name,
+                        "Volume successfully created"
```
tiagolobocastro: this seems inaccurate, volume was already created as above?

control-plane/csi-driver/src/context.rs
```diff
@@ -514,6 +516,106 @@ impl CreateParams {
...
+        let pvc_name = match &self.pvc_name {
+            Some(name) => name,
+            None => return Ok(None),
+        };
+        let pvc_namespace = match &self.pvc_namespace {
+            Some(ns) => ns,
+            None => return Ok(None),
+        };
+
+        let client = Client::try_default()
```
tiagolobocastro, line 541: this should only be enabled if platform is K8s

control-plane/csi-driver/src/context.rs
```diff
@@ -514,6 +516,106 @@ impl CreateParams {
...
+        if let Some(value) = pvc
+            .metadata
+            .annotations
+            .as_ref()
+            .and_then(|a| a.get(ANNOTATION_KEY))
+        {
+            tracing::trace!(
+                "Generated affinity group name: {} from annotation: {}",
+                value,
+                ANNOTATION_KEY
```
```suggestion
                "Found affinity group name: {value} from annotation: {ANNOTATION_KEY}"
```

control-plane/csi-driver/src/context.rs
```diff
@@ -514,6 +516,106 @@ impl CreateParams {
...
+            );
+            return Ok(Some(value.to_string()));
+        }
+
+        let sts_affinity_group_name = generate_sts_affinity_group_name(pvc_name, pvc_namespace);
+        if let Some(ref name) = sts_affinity_group_name {
+            tracing::trace!(
+                "Annotation: {} was not found, generated name: {} using the pvc name",
+                ANNOTATION_KEY,
+                name
```
```suggestion
                "Annotation: {ANNOTATION_KEY} was not found, generated name: {name} using the pvc name"
```

control-plane/csi-driver/src/context.rs
```diff
@@ -514,6 +516,106 @@ impl CreateParams {
...
+
+        let client = Client::try_default()
+            .await
+            .map_err(|e| tonic::Status::aborted(e.to_string()))?;
+
+        let pvc_name = match &self.pvc_name {
+            Some(name) => name,
+            None => {
+                return Err(tonic::Status::invalid_argument(format!(
+                    "Could not patch pvc with {}:{} annotation, pvc name missing",
```
tiagolobocastro: Inline all vars

tiagolobocastro (commented):
Have we lost the error message if name/namespace are not available but sts is enabled on the SC ?
Should there be an enable flag to control this?

control-plane/csi-driver/src/bin/controller/controller.rs
Abhinandan-Purkait: That's fine, it's still a create call from controller perspective. The above debug message already logs the existence.

Abhinandan-Purkait: I will tweak the message.

control-plane/csi-driver/src/context.rs
Abhinandan-Purkait: Done.

Abhinandan-Purkait:
> Have we lost the error message if name/namespace are not available but sts is enabled on the SC ? Should there be an enable flag to control this?

Yeah, I have fixed that now.

control-plane/csi-driver/src/context.rs
```diff
@@ -514,6 +518,107 @@ impl CreateParams {
     pub fn encrypted(&self) -> Option<bool> {
         self.encrypted
     }
+    /// Get the sts_affinity_group name from annotations if exists else generate it.
+    pub async fn sts_affinity_group(&self) -> Result<Option<String>, tonic::Status> {
+        if platform::current_platform_type() != platform::PlatformType::K8s {
```
tiagolobocastro: Should go below the sts_affinity enabled check

control-plane/csi-driver/src/context.rs
```diff
@@ -514,6 +518,107 @@ impl CreateParams {
...
+
+        if !self.sts_affinity_group.unwrap_or(false) {
+            return Ok(());
+        }
+
+        let client = Client::try_default()
+            .await
+            .map_err(|e| tonic::Status::aborted(e.to_string()))?;
+
+        let pvc_name = match &self.pvc_name {
```
tiagolobocastro: api seems a bit clunky here, as we've already validated these things on the `sts_affinity_group`?
Why not change the `sts_affinity_group` to return a type with the required information and add this method there?

control-plane/csi-driver/src/context.rs
```diff
@@ -514,6 +518,107 @@ impl CreateParams {
...
+        let patch = serde_json::json!({
+            "metadata": {
+                "annotations": {
+                    ANNOTATION_KEY: annotation_value
+                }
+            }
+        });
+
+        pvc_api
+            .patch(pvc_name, &PatchParams::default(), &Patch::Merge(&patch))
```
tiagolobocastro: Can there ever be a conflict here with different `annotation_value` on the key?

control-plane/csi-driver/src/context.rs
```diff
@@ -514,6 +518,107 @@ impl CreateParams {
...
+            .await
+            .map_err(|e| tonic::Status::aborted(e.to_string()))?;
+
+        if let Some(value) = pvc
+            .metadata
+            .annotations
+            .as_ref()
+            .and_then(|a| a.get(ANNOTATION_KEY))
+        {
+            tracing::trace!("Found affinity group name: {value} from annotation: {ANNOTATION_KEY}");
```
tiagolobocastro: If we change the return type we can also update it in case the annotation value is already the value that we want as in that case we don't need to issue the patch?
Also can this value != sts_affinity_group_name ?

control-plane/csi-driver/src/context.rs
Abhinandan-Purkait: Not sure if it matters as such, this whole method is supposed to work on k8s, doesn't matter if sts affinity is enabled or not.

control-plane/csi-driver/src/context.rs
Abhinandan-Purkait: Done.

control-plane/csi-driver/src/context.rs
```diff
@@ -575,41 +680,32 @@ impl TryFrom<&HashMap<String, String>> for CreateParams {
...
             }
-            warn!("PVC Name: {pvc_name} is not a valid statefulset pvc naming format, not triggering statefulset volume replica anti-affinity");
-            None
-        }
-        _ => {
-            warn!("Invalid PVC Name: {pvc_name:?} or PVC Namespace: {pvc_ns:?}, not triggering statefulset volume replica anti-affinity");
-            None
         }
     }
+    warn!("PVC Name: {pvc_name} is not a valid statefulset pvc naming format, not triggering statefulset volume replica anti-affinity");
```
dsharma-dc, line 724: Is this possible that these PVC names will be invalid here? The names must have been derived from sts spec and volumeClaimTemplate already?

Abhinandan-Purkait: Yes, it can be if someone enabled this for a deployment rather than a sts

control-plane/csi-driver/src/context.rs
Abhinandan-Purkait: I think if there is a conflict then we can't do anything because there will always be a conflict because the name that we generate will be `<velero-install-ns>/<pvc-naming-prefix>` but the annotation set will taking the backup would be `<backup-pvc-ns>/<pvc-naming-prefix>`.

control-plane/csi-driver/src/context.rs
Abhinandan-Purkait: If it's present we are ignoring patch.

control-plane/csi-driver/src/context.rs
```diff
@@ -514,7 +518,131 @@ impl CreateParams {
...
+            .metadata
+            .annotations
+            .as_ref()
+            .and_then(|a| a.get(ANNOTATION_KEY))
+            .map(|val| {
+                trace!(
+                    "Found affinity group name: {} from annotation: {}",
+                    val,
+                    ANNOTATION_KEY
+                );
```
```suggestion
```

control-plane/csi-driver/src/context.rs
```diff
@@ -575,41 +703,34 @@ impl TryFrom<&HashMap<String, String>> for CreateParams {
...
-                        }
-                    }
+fn generate_sts_affinity_group_name(pvc_name: &String, pvc_ns: &String) -> Option<String> {
+    let re = Regex::from_str(K8S_STS_PVC_NAMING_REGEX);
+    if let Ok(regex) = re {
+        if regex.is_match(pvc_name.as_str()) {
+            if let Some(captures) = regex.captures(pvc_name.as_str()) {
+                if let Some(common_binding) = captures.get(1) {
+                    let name = format!("{pvc_ns}/{}", common_binding.as_str());
+                    trace!("Generated affinity group name: {name}");
```
tiagolobocastro: I think the other traces should cover this?

```suggestion
```

control-plane/csi-driver/src/context.rs
```diff
@@ -514,7 +518,131 @@ impl CreateParams {
...
+    pvc_namespace: String,
+    name: String,
+    annotation_exists: bool,
 }
+
+impl StsAffinityGroupInfo {
+    /// If sts_affinity_group was requested we should patch pvc with it.
+    pub async fn set_sts_affinity_annotation(&self) -> Result<(), tonic::Status> {
+        if self.annotation_exists {
+            trace!("Annotation for STS affinity group already present, skipping patch");
```
```suggestion
            trace!(pvc.name=%self.pvc_name, pvc.stsAffinityGroup=%name, "Annotation for STS affinity group already present, skipping patch");
```

tiagolobocastro (approved):

dsharma-dc (approved):

Abhinandan-Purkait:
bors merge

PR #995, opened 2025-06-05
feat(pool): allow changing blobstore cluster size

dsharma-dc:
The change in cluster size will apply to all the new pools that get created after setting this to desired value. Already created pools keep working fine. It is recommended to do due diligence w.r.t application usage patterns before changing this value. For example - Smaller cluster size provides better storage efficiency and less internal fragmentation, but more metadata overhead. Larger cluster size will provide better performance for large sequential IOs, and low metadata overhead.

### DIFF
--- a/control-plane/agents/src/bin/core/controller/registry.rs
+++ b/control-plane/agents/src/bin/core/controller/registry.rs
@@ -39,6 +39,7 @@ use stor_port::{
     types::v0::{
         store::{
             nexus_persistence::{delete_all_v1_nexus_info, NexusInfo},
+            pool::POOL_BS_CLUSTER_SIZE_DEFAULT,
             registry::{ControlPlaneService, CoreRegistryConfig, NodeRegistration},
             volume::InitiatorAC,
         },
@@ -118,6 +119,9 @@ pub(crate) struct RegistryInner<S: Store> {
     allow_non_persistent_devlinks: bool,
     /// Prefer encrypted pools for volume replicas.
     encrypted_pools_soft_scheduling: bool,
+    /// Blobstore cluster size(in bytes) required for pool creation and replica scheduling. This is
+    /// not per-pool.
+    pool_bs_cluster_size: Option<u32>,
 }
 
 impl Registry {
@@ -146,6 +150,7 @@ impl Registry {
         no_volume_health: bool,
         allow_non_persistent_devlinks: bool,
         encrypted_pools_soft_scheduling: bool,
+        pool_bs_cluster_size: Option<u32>,
     ) -> Result<Self, SvcError> {
         let store_endpoint = Self::format_store_endpoint(&store_url);
         tracing::info!("Connecting to persistent store at {}", store_endpoint);
@@ -207,6 +212,7 @@ impl Registry {
                 etcd_max_page_size,
                 allow_non_persistent_devlinks,
                 encrypted_pools_soft_scheduling,
+                pool_bs_cluster_size: pool_bs_cluster_size.or(Some(POOL_BS_CLUSTER_SIZE_DEFAULT)),
             }),
         };
         registry.init().await?;
@@ -241,6 +247,11 @@ impl Registry {
     /// Check if pool creation using non-persistent devlink is allowed.
     pub(crate) fn allow_non_persistent_devlinks(&self) -> bool {
         self.allow_non_persistent_devlinks
+    }
+
+    /// Get the configured pool cluster_size.
+    pub(crate) fn pool_bs_cluster_size(&self) -> Option<u32> {
+        self.pool_bs_cluster_size
     }
 
     /// Check if the partial rebuilds are disabled.
--- a/control-plane/agents/src/bin/core/controller/registry.rs
+++ b/control-plane/agents/src/bin/core/controller/registry.rs
@@ -121,7 +121,7 @@ pub(crate) struct RegistryInner<S: Store> {
     encrypted_pools_soft_scheduling: bool,
     /// Blobstore cluster size(in bytes) required for pool creation and replica scheduling. This is
     /// not per-pool.
-    pool_bs_cluster_size: Option<u32>,
+    pool_cluster_size: Option<u32>,
 }
 
 impl Registry {
@@ -150,7 +150,7 @@ impl Registry {
         no_volume_health: bool,
         allow_non_persistent_devlinks: bool,
         encrypted_pools_soft_scheduling: bool,
-        pool_bs_cluster_size: Option<u32>,
+        pool_cluster_size: Option<u32>,
     ) -> Result<Self, SvcError> {
         let store_endpoint = Self::format_store_endpoint(&store_url);
         tracing::info!("Connecting to persistent store at {}", store_endpoint);
@@ -212,7 +212,7 @@ impl Registry {
                 etcd_max_page_size,
                 allow_non_persistent_devlinks,
                 encrypted_pools_soft_scheduling,
-                pool_bs_cluster_size: pool_bs_cluster_size.or(Some(POOL_BS_CLUSTER_SIZE_DEFAULT)),
+                pool_cluster_size: pool_cluster_size.or(Some(POOL_BS_CLUSTER_SIZE_DEFAULT)),
             }),
         };
         registry.init().await?;
@@ -250,8 +250,8 @@ impl Registry {
     }
 
     /// Get the configured pool cluster_size.
-    pub(crate) fn pool_bs_cluster_size(&self) -> Option<u32> {
-        self.pool_bs_cluster_size
+    pub(crate) fn pool_cluster_size(&self) -> Option<u32> {
+        self.pool_cluster_size
     }
 
     /// Check if the partial rebuilds are disabled.
--- a/control-plane/agents/src/bin/core/main.rs
+++ b/control-plane/agents/src/bin/core/main.rs
@@ -156,6 +156,12 @@ pub(crate) struct CliArgs {
     /// a non-encryption storage class.
     #[clap(long)]
     encrypted_pools_soft_scheduling: bool,
+
+    /// Blobstore cluster size(in bytes) required for pool creation. This is
+    /// not configurable per-pool here. However, the create pool request can be used
+    /// to configure cluster size per-pool.
+    #[clap(long)]
+    pool_bs_cluster_size: Option<u32>,
 }
 impl CliArgs {
     fn args() -> Self {
@@ -235,6 +241,7 @@ async fn server(cli_args: CliArgs) -> anyhow::Result<()> {
         cli_args.no_volume_health,
         cli_args.allow_non_persistent_devlink,
         cli_args.encrypted_pools_soft_scheduling,
+        cli_args.pool_bs_cluster_size,
     )
     .await?;
 
--- a/control-plane/agents/src/bin/core/main.rs
+++ b/control-plane/agents/src/bin/core/main.rs
@@ -161,7 +161,7 @@ pub(crate) struct CliArgs {
     /// not configurable per-pool here. However, the create pool request can be used
     /// to configure cluster size per-pool.
     #[clap(long)]
-    pool_bs_cluster_size: Option<u32>,
+    pool_cluster_size: Option<u32>,
 }
 impl CliArgs {
     fn args() -> Self {
@@ -241,7 +241,7 @@ async fn server(cli_args: CliArgs) -> anyhow::Result<()> {
         cli_args.no_volume_health,
         cli_args.allow_non_persistent_devlink,
         cli_args.encrypted_pools_soft_scheduling,
-        cli_args.pool_bs_cluster_size,
+        cli_args.pool_cluster_size,
     )
     .await?;
 
--- a/control-plane/agents/src/bin/core/pool/operations_helper.rs
+++ b/control-plane/agents/src/bin/core/pool/operations_helper.rs
@@ -46,7 +46,6 @@ impl OperationGuardArc<PoolSpec> {
             }
             CreatePool::from(spec.deref())
         };
-
         let node = registry.node_wrapper(&request.node).await?;
         if node.pool(&request.id).await.is_none() {
             return Ok(());
--- a/control-plane/agents/src/bin/core/pool/service.rs
+++ b/control-plane/agents/src/bin/core/pool/service.rs
@@ -51,8 +51,12 @@ impl PoolOperations for Service {
         pool: &dyn CreatePoolInfo,
         _ctx: Option<Context>,
     ) -> Result<Pool, ReplyError> {
-        let req = pool.into();
+        let mut req: CreatePool = pool.into();
         let service = self.clone();
+        // Set the cluster_size which is potentially configured via cli args.
+        // Setting it as early as possible otherwise we'll have to make CreatePool
+        // request mutable in the ResourceLifecycle trait.
+        req.cluster_size = pool.cluster_size().or(self.registry.pool_bs_cluster_size());
         let pool = Context::spawn(async move { service.create_pool(&req).await }).await??;
         Ok(pool)
     }
--- a/control-plane/agents/src/bin/core/pool/service.rs
+++ b/control-plane/agents/src/bin/core/pool/service.rs
@@ -56,7 +56,7 @@ impl PoolOperations for Service {
         // Set the cluster_size which is potentially configured via cli args.
         // Setting it as early as possible otherwise we'll have to make CreatePool
         // request mutable in the ResourceLifecycle trait.
-        req.cluster_size = pool.cluster_size().or(self.registry.pool_bs_cluster_size());
+        req.cluster_size = pool.cluster_size().or(self.registry.pool_cluster_size());
         let pool = Context::spawn(async move { service.create_pool(&req).await }).await??;
         Ok(pool)
     }
--- a/control-plane/agents/src/bin/core/tests/pool/mod.rs
+++ b/control-plane/agents/src/bin/core/tests/pool/mod.rs
@@ -59,6 +59,7 @@ async fn pool() {
                 disks: vec!["malloc:///disk0?size_mb=100".into()],
                 labels: None,
                 encryption: None,
+                cluster_size: None,
             },
             None,
         )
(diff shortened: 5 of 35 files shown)


dsharma-dc:
I have not put cluster size in the Pool spec as it's not much value add there. Should we do that?

dsharma-dc:
bors try

Abhinandan-Purkait (approved):

control-plane/agents/src/bin/core/pool/operations_helper.rs
```diff
@@ -39,14 +39,14 @@ impl OperationGuardArc<PoolSpec> {
     /// TODO: Since the data-plane now allows concurrent gRPC we should also modify the
     ///  control-plane to allow this, which would allows to set large timeouts for some gRPCs.
     pub(crate) async fn retry_creating(&mut self, registry: &Registry) -> Result<(), SvcError> {
-        let request = {
+        let mut request = {
             let spec = self.lock();
             if on_create_fail(&spec, registry).is_some() {
                 return Ok(());
             }
             CreatePool::from(spec.deref())
```
tiagolobocastro, line 47: We should take it from the spec itself

control-plane/agents/src/bin/core/main.rs
```diff
@@ -156,6 +156,11 @@ pub(crate) struct CliArgs {
     /// a non-encryption storage class.
     #[clap(long)]
     encrypted_pools_soft_scheduling: bool,
+
+    /// Blobstore cluster size(in bytes) required for pool creation. This is
+    /// not configurable per-pool.
```
tiagolobocastro: Could we make it configurable?

tiagolobocastro (commented):
We also should change the scheduling ensuring that for a given volume only clusters of the same size may be used.

What is the use case, when would we recommend the users to user a non-default cluster size and what are the pros/cons?

control-plane/agents/src/bin/core/pool/operations_helper.rs
dsharma-dc: > We should take it from the spec itself

This: https://github.com/openebs/mayastor-control-plane/pull/995#issuecomment-2943259166

dsharma-dc:
> We also should change the scheduling ensuring that for a given volume only clusters of the same size may be used.
> 
> What is the use case, when would we recommend the users to user a non-default cluster size and what are the pros/cons?

I would suggest refraining from changing the scheduling for this. The setting is too low level to mix with scheduling and if rebuild, snapshots, restore are working for a volume with mixed cluster replicas, then it should be ok.
A use case where disks are huge, this can be recommended depending on application needs. https://github.com/openebs/mayastor-control-plane/pull/995#issue-3120335012

tiagolobocastro:
> I have not put cluster size in the Pool spec as it's not much value add there. Should we do that?

Yes I think would be good to keep there.
Also, we should put it in the state if not there already.

tiagolobocastro:
> I would suggest refraining from changing the scheduling for this. The setting is too low level to mix with scheduling and if rebuild, snapshots, restore are working for a volume with mixed cluster replicas, then it should be ok. A use case where disks are huge, this can be recommended depending on application needs. [#995 (comment)](https://github.com/openebs/mayastor-control-plane/pull/995#issue-3120335012)

It's not just low level, at control-plane we compare replica/snapshot allocated space for example, and they would change if replicas have different cluster size.
As the start I think it's safer to "isolate" the cluster sizes.

dsharma-dc:
> > I would suggest refraining from changing the scheduling for this. The setting is too low level to mix with scheduling and if rebuild, snapshots, restore are working for a volume with mixed cluster replicas, then it should be ok. A use case where disks are huge, this can be recommended depending on application needs. [#995 (comment)](https://github.com/openebs/mayastor-control-plane/pull/995#issue-3120335012)
> 
> It's not just low level, at control-plane we compare replica/snapshot allocated space for example, and they would change if replicas have different cluster size. As the start I think it's safer to "isolate" the cluster sizes.

Can you point me to those comparisons? What are those allocation comparisons done for? We don't compare a volume's replicas among themselves AFAIK.

control-plane/agents/src/bin/core/main.rs
dsharma-dc: > Could we make it configurable?

We could, we'll have to expose via DSP CR. But let's not do that because users will have very little to no understanding of this, and most of the time even global config will be done based on our recommendation.

Abhinandan-Purkait (approved):

control-plane/agents/src/bin/core/main.rs
abhilashshetty04: Shouldn't it be per pool? Like, We have a global default(4Mb) which applies for all pools, For larger pools user can just pass from dsp.

dsharma-dc:
bors try

control-plane/agents/src/bin/core/controller/registry.rs
```diff
@@ -118,6 +119,9 @@ pub(crate) struct RegistryInner<S: Store> {
     allow_non_persistent_devlinks: bool,
     /// Prefer encrypted pools for volume replicas.
     encrypted_pools_soft_scheduling: bool,
+    /// Blobstore cluster size(in bytes) required for pool creation and replica scheduling. This is
+    /// not per-pool.
+    pool_bs_cluster_size: Option<u32>,
```
```suggestion
    /// Default pool cluster size(in bytes) when it is not specified in the pool creation request.
    pool_cluster_size: Option<u32>,
```

control-plane/agents/src/bin/core/controller/registry.rs
```diff
@@ -243,6 +249,11 @@ impl Registry {
         self.allow_non_persistent_devlinks
     }
 
+    /// Get the configured pool cluster_size.
+    pub(crate) fn pool_bs_cluster_size(&self) -> Option<u32> {
```
tiagolobocastro: also rename here and its uses

control-plane/agents/src/bin/core/pool/service.rs
```diff
@@ -51,8 +51,12 @@ impl PoolOperations for Service {
         pool: &dyn CreatePoolInfo,
         _ctx: Option<Context>,
     ) -> Result<Pool, ReplyError> {
-        let req = pool.into();
+        let mut req: CreatePool = pool.into();
         let service = self.clone();
+        // Set the cluster_size which is potentially configured via cli args.
+        // Setting it as early as possible otherwise we'll have to make CreatePool
+        // request mutable in the ResourceOperations trait.
+        req.cluster_size = pool.cluster_size().or(self.registry.pool_bs_cluster_size());
```
tiagolobocastro: move this to the create_pool impl itself

control-plane/agents/src/bin/core/tests/pool/mod.rs
```diff
@@ -270,6 +273,59 @@ async fn pool() {
...
+                    ..Default::default()
+                },
+                None,
+            )
+            .await
+            .unwrap();
+    }
+
+    cluster.composer().restart("io-engine-1").await.unwrap();
+    cluster.wait_pool_online(poolid.clone()).await.unwrap();
```
tiagolobocastro: isn't this missing an intermediate state? Ex: what if we don't catch the pool offline, we might get the "original" online?

control-plane/agents/src/bin/core/controller/registry.rs
tiagolobocastro: or `pool_allocation_unit` but make sure the term is used throughout the code base (ie grpc and all)

control-plane/agents/src/bin/core/tests/pool/mod.rs
```diff
@@ -270,6 +273,59 @@ async fn pool() {
         .is_empty());
 }
 
+// Tests creation and import of pool with bigger blobstore cluster size.
+// Create a few replicas, restart the node. Pool should import again without issue.
+#[tokio::test]
+async fn pool_bigger_cluster_size() {
```
tiagolobocastro:
```suggestion
async fn pool_larget_cluster_size() {
```

control-plane/agents/src/bin/core/tests/pool/mod.rs
```diff
@@ -270,6 +273,59 @@ async fn pool() {
         .is_empty());
 }
 
+// Tests creation and import of pool with bigger blobstore cluster size.
```
tiagolobocastro:
```suggestion
// Tests creation and import of pools with larger cluster sizes.
```

control-plane/rest/openapi-specs/v0_api_spec.yaml
```diff
@@ -2558,6 +2558,15 @@ components:
             type: string
         encryption:
           $ref: '#/components/schemas/Encryption'
+        cluster_size:
+          type: integer
+          format: int64
+          minimum: 4194304
+          multipleOf: 4194304
+          description: |-
+           Blobstore cluster size for the pool, in bytes. Must be >= 4 MiB and in multiples of 4MiB.
```
tiagolobocastro, line 2567: Must it be in multiples of 4MiB?

control-plane/stor-port/src/types/v0/store/pool.rs
```diff
@@ -9,6 +9,8 @@ use crate::types::v0::{
     transport::{self, CreatePool, ImportPool, NodeId, PoolDeviceUri, PoolId},
 };
 
+pub const POOL_BS_CLUSTER_SIZE_DEFAULT: u32 = 4194304;
```
tiagolobocastro, line 12: Let's not do that here, we can default it in the agent itself

control-plane/stor-port/src/types/v0/transport/pool.rs
```diff
@@ -100,6 +100,8 @@ pub struct PoolState {
     pub committed: Option<u64>,
     /// Is the pool encrypted.
     pub encrypted: bool,
+    /// Blobstore cluster size used for this pool.
+    pub bs_cluster_size: u32,
```
tiagolobocastro:
```suggestion
    pub cluster_size: Option<u32>,
```

control-plane/agents/src/bin/core/controller/registry.rs
dsharma-dc: > or `pool_allocation_unit` but make sure the term is used throughout the code base (ie grpc and all)

ok, but a few existing usages in grpc e.g replicaSpaceUsage and pool grpc for io-engine has cluster_size, can't change that.

control-plane/rest/openapi-specs/v0_api_spec.yaml
dsharma-dc: > Must it be in multiples of 4MiB?

It is supposed to be multiples of 1MiB, but for the sake of current default I'm mentioning as multiple of that.

control-plane/stor-port/src/types/v0/store/pool.rs
dsharma-dc: > Let's not do that here, we can default it in the agent itself

do you mean move this constant somewhere in agent code?

control-plane/agents/src/bin/core/controller/registry.rs
tiagolobocastro: oh that's right, then we should probably use `cluster_size` then

dsharma-dc: > oh that's right, then we should probably use `cluster_size` then

o_O  I have been changing all new ones already to say allocation_size, phew. so let's just ignore lvm backend then.

control-plane/agents/src/bin/core/pool/service.rs
dsharma-dc: > move this to the create_pool impl itself

As mentioned in comment, the request isn't mutable beyond this point.

control-plane/agents/src/bin/core/tests/pool/mod.rs
dsharma-dc: added node offline check.

dsharma-dc:
bors try

dsharma-dc:
bors try

control-plane/stor-port/src/types/v0/store/pool.rs
tiagolobocastro: I was thinking it was an agent specific configuration, but actually it's probably not, so maybe it is fine here then

control-plane/grpc/proto/v1/pool/pool.proto
dsharma-dc, line 1: @tiagolobocastro  @Abhinandan-Purkait  Here in grpc, the cluster size should be in PoolSpec or PoolState, or both? These types are only filled by core-agent upstream I guess and not other way round.

tiagolobocastro: In guess in both would be useful. In the state we can populate via the io-engine proto conversion.

dsharma-dc:
bors try

dsharma-dc:
bors try

dsharma-dc:
bors try

control-plane/agents/src/bin/core/main.rs
tiagolobocastro: nit: might be a bit misleading, I guess here it's meant this parameter is not configurable per pool but one can modify the create-pool request itself
perhaps explain that here

control-plane/csi-driver/src/context.rs
```diff
@@ -492,6 +507,7 @@ pub struct CreateParams {
     encrypted: Option<bool>,
     pvc_name: Option<String>,
     pvc_namespace: Option<String>,
+    pool_allocation_unit_size: Option<u64>,
```
tiagolobocastro: Should use cluster_size since we've kept that naming, otherwise with openapi/CLI and SC with different terminology will be confusing.
We'd just need to document what cluster_size means somewhere

```suggestion
    pool_cluster_size: Option<u64>,
```

tiagolobocastro:
> I have not put cluster size in the Pool spec as it's not much value add there. Should we do that?

Yes I think so, so we can show the cluster-size even if the pool is offline

control-plane/csi-driver/src/context.rs
```diff
@@ -692,6 +712,11 @@ impl TryFrom<&HashMap<String, String>> for CreateParams {
                 tonic::Status::invalid_argument("Invalid `Encrypted` value, expected a boolean")
             })?;
 
+        let pool_allocation_unit_size =
+            Parameters::pool_allocation_unit_size(args.get(Parameters::PoolAllocationUnitSize.as_ref())).map_err(|e| {
```
tiagolobocastro: I'm confused, doesn't the function already return a tonic error? Otherwise make the function return a parse error, and map it to tonic here

control-plane/rest/openapi-specs/v0_api_spec.yaml
```diff
@@ -3190,6 +3214,7 @@ components:
         - status
         - used
         - encrypted
+        - cluster_size
```
tiagolobocastro: Should not be required for compatibility, this is potentially a breaking change

control-plane/stor-port/src/types/v0/store/pool.rs
```diff
@@ -124,6 +133,8 @@ pub struct PoolSpec {
     /// Cordon/drain state.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub cordon_drain: Option<CordonDrainState>,
+    /// Blobstore cluster size used for this pool.
+    pub cluster_size: u32,
```
tiagolobocastro, line 137:
```suggestion
    pub cluster_size: Option<u32>,
```

control-plane/stor-port/src/types/v0/store/volume.rs
```diff
@@ -212,6 +213,9 @@ pub struct VolumeSpec {
     /// Data encryption.
     #[serde(default)]
     pub encrypted: bool,
+    /// Pool's blobstore cluster size required for replicas of this volume.
+    #[serde(default = "default_pool_cluster_size")]
+    pub cluster_size: u32,
```
tiagolobocastro, line 218: Let's keep Option, allows us to identify if cluster_size has been explicitly specified or not?

```suggestion
    #[serde(default, skip_serializing_if = "Option::is_none")]
    pub cluster_size: Option<u32>,
```

control-plane/stor-port/src/types/v0/transport/pool.rs
```diff
@@ -100,6 +100,8 @@ pub struct PoolState {
     pub committed: Option<u64>,
     /// Is the pool encrypted.
     pub encrypted: bool,
+    /// Blobstore cluster size used for this pool.
+    pub cluster_size: u32,
```
tiagolobocastro, line 104:
```suggestion
    pub cluster_size: Option<u32>,
```

tiagolobocastro (approved):

control-plane/stor-port/src/types/v0/store/volume.rs
dsharma-dc: > Let's keep Option, allows us to identify if cluster_size has been explicitly specified or not?

No, let's keep it just u32 here. Will be a lot of change again otherwise. If we need to figure that out, it can be figured out from CR?

control-plane/csi-driver/src/context.rs
dsharma-dc: > Should use cluster_size since we've kept that naming, otherwise with openapi/CLI and SC with different terminology will be confusing. We'd just need to document what cluster_size means somewhere

modified

control-plane/csi-driver/src/context.rs
dsharma-dc: > I'm confused, doesn't the function already return a tonic error? Otherwise make the function return a parse error, and map it to tonic here

Propagating the incoming tonic error upstream now.

control-plane/rest/openapi-specs/v0_api_spec.yaml
dsharma-dc: > Should not be required for compatibility, this is potentially a breaking change

Removed from required.

control-plane/stor-port/src/types/v0/store/volume.rs
tiagolobocastro: Ok

abhilashshetty04 (approved):

dsharma-dc:
bors merge

PR #996, opened 2025-06-09
feat(encryption): add check for k8s secret in operator

Abhinandan-Purkait:
### Issue

If encryption is requested for a pool and the k8s secret doesn't exist we get an weird error which doesn't explain the actual issue.

### Solution

This would add a check, where we fetch the k8s secret and proceed iff it exists.

### DIFF
--- a/k8s/operators/src/pool/context.rs
+++ b/k8s/operators/src/pool/context.rs
@@ -3,7 +3,7 @@ use super::{
     diskpool::crd::v1beta3::{CrPoolState, DiskPool, DiskPoolStatus},
     error::Error,
 };
-use crate::diskpool::crd::v1beta3::EncryptionSource;
+use crate::diskpool::crd::v1beta3::{EncryptionSecretConfig, EncryptionSource};
 use openapi::models::{Encryption, EncryptionSecret};
 use openapi::{
     apis::StatusCode,
@@ -188,6 +188,11 @@ impl ResourceContext {
         v1beta3_api(&self.ctx.k8s, &self.namespace().unwrap())
     }
 
+    /// Construct an API handle for the k8s secret
+    fn secret_api(&self) -> Api<k8s_openapi::api::core::v1::Secret> {
+        Api::namespaced(self.ctx.k8s.clone(), &self.namespace().unwrap())
+    }
+
     /// Control plane pool handler.
     fn pools_api(&self) -> &dyn openapi::apis::pools_api::tower::client::Pools {
         self.ctx.http.pools_api()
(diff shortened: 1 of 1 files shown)


k8s/operators/src/pool/context.rs
```diff
@@ -618,4 +621,26 @@ impl ResourceContext {
         .map_err(|e| error!(?e));
         Ok(Action::await_change())
     }
+
+    async fn validate_encryption_secret(
+        &self,
+        config: EncryptionSecretConfig,
+    ) -> Result<Encryption, Error> {
+        if let Err(error) = self.secret_api().get(&config.name).await {
```
tiagolobocastro, line 629: additionally can we check if it's correctly formatted?

tiagolobocastro (approved):

k8s/operators/src/pool/context.rs
Abhinandan-Purkait: I don't think that's possible here. It needs to be done at the parser end.

tiagolobocastro: And we can't use the parser lib here to check?

Abhinandan-Purkait: We can but then we will have to define those types here as well.

Abhinandan-Purkait: I mean the types into which we would want to deserialize it

tiagolobocastro: Alright, then does io-engine return a sane error code and message in case the secret is not formed properly?

Abhinandan-Purkait: Yes it does, here

`
2025-06-09T18:18:46.662918Z ERROR operator_diskpool::context: Unable to create or import pool error in response: status code '400 Bad Request', content: 'RestJsonError { details: "create_pool::status: InvalidArgument, message: \"Failed to parse JSON: missing field `cipher` at line 7 column 1\", details: [], metadata: MetadataMap { headers: {\"content-type\": \"application/grpc\", \"date\": \"Mon, 09 Jun 2025 18:18:46 GMT\", \"content-length\": \"0\"} }", message: "SvcError::GrpcRequestError", kind: InvalidArgument }'   
`

dsharma-dc (approved):

k8s/operators/src/pool/context.rs
tiagolobocastro: Could be tweaked into a nicer user message for sure, but not on this pr

Abhinandan-Purkait: Yeah. Merging this PR for now.

Abhinandan-Purkait:
bors merge

PR #1003, opened 2025-07-07
feat: handle data integrity in republish path

dsharma-dc:
When a new nexus is created in republish path, it can pick up a healthy replica that is local to nexus node. But it is possible that the old nexus could momentarily come back and fails an IO to this replica and in turn marking it unhealthy in etcd. Since new nexus has already chosen this replica, it goes ahead using this as source of truth and hence compromising data integrity. This change lets core-agent indicate to old nexus that it should self shutdown now without modifying updating anything in etcd.

### DIFF
--- a/utils/pstor/src/etcd.rs
+++ b/utils/pstor/src/etcd.rs
@@ -14,9 +14,10 @@ use crate::{
 use async_trait::async_trait;
 use etcd_client::{
     Client, Compare, CompareOp, DeleteOptions, EventType, GetOptions, KeyValue, SortOrder,
-    SortTarget, Txn, TxnOp, WatchStream, Watcher,
+    SortTarget, Txn, TxnOp, TxnOpResponse, WatchStream, Watcher,
 };
 use serde_json::Value;
+
 use snafu::ResultExt;
 use tokio::sync::mpsc::{channel, Receiver, Sender};
 
@@ -343,7 +344,6 @@ impl StoreObj for Etcd {
     async fn put_obj<O: StorableObject>(&mut self, object: &O) -> Result<(), Error> {
         let key = object.key().key();
         let vec_value = serde_json::to_vec(object).context(SerialiseValue)?;
-
         if let Some((lease_id, lock_key)) = self.lease_lock()? {
             let cmp = Compare::lease(lock_key.clone(), CompareOp::Equal, lease_id);
             let put = TxnOp::put(key.to_string(), vec_value, None);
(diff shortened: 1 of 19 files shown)


utils/pstor/src/etcd.rs
```diff
@@ -372,21 +373,66 @@ impl StoreObj for Etcd {
...
 
-    async fn get_obj<O: StorableObject>(&mut self, key: &O::Key) -> Result<O, Error> {
+    /// Compare and swap the value as part of transaction.
+    async fn put_obj_cas<O: StorableObject>(
+        &mut self,
+        object: &O,
+        expected_mod_rev: i64,
+    ) -> Result<(), Error> {
+        let key = object.key().key();
+        let obj_value = to_deterministic_json_vec(&object).context(SerialiseValue)?;
```
tiagolobocastro: If we are using the revision, do we still need this deterministic json?

dsharma-dc: > If we are using the revision, do we still need this deterministic json?

Yeah, because regardless, the serialization here in core-agent puts the keys out of order. It'll still work but the persisted entry will look different from original one(bool keys move to the beginning). So better keep that same. io-engine always automatically persists keys in order, so I've added a CI test there to ensure the ordering is correct.

tiagolobocastro: What's the problem of the keys being out of order? The content is anyway parsed into a typed object, so the ordering shouldn't matter?

dsharma-dc: > What's the problem of the keys being out of order? The content is anyway parsed into a typed object, so the ordering shouldn't matter?

No big deal with being in different order than original as much as I can see. I was anticipating both questions and have addressed `why are we changing the persisted keys order. Is there a way we can keep it same as original?` :)
Shall I remove sorting? Lmk

Ideally, we could've made NexusInfo as a shared type and keeping `skip` fields in some core-agent specific type, but that would be a very widespread refactor.

tiagolobocastro: Yeah let's remove the sorting then.

> Ideally, we could've made NexusInfo as a shared type and keeping skip fields in some core-agent specific type, but that would be a very widespread refactor.

We could and even should perhaps define it in a proto file but I'd still be weary about relying on key ordering since we'd still be writing it as json

dsharma-dc: removed key ordering.
Just to clarify, we weren't relying on ordering anywhere, it's just to keep the persisted ordering same as originally set by io-engine).

tiagolobocastro (approved):

Abhinandan-Purkait (approved):

abhilashshetty04 (approved):

dsharma-dc:
bors merge

dsharma-dc:
bors cancel

dsharma-dc:
bors try

dsharma-dc:
bors try

dsharma-dc:
bors merge

PR #1004, opened 2025-07-08
Add pool cordoned import

tiagolobocastro:
test(pool/cordon): add import test scenario
    
---

    feat(pool/cordon): add imports as cordon constraint
    
    A new constraint is added to the transport apis and to the plugin which allow
    you to block pool imports.
    This is useful for when you're modifying/repairing the pool disk and therefore
    we can't let the control-plane import the pool at the same time.
    
---

    docs(rest-plugin): fix subcommand help
    
    Help was incorrectly claiming we cordon all aspects by default.

### DIFF
--- a/control-plane/plugin/src/resources/pool.rs
+++ b/control-plane/plugin/src/resources/pool.rs
@@ -462,6 +462,47 @@ pub struct CordonReq {
     /// No new restores can be created on the pool.
     #[clap(long)]
     pub restores: bool,
+    /// Pool will not be imported on node/io-engine restart.{n}
+    /// Warning: this may impact existing volume I/O{n}
+    /// This may be useful when repairing the pool metadata.
+    #[clap(long)]
+    pub import: bool,
+    /// Apply all cordon sub-resource scheduling constraints.
+    /// Note: you must selectively enable the import constraint.
+    #[clap(long)]
+    pub all_sub: bool,
+    /// Apply *ALL* cordon constraints.
+    /// Warning: This enables import constraint, which may impact existing volume I/O
+    #[clap(long)]
+    pub all: bool,
+}
+impl Default for CordonReq {
+    fn default() -> Self {
+        Self {
+            replicas: true,
+            snapshots: false,
+            restores: true,
+            import: false,
+            all_sub: false,
+            all: false,
+        }
+    }
+}
+impl From<CordonReq> for models::PoolCordonReq {
+    fn from(value: CordonReq) -> Self {
+        if value.all {
+            models::PoolCordonReq::new_all(true, true, true, false)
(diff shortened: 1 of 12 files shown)


tiagolobocastro:
bors try

niladrih (approved):

control-plane/plugin/src/resources/pool.rs
```diff
@@ -484,36 +547,39 @@ impl Cordoning for Pool {
...
+                    let pool = RestClient::client()
+                        .pools_api()
+                        .get_pool(id)
+                        .await
+                        .map_err(|source| Error::GetPoolError {
+                            id: id.to_string(),
+                            source,
+                        })?;
+                    let constraints = pool_cordon_resources(&pool.into_body());
+                    println!("Pool {id} already cordoned. Current constraints: {constraints:?}");
```
Abhinandan-Purkait: `println!("Pool {id} is already cordoned. Current constraints: {constraints:?}");` ?

control-plane/plugin/src/resources/pool.rs
```diff
@@ -577,12 +634,14 @@ impl Cordoning for Pool {
                     if resources.is_empty() {
                         println!("Pool {id} already uncordoned");
                     } else {
-                        println!("Pool {id} already partially uncordoned. Remaining cordoned resources: {resources:?}");
+                        println!("Pool {id} already partially uncordoned. Remaining constraints: {resources:?}");
```
Abhinandan-Purkait: `println!("Pool {id} is partially uncordoned. Remaining constraints: {resources:?}");` ?

control-plane/plugin/src/resources/pool.rs
```diff
@@ -598,15 +657,20 @@ fn resource(yes: bool, name: &str) -> &str {
...
+        resource(rsc.imports, "imports"),
     ]
     .into_iter()
     .filter(|s| !s.is_empty())
     .collect::<Vec<&str>>()
-    .join(",")
+    .join(",");
+    if cordon.is_empty() {
+        return "Unknown".to_string();
+    }
```
Abhinandan-Purkait, line 677: Should it not be `""` rather than Unknown?

Abhinandan-Purkait (approved):

control-plane/plugin/src/resources/pool.rs
tiagolobocastro: This is in case old plugin is used with newer control-plane.
In this case if node is in the cordon state, but using a  constraint knonwn only to the newer control-plane, then we mark here as unknown.

tiagolobocastro:
bors merge

PR #1005, opened 2025-07-14
fix(ha/multipath): assume multipath as ON

tiagolobocastro:
If the path /sys/module/nvme_core/parameters/multipath is not found
but /sys/module/nvme_core/parameters/iopolicy is then assume multipath
is enabled.

A recent change from RHEL10 made multipath enabled by default, but also
removed the multipath parameter for some reason, which is a breaking
change for us!

On very old kernels it's perhaps possible that multipath parameter is
not found but nvme_core is enabled and so we check the iopolicy which
in the systems I checked is present.

### DIFF
--- a/utils/utils-lib/src/lib.rs
+++ b/utils/utils-lib/src/lib.rs
@@ -22,11 +22,23 @@ pub fn check_nvme_core_ana() -> Result<bool, std::io::Error> {
         Ok(multipath) => Ok(multipath),
         Err(error) if error.kind() == std::io::ErrorKind::NotFound => {
             // check the nvme_core is enabled, otherwise all this is moot.
-            std::fs::exists("/sys/module/nvme_core")?;
+            if !std::fs::exists("/sys/module/nvme_core")? {
+                return Err(std::io::Error::new(
+                    std::io::ErrorKind::Unsupported,
+                    "kernel module nvme_core not found",
+                ));
+            }
+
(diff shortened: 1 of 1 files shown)


Abhinandan-Purkait (approved):

tiagolobocastro:
This would be the change proposed by @gabrielgbs97

dsharma-dc:
Is this still WIP ?

tiagolobocastro:
Yes, waiting for feedback.
I've asked a user for feedback on this, since I remembered that on their clusters `multipath` was not present but `nvme_core` is present.

tiagolobocastro:
Looks like we cannot do this.
On `5.15.72-talos` the multipath parameter is not present, and the system is not multipath enabled :(

tiagolobocastro:
> Looks like we cannot do this. On `5.15.72-talos` the multipath parameter is not present, and the system is not multipath enabled :(

I have a new approach, we check the iopolicy parameter which is present only if `CONFIG_NVME_MULTIPATH` is on, and it's always there on RHEL

Abhinandan-Purkait (approved):
LGTM. Thanks! I hope we don't have any init containers that do this?

tiagolobocastro:
> LGTM. Thanks! I hope we don't have any init containers that do this?

We had some which check the modules, but not the parameters, at least that I know of!

tiagolobocastro:
bors merge

PR #1008, opened 2025-08-06
feat(restore): use cluster size of pool from replica snapshots

dsharma-dc:
While creating a clone volume with snapshot as source, get the pool cluster size from one of the replica snapshot. This gets set as the cluster size constraint in the new restore/clone volume.

### DIFF
--- a/control-plane/agents/src/bin/core/volume/snapshot_operations.rs
+++ b/control-plane/agents/src/bin/core/volume/snapshot_operations.rs
@@ -553,4 +553,41 @@ impl OperationGuardArc<VolumeSnapshot> {
         }
         failed
     }
+
+    /// Get the pool cluster size from one of the replica snapshot. This gets set as the cluster size
+    /// constraint in the new restore/clone volume. Once the correct cluster size constraint is set on the
+    /// volume, the further replica scheduling work as expected in terms of cluster size constraint, for
+    /// the clone volume.
+    pub(crate) async fn pool_cluster_size_for_clone(
+        &self,
+        registry: &Registry,
+    ) -> Result<u32, SvcError> {
+        let repl_snap = self
+            .as_ref()
+            .metadata()
+            .replica_snapshots()
+            .ok_or(SvcError::NoHealthyReplicas {
+                id: self.as_ref().uid_str().clone(),
+            })?
+            .first()
+            .ok_or(SvcError::NoHealthyReplicas {
+                id: self.as_ref().uid_str().clone(),
+            })?;
+
+        let snapshot_replica = registry
+            .snapshot_replica(repl_snap.spec())
+            .await
+            .map_err(|_| SvcError::ReplicaSnapMiss {
+                replica: repl_snap.spec().source_id().replica_id().to_string(),
+            })?;
+
+        Ok(registry
+            .ctrl_pool(snapshot_replica.pool_id())
+            .await?
+            .spec()
+            .ok_or(SvcError::PoolNotFound {
+                pool_id: snapshot_replica.pool_id().clone(),
+            })?
+            .cluster_size)
+    }
 }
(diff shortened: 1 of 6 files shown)


control-plane/agents/src/bin/core/volume/snapshot_operations.rs
```diff
@@ -553,4 +553,37 @@ impl OperationGuardArc<VolumeSnapshot> {
         }
         failed
     }
+
+    pub async fn pool_cluster_size_for_clone(&self, registry: &Registry) -> Result<u32, SvcError> {
+        let repl_snap = self
+            .as_ref()
+            .metadata()
+            .replica_snapshots()
+            .ok_or(SvcError::ReplicaSnapMiss {
```
tiagolobocastro: Is this the same error code which would otherwise be returned? We should return the same one as it's what we tested for csi retries etc

control-plane/agents/src/bin/core/volume/snapshot_operations.rs
```diff
@@ -553,4 +553,37 @@ impl OperationGuardArc<VolumeSnapshot> {
         }
         failed
     }
+
+    pub async fn pool_cluster_size_for_clone(&self, registry: &Registry) -> Result<u32, SvcError> {
```
tiagolobocastro: Also please add a comment explaining why is this needed now

```suggestion
    pub(crate) async fn pool_cluster_size_for_clone(&self, registry: &Registry) -> Result<u32, SvcError> {
```

control-plane/agents/src/bin/core/volume/snapshot_operations.rs
```diff
@@ -553,4 +553,37 @@ impl OperationGuardArc<VolumeSnapshot> {
...
+                        pool_id: r.pool_id().clone(),
+                    })?
+                    .cluster_size
+            }
+            _ => {
+                return Err(SvcError::ReplicaSnapMiss {
+                    replica: repl_snap.spec().source_id().replica_id().to_string(),
+                })
+            }
+        };
```
tiagolobocastro:
```suggestion
        let snapshot_replica = registry.snapshot_replica(repl_snap.spec()).await.map_err(|_| Err(SvcError::ReplicaSnapMiss {
                    replica: repl_snap.spec().source_id().replica_id().to_string(),
                }))?;
        Ok(
                registry
                    .ctrl_pool(r.pool_id())
                    .await?
                    .spec()
                    .ok_or(SvcError::PoolNotFound {
                        pool_id: r.pool_id().clone(),
                    })?
                    .cluster_size
           )
```

control-plane/agents/src/bin/core/volume/snapshot_operations.rs
dsharma-dc: > Is this the same error code which would otherwise be returned? We should return the same one as it's what we tested for csi retries etc

In that case it should be `SvcError::NoHealthyReplicas` as the only other place we do this is here:
 https://github.com/openebs/mayastor-control-plane/blob/41b2b7fb59a8fedb229e530be2bdaec87e707b4a/control-plane/agents/src/bin/core/volume/clone_operations.rs#L182

dsharma-dc:
bors try

Abhinandan-Purkait (approved):

control-plane/agents/src/bin/core/volume/snapshot_operations.rs
tiagolobocastro: Looks like you've updated the error, so is this the right one?

dsharma-dc: > Looks like you've updated the error, so is this the right one?

Yeah, seems like that's what it should be. When there is no replica snapshot available for cloning we return `SvcError::NoHealthyReplicas` as shown on link in my previous comment.

tiagolobocastro (approved):

dsharma-dc:
bors merge

dsharma-dc:
bors merge

PR #1009, opened 2025-08-08
refactor: allow zfs volume paths

Abhinandan-Purkait:
- Allow /dev/zvol paths to be a valid devlink as they are also persistent.

### DIFF
--- a/utils/utils-lib/src/constants.rs
+++ b/utils/utils-lib/src/constants.rs
@@ -186,5 +186,4 @@ pub fn nvme_controller_model_id() -> String {
 pub const ETCD_MAX_PAGE_LIMIT: &str = "500";
 
 /// Regex to match persistent disks.
-pub const DEVLINK_REGEX: &str =
-    r"^(/dev/(disk/by-(id|uuid|label|path|partuuid|partlabel)/[^/\s]+|mapper/[^/\s]+))$";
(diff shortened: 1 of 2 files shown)


utils/utils-lib/src/constants.rs
```diff
@@ -186,5 +186,4 @@ pub fn nvme_controller_model_id() -> String {
 pub const ETCD_MAX_PAGE_LIMIT: &str = "500";
 
 /// Regex to match persistent disks.
-pub const DEVLINK_REGEX: &str =
```
dsharma-dc, line 189: Is this going to allow /dev/mapper/zvol also ? That may not be what is intended.
We only need to allow /dev/zvol ?

Abhinandan-Purkait: So, anything after /dev/mapper should be allowed right? We can't restrict that

Abhinandan-Purkait: Also, /dev/mapper/zvol is not an actual path, I just added to test

dsharma-dc (approved):

tiagolobocastro (approved):

dsharma-dc:
Please cherry-pick to release/2.9 also once merged.

Abhinandan-Purkait:
bors merge

PR #1012, opened 2025-08-15
Refactor Scheduling Sorting

tiagolobocastro:
We're trying to find spots where total order is not implemented, leading to panics by the rust std library.
This PR doesn't solve anything, just some minor refactor and a couple of fixes where sort order was incorrect.
We do switch to the unstable sorter which might "help", since in our simulated testing we could not reproduce
the panics with the unstable sort.

Further we've identified the dual weight scoring as the main cause behind the total order issue, given that by 
nature, it's not a total order!!

### DIFF
--- a/control-plane/agents/src/bin/core/controller/scheduling/mod.rs
+++ b/control-plane/agents/src/bin/core/controller/scheduling/mod.rs
@@ -195,16 +195,12 @@ impl ChildSorters {
 
                     let childa_is_local = !a.spec().share.shared();
                     let childb_is_local = !b.spec().share.shared();
-                    if childa_is_local == childb_is_local {
-                        match a.ag_replicas_on_pool().cmp(&b.ag_replicas_on_pool()) {
-                            Ordering::Less => Ordering::Greater,
-                            Ordering::Equal => Ordering::Equal,
-                            Ordering::Greater => Ordering::Less,
+                    match (childa_is_local, childb_is_local) {
+                        (true, true) | (false, false) => {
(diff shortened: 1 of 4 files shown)


control-plane/agents/src/bin/core/controller/scheduling/mod.rs
```diff
@@ -374,7 +349,7 @@ impl AddReplicaSorters {
                 match (a_healthy, b_healthy) {
                     (true, false) => std::cmp::Ordering::Less,
                     (false, true) => std::cmp::Ordering::Greater,
-                    (_, _) => a.pool().free_space().cmp(&b.pool().free_space()),
+                    (_, _) => b.pool().free_space().cmp(&a.pool().free_space()),
```
Abhinandan-Purkait, line 352: This should have  a.cmp(b) right, the way it was?

Abhinandan-Purkait (approved):

control-plane/agents/src/bin/core/controller/scheduling/mod.rs
tiagolobocastro: Yeah, which was wrong right? We want to prefer `a` if it has a larger free space?

tiagolobocastro: Discussed offline

tiagolobocastro:
bors merge

dsharma-dc (approved):

tiagolobocastro:
bors try

tiagolobocastro:
bors try

tiagolobocastro:
bors merge

PR #1014, opened 2025-08-20
chore: add functionality to generate diskpool crd

dsharma-dc:


k8s/operators/src/pool/main.rs
```diff
@@ -210,10 +214,42 @@ async fn pool_controller(args: ArgMatches) -> anyhow::Result<()> {
     Ok(())
 }
 
+/// Write the DiskPool CRD to a file
+fn write_diskpool_crd(output_dir: &str) -> anyhow::Result<()> {
+    // Generate the CRD
+    let crd = DiskPool::crd();
+
+    // let yaml = serde_yaml::to_string(&crd)?;
```
tiagolobocastro:
```suggestion
```

tiagolobocastro (approved):
nit: might want to consider taking the full output path rather than the directory, like kubectl -o

k8s/operators/src/pool/main.rs
```diff
@@ -210,10 +214,42 @@ async fn pool_controller(args: ArgMatches) -> anyhow::Result<()> {
...
+
+    println!("DiskPool CRD generated successfully at {output_dir:?}/diskpools.crd.yaml");
+    Ok(())
+}
+
 #[tokio::main(flavor = "current_thread")]
 async fn main() -> anyhow::Result<()> {
     let matches = clap::Command::new(utils::package_description!())
         .version(utils::version_info_str!())
+        .arg(
```
abhilashshetty04, line 240: Should we expose this? Its for validations so always generate?

abhilashshetty04 (approved):

k8s/operators/src/pool/main.rs
tiagolobocastro: We could add another binary target, which only does the generation I suppose

dsharma-dc:
> nit: might want to consider taking the full output path rather than the directory, like kubectl -o

Keeping it directory will probably help writing gh runner based tests for CTY easier AFAIU till now, so keeping that for now.

dsharma-dc:
bors try

dsharma-dc:
bors try

k8s/operators/src/pool/main.rs
dsharma-dc: let's keep it in same one for simplicity.

dsharma-dc:
bors merge

dsharma-dc:
bors merge

PR #1015, opened 2025-08-21
CherryPick PR1012

tiagolobocastro:
We're trying to find spots where total order is not implemented, leading to panics by the rust std library.
This PR doesn't solve anything, just some minor refactor and a couple of fixes where sort order was incorrect.
We do switch to the unstable sorter which might "help", since in our simulated testing we could not reproduce
the panics with the unstable sort.

Further we've identified the dual weight scoring as the main cause behind the total order issue, given that by 
nature, it's not a total order!!


Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

PR #1017, opened 2025-08-26
test: crd validation using cty

dsharma-dc:


dsharma-dc:
This is how a good run looks like where the schema validation succeeds while having a new optional field introduced in same version.

```
+--------+----------------------------+---------------+-------+--------------------------------------+
| STATUS | IT                         | MATCHER       | ERROR | TEMPLATE                             |
+--------+----------------------------+---------------+-------+--------------------------------------+
| PASS   | matches DiskPool correctly | matchSnapshot |       | k8s/crd_test/crds/diskpools.crd.yaml |
+--------+----------------------------+---------------+-------+--------------------------------------+

Tests total: 1, failed: 0, passed: 1
```

This is how the run fails if a new mandatory field is added without version bump.

```
+--------+----------------------------+---------------+----------------------------------------------------------------------------------+--------------------------------------+
| STATUS | IT                         | MATCHER       | ERROR                                                                            | TEMPLATE                             |
+--------+----------------------------+---------------+----------------------------------------------------------------------------------+--------------------------------------+
| FAIL   | matches DiskPool correctly | matchSnapshot | matcher returned failure: failed to validate kind DiskPool and version v1beta3:  | k8s/crd_test/crds/diskpools.crd.yaml |
|        |                            |               | failed to validate kind DiskPool and version v1beta3: spec.disktype in body is r |                                      |
|        |                            |               | equired                                                                          |                                      |
+--------+----------------------------+---------------+----------------------------------------------------------------------------------+--------------------------------------+
Tests total: 1, failed: 1, passed: 0
```

.github/workflows/crds.yaml
```diff
@@ -0,0 +1,99 @@
...
+
+            \`\`\`bash
+            cty test k8s/crd_test --update
+            \`\`\`
+
+            and add updated snapshot to the PR patch.
+            `;
+                        await github.rest.issues.createComment({
+                          ...context.repo,
+                          issue_number: prNumber,
```
tiagolobocastro: too many tabs?

.github/workflows/crds.yaml
```diff
@@ -0,0 +1,99 @@
...
+                          body
+                        })
```
tiagolobocastro: add new line

k8s/crd_test/__snapshot__/diskpools.crd-v1beta3.yaml
```diff
@@ -0,0 +1,21 @@
+apiVersion: openebs.io/v1beta3
```
tiagolobocastro, line 1: So is it adding tests based on this?

If so do we need to add more options, ex labelled?

.github/workflows/crds.yaml
```diff
@@ -0,0 +1,99 @@
...
+        run: |
+          # Run cty to test CRDs
+          /usr/local/bin/cty test k8s/crd_test || {
+            echo "Validation failed: Breaking changes detected."
+            exit 1
+          }
+          echo "Validation successful: No breaking changes detected."
+
+      - name: Add PR comment to update snapshot if required
+        if: ${{ success() && steps.crd-test.conclusion == 'success' && github.event_name == 'pull_request' }}
```
tiagolobocastro: steps check seems redundant?

```suggestion
        if: ${{ success() && github.event_name == 'pull_request' }}
```

.github/workflows/crds.yaml
```diff
@@ -0,0 +1,99 @@
...
+on:
+  push:
+    branches:
+      - develop
+    paths:
+      - 'k8s/operators/src/pool/diskpool/**'
+      - 'k8s/operators/src/pool/main.rs'
+      - '.github/workflows/crds.yaml'
+  pull_request:
+    paths:
```
tiagolobocastro: path filtering is bit finecki, it doesn't work well with branch protection settings, I'd say we should always run
Also, running on PR only and not bors, means it's not running against the merged code, is this ok in this case?

k8s/crd_test/__snapshot__/diskpools.crd-v1beta3.yaml
dsharma-dc: > So is it adding tests based on this?
> 
> If so do we need to add more options, ex labelled?

The snapshot directory can contain multiple diskpools e.g based on different versions. All of them will get matched/validated against latest schema.

.github/workflows/crds.yaml
dsharma-dc: > path filtering is bit finecki, it doesn't work well with branch protection settings, I'd say we should always run Also, running on PR only and not bors, means it's not running against the merged code, is this ok in this case?

In this case it should be ok because any crd change will be in PR. snapshot is statically placed in the repo and only updated manually after this workflow succeeds.

dsharma-dc: removed the path matches.

.github/workflows/crds.yaml
```diff
@@ -0,0 +1,91 @@
+name: CRDs
+
+on:
+  push:
+    branches:
+      - develop
```
tiagolobocastro, line 6: on release as well?

k8s/crd_test/__snapshot__/diskpools.crd-v1beta3.yaml
tiagolobocastro: So do we need to add more options or more files here to get fields validated? ex the topology?

.github/workflows/crds.yaml
```diff
@@ -0,0 +1,91 @@
...
+
+            Please update the CRD snapshots by running:
+
+            \`\`\`bash
+            cty test k8s/crd_test --update
+            \`\`\`
+
+            and add updated snapshot to the PR patch.
+            `;
+            await github.rest.issues.createComment({
```
tiagolobocastro, line 96: Do we need to do this all the time? Won't this flood the comments everytime on push?

abhilashshetty04 (approved):

.github/workflows/crds.yaml
dsharma-dc: > on release as well?

added release branch too.

k8s/crd_test/__snapshot__/diskpools.crd-v1beta3.yaml
dsharma-dc: > So do we need to add more options or more files here to get fields validated? ex the topology?

Yes there can be many snapshots for same CRD. So if required later we can keep multiple based on needs. For fields e.g topology, the field is part of current snapshot, but not sure what else we'd want to check on it because apart from that it'll be the internal config for that particular type ?

.github/workflows/crds.yaml
dsharma-dc: Yes it will. That's where path match would've helped. Let me see if it's possible to path match for only commenting step.

tiagolobocastro: You could use the same solution we used on openebs to do equivalent thing to the push filters:
<https://github.com/openebs/openebs/blob/2aec96b8d4aa85af38f591fc1f940bafa7288b4a/.github/workflows/rust.yaml#L24>

dsharma-dc: Added the path filter tool.

k8s/crd_test/__snapshot__/diskpools.crd-v1beta3.yaml
tiagolobocastro: Don't know how this tool works, and whether it needs various examples to determine if the update is valid or not, and if so then we'd need to provide the various configurations options?

.github/workflows/crds.yaml
```diff
@@ -0,0 +1,100 @@
...
+      - name: Check changed files before commenting
+        id: changes
+        uses: dorny/paths-filter@v3
+        with:
+          filters: |
+            crd:
+              - 'k8s/**'
+
+      - name: Add PR comment to update snapshot if required
+        if: ${{ success() && github.event_name == 'pull_request' }}
```
tiagolobocastro: Need to include the check for changes as well

```suggestion
        if: ${{ success() && github.event_name == 'pull_request' && steps.changes.outputs.crd == 'true' }}
```

tiagolobocastro (approved):

k8s/crd_test/__snapshot__/diskpools.crd-v1beta3.yaml
dsharma-dc: > Don't know how this tool works, and whether it needs various examples to determine if the update is valid or not, and if so then we'd need to provide the various configurations options?

I can check with the maintainer about this if required or not to do such checks. I've been in touch with him to understand the behaviour better last week.

dsharma-dc:
bors merge

dsharma-dc:
> Build failed:
> 
> * [bors-ci](https://github.com/openebs/mayastor-control-plane/actions/runs/17383488215/job/49346792055)

Unrelated. NodeStatus::Online wait timed out.

bors merge

PR #1018, opened 2025-08-26
feat(k8s-plugin): add support for kubeconfig context switching

voyvodov:
This feature will allow user to choose between different kubeconfig contexts. 
Useful in case there are multiple clusters in a single kubeconfig file.

### DIFF
--- a/k8s/proxy/src/lib.rs
+++ b/k8s/proxy/src/lib.rs
@@ -16,16 +16,19 @@ pub use proxy::{ConfigBuilder, ForwardingProxy, LokiClient, Scheme};
 /// Get the `kube::Config` from the given kubeconfig file, or the default.
 pub async fn config_from_kubeconfig(
     kube_config_path: Option<PathBuf>,
+    options: KubeConfigOptions,
 ) -> anyhow::Result<kube::Config> {
     let mut config = match kube_config_path {
         Some(config_path) => {
             // NOTE: Kubeconfig file may hold multiple contexts to communicate
-            //       with different kubernetes clusters. We have to pick master
-            //       address of current-context config only
+            //       with different kubernetes clusters. We'll pick either the one
+            //       specified in the options, or the current-context from the
+            //       kubeconfig file.
             let kube_config = kube::config::Kubeconfig::read_from(&config_path)?;
-            kube::Config::from_custom_kubeconfig(kube_config, &Default::default()).await?
+            kube::Config::from_custom_kubeconfig(kube_config, &options).await?
         }
-        None => kube::Config::from_kubeconfig(&KubeConfigOptions::default()).await?,
+
+        None => kube::Config::from_kubeconfig(&options).await?,
     };
     config.apply_debug_overrides();
     Ok(config)
@@ -34,8 +37,21 @@ pub async fn config_from_kubeconfig(
 /// Get the `kube::Client` from the given kubeconfig file, or the default.
 pub async fn client_from_kubeconfig(
     kube_config_path: Option<PathBuf>,
+    context: Option<String>,
 ) -> anyhow::Result<kube::Client> {
+    let options = kubeconfig_options_from_context(context);
     Ok(kube::Client::try_from(
-        config_from_kubeconfig(kube_config_path).await?,
+        config_from_kubeconfig(kube_config_path, options).await?,
     )?)
+}
+
+/// Create a `KubeConfigOptions` from the given context.
+pub fn kubeconfig_options_from_context(context: Option<String>) -> KubeConfigOptions {
+    match context {
+        Some(ctx) => KubeConfigOptions {
+            context: Some(ctx),
+            ..Default::default()
+        },
+        None => KubeConfigOptions::default(),
+    }
 }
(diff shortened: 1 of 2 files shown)


k8s/proxy/src/lib.rs
```diff
@@ -16,16 +16,27 @@ pub use proxy::{ConfigBuilder, ForwardingProxy, LokiClient, Scheme};
...
+    context: Option<String>,
+) -> anyhow::Result<kube::Config> {    
+    let options = match context {
+        Some(ctx) =>  KubeConfigOptions {
+            context: Some(ctx),
+            ..Default::default()
+            },
+        None => KubeConfigOptions {
+        ..Default::default()
+        }
```
tiagolobocastro:
```suggestion
        None => KubeConfigOptions::default()
```

k8s/proxy/src/lib.rs
```diff
@@ -16,16 +16,27 @@ pub use proxy::{ConfigBuilder, ForwardingProxy, LokiClient, Scheme};
...
 pub async fn config_from_kubeconfig(
     kube_config_path: Option<PathBuf>,
-) -> anyhow::Result<kube::Config> {
+    context: Option<String>,
+) -> anyhow::Result<kube::Config> {    
+    let options = match context {
+        Some(ctx) =>  KubeConfigOptions {
+            context: Some(ctx),
+            ..Default::default()
+            },
```
tiagolobocastro: spacing seems off, did you run `./scripts/rust/linter.sh` ?

k8s/proxy/src/lib.rs
```diff
@@ -16,16 +16,27 @@ pub use proxy::{ConfigBuilder, ForwardingProxy, LokiClient, Scheme};
 /// Get the `kube::Config` from the given kubeconfig file, or the default.
 pub async fn config_from_kubeconfig(
     kube_config_path: Option<PathBuf>,
-) -> anyhow::Result<kube::Config> {
+    context: Option<String>,
```
tiagolobocastro: Would it be worth passing `KubeConfigOptions` instead?

tiagolobocastro:
bors try

k8s/proxy/src/lib.rs
voyvodov: Opps...Didn't do that.

Will run to fix everything

k8s/proxy/src/lib.rs
voyvodov: Initially I did it that way, but `config_from_kubeconfig` is used by `CliArgs` too.

So, I need to generate those there too. Probably, can add helper function to generate them and use it in both places if this will be more cleaner

k8s/proxy/src/lib.rs
voyvodov: Fixed!

k8s/proxy/src/lib.rs
tiagolobocastro: Yeah I think that would be great, then we can finally remove the NOTE on L30.
You could use a builder for this too rather than a function of 4 args:
```rust
#[derive(Default, Clone)]
pub struct KubeConfigOptions {
    pub kube_config_path: Option<PathBuf>,
    pub context: Option<String>,
    pub cluster: Option<String>,
    pub user: Option<String>,
}
```
Up to your preference

voyvodov: In fact, I don't think we need function with 4 arguments (e.g. don't see any valid reason to override the server or user), just one function "options from context" which will check and set correct one.

I think I made it (not sure if fn naming is fine). Can you check when you can?

tiagolobocastro:
bors try

k8s/proxy/src/lib.rs
tiagolobocastro: Alright, lgtm

tiagolobocastro (approved):

k8s/proxy/src/lib.rs
```diff
@@ -34,8 +37,21 @@ pub async fn config_from_kubeconfig(
...
 ) -> anyhow::Result<kube::Client> {
+    let options = kubeconfigoptions_from_context(context);
     Ok(kube::Client::try_from(
-        config_from_kubeconfig(kube_config_path).await?,
+        config_from_kubeconfig(kube_config_path, options).await?,
     )?)
 }
+
+/// Create a `KubeConfigOptions` from the given context.
+pub fn kubeconfigoptions_from_context(context: Option<String>) -> KubeConfigOptions {
```
Abhinandan-Purkait:
```suggestion
pub fn kubeconfig_options_from_context(context: Option<String>) -> KubeConfigOptions {
```

Abhinandan-Purkait (approved):

k8s/proxy/src/lib.rs
voyvodov: Changed as suggested.

tiagolobocastro:
Thanks @voyvodov for your first contribution :partying_face: 
bors merge

voyvodov:
> Thanks @voyvodov for your first contribution 🥳 bors merge

I must thank you for the quick review and all the help!

PR #1019, opened 2025-08-26
feat(simulation): add simulator of cluster bundles

tiagolobocastro:
The kubectl-mayastor plugin takes a support bundle which includes various resources, such as pools and volumes.
Sometimes it's helpful to not only visualize but also run commands against a similar configured cluster.
In order to aid this, we add a simulation tool here, which is able to load a bundle and attempts to re-create similar spec and state. The process is not entirely accurate because the bundle is taken using the public api, and here we are using the internal types.

### DIFF
--- a/control-plane/agents/src/bin/core/controller/reconciler/persistent_store.rs
+++ b/control-plane/agents/src/bin/core/controller/reconciler/persistent_store.rs
@@ -1,4 +1,7 @@
-use crate::controller::task_poller::{PollContext, PollResult, PollerState, TaskPoller};
+use crate::controller::{
+    reconciler::PollTriggerEvent,
+    task_poller::{PollContext, PollEvent, PollResult, PollerState, TaskPoller},
+};
 
 /// Reconcile dirty specs in the persistent store.
 /// This happens when we fail to update the persistent store and we have a "live" spec that
@@ -32,5 +35,13 @@ impl TaskPoller for PersistentStoreReconciler {
         }
 
         PollResult::Ok(PollerState::Idle)
+    }
+    async fn poll_event(&mut self, context: &PollContext) -> bool {
+        match context.event() {
+            PollEvent::TimedRun => true,
+            PollEvent::Triggered(PollTriggerEvent::SimulStart) => true,
+            PollEvent::Triggered(_event) => true,
+            PollEvent::Shutdown => true,
+        }
     }
 }
--- a/control-plane/agents/src/bin/core/controller/reconciler/poller.rs
+++ b/control-plane/agents/src/bin/core/controller/reconciler/poller.rs
@@ -69,7 +69,7 @@ impl ReconcilerWorker {
     /// The polling will continue until we receive the shutdown signal
     pub(super) async fn poller(mut self, registry: Registry) {
         // kick-off the first run
-        let mut event = PollEvent::Triggered(PollTriggerEvent::Start);
+        let mut event = PollEvent::Triggered(self.start_trigger(&registry));
         loop {
             let result = match &event {
                 PollEvent::Shutdown => {
@@ -96,6 +96,16 @@ impl ReconcilerWorker {
                     PollEvent::TimedRun
                 }
             };
+        }
+    }
+    fn start_trigger(&self, registry: &Registry) -> PollTriggerEvent {
+        let Some(sim_args) = registry.sim_args() else {
+            return PollTriggerEvent::Start;
+        };
+        if sim_args.no_start_event {
+            PollTriggerEvent::SimulStart
+        } else {
+            PollTriggerEvent::Start
         }
     }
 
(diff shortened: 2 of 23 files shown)


tiagolobocastro:
bors try

utils/bundle-sim/src/bin/bundle-sim.rs
```diff
@@ -0,0 +1,242 @@
...
+            }
+
+            if unpacked_dirs.len() != target_dirs.len() {
+                let mut target_dirs = target_dirs;
+                target_dirs.retain(|t| !unpacked_dirs.contains(t));
+                anyhow::bail!("{target_dirs:?} were not found in the bundle file")
+            }
+            self.bundle = self.untar_path.clone();
+        } else {
+            anyhow::ensure!(self.bundle.is_dir(), "--bundle must be a directory");
```
niladrih:
```suggestion
            anyhow::ensure!(self.bundle.is_dir(), "--bundle arg must be a file/directory");
```

utils/bundle-sim/src/bin/bundle-sim.rs
```diff
@@ -0,0 +1,242 @@
...
+        let mut etcd = Client::connect(["0.0.0.0:2379"], None).await?;
+        etcd_dump.rewind()?;
+        let reader = BufReader::new(etcd_dump);
+
+        let mut lines = reader.lines();
+        let ignore = "ffd20a56-8d97-4f68-8049-1cae2294a690".len();
+        while let (Some(Ok(key)), Some(Ok(value))) = (lines.next(), lines.next()) {
+            if key.len() == ignore {
+                break;
+            }
```
niladrih, line 137: What is special about this uuid?

niladrih (approved):

utils/bundle-sim/src/bin/bundle-sim.rs
tiagolobocastro: indeed, I moved the ensure to the one at the start of the function

utils/bundle-sim/src/bin/bundle-sim.rs
tiagolobocastro: Added a comment, it's to ignore old nexus health entries from mayastor v0

tiagolobocastro:
bors try

control-plane/agents/src/bin/core/controller/reconciler/persistent_store.rs
```diff
@@ -33,4 +36,12 @@ impl TaskPoller for PersistentStoreReconciler {
 
         PollResult::Ok(PollerState::Idle)
     }
+    async fn poll_event(&mut self, context: &PollContext) -> bool {
+        match context.event() {
+            PollEvent::TimedRun => true,
+            PollEvent::Triggered(PollTriggerEvent::SimulStart) => true,
```
Abhinandan-Purkait, line 42: Should we use this just on debug buillds?

control-plane/agents/src/bin/core/controller/reconciler/poller.rs
```diff
@@ -98,6 +98,16 @@ impl ReconcilerWorker {
             };
         }
     }
+    fn start_trigger(&self, registry: &Registry) -> PollTriggerEvent {
+        let Some(sim_args) = registry.sim_args() else {
+            return PollTriggerEvent::Start;
+        };
+        if sim_args.no_start_event {
+            PollTriggerEvent::SimulStart
+        } else {
```
Abhinandan-Purkait, line 107: Likewise here, anything that's on the code maybe a debug_assertion or a feature?

control-plane/agents/src/bin/core/main.rs
```diff
@@ -162,6 +162,18 @@ pub(crate) struct CliArgs {
     /// to configure cluster size per-pool.
     #[clap(long)]
     pool_cluster_size: Option<u32>,
+
+    /// Running in simulation mode.
+    #[clap(long, env = "SIMULATION", requires = "bundle")]
+    simulation: bool,
```
Abhinandan-Purkait, line 168: Or maybe hide this?

Abhinandan-Purkait (approved):

control-plane/agents/src/bin/core/controller/reconciler/persistent_store.rs
tiagolobocastro: We could though this may be needed on release builds, ex a problem we only hit on release builds

control-plane/agents/src/bin/core/main.rs
tiagolobocastro: No one is running agent-core directly like the plugin so I think it's probably fine to show these, but sure I'll hide it then

Abhinandan-Purkait: Upto you, it was more of a nit.

tiagolobocastro:
bors merge

PR #1026, opened 2025-09-19
add pool expand command to plugin

abhilashshetty04:
This adds expand subcommands to the mayastor cli

### DIFF
--- a/control-plane/plugin/src/resources/error.rs
+++ b/control-plane/plugin/src/resources/error.rs
@@ -66,6 +66,9 @@ pub enum Error {
         id: String,
         source: openapi::tower::client::Error<openapi::models::RestJsonError>,
     },
+    /// Error when get pool request fails.
+    #[snafu(display("No state for pool {id}. Please verify if node is online"))]
+    PoolStateError { id: String },
     /// Error when list pools request fails.
     #[snafu(display("Failed to list pools. Error {source}"))]
     ListPoolsError {
@@ -92,6 +95,12 @@ pub enum Error {
     /// Error when scale volume request fails.
     #[snafu(display("Failed to scale volume {id}. Error {source}"))]
     ScaleVolumeError {
+        id: String,
+        source: openapi::tower::client::Error<openapi::models::RestJsonError>,
+    },
+    /// Error when expand pool request fails.
+    #[snafu(display("Failed to expand pool {id}. Error {source}"))]
+    ExpandPoolError {
         id: String,
         source: openapi::tower::client::Error<openapi::models::RestJsonError>,
     },
--- a/control-plane/plugin/src/resources/mod.rs
+++ b/control-plane/plugin/src/resources/mod.rs
@@ -74,6 +74,17 @@ pub enum ScaleResources {
     },
 }
 
+/// The types of resources that support the 'expand' operation.
+#[derive(clap::Subcommand, Debug)]
+pub enum ExpandResources {
+    /// Expand a Pool to cover the entire span of the disk.
+    /// Please refer maxExpansion parameter in the docs which controls maxExpandableSize of the pool.
+    Pool {
+        /// ID of the pool.
+        id: PoolId,
+    },
+}
+
 /// The types of resources that support the 'SetProperty' operation.
 #[derive(clap::Subcommand, Debug)]
 pub enum SetPropertyResources {
--- a/control-plane/plugin/src/resources/pool.rs
+++ b/control-plane/plugin/src/resources/pool.rs
@@ -1,13 +1,12 @@
 extern crate utils as external_utils;
 use super::VolumeId;
 use crate::{
-    operations::{Cordoning, GetWithArgs, Label, ListWithArgs, PluginResult},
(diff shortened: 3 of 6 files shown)


niladrih (approved):

control-plane/plugin/src/resources/utils.rs
```diff
@@ -51,7 +51,9 @@ lazy_static! {
         "ALLOCATED",
         "AVAILABLE",
         "COMMITTED",
-        "ENCRYPTED"
+        "ENCRYPTED",
+        "DISK_CAPACITY",
+        "MAX_EXPANDABLE_SIZE"
```
tiagolobocastro:
```suggestion
        "DISK-CAPACITY",
        "MAX-EXPANDABLE-SIZE"
```

tiagolobocastro: Also hope this is not getting too wide otherwise better to leave only in yaml/json

control-plane/plugin/src/resources/pool.rs
```diff
@@ -548,6 +549,50 @@ impl From<UncordonReq> for models::PoolCordonReq {
...
+                    utils::print_table(output, pool.into_body());
+                }
+                OutputFormat::None => {
+                    // Unlikely that pool won't have state.
+                    let pool_now = pool
+                        .into_body()
+                        .state
+                        .ok_or(Error::PoolStateError { id: id.to_string() })?
+                        .capacity;
+                    println!("Pool expanded from {before_size} to {pool_now}");
```
tiagolobocastro, line 585: Convert bytes to human using `::utils::bytes::into_human`

control-plane/plugin/src/resources/mod.rs
```diff
@@ -74,6 +74,16 @@ pub enum ScaleResources {
     },
 }
 
+/// The types of resources that support the 'expand' operation.
+#[derive(clap::Subcommand, Debug)]
+pub enum ExpandResources {
+    /// Expand Pool.
```
tiagolobocastro: Something like this, you could elaborate

```suggestion
    /// Expand a Pool to cover the entire span of the disk.
    /// Pool is only expandable if it has been created with sufficient expandable slack via `maxExpansion`.
```

control-plane/plugin/src/resources/error.rs
```diff
@@ -66,6 +66,9 @@ pub enum Error {
         id: String,
         source: openapi::tower::client::Error<openapi::models::RestJsonError>,
     },
+    /// Error when get pool request fails.
+    #[snafu(display("Failed to get pool state {id}. Please verify if node is online"))]
```
tiagolobocastro:
```suggestion
    #[snafu(display("No state for pool {id}. Please verify if node is online"))]
```

tiagolobocastro (approved):

control-plane/plugin/src/resources/utils.rs
abhilashshetty04: let me check

abhilashshetty04: disk was too long. aio:///dev/disk/by-id/scsi-SHC_Volume_103171581?uuid=5f936733-38bc-4cbc-b8c6-149b120e4ed6.  uuid seems irrelevant here.

I modified it to take from spec (/dev/disk/by-id/scsi-SHC_Volume_103171581). Now it fits fine.

control-plane/plugin/src/resources/mod.rs
abhilashshetty04: Pool expansion is possible to certain extent without `max_expansion` also. As cluster size increases with default max_expansion also user will see higher `max_expandable_size`.


/// Expand a Pool to cover the entire span of the disk.
/// Please resize underlying disk before executing this cmd.
/// Refer `max_expandable_size` of the pool before resizing disk.

tiagolobocastro: Please also include the bit about creating the Pool with sufficient slack via `maxExpansion` since that's very important

control-plane/plugin/src/resources/utils.rs
tiagolobocastro: hmm the uuid is fine but then you loose the defaulted disk protocol

abhilashshetty04: Yeah...I think we can parse protocol and append or just split on '?'

control-plane/plugin/src/resources/mod.rs
abhilashshetty04: Is it needed here? Its well documented on CRD schema docs.

tiagolobocastro: User may not even read the CRD schema
Since this is a tricky operation, I think we better have good docs at any user endpoint

abhilashshetty04: /// Please refer maxExpansion parameter in the docs which controls maxExpandableSize.

abhilashshetty04:
bors merge

PR #1027, opened 2025-09-30
fix(pool): don't delete pool on abort

tiagolobocastro:
When creating pool, it may timeout. In this case we leave it as is, and allow the caller to retry.
But on the first retry, if the pool has not completed the create, then we see the abort error code instead, and in this case we should also leave as is, and allow caller to retry again.

Also update the existing test to do the retry.

### DIFF
--- a/control-plane/agents/src/bin/core/controller/resources/operations_helper.rs
+++ b/control-plane/agents/src/bin/core/controller/resources/operations_helper.rs
@@ -89,7 +89,7 @@ impl OnCreateFail {
             // 3. dataplane core is shared with other processes
             // TODO: use higher timeout on larger pool sizes or potentially make this
             //  an async operation.
-            tonic::Code::Cancelled => Self::LeaveAsIs,
+            tonic::Code::Cancelled | tonic::Code::Aborted => Self::LeaveAsIs,
             _ => Self::SetDeleting,
         }
     }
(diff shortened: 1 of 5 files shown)


control-plane/agents/src/bin/core/controller/resources/operations_helper.rs
```diff
@@ -89,7 +89,7 @@ impl OnCreateFail {
             // 3. dataplane core is shared with other processes
             // TODO: use higher timeout on larger pool sizes or potentially make this
             //  an async operation.
-            tonic::Code::Cancelled => Self::LeaveAsIs,
+            tonic::Code::Cancelled | tonic::Code::Aborted => Self::LeaveAsIs,
```
abhilashshetty04, line 92: When do we get Aborted ?

abhilashshetty04 (approved):

control-plane/agents/src/bin/core/controller/resources/operations_helper.rs
tiagolobocastro: When pool lock is taken

dsharma-dc (approved):

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

PR #1028, opened 2025-09-30
Cherry-Pick 1027 - don't delete pool on abort

tiagolobocastro:
When creating pool, it may timeout. In this case we leave it as is, and allow the caller to retry.
But on the first retry, if the pool has not completed the create, then we see the abort error code instead, and in this case we should also leave as is, and allow caller to retry again.

Also update the existing test to do the retry.


abhilashshetty04 (approved):

dsharma-dc (approved):

tiagolobocastro:
bors merge

PR #1029, opened 2025-10-09
ha/node: use kernel event monitor

tiagolobocastro:
Switches to latest tokio-udev which has the kernel udev monitor.
Use the kernel monitor which allows the ha node to function in more
environments, such as the deployer where we had to fake it.

### DIFF
--- a/.github/workflows/mod-update.yml
+++ b/.github/workflows/mod-update.yml
@@ -1,0 +1,43 @@
+name: Submodule Update
+on:
+  workflow_dispatch:
+    inputs:
+      submodules:
+        description: 'Override modules to update via comma-separated list of names'
+        required: false
+        default: ''
+
+jobs:
+  submodule:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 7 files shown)


tiagolobocastro:
bors try

tiagolobocastro:
bors try

tiagolobocastro:
bors try

dsharma-dc (approved):
Should you test this on different flavours of linux e.g RHEL, Talos etc if it works seamlessly to subscribe kernel events?

tiagolobocastro:
> Should you test this on different flavours of linux e.g RHEL, Talos etc if it works seamlessly to subscribe kernel events?

Just tested on Talos with `Linux talos-default-worker-1 6.12.25-talos`
Also confirmed multipath is still reported correctly: `~ #  cat /sys/module/nvme_core/parameters/multipath
Y`

tiagolobocastro:
bors merge

PR #1030, opened 2025-10-13
chore: staging ci for cp

pchandra19:


Abhinandan-Purkait:
The changes look good, although I am curious how are we going to map different repos to helm chart, I would say rather place all images in `ghcr.io/openebs/mayastor/dev` rather than individual repos, otherwise we can't change the registry on helm chart based on control-plane, data-plane and extensions.

cc @tiagolobocastro

Abhinandan-Purkait:
Also I think we should move these scripts to common ground, I believe that's already been considered

PR #1031, opened 2025-10-20
Use PreCondition Errors when appropriate

tiagolobocastro:
fix(csi): use correct precondition error
    
    In previous side car versions, non final errors used to lead into a burst
    of retries. Because of this we had "massaged" the returned error codes.
    Now the issue has been fixed upstream, so we can return the appropriate
    error code.
    
---

    refactor(pool/scheduling): use more precise snapshot errors
    
    When a pool is cordoned the returned error should be failed precondition.
    When pools are filtered out, we end up with no choices, and return a
    hardcoded generic error of resource exhausted.
    
    In order to achieve this, we have to remember what reasons were used to
    exclude a pool.
    This changes introduces a list of excluded resources when they are filtered
    out.
    For the moment, this has just been done for the pools, though it could be
    extended to other schedulers.
    
---

    refactor: format debug for stringified id's
    
    Makes debug display properly, example from
    VolumeId("xxxx", "xxxx)
    to
    VolumeId("xxxx")
    
    Implements debug for Service and Watchdog to avoid printing infinite
    loop though these aren't used in practice, it may be useful to print
    while debugging something.
    
    Also remove redundant requests from service handlers.

### DIFF
--- a/control-plane/agents/src/bin/core/controller/scheduling/mod.rs
+++ b/control-plane/agents/src/bin/core/controller/scheduling/mod.rs
@@ -10,8 +10,16 @@ use crate::controller::scheduling::{
     resources::{ChildItem, PoolItem, ReplicaItem},
     volume::{ReplicaResizePoolsContext, VolumeReplicasForNexusCtx},
 };
-use std::{cmp::Ordering, collections::HashMap, future::Future};
+
+use stor_port::transport_api::ResourceKind;
 use weighted_scoring::{Criteria, Value, ValueGrading, WeightedScore};
+
+use snafu::Snafu;
+use std::{
+    cmp::Ordering,
+    collections::{btree_map::Entry, BTreeMap, HashMap},
+    future::Future,
+};
 
 #[async_trait::async_trait(?Send)]
 pub(crate) trait ResourcePolicy<Request: ResourceFilter>: Sized {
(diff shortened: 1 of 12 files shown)


tiagolobocastro:
Part of OEP https://github.com/openebs/openebs/issues/3903

control-plane/agents/src/bin/core/controller/scheduling/mod.rs
```diff
@@ -10,7 +10,13 @@ use crate::controller::scheduling::{
...
     volume::{ReplicaResizePoolsContext, VolumeReplicasForNexusCtx},
 };
-use std::{cmp::Ordering, collections::HashMap, future::Future};
+use snafu::Snafu;
+use std::{
+    cmp::Ordering,
+    collections::{btree_map::Entry, BTreeMap, HashMap},
+    future::Future,
+};
+use stor_port::transport_api::ResourceKind;
```
Abhinandan-Purkait: Please arrange the imports since you are here.

Abhinandan-Purkait (approved):

control-plane/agents/src/bin/core/controller/scheduling/mod.rs
tiagolobocastro: arrange as in?

Abhinandan-Purkait: The way we have been doing at other places, first crate and super then own deps then third party seperated by line and then fmt auto arranges them

tiagolobocastro: Right, so we need to give newlines for that, I'll do it
Also on a related matter, I wonder if we should modify the uses to have 1 per line; recently on cherry-picking I've been finding it difficult when there are conflicts because we have multiple per line

tiagolobocastro:
bors merge

PR #1032, opened 2025-10-23
Bugfixes for spread, snapshot restore and plugin sort

tiagolobocastro:
fix(node-spread): use spread policy on replica add/remove
    
    Use node-spread policy when adding new replicas to a volume.
    This is done by updating the existing filtering logic, ensuring that
    new replicas are created on nodes which contain the spread keys and
    also ensuring that the new values are different from existing replicas.
    
    Also when removing replicas we now use the spread, preferring to remove
    replicas which don't contain the spreads keys (all else being equal).
    If more than 1 replicas don't contain the spread keys, the deadlock is
    broken by picking the one with more spread value collisions.
    
---

    feat(plugin): add --show-labels to volume replica topologies

---

    fix(snapshot): candidates may be less than the snapshots
    
    This is however fine, because user may want to create lesser replicas
    on the cloned volume.
    
    Also fixes a bug whereas the index of the snapshot was used to check
    for completion which is not true when one of the requests fails.
    
---

    fix(snapshot): specify timeout for create volume from snapshot
    
---

    fix(spread): use combinations fine tune candidates
    
    Using the given list directly may not work correctly if the first element
    of the list ends up not being valid for use with the remaining ones.
    
    Instead, we now create all usable combinations and ensure that only valid
    ones are considered before completing the volume creation.
    
    Another bug was found which may leave replicas around until nexus creation.
    This may again lead into using replicas which don't obey the topology. We
    now clean up previous attempts before starting volume creation.
    
---

    fix(plugin): sort node,pool labels and block devlinks
    
---

    fix(plugin): sort tabled output on first column
    
    Add sort_rows method to CreateRows and CreateRow traits.
    The first column is typically the identifier of something so default
    to sorting by row. This may be opted out ofcourse.
    
    Certain outputs such as the topologies are a bit complicated so have
    to be sorted manually prior.
    
    This approach has the downside that the json/yaml output is not sorted,
    but that's probably fine for now.

### DIFF
--- a/control-plane/agents/src/bin/core/volume/clone_operations.rs
+++ b/control-plane/agents/src/bin/core/volume/clone_operations.rs
@@ -121,6 +121,7 @@ impl CreateVolumeExe for SnapshotCloneOp<'_> {
     type Candidates = Vec<SnapshotCloneSpecParams>;
 
     async fn setup<'a>(&'a self, context: &mut Context<'a>) -> Result<Self::Candidates, SvcError> {
+        // todo: topology is not being used here at all
         let clonable_snapshots = self.cloneable_snapshot(context).await?;
         let volume = context.volume.as_ref();
         if volume.num_replicas > clonable_snapshots.len() as u8 {
@@ -142,20 +143,19 @@ impl CreateVolumeExe for SnapshotCloneOp<'_> {
         let volume_replicas = context.volume.as_ref().num_replicas as usize;
         // todo: need to add new replica and do full rebuild, if clonable snapshots
         // count is less than volume replicas count.
-        for (idx, clone_replica) in clone_replicas.iter().enumerate() {
-            match OperationGuardArc::<ReplicaSpec>::create_ext(context.registry, clone_replica)
+        for clone_replica in clone_replicas {
+            match OperationGuardArc::<ReplicaSpec>::create_ext(context.registry, &clone_replica)
                 .await
             {
                 Ok(replica) => {
                     replicas.push(replica);
-                    if idx + 1 == volume_replicas {
+                    if replicas.len() == volume_replicas {
                         break;
                     }
                 }
                 Err(error) => {
                     context.volume.error(&format!(
-                        "Failed to create replica {:?} for volume, error: {}",
-                        clone_replica,
+                        "Failed to create replica {clone_replica:?} for volume, error: {}",
                         error.full_string()
                     ));
                 }
@@ -166,6 +166,7 @@ impl CreateVolumeExe for SnapshotCloneOp<'_> {
 
     async fn undo<'a>(&'a self, _context: &mut Context<'a>, _replicas: Vec<Replica>) {
         // nothing to undo since we only support 1-replica snapshot
+        // todo: we do support multi-replica snapshots, this should have been tweaked :(
     }
 }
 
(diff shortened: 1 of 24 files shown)


tiagolobocastro:
bors try

control-plane/agents/src/bin/core/volume/specs.rs
```diff
@@ -191,7 +191,7 @@ pub(crate) async fn volume_replica_candidates(
 
     volume_spec.trace(&format!(
         "Creation pool candidates for volume: {:?}",
-        pools.iter().map(|p| p.state()).collect::<Vec<_>>()
+        pools.iter().take(10).map(|p| p.state()).collect::<Vec<_>>()
```
dsharma-dc: what's the reason for this `take()` ? To limit log length?

control-plane/agents/src/bin/core/volume/clone_operations.rs
```diff
@@ -142,20 +143,19 @@ impl CreateVolumeExe for SnapshotCloneOp<'_> {
...
-        for (idx, clone_replica) in clone_replicas.iter().enumerate() {
-            match OperationGuardArc::<ReplicaSpec>::create_ext(context.registry, clone_replica)
+        for clone_replica in clone_replicas {
+            match OperationGuardArc::<ReplicaSpec>::create_ext(context.registry, &clone_replica)
                 .await
             {
                 Ok(replica) => {
                     replicas.push(replica);
-                    if idx + 1 == volume_replicas {
+                    if replicas.len() == volume_replicas {
```
dsharma-dc, line 152: So if requested replica count is more than available replicasnaps, we provision that much and return success?
If yes, does that get reconciled automatically afterwards based on spec?

dsharma-dc (approved):

control-plane/agents/src/bin/core/volume/clone_operations.rs
```diff
@@ -121,6 +121,7 @@ impl CreateVolumeExe for SnapshotCloneOp<'_> {
     type Candidates = Vec<SnapshotCloneSpecParams>;
 
     async fn setup<'a>(&'a self, context: &mut Context<'a>) -> Result<Self::Candidates, SvcError> {
+        // todo: topology is not being used here at all
```
abhilashshetty04, line 124: Snapshot clone is restore from snapshot right? It needs to be placed on same pool where snapshot is present?

abhilashshetty04 (approved):

abhilashshetty04:
Does it need test changes? @tiagolobocastro

control-plane/agents/src/bin/core/volume/specs.rs
tiagolobocastro: Yes, otherwise there's way too much information.
Also considered only tracing the names of the pools,  wdyt?

control-plane/agents/src/bin/core/volume/clone_operations.rs
tiagolobocastro: Let's say snapshot has 3 replicas, and we want a 1-replica clone only.
First clone fails, so we continue the loop. Then it succeeds, and we have 1 replica so we can bail out of the loop as we have the replica we need.

control-plane/agents/src/bin/core/volume/clone_operations.rs
tiagolobocastro: Yeah but we still need to validate the topology is a match, otherwise volume may have topology different than its replicas

tiagolobocastro:
> Does it need test changes? @tiagolobocastro

Probably yes but not sure how to do it yet, existing python tests are hardcoding the node labels at the start, and not making it part of the premise

Abhinandan-Purkait (approved):

control-plane/agents/src/bin/core/volume/specs.rs
dsharma-dc: > Yes, otherwise there's way too much information. Also considered only tracing the names of the pools, wdyt?

I think tracing names of `all` pools, instead of tracing complete info about `10` pools will be better.

tiagolobocastro:
bors merge

PR #1033, opened 2025-10-27
Node spread cont

tiagolobocastro:
fix(node-spread): use spread policy on replica add/remove
    
    Use node-spread policy when adding new replicas to a volume.
    This is done by updating the existing filtering logic, ensuring that
    new replicas are created on nodes which contain the spread keys and
    also ensuring that the new values are different from existing replicas.
    
    Also when removing replicas we now use the spread, preferring to remove
    replicas which don't contain the spreads keys (all else being equal).
    If more than 1 replicas don't contain the spread keys, the deadlock is
    broken by picking the one with more spread value collisions.
    
---

    feat(plugin): add --show-labels to volume replica topologies

### DIFF
--- a/control-plane/agents/src/bin/core/controller/scheduling/mod.rs
+++ b/control-plane/agents/src/bin/core/controller/scheduling/mod.rs
@@ -8,7 +8,7 @@ mod volume_policy;
 use crate::controller::scheduling::{
     nexus::GetPersistedNexusChildrenCtx,
     resources::{ChildItem, PoolItem, ReplicaItem},
-    volume::{ReplicaResizePoolsContext, VolumeReplicasForNexusCtx},
+    volume::{GetChildForRemovalContext, ReplicaResizePoolsContext, VolumeReplicasForNexusCtx},
 };
 use std::{cmp::Ordering, collections::HashMap, future::Future};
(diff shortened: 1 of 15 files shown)


control-plane/agents/src/bin/core/controller/scheduling/mod.rs
```diff
@@ -164,26 +164,39 @@ pub(crate) struct ChildSorters {}
...
+        b: &ReplicaItem,
+    ) -> std::cmp::Ordering {
         match Self::sort_by_health(a, b) {
             Ordering::Equal => match Self::sort_by_child(a, b) {
                 Ordering::Equal => {
                     // Remove replicas from nodes which are cordoned with most priority.
                     // remove mismatched topology replicas first
                     if let (Some(a), Some(b)) = (a.valid_node_topology(), b.valid_node_topology()) {
-                        match a.cmp(b) {
+                        match a.cmp(&b) {
```
Abhinandan-Purkait, line 178: `&b -->b`

Abhinandan-Purkait (approved):

control-plane/agents/src/bin/core/controller/scheduling/mod.rs
tiagolobocastro: has to be `&b`, b: bool

PR #1037, opened 2025-11-01
feat(affinity-group): support scale down to 1 replica

tiagolobocastro:
Scaling down affinity groups to 1 replica was previously blocked because we didn't want to get into a situation where replicas would end up sharing the same node, defeating the purpose of Affinity Groups!

We now loosen this restriction by allowing the scale down if we don't get into that scenario :)

This is done by creating a list of restricted nodes which are nodes where a 1-replica volume part of the same affinity group places its replica. In such case, we don't want to scale down in such a way that the last remaining replica would be placed in restricted nodes!

This isn't the full monty as replicas may be sharing nodes when Scaling back from 3 to 2, which would end up impeding further scaling down to 1! For this we'd have to implement some logic to improve the sorting. This limitation is fine for now, we can document scaling up and down as a way of moving replicas around!

### DIFF
--- a/control-plane/agents/src/bin/core/controller/scheduling/affinity_group.rs
+++ b/control-plane/agents/src/bin/core/controller/scheduling/affinity_group.rs
@@ -1,5 +1,5 @@
 use crate::controller::{registry::Registry, resources::ResourceUid};
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use stor_port::types::v0::{
     store::volume::{AffinityGroupSpec, VolumeOperation, VolumeSpec},
     transport::{NodeId, PoolId},
@@ -59,6 +59,27 @@ pub(crate) async fn get_pool_ag_replica_count(
         }
     }
     pool_ag_replica_count
+}
+
+/// Get a set of nodes which contain the replica of a single-replica volume, part of the affinity group.
+pub(crate) async fn ag_restricted_nodes(
+    affinity_group_spec: &AffinityGroupSpec,
+    registry: &Registry,
+) -> HashSet<NodeId> {
+    let mut nodes = HashSet::<NodeId>::new();
+    let specs = registry.specs();
+    for volume_id in affinity_group_spec.volumes() {
+        let Some(volume) = registry.specs().volume_rsc(volume_id) else {
+            // should not really happen
+            continue;
+        };
+        if volume.lock().num_replicas != 1 {
+            continue;
+        }
+
+        nodes.extend(specs.volume_replica_nodes(volume_id));
+    }
+    nodes
 }
 
 /// Get the map of node to the number of the Affinity Group nexuses on the node.
--- a/control-plane/agents/src/bin/core/controller/scheduling/volume.rs
+++ b/control-plane/agents/src/bin/core/controller/scheduling/volume.rs
@@ -22,11 +22,12 @@ use stor_port::types::v0::{
     transport::{NodeId, PoolId, VolumeState},
 };
 
-use crate::controller::scheduling::ResourceExcReason;
+use crate::controller::scheduling::{affinity_group::ag_restricted_nodes, ResourceExcReason};
 use std::{
-    collections::{BTreeMap, HashMap},
+    collections::{BTreeMap, HashMap, HashSet},
     ops::Deref,
 };
+use stor_port::types::v0::store::volume::AffinityGroupSpec;
 
 /// Move replica to another pool.
 #[derive(Default, Clone)]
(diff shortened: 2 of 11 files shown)


tiagolobocastro:
Resolves https://github.com/openebs/openebs/issues/3795

control-plane/agents/src/bin/core/controller/scheduling/affinity_group.rs
```diff
@@ -61,6 +61,28 @@ pub(crate) async fn get_pool_ag_replica_count(
...
+    for volume_id in affinity_group_spec.volumes() {
+        let Some(volume) = registry.specs().volume_rsc(volume_id) else {
+            // should not really happen
+            continue;
+        };
+        if volume.lock().num_replicas != 1 {
+            continue;
+        }
+
+        tracing::info!("Adding {volume_id}");
```
Abhinandan-Purkait: Can you add a bit more info on this log, just `Adding 9c39458e-5cbc-4ba2-bc2d-f040b79a8850` might look a bit vague?

control-plane/agents/src/bin/core/controller/scheduling/volume.rs
```diff
@@ -448,6 +480,24 @@ impl ReplicaRemovalCandidates {
...
+        let ag = self.context.affinity_group.as_ref()?;
+        if ag.restricted_nodes.is_empty() {
+            return None;
+        }
+        //     N
+        // R 1 2 3
+        // 1 x x
+        // 2 x x
+        // 3 x x x
+        // todo: In this case, shouldn't we block removal or r3n3 ?
```
Abhinandan-Purkait, line 493: Indeed! But then how do we do it? Non single replicas also in ag_restricted_nodes? But that might not work well

control-plane/agents/src/bin/core/tests/volume/affinity_group.rs
```diff
@@ -191,6 +220,73 @@ async fn scale_up_test(cluster: &Cluster) {
...
             .await
-            .expect_err("Scale down should fail");
+            .expect("Scale down should fail");
+    }
+
+    // Replica location now looks like this:
+    // 1 2 3
+    // x
+    //   x
+    // x   x
```
Abhinandan-Purkait, line 230: Remove this? And rather put a quick steps as a doc comment to the scale_up_down_test method?

control-plane/agents/src/bin/core/volume/specs.rs
```diff
@@ -1090,10 +1088,11 @@ impl SpecOperationsHelper for VolumeSpec {
...
                     // Validation for Affinity Group volume's replica count cannot go below 2.
                     if self.affinity_group.is_some() && *replica_count < 2 {
-                        Err(SvcError::RestrictedReplicaCount {
-                            resource: ResourceKind::AffinityGroup,
-                            count: *replica_count,
-                        })
+                        // Err(SvcError::RestrictedReplicaCount {
+                        //     resource: ResourceKind::AffinityGroup,
+                        //     count: *replica_count,
+                        // })
```
Abhinandan-Purkait: Remove this?

Abhinandan-Purkait (approved):
LGTM. Thanks! Although I am a bit concerned about the scenario you mentioned when we end up removing a replica which can potentially disrupt ag group, when all ag group volumes are non single.

I would say it would be good to restrict the last volume from scaling down and fail saying ag group violation --> Scale up again and fix it gradually?

control-plane/agents/src/bin/core/controller/scheduling/affinity_group.rs
tiagolobocastro: Oops, that was for debug, I'll remove it

control-plane/agents/src/bin/core/volume/specs.rs
tiagolobocastro: Yeah, can remove the first if completely

control-plane/agents/src/bin/core/controller/scheduling/volume.rs
tiagolobocastro: We could consider nodes containing replicas from more than 1 volume to be restricted nodes? Though this may be a bit premature

Abhinandan-Purkait: This might happen easily though so would it make sense to do that?

tiagolobocastro: Possibly, though it's probably still fine as long as the user doesn't intend to reduce replicas down to 1.
This needs some more consideration, I'll leave it out of this PR since it can already happen anyway.

tiagolobocastro:
bors merge

tiagolobocastro:
bors cancel

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

tiagolobocastro:
@mergify backport release/2.10

tiagolobocastro:
@mergify backport release/2.10

PR #1039, opened 2025-11-04
test(ag): scale correct volume

tiagolobocastro:
The scale down of the affinity group is ran in parallel, so it might not be the last volume in the list which gets attempted last. Fetch the correct volume when doing the scale up/down trick.


Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

PR #1049, opened 2025-12-15
fix: don't fail node unpublish if target path is an empty file

Abhinandan-Purkait:


control-plane/csi-driver/src/bin/node/node.rs
```diff
@@ -393,14 +393,20 @@ impl node_server::Node for Node {
...
+                // and return success.
                 if target_path.is_file() {
-                    return Err(Status::new(
-                        Code::Unknown,
-                        format!(
-                            "Failed to unpublish volume {}: {} is a file.",
-                            &msg.volume_id, &msg.target_path
-                        ),
-                    ));
+                    return match fs::remove_file(target_path) {
```
tiagolobocastro: We're not the ones responsibly for its creation or are we? If not then perhaps we should not delete it

Abhinandan-Purkait: We create the target path file. So we should remove it. That's how it's at the next step

ref: https://github.com/openebs/mayastor-control-plane/blob/e04b6966068b59edeb4cfe3239fa08ebb9606c62/control-plane/csi-driver/src/bin/node/block_vol.rs#L131

tiagolobocastro: sounds good then
btw we're not erroring out when failing to remove it on the function you've pasted above, might want to keep the same behavior

tiagolobocastro (approved):

Abhinandan-Purkait:
bors merge

PR #1051, opened 2025-12-18
Add pool diagnostics to the internal transport and agent-core

tiagolobocastro:
refactor(pstor/pool): split pool spec types
    
    As with other types, the pool pstor info was incorrectly stored as the spec.
    This makes it harder to separate user spec and other information.
    
    For now we simply create an innter type PoolUSpec which is basically the user
    specification of a pool.
    This allows us to transport pool spec and pool diag separately.
    
---

    test(pool/diag): add pool diag info check to pool reconciler
    
    Tests both disk not found and invalid superblock.
    
---

    feat(pool/diag): collect import errors during pool import
    
    When pool import fails, collect error reason.
    This is stored in runtime information within the PoolSpec.
    
    A downside of this is that the PoolSpec in the pstor, alike the volume,
    is now a misnomer.
    In the future we should introduce an upgrade way of updating the entries.
    
---

    feat(pool/diag): add grpc proto and transport interface

### DIFF
--- a/control-plane/agents/src/bin/core/pool/operations_helper.rs
+++ b/control-plane/agents/src/bin/core/pool/operations_helper.rs
@@ -1,8 +1,6 @@
-use crate::controller::io_engine::HostApi;
-use crate::node::wrapper::NodeWrapper;
 use crate::{
     controller::{
-        io_engine::PoolApi,
+        io_engine::{HostApi, PoolApi},
         registry::Registry,
         resources::{
             operations::ResourceLifecycle,
@@ -10,25 +8,27 @@ use crate::{
             OperationGuardArc,
         },
     },
-    node::wrapper::GetterOps,
+    node::wrapper::{GetterOps, NodeWrapper},
 };
 use agents::{errors, errors::SvcError};
-use stor_port::transport_api::ResourceKind;
-use stor_port::types::v0::transport::{GetBlockDevices, PoolDeviceUri};
-use stor_port::types::v0::{
-    store::{
-        pool::PoolSpec,
-        replica::{PoolRef, ReplicaSpec},
+use stor_port::{
+    transport_api::ResourceKind,
+    types::v0::{
+        store::{
+            pool::{PoolImportOp, PoolOperation, PoolSpec},
+            replica::{PoolRef, ReplicaSpec},
+        },
+        transport::{
+            CreatePool, DestroyReplica, GetBlockDevices, ImportPool, NodeId, PoolDeviceUri,
+            PoolDiag, PoolDiskError, PoolError, PoolErrorCode, PoolState, ReplicaOwners,
+        },
     },
-    transport::{CreatePool, DestroyReplica, NodeId, ReplicaOwners},
 };
 
 use itertools::Itertools;
 use regex::Regex;
 use snafu::OptionExt;
-use std::collections::HashSet;
-use std::ops::Deref;
-use std::sync::Arc;
+use std::{collections::HashSet, ops::Deref, sync::Arc};
 use tokio::sync::RwLock;
 
 impl OperationGuardArc<PoolSpec> {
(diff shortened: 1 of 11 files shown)


control-plane/grpc/proto/v1/pool/pool.proto
```diff
@@ -5,6 +5,36 @@ import "google/protobuf/wrappers.proto";
 
 package v1.pool;
 
+message PoolHealth {
```
```suggestion
message PoolDiagnostics {
```

Wondering if this would be a better naming?

abhilashshetty04: This will always have import errors in them. If yes, we can just call it as ImportError.

struct ImportError {
errors: Vec<DiskPoolError>
}

control-plane/agents/src/bin/core/pool/operations_helper.rs
```diff
@@ -60,6 +60,56 @@ impl OperationGuardArc<PoolSpec> {
...
+                    _ => PoolErrorCode::Unknown,
+                };
+                // todo: this is rather long as it includes details and metadata... trim it?
+                let msg = error.to_string();
+                PoolError { code, msg }
+            };
+            match error.tonic_code() {
+                tonic::Code::NotFound | tonic::Code::InvalidArgument => {
+                    let disks = pool_spec.disks.first().map(|d| d.to_string());
+                    *reporter.lock().expect("not poisoned") = Some(PoolHealth {
```
abhilashshetty04: Will we clear this if import passes eventually?

abhilashshetty04:
@tiagolobocastro, Can we write unit test for this?

control-plane/grpc/proto/v1/pool/pool.proto
tiagolobocastro: Not necessarily, this may be expanded for something like keep alives on a timer, smart data etc

control-plane/agents/src/bin/core/pool/operations_helper.rs
tiagolobocastro: Yes that's already here on the success complete case

tiagolobocastro:
> @tiagolobocastro, Can we write unit test for this?

I tried but it's tricky, when a device is missing on container restarts it creates a directory in its place
I could try with a file instead of a block device

abhilashshetty04:
nit: commits have different type.. feat(pool/health) and feat(pool-health) was it intended?

control-plane/stor-port/src/types/v0/transport/pool.rs
```diff
@@ -80,6 +80,52 @@ impl Deref for CtrlPoolState {
...
+    /// Error encountered.
+    pub error: PoolError,
+}
+
+/// Pool state information - as reported by the io-engine.
+#[derive(Serialize, Deserialize, Default, Debug, Clone, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct PoolHealth {
+    /// Errors encountered when trying to import the pool.
+    pub import_errors: Vec<PoolDiskError>,
```
abhilashshetty04, line 126: Does it need to be vector? Are we going to store more then 1 errors or just the latest failure information.

tiagolobocastro: In case there's more than 1 error detected.

tiagolobocastro:
> nit: commits have different type.. feat(pool/health) and feat(pool-health) was it intended?

no :D 
I'll fix that, but also would like to check wether to use term health or diagnostic (diag) ?
CC @Abhinandan-Purkait

control-plane/stor-port/src/types/v0/transport/pool.rs
abhilashshetty04: Suppose we face error during request no 1: DiskNotFound. During request no 3 Disk was there but failed due to crc mismatch. So will we have both of them in import errors vec?

Is it possible that we may see more then 1 error in one request?

tiagolobocastro: On the same request; say you have 2 disks, 1 not found, another is ReadIoError;
Also in the future we may have other tests which we perform before import, ex: latency tests, etc

tiagolobocastro:
> > @tiagolobocastro, Can we write unit test for this?
> 
> I tried but it's tricky, when a device is missing on container restarts it creates a directory in its place I could try with a file instead of a block device

@abhilashshetty04 the reconciler tests were already doing something similar to enhanced that to check for invalid super block and not found errors.

abhilashshetty04:
> > > @tiagolobocastro, Can we write unit test for this?
> > 
> > 
> > I tried but it's tricky, when a device is missing on container restarts it creates a directory in its place I could try with a file instead of a block device
> 
> @abhilashshetty04 the reconciler tests were already doing something similar to enhanced that to check for invalid super block and not found errors.

Is it there on the same PR?

tiagolobocastro:
> > > > @tiagolobocastro, Can we write unit test for this?
> > > 
> > > 
> > > I tried but it's tricky, when a device is missing on container restarts it creates a directory in its place I could try with a file instead of a block device
> > 
> > 
> > @abhilashshetty04 the reconciler tests were already doing something similar to enhanced that to check for invalid super block and not found errors.
> 
> Is it there on the same PR?

Yes it's here

control-plane/agents/src/bin/core/pool/operations_helper.rs
Abhinandan-Purkait: Do we need the mutex here? I couldn't spot any critical section here?

control-plane/grpc/proto/v1/pool/pool.proto
Abhinandan-Purkait: Actually it's a difficult choice here, the process is diagnostic but the outcome comes as a health metric. So, maybe health makes sense?

Abhinandan-Purkait (approved):

control-plane/agents/src/bin/core/pool/operations_helper.rs
tiagolobocastro: Yes we need mutex because the reporter is shared, so we need mutex to get a mutable reference

control-plane/grpc/proto/v1/pool/pool.proto
tiagolobocastro: Yeah it is related to health, but it's not exactly the pool health, this doesn't tell you if the pool is healthy, only reports diagnostic information?

abhilashshetty04 (approved):

control-plane/grpc/proto/v1/pool/pool.proto
abhilashshetty04: I think we can keep it as Diagnostics.

tiagolobocastro:
bors merge

PR #1052, opened 2025-12-23
Support MultiNodeMultiWriter for block volumes (kubevirt)

tiagolobocastro:
test(csi/rwx): extend bdd with rwx block
    
    Attempts to republish volume to another node with rwx block
    todo: add negative tests
    
---

    feat(csi/controller): add rwx block enablement via volume context
    
    A new SC parameter "rwxBlock: true", may be used to enable rwx block.
    
---

    feat(kubevirt/live-migrate): support rwx for block volumes
    
    Allows multi-node writer access modes but only if the access type
    is for Block.
    This should only be enabled for applications which can handle
    having the same block device connected by multiple initiator nodes!
    
    todo: add kube-virt flag to the storage class parameters which will
    enable this feature (or global CLI flag)
    
    Allows multiple frontend nodes for the same volume target.
    Each node's nqn is added to the allowed hosts lists in the pstor and
    in the volume's nvmf target.
    
    When unpublishing, if all frontend nodes would be removed, then the
    nexus is also destroyed.
    Note that eventually when we switch to offline rebuilding, this will
    likely have to change. Nevertheless, this is equivalent to the current
    behaviour where the unpublish destroys the nexus as well.
    
---

    test(bdd): expect no error on unpublish from different node

### DIFF
--- a/control-plane/agents/src/bin/core/nexus/specs.rs
+++ b/control-plane/agents/src/bin/core/nexus/specs.rs
@@ -51,11 +51,17 @@ impl SpecOperationsHelper for NexusSpec {
         op: Self::UpdateOp,
     ) -> Result<(), SvcError> {
         match &op {
-            NexusOperation::Share(_, _) if state.share.shared() => Err(SvcError::AlreadyShared {
-                kind: ResourceKind::Nexus,
-                id: self.uuid_str(),
-                share: state.share.to_string(),
-            }),
+            NexusOperation::Share(_, a)
+                if state.share.shared()
+                    && &state.allowed_hosts == a
+                    && &self.allowed_hosts == a =>
+            {
+                Err(SvcError::AlreadyShared {
+                    kind: ResourceKind::Nexus,
+                    id: self.uuid_str(),
+                    share: state.share.to_string(),
+                })
+            }
             NexusOperation::Share(_, _) => Ok(()),
             NexusOperation::Unshare if !state.share.shared() => Err(SvcError::NotShared {
                 kind: ResourceKind::Nexus,
--- a/control-plane/agents/src/bin/core/volume/operations.rs
+++ b/control-plane/agents/src/bin/core/volume/operations.rs
@@ -32,16 +32,20 @@ use stor_port::{
         store::{
             nexus_persistence::NexusInfoKey,
             replica::ReplicaSpec,
-            volume::{PublishOperation, RepublishOperation, VolumeOperation, VolumeSpec},
+            volume::{
+                PublishOperation, RepublishOperation, UnpublishOperation, VolumeOperation,
+                VolumeSpec,
+            },
         },
         transport::{
             CreateReplica, CreateVolume, DestroyNexus, DestroyReplica, DestroyShutdownTargets,
             DestroyVolume, NodeTopology, Protocol, PublishVolume, Replica, ReplicaId,
             ReplicaOwners, RepublishVolume, ResizeVolume, SetVolumeProperty, SetVolumeReplica,
             ShareNexus, ShareVolume, ShutdownNexus, UnpublishVolume, UnshareNexus, UnshareVolume,
-            Volume,
+            Volume, VolumeShareProtocol,
         },
     },
+    HostAccessControl,
 };
 
 use itertools::Itertools;
@@ -321,6 +325,46 @@ impl ResourcePublishing for OperationGuardArc<VolumeSpec> {
         request: &Self::Publish,
     ) -> Result<Self::PublishOutput, SvcError> {
         let state = registry.volume_state(&request.uuid).await?;
+
+        if let Some(target_cfg) = self.as_ref().target_cfg() {
+            let target = target_cfg.target();
+
+            let host_acl =
+                registry.host_acl_nodename(HostAccessControl::Nexuses, &request.frontend_nodes);
+            if target_cfg.frontend().needs_update(&host_acl) {
+                let mut nexus = registry.specs().nexus(target.nexus()).await?;
+                let nexus_state = registry.nexus(target.nexus()).await?;
+
+                let mut target_cfg = target_cfg.clone();
+                target_cfg.frontend_mut().add_acls(host_acl);
+
+                let operation = VolumeOperation::Publish(PublishOperation::new(
+                    target_cfg.clone(),
+                    request.publish_context.clone(),
+                ));
+                let spec_clone = self.start_update(registry, &state, operation).await?;
+
+                let result = nexus
+                    .share(
+                        registry,
+                        &ShareNexus::new(
+                            &nexus_state,
+                            VolumeShareProtocol::Nvmf,
+                            target_cfg.frontend().node_nqns(),
+                        ),
+                    )
+                    .await;
+
+                self.complete_update(registry, result, spec_clone).await?;
+
+                let volume = registry.volume(&request.uuid).await?;
+                registry
+                    .notify_if_degraded(&volume, PollTriggerEvent::VolumeDegraded)
+                    .await;
+                return Ok(volume);
+            }
+        }
+
         let nexus_node = self
             .next_target_node(registry, request, &state, false)
             .await?;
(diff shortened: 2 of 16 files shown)


control-plane/agents/src/bin/core/nexus/specs.rs
```diff
@@ -51,11 +51,11 @@ impl SpecOperationsHelper for NexusSpec {
...
-            NexusOperation::Share(_, _) if state.share.shared() => Err(SvcError::AlreadyShared {
-                kind: ResourceKind::Nexus,
-                id: self.uuid_str(),
-                share: state.share.to_string(),
-            }),
+            // NexusOperation::Share(_, _) if state.share.shared() => Err(SvcError::AlreadyShared {
+            //     kind: ResourceKind::Nexus,
+            //     id: self.uuid_str(),
+            //     share: state.share.to_string(),
+            // }),
```
Abhinandan-Purkait: Remove this?

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -321,6 +325,48 @@ impl ResourcePublishing for OperationGuardArc<VolumeSpec> {
...
+
+                self.complete_update(registry, result, spec_clone).await?;
+
+                let volume = registry.volume(&request.uuid).await?;
+                registry
+                    .notify_if_degraded(&volume, PollTriggerEvent::VolumeDegraded)
+                    .await;
+                return Ok(volume);
+            } else {
+                // error out?
```
Abhinandan-Purkait: Don't need to?

control-plane/csi-driver/src/bin/controller/controller.rs
```diff
@@ -59,22 +59,55 @@ impl CsiControllerSvc {
...
+    };
+    match access_mode {
+        volume_capability::access_mode::Mode::SingleNodeWriter => Ok(access_mode),
+        volume_capability::access_mode::Mode::MultiNodeMultiWriter => {
+            if matches!(access_type, AccessType::Block(_)) {
+                if rwx_block == &Some(true) {
+                    Ok(access_mode)
+                } else {
+                    Err(Status::invalid_argument(
+                        "MultiNodeMultiWriter but RWX Block is disabled".to_string(),
```
Abhinandan-Purkait:
```suggestion
                        "MultiNodeMultiWriter for RWX Block is disabled".to_string(),
```
?

control-plane/csi-driver/src/bin/controller/controller.rs
```diff
@@ -661,18 +718,16 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
         let volume = match RestApiClient::get_client().get_volume(&volume_uuid).await {
             Ok(volume) => volume,
             Err(ApiClientError::ResourceNotExists { .. }) => {
-                debug!("Volume {} does not exist, not unpublishing", args.volume_id);
+                // todo: shouldn't this be a grpc NOT_FOUND?
+                debug!("Volume does not exist, not unpublishing");
```
Abhinandan-Purkait, line 722: You mean return an error with 404?

control-plane/csi-driver/src/bin/controller/controller.rs
```diff
@@ -708,23 +763,22 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
         let _volume = RestApiClient::get_client()
             .get_volume(&volume_uuid)
             .await
+            // todo: this seems wrong...
```
Abhinandan-Purkait, line 766: Indeed, should have been Unavailable rather than unimplemented, or maybe aborted with xyz reason.

control-plane/csi-driver/src/bin/controller/controller.rs
```diff
@@ -708,23 +763,22 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
...
-                    }
-                }
-                false
+                let cap_mode = cap.access_mode.as_ref();
+                let access_mode = cap_mode.map(|m| m.mode()).unwrap_or_default();
+                supported_caps.contains(&access_mode)
             })
             .collect();
 
+        // todo: this seems wrong? See https://github.com/container-storage-interface/spec/blob/e981e2a057ca10f4a7f81289c97a4e829fd69152/spec.md?plain=1#L1502
```
Abhinandan-Purkait: Should have been none, rather than empty defaults?

control-plane/csi-driver/src/bin/controller/controller.rs
```diff
@@ -771,6 +825,7 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
                 let volume = rpc::csi::Volume {
                     volume_id: v.spec.uuid.to_string(),
                     capacity_bytes: v.spec.size as i64,
+                    // todo: these 2 seem wrong?
```
Abhinandan-Purkait, line 819: Probably because we don't store the context? Or maybe we could have just generated it, I believe it's only the device uri that goes here?

Ahh we can definitely update the content source now that we have implemented the snaps.

control-plane/csi-driver/src/bin/controller/controller.rs
```diff
@@ -799,8 +854,10 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
         let _ = csi_driver::trace::CsiRequest::new_trace("Get Node(s) Capacity");
 
         // Check capabilities.
-        check_volume_capabilities(&args.volume_capabilities)?;
+        let context = CreateParams::try_from(&args.parameters)?;
+        check_volume_capabilities(&args.volume_capabilities, context.rwx_block())?;
 
+        // todo: this seems to be conflating app node vs storage node?
```
Abhinandan-Purkait, line 851: I am a bit confused here, what exactly are we trying to do here? IIRC if we set allowedTopologies on the storage class this get's populated, and it's expected to be a subset of the csi-nodes only.

So if someone does that and that csi-node happens to be a non mayastor node we return an error saying node not found?

This probably needs to be re-evaluated, the accessible topology has been mixed with data placement topology here.

control-plane/csi-driver/src/context.rs
```diff
@@ -511,6 +518,10 @@ pub struct CreateParams {
     pool_cluster_size: Option<u64>,
 }
 impl CreateParams {
+    /// Get the `Parameters::RwxBlock` value.
+    pub fn rwx_block(&self) -> &Option<bool> {
```
Abhinandan-Purkait, line 522: Just wondering do we need an explicit knob as well? What happens if someone says ReadWriteMany and sets rwxBlock to false?

control-plane/agents/src/bin/core/volume/operations.rs
tiagolobocastro: This is in case we don't need an update, wasn't sure if we should error out here, wdyt?

control-plane/csi-driver/src/bin/controller/controller.rs
tiagolobocastro: hmm, but this is to say "RWX Block" is disabled? maybe

```suggestion
                        "MultiNodeMultiWriter requested but RWX Block is disabled".to_string(),
```

control-plane/csi-driver/src/bin/controller/controller.rs
tiagolobocastro: Well this is the CSI api here, so there's no 404, only gRPC status code of `NOT_FOUND`.

control-plane/csi-driver/src/bin/controller/controller.rs
tiagolobocastro: Yes something like that indeed... I think I'll leave as is for now, maybe can fix this on another PR

control-plane/csi-driver/src/bin/controller/controller.rs
tiagolobocastro: As per the docs: `This RPC call SHALL return confirmed only if all the volume capabilities specified in the request are supported (see caveat below).`
But here we return ok as long as one capability is ok, which seems to violate the above?

control-plane/csi-driver/src/bin/controller/controller.rs
tiagolobocastro: For the context yes, we'd have to start storing it, and no it's not the uri, that's the publish context.
Probably we'd need clarification here...

For content source, yes we should be able to set it correctly here

control-plane/csi-driver/src/bin/controller/controller.rs
tiagolobocastro: I don't know what we are trying to do here...

control-plane/csi-driver/src/context.rs
tiagolobocastro: Then we'll fail the validation check on `check_volume_capability`.
This is needed to ensure users enable this explicitly as this should only be used with applications which handle this properly.

control-plane/agents/src/bin/core/volume/operations.rs
Abhinandan-Purkait: I think we should not? Just make add a log and move on?

control-plane/csi-driver/src/bin/controller/controller.rs
Abhinandan-Purkait: Sounds good.

control-plane/csi-driver/src/bin/controller/controller.rs
Abhinandan-Purkait: Yes, I meant that only.

control-plane/csi-driver/src/bin/controller/controller.rs
Abhinandan-Purkait: Yes I think so.

control-plane/csi-driver/src/bin/controller/controller.rs
Abhinandan-Purkait: Yeah maybe you can just add the content source. I am always a bit skeptical about changing any error handling in csi, changes the behaviour :-;

control-plane/csi-driver/src/bin/controller/controller.rs
Abhinandan-Purkait: Maybe leave it for now?

control-plane/csi-driver/src/context.rs
Abhinandan-Purkait: Makes sense.

Abhinandan-Purkait (approved):
LGTM

control-plane/agents/src/bin/core/nexus/specs.rs
tiagolobocastro: I rewrote it to take into account the allowed hosts

control-plane/agents/src/bin/core/volume/operations.rs
tiagolobocastro: I've removed it, since we weren't checking it prior to  these changes anyway

control-plane/csi-driver/src/bin/controller/controller.rs
tiagolobocastro: I think we should not not implement this method

control-plane/csi-driver/src/bin/controller/controller.rs
tiagolobocastro: I'll leave for another PR

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

PR #1056, opened 2026-01-12
RWX: fix unpublish/republish loop

tiagolobocastro:
fix(rwx/stage): connect to volume using node's own hostnqn
    
    Since we now publish the volume for multiple nodes, the share url contains
    many hostnqn's.
    On the node stage, we must connect using the node's hostnqn!
    
    It would be nice to start connecting using the target configuration, instead
    of the share uri, but that would require more changes and thought.
    So for now we just clip the nqn's which belong to other nodes from the share
    uri.
    The downside if this change is that we can no longer use custom nqn, but I
    think this is ok, because we're already using the nqn for the ha logic, so
    this is already expected.
    
---

    fix(rwx): reconcile due to unsorted host nqn list
    
    Always ensure the hostnqn lists are sorted, otherwise we may
    run reconcile to fix the hostnqn when it's not needed.
    
---

    fix: validate volume access mode on publish
    
    Publishing the volume on multiple nodes should only be allowed
    if RWX specified via the request.
    
---

    refactor: include access mode in the volume publish operation
    
    This will allow the core agent to evaluate and return appropriate errors.
    This is now required because we've modified the publish to support many
    initiators but we did not account for the error cases and idempotency for
    when rwx is on or off.
    
---

    chore: update terraform to K8s v1.33
    
    This is used out of band from CI, only for local isolated testing.
    
---

    fix(rwx): keep publish context until full unpublish
    
    Now that we have multiple publishes, we have to keep the context until
    the last unpublish completes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1048,6 +1048,7 @@ dependencies = [
  "nvmeadm",
  "once_cell",
  "parse-size",
+ "percent-encoding",
  "prost",
  "prost-types",
  "regex",
--- a/control-plane/agents/src/bin/core/controller/io_engine/v1/translation.rs
+++ b/control-plane/agents/src/bin/core/controller/io_engine/v1/translation.rs
@@ -22,6 +22,7 @@ use stor_port::{
     IntoOption,
 };
 
+use itertools::Itertools;
 use std::{convert::TryFrom, time::UNIX_EPOCH};
 
 /// Trait for converting agent messages to io-engine messages.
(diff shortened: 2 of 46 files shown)


Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

PR #1059, opened 2026-02-17
Add support for offline pool obj delete

niladrih:


niladrih:
bors try

niladrih:
bors try

niladrih:
bors try

niladrih:
bors try

niladrih:
bors try

control-plane/agents/src/bin/core/pool/pool_operations.rs
```diff
@@ -286,3 +279,358 @@ impl ResourceCordon for OperationGuardArc<PoolSpec> {
...
+            let key = replica.key();
+            if let Err(e) = registry.delete_kv(&key.key()).await {
+                tracing::warn!(
+                    replica.uuid = %replica.uuid,
+                    error = %e,
+                    "Failed to delete replica spec from store during pool purge"
+                );
+            }
+            // Remove from in-memory specs
+            registry.specs().remove_replica(&replica.uuid);
```
Abhinandan-Purkait: So if delete from kv fails we still delete from memory? But on restart wouldn't this come back again in memory and at that point this spec just becomes an orphan but the pool is gone? I feel this would make it leak replica specs. cc @tiagolobocastro wdyt?

control-plane/agents/src/bin/core/pool/pool_operations.rs
```diff
@@ -286,3 +279,358 @@ impl ResourceCordon for OperationGuardArc<PoolSpec> {
...
+
+            let total_replicas = volume_replicas.len() as u32;
+            let replicas_on_pool = volume_replicas
+                .iter()
+                .filter(|r| r.pool_name() == pool_id)
+                .count() as u32;
+
+            let healthy_before = total_replicas;
+            let healthy_after = total_replicas.saturating_sub(replicas_on_pool);
+
```
Abhinandan-Purkait: I am a bit confused on what is happening here? What do you mean by healthy_before if you assign all total replicas to it? All replicas might not be healthy?

control-plane/agents/src/bin/core/pool/pool_operations.rs
```diff
@@ -286,3 +279,358 @@ impl ResourceCordon for OperationGuardArc<PoolSpec> {
...
+            // Node is online - check pool state via ctrl_pool_state
+            if let Ok(ctrl_state) = registry.ctrl_pool_state(pool_id).await {
+                if ctrl_state.status != PoolStatus::Unknown {
+                    return Err(SvcError::PoolStateNotUnknown {
+                        pool_id: pool_id.clone(),
+                        state: format!("{:?}", ctrl_state.status),
+                    });
+                }
+            }
+            // If ctrl_pool_state fails, pool state is unknown - allow purge
```
Abhinandan-Purkait, line 483: Isn't this wrong? Why are we allowing purge if pool state is not there but the node is online. This seems risky. If the pool is not getting listed for whatever reason we shouldn't just purge it like that. @tiagolobocastro wdyt? For example a simple import faliure or something.

control-plane/rest/openapi-specs/v0_api_spec.yaml
```diff
@@ -843,9 +843,19 @@ paths:
...
+              $ref: '#/components/schemas/DeletePoolBody'
       responses:
-        '204':
-          description: OK
+        '200':
+          description: Pool deleted successfully
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/PoolDeleteResult'
```
Abhinandan-Purkait, line 859: Yeah we shouldn't do that. But having said that I am not sure what is the way to do this. We cannot return a body with 204 either way right?

control-plane/agents/src/bin/core/pool/pool_operations.rs
```diff
@@ -286,3 +279,358 @@ impl ResourceCordon for OperationGuardArc<PoolSpec> {
...
+            }))
+        }
+    }
+
+    /// Analyze snapshot loss impact for replica snapshots being deleted.
+    fn analyze_snapshot_loss(
+        &self,
+        registry: &Registry,
+        pool_id: &PoolId,
+        _replica_snapshots: &[ReplicaSnapshot],
```
Abhinandan-Purkait: Why do we need this param? Doesn't seem to be used.

control-plane/plugin/src/resources/pool.rs
```diff
@@ -243,6 +243,261 @@ impl GetWithArgs for Pool {
...
+
+#[derive(Debug, Clone, clap::Args)]
+pub struct PreparePurgeArgs {
+    /// Id of the pool to prepare for purge.
+    pool_id: PoolId,
+}
+
+impl PreparePurgeArgs {
+    /// Show what would happen if this pool is purged.
+    pub async fn show_impact(&self, output: &OutputFormat) -> PluginResult {
```
Abhinandan-Purkait: Where are we calling this? I don't see a caller

control-plane/stor-port/src/types/v0/transport/pool.rs
```diff
@@ -440,6 +495,141 @@ impl From<CreatePool> for DestroyPool {
...
+            data_loss: src.data_loss.map(|dl| models::DataLossInfo {
+                volumes: dl
+                    .volumes
+                    .into_iter()
+                    .map(|v| models::VolumeLossDetail {
+                        volume_id: v.volume_id.to_string(),
+                        replicas_before: v.replicas_before as i32,
+                        healthy_before: v.healthy_before as i32,
+                        lost_on_pool: v.lost_on_pool as i32,
+                        healthy_after: v.healthy_after as i32,
```
Abhinandan-Purkait: Hmm, can't this be u32? Count can't be negative right? We would not need this conversion?

Abhinandan-Purkait (commented):
Overall looks good. Can we add a test to test the reconciler bits which cleans up later like the volume snaphot one I guess.

control-plane/agents/src/bin/core/tests/pool/purge.rs
```diff
@@ -0,0 +1,382 @@
...
+    let result = pool_client.destroy(&destroy, None).await;
+    assert!(result.is_err());
+    let err = result.unwrap_err();
+    assert!(err.to_string().contains("PoolStateNotUnknown"));
+}
+
+/// Test: Purge rejected when pool is not cordoned
+#[tokio::test]
+async fn purge_rejected_pool_not_cordoned() {
+    let cluster = ClusterBuilder::builder()
```
tiagolobocastro: Is it possible to combine some of these into regular functions with less test functions? Adding more test functions increases the setup/teardown duration.

control-plane/agents/src/bin/core/tests/pool/purge.rs
```diff
@@ -0,0 +1,382 @@
...
+        .into_inner();
+    let pool = pools.first().unwrap();
+
+    // Pool is online - purge should be rejected
+    let destroy = DestroyPool::purge(pool.node(), pool.id().clone()).with_confirm(true);
+
+    let result = pool_client.destroy(&destroy, None).await;
+    assert!(result.is_err());
+    let err = result.unwrap_err();
+    assert!(err.to_string().contains("PoolStateNotUnknown"));
```
tiagolobocastro: Compare the actual type

control-plane/agents/src/bin/core/tests/pool/purge.rs
```diff
@@ -0,0 +1,382 @@
...
+        .with_pools(1)
+        .build()
+        .await
+        .unwrap();
+
+    // Stop the io-engine to make pool state Unknown
+    cluster.composer().stop("io-engine-1").await.unwrap();
+
+    // Wait for node to go offline
+    tokio::time::sleep(std::time::Duration::from_secs(5)).await;
```
tiagolobocastro: please don't hardcode sleeps unless last resort, prefer to use retry loops with timeout
we already have functions for this in the Cluster type IIRC, or otherwise other examples spread around

control-plane/agents/src/bin/core/tests/pool/purge.rs
```diff
@@ -0,0 +1,382 @@
+use deployer_cluster::ClusterBuilder;
+use grpc::operations::pool::traits::{PoolCordonRequest, PoolOperations};
+use stor_port::types::v0::openapi::models::{CreateVolumeBody, VolumePolicy};
+use stor_port::types::v0::transport::{DestroyPool, Filter, VolumeId};
+
+/// Test: Purge rejected when pool state is Online
+#[tokio::test]
+async fn purge_rejected_pool_state_online() {
+    let cluster = ClusterBuilder::builder()
+        .with_rest(true)
```
tiagolobocastro: you're not using rest though?

control-plane/agents/src/bin/core/tests/pool/purge.rs
```diff
@@ -0,0 +1,382 @@
...
+        .with_agents(vec!["core"])
+        .with_io_engines(1)
+        .with_pools(1)
+        .build()
```
tiagolobocastro, line 28: if you rely on the reconciler or cache, please override the default values to x miliseconds

control-plane/agents/src/common/errors.rs
```diff
@@ -448,6 +449,56 @@ pub enum SvcError {
     DiskNotExtended { name: PoolId },
     #[snafu(display("Pool {name} underlying disk rescan failed"))]
     DiskRescanFailed { name: PoolId },
+    #[snafu(display(
+        "Cannot purge pool '{}': pool state is '{}', purge only allowed when state is Unknown",
+        pool_id,
+        state
```
tiagolobocastro: Use this where possible

```suggestion
        "Cannot purge pool '{pool_id}': pool state is '{state}', purge only allowed when state is Unknown"
```

control-plane/grpc/proto/v1/pool/pool.proto
```diff
@@ -163,6 +163,14 @@ message DestroyPoolRequest {
   string pool_id = 1;
   // node id of the pool
   string node_id = 2;
+  // Purge mode - delete control plane state without contacting io-engine
+  bool purge = 3;
```
tiagolobocastro: mark them as optional for clarity since it's a modification of the api

control-plane/grpc/proto/v1/pool/pool.proto
```diff
@@ -187,9 +195,50 @@ message ExpandPoolReply {
...
+
+// Information about volumes that will lose all healthy replicas
+message DataLossInfo {
+  repeated VolumeLossDetail volumes = 1;
+}
+
+// Details about data loss for a specific volume
+message VolumeLossDetail {
+  string volume_id = 1;
+  uint32 replicas_before = 2;
```
tiagolobocastro, line 240: what does before and after mean?
please document this

control-plane/grpc/src/operations/pool/traits.rs
```diff
@@ -406,6 +407,14 @@ pub trait DestroyPoolInfo: Sync + Send + std::fmt::Debug {
     fn pool_id(&self) -> PoolId;
     /// Id of the IoEngine instance
     fn node_id(&self) -> NodeId;
+    /// Whether to purge (delete specs without contacting io-engine)
```
tiagolobocastro: End with `.`
The existing ones missing probably predate this practice

```suggestion
    /// Whether to purge (delete specs without contacting io-engine).
```

control-plane/plugin/src/resources/pool.rs
```diff
@@ -243,6 +243,261 @@ impl GetWithArgs for Pool {
...
+            Err(e) => {
+                // Handle not found with ignore flag
+                if ignore_not_found && e.status() == Some(StatusCode::NOT_FOUND) {
+                    return Ok(());
+                }
+
+                // Handle specific error cases with helpful messages
+                let error_str = format!("{e:?}");
+
+                if error_str.contains("PoolStateNotUnknown") {
```
tiagolobocastro: Use the type system

control-plane/plugin/src/resources/pool.rs
```diff
@@ -243,6 +243,261 @@ impl GetWithArgs for Pool {
...
+    }
+}
+
+impl CreateRow for models::PoolDeleteResult {
+    fn row(&self) -> Row {
+        let data_loss = self
+            .data_loss
+            .as_ref()
+            .map(|dl| format!("{} volume(s)", dl.volumes.len()))
+            .unwrap_or_else(|| "none".to_string());
```
tiagolobocastro: we already have helper for optional

control-plane/plugin/src/resources/pool.rs
```diff
@@ -243,6 +243,261 @@ impl GetWithArgs for Pool {
...
+        ]
+    }
+}
+
+impl CreateRow for PoolPurgeImpact {
+    fn row(&self) -> Row {
+        row![
+            self.pool_id,
+            self.state,
+            if self.is_cordoned { "Yes" } else { "No" },
```
tiagolobocastro: We use true/false in existing things IIRC

```suggestion
            self.is_cordoned.to_string(),
```

control-plane/plugin/src/resources/pool.rs
```diff
@@ -243,6 +243,261 @@ impl GetWithArgs for Pool {
...
+
+        // Collect affected volumes
+        let mut volumes: std::collections::HashSet<String> = std::collections::HashSet::new();
+        for replica in &replicas {
+            if let Some(vol_id) = &replica.owners.volume {
+                volumes.insert(vol_id.to_string());
+            }
+        }
+
+        let ready_for_purge = state_str == "Unknown" && cordon_ok;
```
tiagolobocastro: Use the type system please

control-plane/plugin/src/resources/pool.rs
```diff
@@ -243,6 +243,261 @@ impl GetWithArgs for Pool {
...
+            .await
+        {
+            Ok(p) => p.into_body(),
+            Err(e) => {
+                return Err(Error::GetPoolError {
+                    id: self.pool_id.to_string(),
+                    source: e,
+                });
+            }
+        };
```
tiagolobocastro: nit:

```suggestion
        {
            Ok(p) => Ok(p.into_body()),
            Err(source) => Err(Error::GetPoolError {
                    id: self.pool_id.to_string(),
                    source,
                }),
        }?;
```

control-plane/plugin/src/resources/pool.rs
```diff
@@ -243,6 +243,261 @@ impl GetWithArgs for Pool {
...
+        let state_str = pool
+            .state
+            .as_ref()
+            .map(|s| format!("{:?}", s.status))
+            .unwrap_or_else(|| "Unknown".to_string());
+
+        // Check cordon status
+        let is_cordoned = pool
+            .spec
+            .as_ref()
```
tiagolobocastro: might be able to reduce some of the lines by "caching" the spec?

control-plane/plugin/src/resources/pool.rs
```diff
@@ -243,6 +243,261 @@ impl GetWithArgs for Pool {
...
+            .and_then(|s| s.cordon_drain.as_ref())
+            .map(|cd| match cd {
+                PoolCordonDrain::cordoned(state) => state.replicas && state.snapshots,
+            })
+            .unwrap_or(false);
+
+        // Get replica specs from Specs API
+        let specs = RestClient::client()
+            .specs_api()
```
tiagolobocastro: hmm, is the specs api even maintained?
Also this could be an incredibly large number of things, this fetches everything! There's no paging either... 
CC @Abhinandan-Purkait

control-plane/rest/openapi-specs/v0_api_spec.yaml
tiagolobocastro: We'd need to pass here different responses, one for 204 and another for 200 ?

control-plane/rest/openapi-specs/v0_api_spec.yaml
```diff
@@ -2603,6 +2623,131 @@ components:
...
+
+    PoolDeleteResult:
+      type: object
+      required:
+        - pool_id
+      description: Result of a pool delete operation
+      properties:
+        pool_id:
+          type: string
+          description: The deleted pool ID
```
```suggestion
          $ref: '#/components/schemas/PoolId'
```

control-plane/rest/openapi-specs/v0_api_spec.yaml
```diff
@@ -2603,6 +2623,131 @@ components:
...
+          default: false
+          description: |
+            Purge pool without contacting io-engine.
+            Use when pool's node is offline or pool state is Unknown.
+            Requires pool to be cordoned with replicas and snapshots blocked.
+        confirm:
+          type: boolean
+          default: false
+          description: Confirm deletion when pool has replicas
+        confirm_data_loss:
```
tiagolobocastro: use camel case on the openapi

control-plane/rest/openapi-specs/v0_api_spec.yaml
```diff
@@ -2603,6 +2623,131 @@ components:
...
+        confirm_snapshot_loss:
+          type: boolean
+          default: false
+          description: |
+            Confirm snapshot loss. Required when purge would cause
+            snapshots to lose their last replica snapshot.
+
+    PoolDeleteResult:
+      type: object
+      required:
```
tiagolobocastro: move the required to the end, that's how we generally have it here

control-plane/rest/openapi-specs/v0_api_spec.yaml
```diff
@@ -2603,6 +2623,131 @@ components:
...
+        - lost_on_pool
+        - healthy_after
+      description: Details about a volume affected by pool deletion
+      properties:
+        volume_id:
+          type: string
+          description: The affected volume ID
+        replicas_before:
+          type: integer
+          format: int32
```
tiagolobocastro, line 2729: Make this unsigned like so:

```suggestion
          format: int32
          minimum: 0
```

niladrih:
bors try

niladrih:
bors try

niladrih:
bors try

control-plane/rest/openapi-specs/v0_api_spec.yaml
niladrih: That's a good idea.

control-plane/rest/openapi-specs/v0_api_spec.yaml
niladrih: Okay, done.

control-plane/rest/openapi-specs/v0_api_spec.yaml
niladrih: Done.

control-plane/rest/openapi-specs/v0_api_spec.yaml
niladrih: Right.

control-plane/agents/src/bin/core/tests/pool/purge.rs
niladrih: I've put the exact numbers that are on the mod.rs test file (on the same directory). Seems to work reliably so far.

control-plane/agents/src/common/errors.rs
niladrih: Okay.

control-plane/grpc/proto/v1/pool/pool.proto
niladrih: It means the no. of replicas based on observations from before the pool delete.
I've added doc comments to these.

control-plane/plugin/src/resources/pool.rs
niladrih: I've added the types to the assert. However, the members in the error types has limitations to what exact error is being displayed. We can do one of two things
1. Not test all that, i. e. leave it at just the higher level error variant, without going and looking for the exact error.
2. Add a new member, viz. `reason` on the error types, that lets us document the reason why, say, PreconditionFailed happened.

control-plane/grpc/proto/v1/pool/pool.proto
niladrih: Made this optional, this doesn't change the generated code, but gives it clarity towards the API stability, as you've already mentioned.

control-plane/agents/src/bin/core/tests/pool/purge.rs
niladrih: Yep.

control-plane/agents/src/bin/core/tests/pool/purge.rs
niladrih: Yep.

control-plane/agents/src/bin/core/tests/pool/purge.rs
niladrih: Yes, but... the members in the error types have limitations to what exact error is being displayed. We can do one of two things
1. Not test all that, i. e. leave it at just the higher level error variant, without going and looking for the exact error.
2. Add a new member, viz. reason on the error types, that lets us document the reason why, say, PreconditionFailed happened.

control-plane/agents/src/bin/core/tests/pool/purge.rs
niladrih: Yes, pool purge is 1 test now. All on one cluster. There are 2 different pools to test regular delete API integrity and actual purge API.

control-plane/stor-port/src/types/v0/transport/pool.rs
niladrih: Yes, I've changed it all to u32.

control-plane/plugin/src/resources/pool.rs
niladrih: That's a miss. I've added a `--show-impact` flag to use this function. This was the original intention, missed the actual flag codepath implementation.

control-plane/agents/src/bin/core/pool/pool_operations.rs
niladrih: Removed.

niladrih: The section has been refactored heavily, doesn't need similar args now.

control-plane/rest/openapi-specs/v0_api_spec.yaml
niladrih: I've added both 200 and 204 to the responses map. 200 compulsorily comes with PoolDeleteResult body. 204 is empty, as it was earlier. This is enabled by changing the openapi v3 template to generate an enum against this exact openapi spec configuration.

control-plane/agents/src/bin/core/pool/pool_operations.rs
niladrih: I'd require more input on this. Isn't pool purge also meant to support online node, but bad disk, etc. cases?
Tagging @tiagolobocastro.

control-plane/agents/src/bin/core/pool/pool_operations.rs
niladrih: This is wrong. I've changed this logic to use info from the nexus and node status to judge if a replica is healthy. Please let me know if the existing method is sane and correct.

control-plane/agents/src/bin/core/pool/pool_operations.rs
niladrih: More input on why is flow is undesirable would be appreciated.
Also tagging @tiagolobocastro.

niladrih:
@tiagolobocastro @Abhinandan-Purkait -- I've made most of the changes requested in the comments. I require clarity on some of the changes though.

bors try

control-plane/plugin/src/resources/pool.rs
niladrih: Done.

control-plane/plugin/src/resources/pool.rs
niladrih: Yes, that's right. Done.

control-plane/plugin/src/resources/pool.rs
niladrih: Changed this to use replicas, volumes APIs instead.

niladrih:
bors try

control-plane/agents/src/bin/core/pool/pool_operations.rs
tiagolobocastro: Yeah for sure we can't ignore this error
Also not sure if this is the right way to do it, shouldn't this be done by replica.destroy?

control-plane/agents/src/bin/core/pool/pool_operations.rs
tiagolobocastro: hmm you don't need that, you can use the volume, which has the health information with it already

control-plane/agents/src/bin/core/pool/pool_operations.rs
tiagolobocastro: Yes we do want to purge "bad" diskpools from online nodes as well

But if the pool is still being reported, it may be for 3 reasons:
- [ ] the pools is still good -> we don't want to purge in this case
- [x] the pool is getting I/O errors -> we want to purge
- [x] the pool is having I/O stalls -> we want to purge

For I/O errors, we should be able to issue destroy against the pool, and it will complete fine
For I/O stalls, we're out of luck... I think we'll have to fail and tell user the pool is stalled, please restart io-engine and retry

Btw, we should cordon the pool at this point or earlier to avoid import retries should io-engine restart.

control-plane/agents/src/bin/core/tests/pool/purge.rs
tiagolobocastro: humm, we should be able to get an actual error from the ReplyError?

control-plane/agents/src/bin/core/pool/pool_operations.rs
niladrih: For the replicas on the pool, I'm checking volume health.

For the snapshots, I am just checking to see if they exist on the pool or not. Because, the pool is unknown, so it's inaccessible.

control-plane/agents/src/bin/core/pool/pool_operations.rs
niladrih: Cordon check is the very next check in sequence at the caller side.

control-plane/agents/src/bin/core/pool/pool_operations.rs
niladrih: Fine, I'm trying .destroy(), and resorting to spec delete if it fails
```rust
            if let Err(error) = replica.destroy(registry, &destroy_request).await {
                tracing::warn!(
                    replica.uuid = %replica_spec.uuid,
                    %error,
                    "Failed to destroy replica via io-engine, removing spec only"
                );
                replica.destroy_for_purge(registry).await?;
            }
```

control-plane/agents/src/bin/core/tests/pool/purge.rs
niladrih: I completely overlooked the ReplyErrorKind type. Fixed this.

control-plane/plugin/src/resources/pool.rs
niladrih: I had missed the ReplyErrorKind type. Using that to compare against error types, instead of strings.

control-plane/agents/src/bin/core/pool/pool_operations.rs
Abhinandan-Purkait: I am not sure if this solves it. But I don't have a better way in mind.

control-plane/plugin/src/resources/pool.rs
Abhinandan-Purkait: It still seems to be the same, so it's not resolved?

control-plane/plugin/src/resources/pool.rs
Abhinandan-Purkait: Sorry I don't see the change?

Abhinandan-Purkait (approved):
Mostly LGTM. Please address the unresolved comments.

control-plane/agents/src/common/errors.rs
```diff
@@ -448,6 +449,46 @@ pub enum SvcError {
     DiskNotExtended { name: PoolId },
     #[snafu(display("Pool {name} underlying disk rescan failed"))]
     DiskRescanFailed { name: PoolId },
+    #[snafu(display(
+        "Cannot purge pool '{pool_id}': pool state is '{state}', purge only allowed when state is Unknown",
```
tiagolobocastro: nit: Do we need all these `'` around `'` ?

```suggestion
        "Cannot purge pool {pool_id}: pool state is {state}, purge only allowed when state is Unknown",
```

control-plane/grpc/proto/v1/pool/pool.proto
tiagolobocastro: Don't see the comment here?

control-plane/plugin/src/resources/mod.rs
```diff
@@ -249,6 +249,8 @@ pub struct DeleteArgs {
 /// The type of resources which support the delete operation.
 #[derive(clap::Subcommand, Debug)]
 pub enum DeleteResources {
+    /// Deletes the specified pool resource.
+    Pool(pool::DeletePoolArgs),
```
tiagolobocastro, line 253: good to see this implemented finally :)

control-plane/plugin/src/resources/pool.rs
```diff
@@ -335,6 +337,255 @@ impl GetWithArgs for Pool {
...
+
+    /// Show what would happen if this pool is purged, without actually deleting.
+    /// Displays pool state, cordon status, replica count, and affected volumes.
+    #[arg(long)]
+    pub show_impact: bool,
+
+    /// Confirm deletion when pool has replicas.
+    /// Required when using --purge on a pool with replicas.
+    #[arg(long)]
+    confirm: bool,
```
tiagolobocastro: Can't remember if this was previously discussed, just reading it now, would `accept` be more suited?

```suggestion
    accept: bool,
```

control-plane/plugin/src/resources/pool.rs
tiagolobocastro: Yeah still here

control-plane/plugin/src/resources/pool.rs
```diff
@@ -335,6 +337,255 @@ impl GetWithArgs for Pool {
...
+                // Handle not found with ignore flag
+                if ignore_not_found && source.status() == Some(StatusCode::NOT_FOUND) {
+                    return Ok(());
+                }
+
+                // Handle specific error cases with helpful messages
+                let error_str = format!("{source:?}");
+
+                if error_str.contains("PoolStateNotUnknown") {
+                    eprintln!("Error: Cannot purge pool - pool state is known.");
```
tiagolobocastro: Should this not be part of the error instead?

control-plane/plugin/src/resources/pool.rs
tiagolobocastro: Yeah I still see it...

openapi/templates/default/tower-hyper/client/api_clients.mustache
```diff
@@ -213,9 +213,18 @@ impl {{{classname}}} for {{{classname}}}Client {
         {{/isMultipart}}
         {{#hasBodyParam}}
         {{#bodyParam}}
+        {{#required}}
         let body = Body::from(
             serde_json::to_vec(&{{{paramName}}}).map_err(RequestError::Serde)?,
         );
+        {{/required}}
```
tiagolobocastro, line 220: Why do I still see this, didn't this get merged on a different PR?
Missing a rebase?

niladrih: This is a commit in this PR -- d66835b0e26d04c6834046bcc7c1e05825adafc4  
This is the commit message, and this is to handle the case where the pytest client requires an absent body, and not an empty one.
```bash
`Json<Option<T>>` in actix-web requires a valid JSON body to be present (it accepts `null` but not
an empty body). The Python tests send `DELETE /pools/{id}` with `Content-Type: application/json`
and no body — actix-web's Json extractor tries to parse empty bytes as JSON and fails with 400.

Changes:
- For optional (`required: false`) request bodies: parameter changed from
  `Json(param): Json<Option<T>>` and trait call from `Body(param)` to
  `Body(param.map(|Json(b)| b))`.
- For optional bodies: skip serialization when `None` instead of `null`.
```

control-plane/plugin/src/resources/pool.rs
niladrih: Accept works just as well. I'll make the change.

control-plane/plugin/src/resources/pool.rs
niladrih: Logging error body messages instead of writing them manually.

control-plane/plugin/src/resources/mod.rs
niladrih: 🏁

control-plane/agents/src/common/errors.rs
niladrih: Removed these. However, there a mix of quotations and no quotations in the list.

control-plane/grpc/proto/v1/pool/pool.proto
niladrih: I've added comments here.

control-plane/plugin/src/resources/pool.rs
niladrih: Right, I'd forgotten to get the plugin ones, made the change here as well.

control-plane/plugin/src/resources/pool.rs
niladrih: I've removed the specs_api use.

niladrih:
bors try

niladrih:
bors try

control-plane/plugin/src/resources/pool.rs
```diff
@@ -335,6 +337,223 @@ impl GetWithArgs for Pool {
     }
 }
 
+/// Arguments for deleting a pool.
+#[derive(Debug, Clone, clap::Args)]
+pub struct DeletePoolArgs {
+    /// Id of the pool to delete.
+    pool_id: PoolId,
+
+    /// Purge pool without contacting io-engine.
```
tiagolobocastro: You need to add this if you want clap to add a newline

```suggestion
    /// Purge pool without contacting io-engine.{n}
```

control-plane/plugin/src/resources/pool.rs
```diff
@@ -335,6 +337,223 @@ impl GetWithArgs for Pool {
...
+
+    /// Accept data loss for volumes losing their last healthy replica.
+    /// Required when --purge would cause data loss.
+    #[arg(long)]
+    accept_data_loss: bool,
+
+    /// Accept snapshot loss for snapshots losing their last replica snapshot.
+    /// Required when --purge would cause snapshot loss.
+    #[arg(long)]
+    accept_snapshot_loss: bool,
```
tiagolobocastro: Seems a little funny to use "data_loss" for replicas and "snapshot_loss" for snapshots, as if snapshot is not data loss...
Perhaps we could use "accept_volume_loss" instead?
And perhaps `accept_data_loss` could be a general, accept all  loss override?

control-plane/plugin/src/resources/pool.rs
```diff
@@ -335,6 +337,223 @@ impl GetWithArgs for Pool {
...
+                });
+            }
+        };
+
+        // Check pool state.
+        let state_str = pool
+            .state
+            .as_ref()
+            .map(|s| format!("{:?}", s.status))
+            .unwrap_or_else(|| "Unknown".to_string());
```
tiagolobocastro: Can't we use the type?

```suggestion
            .unwrap_or_default();
```

control-plane/plugin/src/resources/pool.rs
```diff
@@ -335,6 +337,223 @@ impl GetWithArgs for Pool {
...
+                    affected_volumes.insert(volume.spec.uuid.to_string());
+                }
+            }
+        }
+
+        let ready_for_purge = state_str == "Unknown" && cordon_ok;
+
+        let impact = PoolPurgeImpact {
+            pool_id: pool_id.to_string(),
+            state: state_str,
```
tiagolobocastro:
```suggestion
            state: state.to_string(),
```

control-plane/plugin/src/lib.rs
```diff
@@ -321,17 +321,33 @@ impl ExecuteOperation for DeleteArgs {
...
+                pool::Pool::del(args, self.ignore_not_found, &cli_args.output).await
             }
-            DeleteResources::VolumeSnapshot(args) => {
-                snapshot::VolumeSnapshots::del(args, self.ignore_not_found, &cli_args.output).await
+            _ => {
+                resources::confirm(
+                    "Are you sure you want to delete the resource?",
+                    "This operation cannot be undone!",
+                    self.yes,
+                )?;
```
tiagolobocastro, line 334: If we have the --yes, perhaps we don't need the --accept?

control-plane/stor-port/src/types/v0/transport/pool.rs
```diff
@@ -662,6 +717,150 @@ impl From<CreatePool> for DestroyPool {
     }
 }
 
+/// Result of a pool purge operation.
+///
+/// Only returned by purge operations (not normal deletes). The `data_loss` and `snapshot_loss`
+/// fields are always present — empty lists indicate no loss occurred.
+#[derive(Serialize, Deserialize, Default, Debug, Clone, Eq, PartialEq)]
```
tiagolobocastro: Do these need the serialize/deser implemented?

control-plane/plugin/src/resources/pool.rs
```diff
@@ -335,6 +337,223 @@ impl GetWithArgs for Pool {
...
+impl Delete for Pool {
+    type ID = DeletePoolArgs;
+
+    async fn del(id: &Self::ID, ignore_not_found: bool, output: &OutputFormat) -> PluginResult {
+        // If --show-impact is set, show purge impact analysis and return without deleting.
+        if id.show_impact {
+            return Self::show_impact_analysis(&id.pool_id, output).await;
+        }
+
+        // Build request body
```
tiagolobocastro:
```suggestion
```

control-plane/plugin/src/resources/pool.rs
```diff
@@ -335,6 +337,223 @@ impl GetWithArgs for Pool {
...
+            }
+            Err(source) => {
+                // Handle not found with ignore flag
+                if ignore_not_found && source.status() == Some(StatusCode::NOT_FOUND) {
+                    return Ok(());
+                }
+
+                // Print the server-provided error message for pool purge errors.
+                if let Some(body) = source.error_body() {
+                    eprintln!("{}", body.message);
```
tiagolobocastro: what I had meant was not for this error, rather the error in the plugin error itself, like [this](https://github.com/openebs/mayastor-control-plane/blob/a0c0b5fa837a46c07807e588aa5d7762fd5cfb6a/control-plane/plugin/src/resources/error.rs#L155)

control-plane/plugin/src/resources/pool.rs
```diff
@@ -335,6 +337,223 @@ impl GetWithArgs for Pool {
...
+        );
+
+        row![self.pool_id, data_loss, snapshot_loss,]
+    }
+}
+
+/// Result of pool purge impact analysis.
+#[derive(Serialize, Default, Debug, Clone)]
+pub struct PoolPurgeImpact {
+    pub pool_id: String,
```
tiagolobocastro: Use the types

control-plane/plugin/src/resources/pool.rs
tiagolobocastro: Was marked as resolved, but still using strings here and in the impact in general...

control-plane/plugin/src/resources/pool.rs
niladrih: Okay 👍🏾

control-plane/stor-port/src/types/v0/transport/pool.rs
niladrih: That's right, I didn't pay attention to that at all. That and the lower order types embedded in it don't require the Serialize and Deserialize.

control-plane/plugin/src/lib.rs
niladrih: I've removed the `--accept` entirely from the plugin options. The REST API and the gRPC still have it.

control-plane/plugin/src/resources/pool.rs
niladrih: Using unwrap_or(Unknown) type, not raw string.

control-plane/plugin/src/resources/pool.rs
niladrih: Changed all `accept_data_loss` to `accept_volume_loss` across all of the APIs. Added the CLI option `accept_data_loss` to mean both `accept_volume_loss` and `accept_snapshot_loss`.

control-plane/plugin/src/resources/pool.rs
niladrih: Oh, TIL.

control-plane/plugin/src/resources/pool.rs
niladrih: Apologies, I didn't realize we'd moved to this way of doing things. I recall you'd made a large refactor across this codebase a long while back to implement errors this way. I was under the impression we still print things outright on the plugin.

control-plane/plugin/src/resources/pool.rs
niladrih: Okay.

control-plane/plugin/src/resources/pool.rs
niladrih: I've removed these requating and matching lines which act on raw strings, in favour of Error types.

niladrih:
bors try

control-plane/agents/src/bin/core/pool/pool_operations.rs
```diff
@@ -314,3 +295,378 @@ impl ResourceCordon for OperationGuardArc<PoolSpec> {
...
+            Err(error) => Err(error),
+        };
+        self.complete_destroy(result, registry).await?;
+        Ok(None)
+    }
+
+    /// Start destroy operation for purge - skips validate_destroy since purge
+    /// explicitly handles pools with replicas/snapshots.
+    async fn start_destroy_purge(&mut self, registry: &Registry) -> Result<(), SvcError> {
+        let spec_clone = {
```
tiagolobocastro: This should be tied into the existing destroy impl, we should avoid this this manually here

control-plane/agents/src/bin/core/pool/pool_operations.rs
```diff
@@ -314,3 +295,378 @@ impl ResourceCordon for OperationGuardArc<PoolSpec> {
...
+                &pool_spec.node,
+                ReplicaOwners::new_disown_all(),
+            );
+            if let Err(error) = replica.destroy(registry, &destroy_request).await {
+                tracing::warn!(
+                    replica.uuid = %replica_spec.uuid,
+                    %error,
+                    "Failed to destroy replica via io-engine, removing spec only"
+                );
+                replica.destroy_for_purge(registry).await?;
```
tiagolobocastro: Pool should be started for deletion at this point, otherwise we're starting to break before it knows, this could be confusing

control-plane/agents/src/bin/core/pool/pool_operations.rs
```diff
@@ -314,3 +295,378 @@ impl ResourceCordon for OperationGuardArc<PoolSpec> {
...
+    ) -> Result<(), SvcError> {
+        let node_online = match registry.node_wrapper(&pool_spec.node).await {
+            Ok(node) => node.read().await.is_online(),
+            Err(_) => false,
+        };
+
+        if node_online {
+            // Node is online - check pool state via ctrl_pool_state
+            if let Ok(ctrl_state) = registry.ctrl_pool_state(pool_id).await {
+                if ctrl_state.status != PoolStatus::Unknown {
```
tiagolobocastro: should allow offline as well

control-plane/agents/src/bin/core/pool/replica_operations.rs
```diff
@@ -228,6 +233,24 @@ impl ResourceOwnerUpdate for OperationGuardArc<ReplicaSpec> {
 }
 
 impl OperationGuardArc<ReplicaSpec> {
+    /// Destroy this replica's spec as part of a pool purge.
+    ///
+    /// Unlike normal destroy, this:
+    /// - Skips the io-engine RPC (the node is offline during purge)
+    /// - Force-disowns the replica regardless of current ownership
+    /// - Overwrites any stale pending operations (node went down mid-op)
+    pub(crate) async fn destroy_for_purge(&mut self, registry: &Registry) -> Result<(), SvcError> {
```
tiagolobocastro: same here, should be integrated with the `GuardedOperationsHelper`

niladrih: Done.

control-plane/agents/src/bin/core/pool/pool_operations.rs
niladrih: Both Unknown and Offline Pools are allowed inside the block now.

control-plane/agents/src/bin/core/pool/pool_operations.rs
niladrih: Right. The start is marked too late for it to be correct. Moved the start to before this step.

niladrih:
bors try

niladrih:
bors try

control-plane/agents/src/bin/core/controller/reconciler/pool/mod.rs
```diff
@@ -173,16 +175,24 @@ async fn missing_pool_state_reconciler(
...
+/// If the pool is in Purging state (from a pool purge operation that failed mid-way),
+/// complete the spec-only deletion without waiting for the node.
 #[tracing::instrument(skip(pool, context), level = "trace", fields(pool.id = %pool.id(), request.reconcile = true))]
 async fn deleting_pool_spec_reconciler(
     pool: &mut OperationGuardArc<PoolSpec>,
     context: &PollContext,
 ) -> PollResult {
-    if !pool.as_ref().status().deleting() {
+    let status = pool.as_ref().status();
+    if !status.deleting() && !status.purging() {
```
tiagolobocastro: We could merge without these spec deletion changes

control-plane/agents/src/bin/core/controller/reconciler/pool/mod.rs
```diff
@@ -162,8 +164,8 @@ async fn missing_pool_state_reconciler(
                 }
             }
         }
-        .instrument(tracing::info_span!("missing_pool_state_reconciler", pool.id = %pool_spec.id, request.reconcile = true))
-        .await
+            .instrument(tracing::info_span!("missing_pool_state_reconciler", pool.id = %pool_spec.id, request.reconcile = true))
+            .await
```
tiagolobocastro, line 168: this seems wrong tabbing?

tiagolobocastro (approved):
We could merge this without the specstatus changes for now and add that in subsequent PR
Please add description to the PR itself

niladrih:
bors merge

control-plane/agents/src/bin/core/controller/reconciler/pool/mod.rs
niladrih: This is not going in, removed these changes.

PR #1061, opened 2026-02-18
Refactor Dsp Operator

tiagolobocastro:
Some behaviors in the operator, in the way api versions are handled seemed incorrect so I've refactored the code.
This aims to fix and also ensure correctness by leveraging the type system.

    fix(k8s/dsp): handle mixed stored versions
    
    If the operator crashes between merging the crd and discarding
    the older schema, we end up with 2 stored versions when starting.
    
    The previous code would just pick the first element, which is not
    clear and may potentially be incorrect.
    
    Enforce the stored version restrictions, allowing it to crash
    rather than do something incorrectly.
    
    todo: we could also skip the merge when we have already merged!
    
----

    fix(k8s/dsp): migrate crd only if necessary
    
    If the stored version is already the latest, then we can simply
    patch to the latest crd, and avoid the large dump.
    
---

    refactor(k8s/dsp): remove hardcoded versions
    
    Use latest where applicable instead of v1beta3.
    Replace explicit names with deprecated/latest lingo.
    
---

    chore: don't warn when no msp is found
    
    We're way past someone having an msp, so we don't need to make
    a glaring notification.
    
---

    fix(k8s/dsp): merge crd with previous version only if required
    
    In case the current version of the CRD is already the latest, then we
    can simply patch to the current schema.
    
    This is possible now that the api is clear, and we know we're on the
    latest version.
    
---

    refactor(k8s/dsp): clarify api versions
    
    There's currently a bug which causes the operator to merge v1beta2 with
    v1beta3 even if the v1beta2 was never installed in the cluster.
    
    This change makes use of the type system to clarify the distinction
    between the previous and the current/latest api version.
    
    Also removes a lot of redundant code passing the latest version
    as a raw string.
    
---

    refactor(k8s/dsp): tidy up some of the logging
    
    Removes the pretty json (devs can just copy and use jq).
    Adds missing logs after CRD creation.
    Use correct verbs after update, etc.

### DIFF
--- a/k8s/operators/src/pool/diskpool/client.rs
+++ b/k8s/operators/src/pool/diskpool/client.rs
@@ -48,7 +48,7 @@ pub(crate) async fn create_crd(k8s: Client) -> Result<CustomResourceDefinition, 
 }
 
 /// This discards older unserved schema from the crd.
-pub(crate) async fn discard_older_schema(k8s: &Client, new_version: &str) -> Result<(), Error> {
+pub(crate) async fn discard_older_schema(k8s: &Client) -> Result<(), Error> {
     let crd_api: Api<CustomResourceDefinition> = Api::all(k8s.clone());
     let mut new_crd = DiskPool::crd();
     let crd_name = new_crd
@@ -63,7 +63,7 @@ pub(crate) async fn discard_older_schema(k8s: &Client, new_version: &str) -> Res
             "Replacing CRD: {}",
             serde_json::to_string_pretty(&new_crd).unwrap()
         );
-        update_stored_version(k8s, crd_name, new_version).await?;
+        update_stored_version(k8s, crd_name).await?;
         if let Ok(modified_crd) = crd_api.get(crd_name).await {
             new_crd.metadata.resource_version = modified_crd.resource_version();
             let pp = PostParams::default();
@@ -122,19 +122,16 @@ pub(crate) async fn create_missing_cr(
     Ok(())
 }
 
-/// Updates stored version in CRD status to the new version passed as arg,
+/// Updates stored version in CRD status to the latest version.
 /// Please ensure that older versions are served:false, stored:false before calling this method,
 /// This would allow us to remove older schema from CRD versions.
-async fn update_stored_version(
-    k8s: &Client,
-    crd_name: &str,
-    new_version: &str,
-) -> Result<(), Error> {
+/// todo: Shouldn't we just ensure that here rather leave it for chance??
+async fn update_stored_version(k8s: &Client, crd_name: &str) -> Result<(), Error> {
     let crd_api: Api<CustomResourceDefinition> = Api::all(k8s.clone());
     if let Ok(mut crd) = crd_api.get_status(crd_name).await {
         let param = PatchParams::apply("status_patch").force();
         if let Some(status) = crd.status.as_mut() {
-            status.stored_versions = Some(vec![new_version.to_string()]);
+            status.stored_versions = Some(vec![ApiVersion::Latest.to_string()]);
         } else {
             return Err(Error::CrdFieldMissing {
                 name: crd_name.to_string(),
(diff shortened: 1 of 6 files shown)


k8s/operators/src/pool/diskpool/crd/migration.rs
```diff
@@ -24,21 +24,20 @@ const PAGINATION_LIMIT: u32 = 100;
...
     k8s: Client,
     namespace: &str,
-    api_version: &ApiVersion,
-    target_schema: &str,
+    api_version: ApiVersion,
 ) -> Result<(), Error> {
     match ensure_crd(&k8s, api_version).await {
         Ok(o) => {
-            info!(crd = ?o.name_any(), "Created");
+            info!(crd = ?o.name_any(), "Updated DiskPool CRD");
```
niladrih, line 31: It also creates the CRD, for fresh installs.

k8s/operators/src/pool/main.rs
```diff
@@ -88,39 +84,75 @@ async fn reconcile(dsp: Arc<DiskPool>, ctx: Arc<OperatorContext>) -> Result<Acti
...
     V1Beta2,
-    /// Represents v1beta3
-    V1Beta3,
 }
 
+/// Current Api versions for the [`DiskPool`].
+#[derive(Debug, Clone, Copy, Eq, PartialEq, Ord, PartialOrd)]
+pub enum ApiVersion {
+    Deprecated(PrevApiVersion),
+    Latest,
```
niladrih, line 103: Maybe call it 'Current'?

niladrih (approved):

k8s/operators/src/pool/diskpool/crd/migration.rs
tiagolobocastro: hmm no that's done on `create_crd` called by `pool_controller` ?

k8s/operators/src/pool/main.rs
tiagolobocastro: I thought about it but somehow `Current` was making me confuse it with the currently installed, which is why I came up with `Latest` instead, wdyt?

k8s/operators/src/pool/diskpool/crd/migration.rs
niladrih: Oh, this one used to be the one to do it, in the olden days :P 👴🏾.

k8s/operators/src/pool/main.rs
niladrih: Sure, against Deprecated, both Current and Latest set themselves apart. It's fine, nothing absurd IMO.

Abhinandan-Purkait (approved):

k8s/operators/src/pool/diskpool/client.rs
```diff
@@ -122,31 +117,44 @@ pub(crate) async fn create_missing_cr(
...
+
+    warn!(
+        stored_versions = ?status.stored_versions,
+        latest = ?set_latest,
+        "Replacing stored version with the latest version"
+    );
+
+    status.stored_versions = set_latest;
+    // todo: why is this being cleared?
+    crd.metadata.managed_fields = None;
```
Abhinandan-Purkait, line 150: Indeed, we should not do that. This helps us see the history right iirc? cc @niladrih ?

k8s/operators/src/pool/main.rs
```diff
@@ -88,39 +84,75 @@ async fn reconcile(dsp: Arc<DiskPool>, ctx: Arc<OperatorContext>) -> Result<Acti
         None => dsp.init_cr().await,
     }
 }
-
-/// Api version of DSP.
-#[derive(Debug, Clone, PartialEq)]
-pub enum ApiVersion {
+/// Previous Api versions for the [`DiskPool`].
+#[derive(Debug, Clone, Copy, Eq, PartialEq, Ord, PartialOrd, Display, EnumString, VariantNames)]
+#[strum(serialize_all = "lowercase")]
```
Abhinandan-Purkait: VariantNames are we using this anywhere?

k8s/operators/src/pool/main.rs
```diff
@@ -403,3 +435,53 @@ pub(crate) async fn migrate_and_clean_msps(k8s: &Client, namespace: &str) -> Res
...
+        assert_eq!(
+            "v1alpha1".parse::<ApiVersion>().unwrap(),
+            ApiVersion::Deprecated(PrevApiVersion::V1Alpha1)
+        );
+        assert_eq!(
+            ApiVersion::Deprecated(PrevApiVersion::V1Alpha1).to_string(),
+            PrevApiVersion::V1Alpha1.to_string()
+        );
+        let latest_v = "v1beta3";
+        assert_eq!(latest_v, DiskPool::version(&()));
```
Abhinandan-Purkait: assert_eq!(ApiVersion::Latest.to_string(), DiskPool::version(&()))  ? Can this be done like this? Just to avoid updating the test in future?

k8s/operators/src/pool/diskpool/client.rs
tiagolobocastro: cc @abhilashshetty04 ?

k8s/operators/src/pool/main.rs
tiagolobocastro: That's right, we're not, I ended up doing without that, removed

k8s/operators/src/pool/main.rs
tiagolobocastro: Done

tiagolobocastro:
bors merge

PR #1064, opened 2026-02-23
Handle target failover for RWX block

tiagolobocastro:
feat(rwx): handle failure reports from multiple nodes
    
    With RWX the cluster-agent may receive failovers from multiple nodes
    for the same volume.
    We now record this this per node, allowing us to keep track of them
    separately.
    
    Ideally we should modify the switchover engine to process the requests
    for the same volume as a single operation, but this would require further
    refactoring of the switchover request.
    Since the core agents has now some tolerance ensuring that at least
    one republish will not trigger a move, this should sufficient.
    
    Also the switchover request on the pstor now has to contain the frontend
    node in the key, allowing us to store the requests independently.
    A new rwx parameter is stored, allowing us to store in either way, thus
    keeping back compatibility with older requests during upgrade.
    
---

    feat(rwx/ha): keep track of republished volume frontends
    
    When ha triggers failover, a republish of the volume is triggerd by the
    cluster agent.
    In case of RWX, multiple nodes may request the failover, but we should
    not allow each one to trigger a new failover, otherwise each request
    could undo the previous and potentially lead into a deadlock.
    
    To avoid this we ensure that each frontend must republish with no target
    changes at least once.
    
    Note that it's still possible to end up in disarray in case of a network
    partition affecting only some of the frontends.
    
---

    fix(rwx/ha): ensure hostnqn is used when replacing the failed path
    
    This fix is similar to another which had been done on the csi-node, but
    that was not sufficient as it only covered the CSI workflow.
    The ha path replacement is using another entrypoint, which did not use
    the corrected uri, meaning the subsystem could be connected without
    using the hostnqn.
    This would lead to infinite reconnect attempts...

### DIFF
--- a/control-plane/agents/src/bin/core/tests/volume/switchover.rs
+++ b/control-plane/agents/src/bin/core/tests/volume/switchover.rs
@@ -846,6 +846,7 @@ async fn republished_nexus_io_engine_txn_fail() {
         .unwrap();
 
     let node_idx0 = cluster.node(0);
+    let app_node_idx0 = cluster.csi_node(0);
     let node_idx1 = cluster.node(1);
     let app_id = cluster.csi_node(0);
     let vol_cli = cluster.grpc_client().volume();
@@ -869,7 +870,7 @@ async fn republished_nexus_io_engine_txn_fail() {
                 share: Some(VolumeShareProtocol::Nvmf),
                 target_node: Some(node_idx0.clone()),
                 publish_context: HashMap::new(),
-                frontend_nodes: vec![node_idx0.to_string()],
+                frontend_nodes: vec![app_node_idx0.to_string()],
                 access_mode: VolumeAccessMode::SingleNodeWriter,
             },
             None,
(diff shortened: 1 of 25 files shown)


Abhinandan-Purkait (approved):

niladrih (approved):
Hey Tiago, if I gathered the context correctly here: with RWX, there's (say) three frontend nodes trying to connect to a single nvme-of device. And if this device were to disappear, then all of these three hosts would request failovers for this volume. And, this will result in at least three consecutive failovers.

The change here requires that the frontend node request failovers with some host identification. And once a republish is done, all of the other hosts must also do a republish, before they're allowed to request failovers. Cluster agent, now, instead of tracking failovers per volume, sees them as failovers per node per volume.  

It looks fine to me, but I don't have the most context here. So apply 'grain of salt' :P. I assume you've considered network partition scenarios for this workflow, those are usually hairy for these kind of sequences.

tiagolobocastro:
> It looks fine to me, but I don't have the most context here. So apply 'grain of salt' :P. I assume you've considered network partition scenarios for this workflow, those are usually hairy for these kind of sequences.

As per the commit description, network partitions are not handled we'd need to come up with an holistic approach for determining when to accept 1+ initiator cannot reconnect and prefer to keep the other ones connected.

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

PR #1065, opened 2026-02-27
fix(csi/node): more robust on bad udev properties and disconnects

tiagolobocastro:
When the udev properties don't match, try to use the sysfs attributes instead.

Also, when disconnecting try harder to wait for controller deletion and log the steps for better debug support.
This is done by fetching all the controllers from the subsystem and trying to issue controller delete 1 by 1.
We then wait for all these controllers to at least enter their deletion phase before completing the csi call.

todo: do we need to use udev at all?

### DIFF
--- a/control-plane/csi-driver/src/bin/node/dev/nvmf.rs
+++ b/control-plane/csi-driver/src/bin/node/dev/nvmf.rs
@@ -1,6 +1,6 @@
 use nvmeadm::{
     error::NvmeError,
-    nvmf_discovery::{disconnect, ConnectArgsBuilder, TrType},
+    nvmf_discovery::{ConnectArgsBuilder, TrType},
 };
 use std::{
     collections::HashMap,
@@ -9,6 +9,14 @@ use std::{
     str::FromStr,
 };
 
+use super::{Attach, AttachParameters, Detach, DeviceError, DeviceName};
+use crate::{
+    config::{config, NvmeConfig, NvmeParseParams},
+    dev::util::extract_uuid,
+    match_dev::match_nvmf_device,
+    node::RDMA_CONNECT_CHECK,
+    runtime,
+};
 use csi_driver::PublishParams;
 use glob::glob;
 use nvmeadm::nvmf_subsystem::Subsystem;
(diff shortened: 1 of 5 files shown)


tiagolobocastro:
bors try

control-plane/csi-driver/src/bin/node/dev/nvmf.rs
```diff
@@ -291,30 +289,143 @@ impl Attach for NvmfAttach {
...
+                }
+                tracing::debug!(name, traddr, state = %controller.state, "Controller state sync");
+
+                if start.elapsed() >= deadline {
+                    return Err(DeviceError::new(format!(
+                        "Timeout waiting for {name}/{nqn}/{traddr} to get removed"
+                    )));
+                }
+                tokio::time::sleep(std::time::Duration::from_millis(100)).await;
+            }
```
Abhinandan-Purkait, line 425: Since this is a nested loop and the deadline and start both defined outside it can happen that the deadline is exceeded when we iterate over only a few controllers right? We are not waiting enough to check for all the controllers? Consider if there are a lot of controllers?

Abhinandan-Purkait (approved):
LGTM. Just one thought!

control-plane/csi-driver/src/bin/node/dev/nvmf.rs
tiagolobocastro: We shouldn't have too many controllers connected, but it could happen I guess.

We've already done the needful by issuing delete against all controllers, so we just retry when csi retries the call?

Abhinandan-Purkait: Yeah, csi retry might work, but I was just wondering instead of a 15 sec deadline overall, we should do a per controller deadline?

Abhinandan-Purkait: But even that might lead to timeout on the caller side. Yeah not sure.

tiagolobocastro: But how will that help? (remember, we've already issued the delete, we're only waiting)

Abhinandan-Purkait: Right. Okay

abhilashshetty04 (approved):

tiagolobocastro:
bors merge

tiagolobocastro:
@mergify backport develop

PR #1066, opened 2026-03-06
feat: use new io-engine status error codes

tiagolobocastro:
feat: use new io-engine status error codes
    
    New io-engine will now return DataLoss with errno set to EIO or EILSEQ
    depending on whether import failed with I/O error or invalid superblock.
    
---

    fix: adjust to new name for the io error threshold env var
    
    The variable has been renamed to include the pool prefix.


tiagolobocastro:
This should work as is against old versions. Since io-engine PR has not merged yet, let's try it:
bors try

niladrih (approved):

tiagolobocastro:
bors merge

PR #1067, opened 2026-03-09
Handle graceful node shutdown

tiagolobocastro:
feat: add perceived node status to internal and public api
    
    This allows caller to quickly glance the perceived node status,
    without having to check state is missing and infer by itself
    that the node is offline/unknown.
    
    Instead the api fills the correct status by checking the state
    and other checking if the node was shutdown or otherwise.
    
---

    test: take into account new node shutdown state changes
    
    When a node is gracefully shutdown its state now transitions to Offline
    in the registry and also shutdown is persisted to the pstor and then
    exposed via the public api.
    
---

    feat: expose node shutdown via openapi
    
---

    feat: shutdown flag to the io-engine node
    
    When a node deregisters itself, mark it as shutdown.
    
---

    fix(diag): when node is unknown mark pool as unknown as well

### DIFF
--- a/control-plane/agents/src/bin/core/pool/operations_helper.rs
+++ b/control-plane/agents/src/bin/core/pool/operations_helper.rs
@@ -75,7 +75,7 @@ impl OperationGuardArc<PoolSpec> {
         let error = match registry.node_wrapper(node).await {
             Ok(node) => {
                 let node_guard = node.read().await;
-                if !node_guard.is_online() {
+                if node_guard.is_offline() {
                     self.mark_as_offline(PoolError {
                         code: PoolErrorCode::NodeIsOffline,
                         msg: "".to_string(),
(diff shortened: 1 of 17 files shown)


control-plane/agents/src/bin/core/pool/operations_helper.rs
```diff
@@ -88,7 +88,17 @@ impl OperationGuardArc<PoolSpec> {
...
+            if node.is_shutdown() {
+                self.mark_as_offline(PoolError {
+                    code: PoolErrorCode::NodeIsOffline,
+                    msg: "".to_string(),
+                });
+                return Ok(Err(NodeStatus::Offline));
+            }
+        }
+
+        self.mark_as_unknown(PoolError {
```
tiagolobocastro, line 101: In case the node is unknown, should we keep pool state as offline or unknown
CC @abhilashshetty04 @niladrih 

As of this change it will be `Unknown` if the node state is `Unknown` but `Offline` if the node is `Offline` (ie graceful shutdown).

the main intent of unknown was because we didn't really know if the node was offline or in network partition for example

Perhaps we should leave as `Unknown` for now, and come up with a plan to refresh this into better naming (as a separate issue/pr)?

niladrih: This seems reasonable, and symmetric. While I have less context, I don't see a strong reason why this should be the other way round. I approve of this approach.

niladrih (approved):

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

PR #1070, opened 2026-03-13
feat(csi): allow restoring snapshots to larger PVCs

azgms:
## Problem

The CSI spec ([section 5.5, CreateVolume](https://github.com/container-storage-interface/spec/blob/master/spec.md#createvolume)) permits restoring a snapshot to a volume of equal **or greater** size. However, the control plane currently enforces an exact size match between the new volume and the snapshot, causing `CreateVolume` to fail whenever a PVC requests more capacity than the source snapshot.

This is a common scenario when users want to restore data into a larger volume (e.g., expanding storage during a migration or restore).

## Solution

1. **Relax the pre-flight validation** in `clone_operations.rs` from `==` to `>=`, so requests for larger volumes are not rejected upfront.

2. **Create-then-resize strategy** in `service.rs`: when the requested size exceeds the snapshot size, create the clone at the snapshot size first (so the volume spec matches the replica), then resize both spec and replicas to the requested size. This avoids the mismatch where `create_clone` sets the spec size but the replica remains at snapshot size.

3. **Feature flag**: the resize-after-clone logic is gated behind a new `--enable-snapshot-resize-on-clone` CLI flag (env: `ENABLE_SNAPSHOT_RESIZE_ON_CLONE`) that defaults to **disabled**, preserving current behavior for existing deployments.

## Changes

| File | Change |
|------|--------|
| `clone_operations.rs` | `==` -> `>=` size check |
| `service.rs` | Create clone at snapshot size, then resize if flag enabled |
| `registry.rs` | Store and expose the feature flag |
| `main.rs` | Add `--enable-snapshot-resize-on-clone` CLI arg |
| `errors.rs` | Update error message to reflect `>=` semantics |

## Testing

- Tested on a production Mayastor cluster with a snapshot from a 5Gi PVC restored to a 10Gi PVC
- Verified the cloned volume is created at snapshot size, then resized to the requested size
- Confirmed that without the flag, existing behavior (exact match only) is preserved
- Confirmed that restoring to a PVC of equal size still works as before

----

### DIFF
--- a/control-plane/agents/src/bin/core/volume/service.rs
+++ b/control-plane/agents/src/bin/core/volume/service.rs
@@ -598,8 +598,27 @@ impl Service {
         let mut snapshot = self.specs().volume_snapshot(snap_uuid).await?;
         let cluster_size_for_clone_vol =
             snapshot.pool_cluster_size_for_clone(&self.registry).await?;
+        let snapshot_size = snapshot.as_ref().metadata().spec_size();
+        let requested_size = request.params().size;
         request.params_mut().cluster_size = Some(cluster_size_for_clone_vol);
+
+        let resize_after_clone = self.registry.snapshot_resize_on_clone_enabled()
+            && requested_size > snapshot_size;
+
+        if resize_after_clone {
+            request.params_mut().size = snapshot_size;
+        }
         snapshot.create_clone(&self.registry, request).await?;
+
+        if resize_after_clone {
+            let resize_req = ResizeVolume::new(
+                request.params().uuid.clone(),
+                requested_size,
+                None,
+            );
+            self.resize_volume(&resize_req).await?;
+        }
+
         self.registry.volume(&request.params().uuid).await
     }
 
--- a/control-plane/agents/src/bin/core/controller/registry.rs
+++ b/control-plane/agents/src/bin/core/controller/registry.rs
@@ -122,6 +122,8 @@ pub(crate) struct RegistryInner<S: Store> {
     /// Blobstore cluster size(in bytes) required for pool creation and replica scheduling. This is
     /// not per-pool.
     pool_cluster_size: Option<u32>,
+    /// Enable snapshot clone resize to larger PVC sizes.
+    enable_snapshot_resize_on_clone: bool,
     deprecated_access_mode: bool,
     /// Running in simulation mode.
     sim_args: Option<SimArgs>,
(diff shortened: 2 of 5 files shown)


control-plane/agents/src/bin/core/volume/service.rs
```diff
@@ -598,8 +598,27 @@ impl Service {
...
+
+        let resize_after_clone = self.registry.snapshot_resize_on_clone_enabled()
+            && requested_size > snapshot_size;
+
+        if resize_after_clone {
+            request.params_mut().size = snapshot_size;
+        }
         snapshot.create_clone(&self.registry, request).await?;
+
+        if resize_after_clone {
```
tiagolobocastro, line 613: Is this idempotent? If for example clone creation completes but we crash before the resize, would csi pick up the existing volume with the snapshot size?

control-plane/agents/src/bin/core/volume/service.rs
```diff
@@ -598,8 +598,27 @@ impl Service {
...
+            let resize_req = ResizeVolume::new(
+                request.params().uuid.clone(),
+                requested_size,
+                None,
+            );
+            self.resize_volume(&resize_req).await?;
```
tiagolobocastro, line 619: Another potential gotcha here is that we're creating the volume without ensuring the pools have capacity for the ~cloned~ larger restore size?
If you find `ResourcePolicy<CloneVolumeSnapshot> for SimplePolicy` there should be some place there where we check for capacity for request.size or something like that

This would leave the volume potentially leaked if we can't resize?

tiagolobocastro (commented):
Thanks for the PR, I think it's a great addition, just a few questions regarding idempotence and capacity checking

Could you please look whether we could resize the replicas in `CreateVolumeExe for SnapshotCloneOp<'_>` and if not then fail the request?

Also does the dataplane allow cloning to a greater size directly? Though I suspect it probably doesn't

tiagolobocastro:
> Also does the dataplane allow cloning to a greater size directly? Though I suspect it probably doesn't

Seems not, as there's no size [here](https://github.com/openebs/mayastor-dependencies/blob/8f91faeb4b32fe6099760f203611259a7c070d1a/apis/io-engine/protobuf/v1/snapshot.proto#L106).

tiagolobocastro:
@azgms did you get a chance to take a look at the comments?

abhilashshetty04:
@azgms, just checking in—are you still planning to work on this PR? It's been inactive for a while.

PR #1071, opened 2026-03-16
Pool probe

tiagolobocastro:
refactor: use new mapping of pool error code to status
    
    We've added the mapping on previous commit, so we can use make
    use of that for the node_wrapper.
    
---

    feat: probe pool before trying to import import
    
    This allows us to avoid many levels of error logs when a pool backend is
    not ready for whatever reason.
    The metadata/superblock check is not part of the probe api yet, but in
    the future it'll also allow us to avoid importing over and over again
    in case of superblock corruption.
    
    This probe also allows us to expose better errors out of the pool
    diagnostics api.
    
    If the target io-engine does not implement this api then we ignore it
    and proceed to itrying to import the pool. This keeps the existing
    behaviour as is during an upgrade.
    
---

    feat: expose probe errors via internal and external api
    
---

    feat: add pool probe to core agent node wrapper and pool api
    
    This will allow any part of the agent to probe pool storage for errors.
    
---

    chore: update dependencies submodule to include probing


niladrih (approved):

control-plane/agents/src/bin/core/node/wrapper.rs
```diff
@@ -1328,6 +1330,21 @@ impl PoolApi for Arc<tokio::sync::RwLock<NodeWrapper>> {
...
+    async fn probe_pool(&self, request: &ProbePoolRequest) -> Result<ProbePoolResponse, SvcError> {
+        let dataplane = self
+            .grpc_client_locked(MessageIdVs::ProbePool.into())
+            .await?;
+        match dataplane.probe_pool(request).await {
+            Ok(resp) => Ok(resp),
+            Err(SvcError::GrpcRequestError { source, .. })
+                if source.code() == tonic::Code::Unimplemented =>
+            {
+                Ok(ProbePoolResponse::new_unimpl())
```
Abhinandan-Purkait, line 1344: Let's add a warn or info maybe to notify this is older version and doesn't support? Just for visibility

control-plane/agents/src/bin/core/pool/operations_helper.rs
```diff
@@ -118,20 +123,69 @@ impl OperationGuardArc<PoolSpec> {
...
+        for (_, error) in probed.errors {
+            for probe in error.error {
+                diag.import_errors.push(probe);
+            }
+        }
+        diag.error = diag.import_errors.first().map(|e| e.error.clone());
+        diag.status =
+            self.pool_error_to_status(diag.error.as_ref().map(|e| e.code).unwrap_or_default());
+        let mut pool = self.lock();
+        pool.metadata.runtime.diag = Some(diag);
```
Abhinandan-Purkait, line 192: Shouldn't we merge rather than replace? Line 175 will always reset to default in every probe and that will remove the other fields. I mean might be okay but just wanted to confirm is this intended?

control-plane/agents/src/bin/core/pool/operations_helper.rs
```diff
@@ -118,20 +123,69 @@ impl OperationGuardArc<PoolSpec> {
...
+        }
+
+        let mut diag = PoolDiag::default();
+
+        for (_, error) in probed.errors {
+            for probe in error.error {
+                diag.import_errors.push(probe);
+            }
+        }
+        diag.error = diag.import_errors.first().map(|e| e.error.clone());
```
Abhinandan-Purkait, line 188: Should we not join all the errors and set rather than just the first one?

control-plane/agents/src/bin/core/pool/operations_helper.rs
```diff
@@ -118,20 +123,69 @@ impl OperationGuardArc<PoolSpec> {
...
+            PoolErrorCode::NodeIsUnknown => PoolStatus::Unknown,
+            PoolErrorCode::NodeIsOffline => PoolStatus::Offline,
+            PoolErrorCode::ImportDisabled => PoolStatus::Offline,
+            PoolErrorCode::TimeOut => PoolStatus::Unknown,
+            PoolErrorCode::DiskClaimed => PoolStatus::Offline,
+            PoolErrorCode::PCIDriverUnsupported => PoolStatus::Offline,
+            PoolErrorCode::PCIKernelBound => PoolStatus::Offline,
+            PoolErrorCode::PCINotNvme => PoolStatus::Offline,
+            PoolErrorCode::InvalidDiskUri => PoolStatus::Offline,
+        }
```
Abhinandan-Purkait, line 153: Can you please add a doc comment difference in the logic of calling something as faulted vs offline vs online. Will help in adding new variants.

control-plane/agents/src/bin/core/controller/io_engine/mod.rs
```diff
@@ -71,6 +72,9 @@ pub(crate) trait PoolApi {
     async fn expand_pool(&self, request: &ExpandPool) -> Result<PoolState, SvcError>;
     /// Clear the pool errors.
     async fn clear_errors(&self, request: &ClearErrorsRequest) -> Result<PoolState, SvcError>;
+    /// Probe pool's disks for errors.
+    #[allow(unused)]
```
Abhinandan-Purkait: Is this not used?

Abhinandan-Purkait (approved):

control-plane/agents/src/bin/core/controller/io_engine/mod.rs
tiagolobocastro: It wasn't when interface was added, subsequent commit started using it...
Removed

control-plane/agents/src/bin/core/pool/operations_helper.rs
tiagolobocastro: Indeed, I've added it

control-plane/agents/src/bin/core/pool/operations_helper.rs
tiagolobocastro: If you want all errors, then you just loop all the probed errors
This is just a high-level error representing the entire pool. Since we only have 1 disk per pool so far, we don't need anything different, we just pick the first error.

control-plane/agents/src/bin/core/pool/operations_helper.rs
tiagolobocastro: No, here we have to replace the errors, since this is the new probe result, otherwise we'd never get rid of the previous errors.

control-plane/agents/src/bin/core/pool/operations_helper.rs
tiagolobocastro: If you have suggestions on how to improve this, we can take this up on separate PR, please let me know (I'll resolve this to proceed with the merge)

tiagolobocastro:
bors merge

PR #1072, opened 2026-03-18
feat(k8s/operator): add support for pool Purge via the DiskPool operator

niladrih:
Changes:
- Adds a new state 'Deleted' to the CRD for DiskPool
- Adds Pool delete (with purge options) functionality to DiskPool CR delete reconciliation
- Adds code to deserialize openebs.io/delete-opts from the DiskPool CR annotation
- Adds tests for pool delete option handling behaviour
- Adds guiding error messages for purge failures

### DIFF
--- a/k8s/operators/src/pool/context.rs
+++ b/k8s/operators/src/pool/context.rs
@@ -9,9 +9,11 @@ use crate::diskpool::crd::v1beta3::{
 use openapi::{
     apis::StatusCode,
     clients, models,
-    models::{rest_json_error::Kind, CreatePoolBody, Encryption, EncryptionSecret, Pool},
+    models::{
+        rest_json_error::Kind, CreatePoolBody, DeletePoolBody, Encryption, EncryptionSecret, Pool,
+    },
 };
-use tracing::{debug, error, info};
+use tracing::{debug, error, info, trace};
 use utils::{disk::normalize_disk, dsp_created_by_key};
 
 use chrono::Utc;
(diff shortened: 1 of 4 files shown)


k8s/operators/src/pool/diskpool/crd/v1beta3.rs
```diff
@@ -273,6 +273,9 @@ pub enum CrPoolState {
     Created,
     /// This state is set when we receive delete event on the dsp cr.
     Terminating,
+    /// The pool spec was purged from the control plane. The CR is orphaned
+    /// and can be safely deleted by the user.
+    Deleted,
```
Abhinandan-Purkait: Deleted might be a bit ambiguous, purge is not deleting, it just cleans up control plane's view on when things are not reachable. This kind of bundles both, not sure it this portrays the correct state.

k8s/operators/src/pool/context.rs
```diff
@@ -509,23 +616,28 @@ impl ResourceContext {
...
-                    "Destroy",
-                    "Destroyed",
-                    "The pool has been destroyed",
-                    "Normal",
-                )
-                .await;
+                let (reason, message) = if is_purge {
+                    ("Purged", "The pool has been purged")
+                } else {
+                    ("Destroyed", "The pool has been destroyed")
```
Abhinandan-Purkait, line 1046: What is destroyed? Let's not keep introducing different words, just stick to Purge and Delete.

k8s/operators/src/pool/context.rs
```diff
@@ -781,28 +893,72 @@ impl ResourceContext {
...
+                        let delete_body = match self.parse_delete_opts() {
+                            Ok(body) => body,
+                            Err(e) => {
+                                warn!(
+                                    name = self.name_any(),
+                                    %e,
+                                    "ignoring malformed delete-opts annotation, \
+                                     falling back to normal deletion"
+                                );
+                                None
```
Abhinandan-Purkait: Should we rather return an error here?

k8s/operators/src/pool/context.rs
```diff
@@ -781,28 +893,72 @@ impl ResourceContext {
...
+                                if response.status() == StatusCode::NOT_FOUND =>
+                            {
+                                Self::delete_finalizer(self.clone(), false, None).await
+                            }
+                            Err(error) if is_purge => {
+                                // When purge is requested and we cannot reach the pool via
+                                // get_node_pool (e.g. the node is offline and the control
+                                // plane returns an error), we still attempt the delete with
+                                // purge since the REST API can handle purge without the
+                                // io-engine being reachable.
```
Abhinandan-Purkait: get_node_pool returns an error if the node is offline? Will it not just return the pool with only spec and no state? Please double check, this seems incorrect.

k8s/operators/src/pool/context.rs
```diff
@@ -33,6 +35,78 @@ use std::{
 const WHO_AM_I: &str = "DiskPool Operator";
 const WHO_AM_I_SHORT: &str = "dsp-operator";
 
+/// Annotation key for declarative pool deletion options.
+/// The value is a YAML map that gets deserialized into `DeletePoolBody`.
+///
+/// Example:
+/// ```yaml
+/// openebs.io/delete-opts: |
```
tiagolobocastro: how easy is it to add a label like like?
would it be easier to have many key/vals instead?

Also to consider, do we need such fine tuning here, or should we just have a single purge which would accept all? And if user wants more config, then they can use plugin?

k8s/operators/src/pool/context.rs
```diff
@@ -781,28 +893,72 @@ impl ResourceContext {
...
-                        }
-                        Err(clients::tower::Error::Response(response))
-                            if response.status() == StatusCode::NOT_FOUND =>
+                        };
+
+                        let is_purge = delete_body.as_ref().and_then(|b| b.purge).unwrap_or(false);
+
+                        match self
+                            .pools_api()
+                            .get_node_pool(&self.spec.node(), &self.name_any())
```
tiagolobocastro: Why do we need the get request here, can't we just issue the delete and inspect error code?

k8s/operators/src/pool/context.rs
niladrih: These are one-liners for these:
```
kubectl annotate diskpool <pool-name> openebs.io/delete-opts=$'purge: true\nconfirm: true\nconfirm_data_loss: true\nconfirm_snapshot_loss: true' --overwrite
```
```
kubectl patch diskpool <pool-name> --type=merge -p '{"metadata":{"annotations":{"openebs.io/delete-opts":"purge: true\nconfirm: true\nconfirm_data_loss: true\nconfirm_snapshot_loss: true"}}}'
```
```
# Edit the diskpool and make the change by hand on the shell
kubectl edit diskpool <pool-name>
```

It's not something that you could do without copying this command from documentation.
It's not an intuitive API. However, during the design phase, it was not meant to be. It was meant to be an alternative, for an advanced user, who wanted access via the CR exclusively. I think we should have a complete API via the CR, without absolute reliance on the kubectl plugin. We might make the documentation for this API obscure, maybe in troubleshooting, or an 'advanced' section, so that this isn't something most people know about.

k8s/operators/src/pool/context.rs
niladrih: It was already there. It's not being 'introduced'. But we can reason about if it still fits today in the new context.

k8s/operators/src/pool/diskpool/crd/v1beta3.rs
niladrih: This is not exclusive to Purged. As it stands today, there's no way to reason if a downstream pool is missing because of Purged or Destroyed. This just means that it's not there. It's not a hint towards suggesting how it happened to not be there.

k8s/operators/src/pool/context.rs
niladrih: Removed get_node_pool call here. It's an unnecessary round trip from the REST, instead just deleting directly and handling the error.

k8s/operators/src/pool/context.rs
niladrih: Yes it should. I've changed this behaviour to return the error.

k8s/operators/src/pool/context.rs
niladrih: Removed the GET, just tries the delete directly now.

k8s/operators/src/pool/diskpool/crd/v1beta3.rs
```diff
@@ -463,6 +440,21 @@ impl DiskPoolStatus {
...
+        Self {
+            cr_state: CrPoolState::Deleted,
+            pool_status: None,
+            status: Some(PoolStatus::Unknown),
+            error: Some(PoolError {
+                code: PoolErrorCode::PoolDeleted,
+                message: Some("pool spec was removed from the control plane".to_string()),
+            }),
+            ..Default::default()
+        }
```
tiagolobocastro, line 483: missing conditions?

k8s/operators/src/pool/diskpool/crd/v1beta3.rs
```diff
@@ -463,6 +440,21 @@ impl DiskPoolStatus {
         }
     }
 
+    /// Set when the pool spec has been removed from the control plane.
+    /// The CR is orphaned and can be safely deleted by the user.
+    pub fn deleted() -> Self {
+        Self {
+            cr_state: CrPoolState::Deleted,
+            pool_status: None,
+            status: Some(PoolStatus::Unknown),
```
tiagolobocastro:
```suggestion
            status: Some(PoolStatus::Offline),
```

k8s/operators/src/pool/context.rs
tiagolobocastro: Seeing it now, those are very complicate, tempted to say it'd be more straightforward to add multiple keys.

Also we should have also a simple key that purge/acceptsAll to make it simple when you don't care which is the most common case for purge.

k8s/operators/Cargo.toml
```diff
@@ -39,5 +39,6 @@ tokio = { workspace = true, features = ["full"], optional = true }
 humantime = { workspace = true, optional = true }
 tracing = { workspace = true, optional = true }
 parse-size = { workspace = true }
+serde_yaml = { workspace = true }
 strum_macros = { workspace = true, optional = true }
-strum = { workspace = true, optional = true }
+strum = { workspace = true, optional = true }
```
tiagolobocastro: line

Abhinandan-Purkait (approved):
Looks good to me apart from the key value comment from @tiagolobocastro

k8s/operators/src/pool/diskpool/crd/v1beta3.rs
niladrih: I am chaining it at the caller site
```rust
    async fn mark_deleted(&self) -> Result<Action, Error> {
        self.patch_status(DiskPoolStatus::deleted().with_conditions(self))
            .await?;
        Ok(Action::requeue(Duration::from_secs(self.ctx.interval)))
    }
```
... just like how it's being done for terminating
```rust
    async fn mark_terminating_when_unknown(&self) -> Result<Action, Error> {
        self.patch_status(DiskPoolStatus::terminating_when_unknown().with_conditions(self))
            .await?;
        Ok(Action::requeue(Duration::from_secs(self.ctx.interval)))
    }
```

k8s/operators/src/pool/context.rs
niladrih: As we've discussed, I've moved to a different format for inputting the delete options. E.g.
`openebs.io/delete-opts: purge=true,accept=true,accept-data-loss=true`

niladrih: I have not added documentation for `openebs.io/delete-opts=purge-accept-all` in the error logs because I prefer that users see the lists of volumes, snapshots, etc. being lost before they make the call to delete the pool. WDYT?

We can document it on the documentation website. Since all of the Diskpool documentation comes from the website, and users can't `--help` all the documentation on their console, they are sure to not miss it.

niladrih: Also, in the options list, latter options take precedence over earlier options.

tiagolobocastro: > I have not added documentation for openebs.io/delete-opts=purge-accept-all in the error logs

Sorry didn't understand this, what do you mean by that?

Also btw, we have:

`openebs.io/delete-opts=purge,accept,....`

Could purge be implied in the key name:

`openebs.io/purge-opts=accept`

?

niladrih: @tiagolobocastro
> Sorry didn't understand this, what do you mean by that?

There are error logs which show up on the operator which tell you why the purge attempt failed, and what you could do (set an option) to proceed.

I have not documented the purge-accept-all in the error logs.

niladrih: > Could purge be implied in the key name:
>
> `openebs.io/purge-opts=accept`
>
> ?

Sure yeah

niladrih: > Could purge be implied in the key name:
> 
> `openebs.io/purge-opts=accept`
> 
> ?

This would however add an awkward configuration like so...

```yaml
metadata:
  annotations:
    openebs.io/purge-opts: ""
```

Is this saying that they want to purge, or that they want to perform normal delete via io-engine?
`delete-opts` would interpret this as a normal delete request.

This use of the interface is not very well-formed, as it involves using this kubernetes annotation key:value interface is a way that's not very often used.

k8s/operators/src/pool/context.rs
```diff
@@ -33,6 +35,176 @@ use std::{
...
+        String::new()
+    } else {
+        format!("\n  {}", body.details)
+    };
+
+    match body.kind {
+        Kind::Unavailable => Some(format!(
+            "Pool is offline, normal deletion is not possible. \
+             To purge, cordon the pool then annotate the DiskPool CR with delete options:\n  \
+             kubectl mayastor -n {ns} cordon pool {pool_name} --replicas --snapshots\n  \
```
tiagolobocastro: How would this work for the umbrella or any other dependents?

k8s/operators/src/pool/context.rs
```diff
@@ -276,6 +446,84 @@ impl ResourceContext {
...
+            Some(a) => a,
+            None => return Ok(None),
+        };
+        let value = match annotations.get(DELETE_OPTS_ANNOTATION) {
+            Some(v) => v,
+            None => return Ok(None),
+        };
+        let body = parse_delete_opts_value(value)?;
+        if let Some(ref b) = body {
+            debug!(name = dsp.name_any(), ?b, "parsed delete-opts annotation");
```
tiagolobocastro:
```suggestion
            trace!(name = dsp.name_any(), ?b, "parsed delete-opts annotation");
```

k8s/operators/src/pool/context.rs
```diff
@@ -276,6 +446,84 @@ impl ResourceContext {
...
+    fn parse_delete_opts(dsp: &DiskPool) -> Result<Option<DeletePoolBody>, Error> {
+        let annotations = match dsp.metadata.annotations.as_ref() {
+            Some(a) => a,
+            None => return Ok(None),
+        };
+        let value = match annotations.get(DELETE_OPTS_ANNOTATION) {
+            Some(v) => v,
+            None => return Ok(None),
+        };
+        let body = parse_delete_opts_value(value)?;
```
tiagolobocastro, line 683: How would a user know they've passed wrong annotations? Should we add this to the events?

tiagolobocastro (approved):
Seems fine but how does this workflow work with the DiskPool CR if we still have to cordon it which would require the kubectl-plugin anyhow?

Could purge command itself also cordon the pool? Then it doesn't need to be a manual pre-conditon, but handled by the system

k8s/operators/src/pool/context.rs
```diff
@@ -33,6 +35,176 @@ use std::{
...
+
+        let bool_val = value.parse::<bool>().map_err(|_| Error::Generic {
+            message: format!(
+                "invalid value '{value}' for key '{key}' in '{DELETE_OPTS_ANNOTATION}' \
+                 annotation: expected 'true' or 'false'"
+            ),
+        })?;
+
+        match key {
+            "purge-accept-all" => {
```
tiagolobocastro: these should be an enum variant, rather than hardcoded

k8s/operators/src/pool/context.rs
```diff
@@ -781,28 +1055,37 @@ impl ResourceContext {
...
+                        // normal deletion could fail (503 for offline pool) or destroy
+                        // data the user wanted to preserve via the accept flags.
+                        let delete_body = Self::parse_delete_opts(&dsp)?;
+
+                        if dsp.status.as_ref().map(|d| d.cr_state) != Some(CrPoolState::Terminating)
                         {
-                            Self::delete_finalizer(self.clone(), false).await
+                            let _ = self
+                                .patch_status(
+                                    DiskPoolStatus::terminating_when_unknown()
```
tiagolobocastro: hmm this doesn't seem right, why using this instead of the terminating method?

tiagolobocastro (changes requested):

k8s/operators/src/pool/context.rs
niladrih: The umbrella plugin commands are on the next line. Other dependents... I'm not sure. Do you have any suggestions?

k8s/operators/src/pool/context.rs
```diff
@@ -33,6 +35,176 @@ use std::{
...
+             To purge, cordon the pool then annotate the DiskPool CR with delete options:\n  \
+             kubectl mayastor -n {ns} cordon pool {pool_name} --replicas --snapshots\n  \
+             # (or: kubectl openebs mayastor -n {ns} cordon pool {pool_name} --replicas --snapshots)\n\
+             Then apply the annotation:\n  \
+             {DELETE_OPTS_ANNOTATION}: purge=true\n  \
+             kubectl annotate diskpool {pool_name} -n {ns} \
+             {DELETE_OPTS_ANNOTATION}=purge=true --overwrite"
+        )),
+        Kind::PoolNotPurgeable => Some(format!(
+            "Cannot purge: pool state is known (Online/Faulted/Degraded). \
```
abhilashshetty04: If pool has too many w.r.t IO failing/stalls etc. We may mark it as Suspected.  Would we want an Option to discard such pools safely. Node may be Online and other pools on the same node may be working Fine.

k8s/operators/src/pool/context.rs
niladrih: Right this should definitely be terminating instead of terminating_when_unknown.

k8s/operators/src/pool/context.rs
niladrih: I've added an eventing framework sort of. It validates the length of the unexpected messages at runtime, and also validates the length of the raw strings at compile time (assuming k8s names and namespaces as 63 characters).

niladrih:
> Seems fine but how does this workflow work with the DiskPool CR if we still have to cordon it which would require the kubectl-plugin anyhow?
>
> Could purge command itself also cordon the pool? Then it doesn't need to be a manual pre-conditon, but handled by the system

I've added pool cordon for replicas and snapshots to the purge workflow. I had a question -- if we fail purge, should we uncordon?

If we don't uncordon, in the next reconciliation attempt the error is fixed, we succeed the purge, or if we did uncordon earlier, we cordon + attempt purge.

If we don't uncordon, and in the next reconciliation attempt we notice that the purge annoation is gone, we will have to try normal delete. Which is fine, because normal delete is not affected by cordon.

I am of the opinion that we should not uncordon, we should leave it cordoned, on failure.
WDYT @tiagolobocastro @abhilashshetty04 @Abhinandan-Purkait?

k8s/operators/src/pool/context.rs
niladrih: Added an Enum mapped to these variants.

niladrih:
> > Seems fine but how does this workflow work with the DiskPool CR if we still have to cordon it which would require the kubectl-plugin anyhow?
> > Could purge command itself also cordon the pool? Then it doesn't need to be a manual pre-conditon, but handled by the system
> 
> I've added pool cordon for replicas and snapshots to the purge workflow. I had a question -- if we fail purge, should we uncordon?
> 
> If we don't uncordon, in the next reconciliation attempt the error is fixed, we succeed the purge, or if we did uncordon earlier, we cordon + attempt purge.
> 
> If we don't uncordon, and in the next reconciliation attempt we notice that the purge annoation is gone, we will have to try normal delete. Which is fine, because normal delete is not affected by cordon.
> 
> I am of the opinion that we should not uncordon, we should leave it cordoned, on failure. WDYT @tiagolobocastro @abhilashshetty04 @Abhinandan-Purkait?

Leaving it cordoned.

k8s/operators/src/pool/context.rs
niladrih: I've added the Suspended state to the log message. The control plane doesn't allow for purge if the pool state is not Unknown or Offline.

tiagolobocastro (approved):

niladrih:
bors merge

PR #1073, opened 2026-03-19
fix: don't thrash diskpool cr on "migration"

tiagolobocastro:
DiskPool migration was brought to upgrade older api versions to the latest. This is erasing all metadata a user may have added!

Also, on v2.10 migration began to run even from same version as optional new fields were added.
This was also erasing clusterSize, maxExpansion and encryptionConfig!

### DIFF
--- a/k8s/operators/src/pool/diskpool/crd/migration.rs
+++ b/k8s/operators/src/pool/diskpool/crd/migration.rs
@@ -1,8 +1,6 @@
 use super::{
-    v1alpha1::DiskPool as AlphaDiskPool,
-    v1beta1::DiskPool as Beta1DiskPool,
-    v1beta2::DiskPool as Beta2DiskPool,
-    v1beta3::{DiskPool, DiskPoolSpec},
+    v1alpha1::DiskPool as AlphaDiskPool, v1beta1::DiskPool as Beta1DiskPool,
+    v1beta2::DiskPool as Beta2DiskPool, v1beta3::DiskPool,
 };
 use crate::{
     diskpool::client::{discard_older_schema, dsp_api, list_existing_cr},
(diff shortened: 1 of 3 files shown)


k8s/operators/src/pool/diskpool/crd/migration.rs
```diff
@@ -115,62 +113,44 @@ pub(crate) async fn migrate_to_latest(
...
-                    name: dsp.name_any(),
-                    field: "resource_version".to_string(),
-                });
-            }
-        }
-    } else {
-        return Err(Error::Generic {
+    let pools = list_existing_cr(&k8s, ns, pagination_limit)
+        .await
+        .map_err(|_| Error::Generic {
```
niladrih, line 118: Should we drop the error here?

niladrih (approved):

k8s/operators/src/pool/diskpool/crd/migration.rs
tiagolobocastro: Probably not, though that was already here

k8s/operators/src/pool/diskpool/crd/migration.rs
```diff
@@ -115,62 +113,44 @@ pub(crate) async fn migrate_to_latest(
...
-                let name = dsp.name_any();
-                let node = dsp.spec.node();
-                let disk = dsp.spec.disks();
-                let topology = dsp.spec.topology();
-                replace_with_latest(
-                    &k8s,
-                    &name,
-                    ns,
-                    Some(res_ver.clone()),
-                    DiskPoolSpec::new(node, disk, topology, None, None, None),
```
abhilashshetty04, line 130: @tiagolobocastro , I think these were set to None during replace as these were added as part of v1beta3. So, older scheme will not be having these configs. Migration code was added when we introduced v1beta1. 

But, we did thrash metadata (labels, annotations).

Abhinandan-Purkait (approved):

k8s/operators/src/pool/diskpool/crd/migration.rs
tiagolobocastro: Yeah.. but as I said since v2.10.0 we started running this code for same version, so that thrashes the other 3 fields there!!

tiagolobocastro: So is what I'm doing here of keeping the status as is, is ok or not?

abhilashshetty04: I think would be safer to default it. status will be updated anyway later on. We had breaking changes on v1alpha1 to v1beta1 migration. `status.state` was removed.

tiagolobocastro: But the type which we're getting is of the new type `DiskPool`, so it should already be "corrected" ?

abhilashshetty04: Yes, we do use new client (v1beta3) to list old CRs. Upto you if you want to update status it should be fine.

abhilashshetty04 (approved):

tiagolobocastro:
bors merge

PR #1074, opened 2026-03-23
feat: add offline node delete

niladrih:
Add DELETE /v0/nodes/{node_id} API for removing offline nodes and all
their resources (pools, replicas, nexuses, snapshots) without contacting
io-engine. Follows the pool purge pattern across all layers.

API:
  DELETE /v0/nodes/{id} with body {purge, accept, acceptVolumeLoss,
  acceptSnapshotLoss} returns NodeDeleteResult with aggregated volume
  and snapshot loss information.

CLI:
  kubectl mayastor delete node <id> --purge [--accept-data-loss]
  --accept-data-loss is shorthand for --accept-volume-loss
  --accept-snapshot-loss. The --yes prompt maps to accept=true.

Flow:
  1. Validate node is offline and cordoned
  2. Check purge/accept flags
  3. Mark node as Purging (crash-safe checkpoint)
  4. Auto-cordon each pool (replicas + snapshots) on behalf of the user
  5. Delegate to pool purge for each pool (replicas, snapshots, loss
     analysis all handled by existing pool purge code)
  6. On pool rejection: complete_destroy(Err) clears pending op but
     keeps Purging status for reconciler resume
  7. On success: complete_destroy removes node spec from store
  8. Remove node runtime wrapper

Crash recovery:
  - NodePurgeReconciler detects SpecStatus::Purging on startup, resumes
    with all accepts=true (validation was done before Purging was set)
  - PersistentStoreReconciler clears stale pending ops first
  - Individual pool crashes handled by existing pool purge reconciler

-------------------------

CLI:
  Add --show-impact to dry-run Node Purge

### DIFF
--- a/control-plane/agents/src/bin/core/controller/reconciler/node/purge.rs
+++ b/control-plane/agents/src/bin/core/controller/reconciler/node/purge.rs
@@ -1,0 +1,74 @@
+use crate::controller::{
+    resources::{operations::ResourceLifecycle, OperationGuardArc, ResourceUid},
+    task_poller::{PollContext, PollResult, PollTimer, PollerState, TaskPoller},
+};
+use stor_port::types::v0::{store::node::NodeSpec, transport::DestroyNode};
+
+/// Node purge reconciler — resumes interrupted node purge operations.
+///
+/// When a node delete (purge) is interrupted mid-way (e.g. control-plane restart),
+/// the node spec remains in `Purging` state. This reconciler detects such nodes
+/// and resumes the purge with all accepts set to true (matching the pool purge
+/// reconciler pattern).
+#[derive(Debug)]
+pub(super) struct NodePurgeReconciler {
+    counter: PollTimer,
+}
+
+impl NodePurgeReconciler {
+    /// Return a new `Self`.
+    pub(super) fn new() -> Self {
+        Self {
+            counter: PollTimer::from(1),
+        }
+    }
+}
+
+#[async_trait::async_trait]
+impl TaskPoller for NodePurgeReconciler {
+    async fn poll(&mut self, context: &PollContext) -> PollResult {
+        let nodes = context.specs().nodes();
+        let mut results = Vec::with_capacity(nodes.len());
+
+        for node in nodes {
+            if !node.status.purging() {
+                continue;
+            }
+            let mut node_guard = context.registry().specs().guarded_node(node.id()).await?;
+            results.push(resume_node_purge(context, &mut node_guard).await);
+        }
+        Self::squash_results(results)
+    }
+
+    async fn poll_timer(&mut self, _context: &PollContext) -> bool {
+        self.counter.poll()
+    }
+}
+
+/// Resume an interrupted node purge.
+///
+/// The node is already in Purging state. `purge_node` detects the resume case,
+/// skips pre-flight validation, and re-runs the destroy lifecycle with all
+/// accepts set to true.
+#[tracing::instrument(skip_all, level = "trace", fields(node.id = %node_guard.uid(), request.reconcile = true))]
+async fn resume_node_purge(
+    context: &PollContext,
+    node_guard: &mut OperationGuardArc<NodeSpec>,
+) -> PollResult {
+    let node_id = node_guard.uid().clone();
+
+    // Build a request with all accepts=true — validation was already done
+    // on the original request before the crash.
+    let request = DestroyNode::purge(node_id.clone())
+        .with_accept(true)
+        .with_accept_volume_loss(true)
+        .with_accept_snapshot_loss(true);
+
+    node_guard.destroy(context.registry(), &request).await?;
+
+    tracing::info!(
+        node.id = %node_id,
+        "Purging node reconciled successfully"
+    );
+    PollResult::Ok(PollerState::Idle)
+}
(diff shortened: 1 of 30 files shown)


control-plane/plugin/src/resources/node.rs
```diff
@@ -764,3 +764,84 @@ impl Label for Node {
...
+impl Delete for Node {
+    type ID = DeleteNodeArgs;
+
+    async fn del(id: &Self::ID, _ignore_not_found: bool, output: &OutputFormat) -> PluginResult {
+        // --accept-data-loss is shorthand for both volume and snapshot loss.
+        let accept_volume_loss = id.accept_volume_loss || id.accept_data_loss;
+        let accept_snapshot_loss = id.accept_snapshot_loss || id.accept_data_loss;
+        let body = openapi::models::DeleteNodeBody {
+            purge: Some(id.purge),
+            accept: Some(true),
```
Abhinandan-Purkait, line 816: Why is this hardcoded to true? Is this not driven by any cli?

control-plane/agents/src/bin/core/controller/reconciler/node/purge.rs
```diff
@@ -0,0 +1,129 @@
...
+        };
+        pool_guard
+            .cordon(context.registry(), cordon_request)
+            .await
+            .ok(); // Ignore if already cordoned.
+
+        let destroy_pool = DestroyPool::purge(pool.node(), pool.id().clone())
+            .with_accept(true)
+            .with_accept_volume_loss(true)
+            .with_accept_snapshot_loss(true);
```
Abhinandan-Purkait: How do you know the original request validated this for all pools? If the original request went half way through all pools and then it crashed?

control-plane/agents/src/bin/core/node/service.rs
```diff
@@ -450,4 +460,158 @@ impl Service {
...
+                }
+            }
+        }
+
+        // 7. All pools purged — mark node as Purging and complete deletion.
+        //    We mark Purging only after pools are gone so that a pre-flight
+        //    rejection (e.g. missing accept_volume_loss) doesn't leave the
+        //    node in a dirty Purging state.
+        //    In the resume case the node is already Purging — start_destroy_for_purge
+        //    re-logs the pending op so complete_destroy can commit.
```
Abhinandan-Purkait: Who will resume this if it fails above and we never marked the state purging until this point?

control-plane/agents/src/common/errors.rs
```diff
@@ -489,6 +489,25 @@ pub enum SvcError {
...
+    #[snafu(display("Cannot delete node {node_id}: node must be cordoned first"))]
+    NodeNotCordoned { node_id: NodeId },
+
+    #[snafu(display(
+        "Cannot delete node {node_id}: node has resources. Use purge=true to delete all resources"
+    ))]
+    NodeHasResources { node_id: NodeId },
+
+    #[snafu(display(
+        "Cannot delete node {node_id}: node has {pool_count} pool(s), confirm=true required"
```
Abhinandan-Purkait: Is that accept=true or confirm=true?

Abhinandan-Purkait: Also, I did not see that option in the plugin

niladrih: That's a miss, it'll be accept=true.

The plugin delete flow now has a Y/n question (even for volume delete). If that question is answered with a Yes (or user uses the `--yes` option), we assume it's an accept=true for pools may have replicas and the user accepts it.

control-plane/plugin/src/resources/node.rs
niladrih: The plugin delete flow now has a Y/n question (even for volume delete). If that question is answered with a Yes (or user uses the `--yes` option), we assume it's an accept=true for pools may have replicas and the user accepts it. If the code has reached this stage, it's already passed the Y/n questions and the user has answered with a yes.

tiagolobocastro:
I feel like I keep repeating myself but please add description to the commit and PR...

control-plane/agents/src/bin/core/node/service.rs
niladrih: The pools are set to SpecStatus::Purging, so once the agent-core restarts, the pool reconciler should be able to finish the purge.

control-plane/agents/src/common/errors.rs
niladrih: Made this change.

Abhinandan-Purkait: This errror message needs to be fixed right? accept=true

Abhinandan-Purkait (approved):
Looks mostly good to me, one thing I feel could be changed is error mapping, everything seems to result in CONFLICT.

control-plane/agents/src/common/errors.rs
niladrih: Yes I have done that?

control-plane/agents/src/bin/core/controller/reconciler/node/purge.rs
niladrih: Pools are marked SpecStatus::Purged. The pool reconciler will pick up pending purges after we recover from the crash.

niladrih:
This change has been added, and is no longer required.

PR #1079, opened 2026-04-07
feat(control-plane): add Purge logic to Node service

niladrih:
Flow:
  - Validate node is offline and cordoned
  - Check purge/accept flags
  - Mark node as Purging (crash-safe checkpoint)
  - Auto-cordon each pool (replicas + snapshots) on behalf of the user
  - Delegate to pool purge for each pool (replicas, snapshots, loss analysis all handled by existing pool purge code)
  - Disown replicas from the nexus children and remove the nexus specs
  - On pool rejection: complete_destroy(Err) clears pending op but keeps Purging status for reconciler resume
  - On success: complete_destroy removes node spec from store
  - Remove node runtime wrapper

### DIFF
--- a/control-plane/agents/src/bin/core/node/operations.rs
+++ b/control-plane/agents/src/bin/core/node/operations.rs
@@ -1,16 +1,20 @@
 use agents::errors::SvcError;
 
-use stor_port::types::v0::store::node::{DrainingVolumes, NodeOperation, NodeSpec};
+use stor_port::types::v0::{
+    store::node::{DrainingVolumes, NodeOperation, NodeSpec},
+    transport::{DestroyNode, DestroyPool, NodeDeleteResult, NodeId, NodeStatus},
+};
 
 use crate::controller::{
     registry::Registry,
     resources::{
-        operations::{ResourceCordon, ResourceDrain, ResourceLabel},
-        operations_helper::GuardedOperationsHelper,
+        operations::{ResourceCordon, ResourceDrain, ResourceLabel, ResourceLifecycle},
+        operations_helper::{analyze_snapshot_loss, analyze_volume_loss, GuardedOperationsHelper},
         OperationGuardArc,
     },
 };
-use std::collections::HashMap;
+use grpc::operations::pool::traits::PoolCordonRequest;
+use std::collections::{HashMap, HashSet};
 
 /// Resource Cordon Operations.
 #[async_trait::async_trait]
@@ -128,6 +132,39 @@ impl ResourceDrain for OperationGuardArc<NodeSpec> {
     }
 }
 
+#[async_trait::async_trait]
+impl ResourceLifecycle for OperationGuardArc<NodeSpec> {
+    type Create = ();
+    type CreateOutput = ();
+    type Destroy = DestroyNode;
+    type DestroyOutput = NodeDeleteResult;
+
+    async fn create(
+        _registry: &Registry,
+        _request: &Self::Create,
+    ) -> Result<Self::CreateOutput, SvcError> {
+        unimplemented!(
+            "Nodes self-register via keep-alives; they are not created via ResourceLifecycle"
+        )
+    }
+
+    async fn destroy(
+        &mut self,
+        registry: &Registry,
+        request: &Self::Destroy,
+    ) -> Result<Self::DestroyOutput, SvcError> {
+        let purge_result = self.purge_node(registry, request).await?;
+
+        // Remove the node's runtime state (wrapper).
+        {
+            let mut nodes = registry.nodes().write().await;
+            nodes.remove(request.id.as_ref());
+        }
+
+        Ok(purge_result)
+    }
+}
+
 /// Node drain Operations.
 impl OperationGuardArc<NodeSpec> {
     /// Drain the set of draining volumes to the stored set.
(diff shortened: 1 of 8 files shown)


Abhinandan-Purkait (approved):

control-plane/agents/src/bin/core/node/service.rs
```diff
@@ -456,4 +457,205 @@ impl Service {
...
+        use crate::controller::resources::{
+            operations::{ResourceCordon, ResourceLifecycle},
+            operations_helper::{GuardedOperationsHelper, OperationSequenceGuard},
+        };
+        use crate::pool::pool_operations::{analyze_snapshot_loss, analyze_volume_loss};
+        use grpc::operations::pool::traits::PoolCordonRequest;
+        use std::collections::HashSet;
+        use stor_port::types::v0::transport::DestroyPool;
+
+        let node_id = &request.id;
```
tiagolobocastro, line 471: this implementation should not be in the service
this should be in a destroy operation, just like for pools

control-plane/agents/src/bin/core/node/service.rs
```diff
@@ -456,4 +457,205 @@ impl Service {
...
+        //    pool purge will reject if loss exists and isn't accepted.
+        //    We cordon each pool (replicas + snapshots) before purging — the
+        //    user has already cordoned the node, so we handle pool cordoning
+        //    on their behalf.
+        //    Re-collect pools here since in the resume case the pre-flight
+        //    collection was skipped.
+        let node_pools = self
+            .registry
+            .get_node_opt_pools(Some(node_id.clone()))
+            .await?;
```
tiagolobocastro: we can't use `?` right, need to clean up operation in case of any failure?

control-plane/agents/src/bin/core/node/service.rs
```diff
@@ -456,4 +457,205 @@ impl Service {
...
+            // Cordon pool for purge (replicas + snapshots) if not already cordoned.
+            let cordon_request = PoolCordonRequest {
+                node_id: None,
+                pool_id: pool.id().clone(),
+                replicas: true,
+                snapshots: true,
+                restores: false,
+                import: false,
+            };
+            pool_guard.cordon(&self.registry, cordon_request).await.ok(); // Ignore if already cordoned.
```
tiagolobocastro: this is ignoring other failures?

control-plane/agents/src/bin/core/pool/pool_operations.rs
```diff
@@ -598,48 +556,118 @@ impl OperationGuardArc<PoolSpec> {
...
 
-        for vol_snapshot in registry.specs().volume_snapshots_rsc() {
-            let vol_snap = vol_snapshot.lock();
-            let vol_snap_id = vol_snap.spec().uuid().clone();
+/// Analyze volume loss impact across one or more pools being purged.
+///
+/// For each volume that has replicas on the given pool(s), fetches the
+/// `Volume` object and uses its `replica_topology` to determine how many
+/// healthy replicas exist before and after the deletion.
+pub(crate) async fn analyze_volume_loss(
```
tiagolobocastro: This seems wrong, these should probably reside be in the operations_helper, not here

control-plane/agents/src/bin/core/node/service.rs
niladrih: Yes!!!!!!! That's a stupid miss.

control-plane/agents/src/bin/core/node/service.rs
niladrih: I am logging errors. And leaving it up to pool purge to fail and return an error.

control-plane/agents/src/bin/core/pool/pool_operations.rs
niladrih: Moved it to controller::resources::operations_helper.rs. Added a thin helper on pool::operations_helper.rs inside the OperationGuardArc<PoolSpec>.

control-plane/agents/src/bin/core/node/service.rs
niladrih: Put Node purge inside a ResourceLifecycle destroy implementation.

control-plane/agents/src/bin/core/node/operations.rs
```diff
@@ -185,4 +217,204 @@ impl OperationGuardArc<NodeSpec> {
...
+                Ok(result) => (Some(result), Ok(())),
+                Err(error) => (None, Err(error)),
+            };
+
+        // Always complete the op:
+        // - Ok(()) deletes the node spec from etcd and memory.
+        // - Err(error) clears the pending op, keeps Purging for reconciler retry.
+        self.complete_destroy(destroy_result, registry).await?;
+
+        Ok(node_result.expect("purge_node_resources succeeded"))
```
tiagolobocastro: Let's not add any panics please
Also seems like we made things harder, won't this just work?

```suggestion
        let result = Self::purge_node_resources(registry, &node_id, request).await;

        // Always complete the op:
        // - Ok(_) deletes the node spec from etcd and memory.
        // - Err(error) clears the pending op, keeps Purging for reconciler retry.
        self.complete_destroy(destroy_result, registry).await
```

control-plane/agents/src/bin/core/node/operations.rs
```diff
@@ -185,4 +217,204 @@ impl OperationGuardArc<NodeSpec> {
...
+        // These nexuses can never be recreated since the node is gone.
+        let nexuses = registry.specs().nexuses();
+        let mut nexuses_deleted = 0usize;
+        for nexus_rsc in &nexuses {
+            if nexus_rsc.lock().node != *node_id {
+                continue;
+            }
+            let mut nexus_guard = nexus_rsc.operation_guard_wait().await?;
+            registry.specs().on_delete_disown_replicas(&nexus_guard);
+            nexus_guard.start_destroy_for_purge(registry).await?;
```
tiagolobocastro: wait, is this safe, won't this cause issues for volumes if we take the nexus without "telling" the volume?

control-plane/agents/src/bin/core/node/service.rs
```diff
@@ -456,4 +458,27 @@ impl Service {
...
+        request: &DestroyNode,
+    ) -> Result<NodeDeleteResult, SvcError> {
+        let node_id = &request.id;
+
+        let mut node_guard = self.registry.specs().guarded_node(node_id).await?;
+        let result = node_guard.destroy(&self.registry, request).await?;
+
+        // Remove the node's runtime state (wrapper).
+        {
+            let mut nodes = self.registry.nodes().write().await;
```
tiagolobocastro: hmm should the destroy itself do this?

tiagolobocastro (commented):
lgtm, but I've got some doubts on the nexus destroy logic

control-plane/agents/src/bin/core/node/service.rs
niladrih: Moved this into destroy.

control-plane/agents/src/bin/core/node/operations.rs
niladrih: Simplified this.

control-plane/agents/src/bin/core/node/operations.rs
niladrih: Removed nexus delete. Requires more testing to know how a volume is affected when a node is lost.

tiagolobocastro (approved):

niladrih:
bors merge

PR #1080, opened 2026-04-09
feat: add simple backoff to failed pool imports

tiagolobocastro:
feat: add simple backoff to failed pool imports
    
    Avoids retrying pool import too often, especially if t here's something
    bad with the backend.
    
    Ideally this should be solved entirely by the proving, but as of now the
    probing doesn't do superblock probing...
    
---

    test: fix race condition on missing pool reconciler
    
    Kill the io-engine container to ensure there's no race between
    restoring the correct backend disk and checking for pool status.
    
---

    chore: limit GOMAXPROCS for jaeger
    
    Limit the memory used during tests.

### DIFF
--- a/control-plane/agents/src/bin/core/controller/reconciler/pool/mod.rs
+++ b/control-plane/agents/src/bin/core/controller/reconciler/pool/mod.rs
@@ -145,6 +145,10 @@ async fn missing_pool_state_reconciler(
             Ok(Ok(node)) => node,
         };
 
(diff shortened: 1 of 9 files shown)


Abhinandan-Purkait (approved):

niladrih (approved):
LGTM

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

PR #1081, opened 2026-04-10
Node delete reconciler and tests

niladrih:
feat(agent/core): add node purge reconciler

This adds a reconciler to handle Node resources in SpecStatus::Purging.

---

test: add node purge tests

### DIFF
--- a/control-plane/agents/src/bin/core/controller/reconciler/node/purge.rs
+++ b/control-plane/agents/src/bin/core/controller/reconciler/node/purge.rs
@@ -1,0 +1,74 @@
+use crate::controller::{
+    resources::{operations::ResourceLifecycle, OperationGuardArc, ResourceUid},
+    task_poller::{PollContext, PollResult, PollTimer, PollerState, TaskPoller},
+};
+use stor_port::types::v0::{store::node::NodeSpec, transport::DestroyNode};
+
+/// Node purge reconciler — resumes interrupted node purge operations.
+///
+/// When a node delete (purge) is interrupted mid-way (e.g. control-plane restart),
+/// the node spec remains in `Purging` state. This reconciler detects such nodes
+/// and resumes the purge with all accepts set to true (matching the pool purge
+/// reconciler pattern).
+#[derive(Debug)]
+pub(super) struct NodePurgeReconciler {
+    counter: PollTimer,
+}
+
+impl NodePurgeReconciler {
+    /// Return a new `Self`.
+    pub(super) fn new() -> Self {
+        Self {
+            counter: PollTimer::from(1),
+        }
+    }
+}
+
+#[async_trait::async_trait]
+impl TaskPoller for NodePurgeReconciler {
+    async fn poll(&mut self, context: &PollContext) -> PollResult {
+        let nodes = context.specs().nodes();
+        let mut results = Vec::with_capacity(nodes.len());
+
+        for node in nodes {
+            if !node.status.purging() {
+                continue;
+            }
+            let mut node_guard = context.registry().specs().guarded_node(node.id()).await?;
+            results.push(resume_node_purge(context, &mut node_guard).await);
+        }
+        Self::squash_results(results)
+    }
+
+    async fn poll_timer(&mut self, _context: &PollContext) -> bool {
+        self.counter.poll()
+    }
+}
+
+/// Resume an interrupted node purge.
+///
+/// The node is already in Purging state. `purge_node` detects the resume case,
+/// skips pre-flight validation, and re-runs the destroy lifecycle with all
+/// accepts set to true.
+#[tracing::instrument(skip_all, level = "trace", fields(node.id = %node_guard.uid(), request.reconcile = true))]
+async fn resume_node_purge(
+    context: &PollContext,
+    node_guard: &mut OperationGuardArc<NodeSpec>,
+) -> PollResult {
+    let node_id = node_guard.uid().clone();
+
+    // Build a request with all accepts=true — validation was already done
+    // on the original request before the crash.
+    let request = DestroyNode::purge(node_id.clone())
+        .with_accept(true)
+        .with_accept_volume_loss(true)
+        .with_accept_snapshot_loss(true);
+
+    node_guard.destroy(context.registry(), &request).await?;
+
+    tracing::info!(
+        node.id = %node_id,
+        "Purging node reconciled successfully"
+    );
+    PollResult::Ok(PollerState::Idle)
+}
--- a/control-plane/agents/src/bin/core/node/specs.rs
+++ b/control-plane/agents/src/bin/core/node/specs.rs
@@ -137,6 +137,28 @@ impl ResourceSpecsLocked {
             .collect()
     }
 
+    /// Worker that reconciles dirty NodeSpec's with the persistent store.
+    /// This is useful when node operations are performed but we fail to
+    /// update the spec with the persistent store.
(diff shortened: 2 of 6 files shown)


control-plane/agents/src/bin/core/controller/reconciler/node/purge.rs
```diff
@@ -0,0 +1,76 @@
...
+        self.counter.poll()
+    }
+}
+
+/// Resume an interrupted node purge.
+///
+/// The node is already in Purging state. `purge_node` detects the resume case,
+/// skips pre-flight validation, and re-runs the destroy lifecycle with all
+/// accepts set to true.
+#[tracing::instrument(skip(context), level = "trace", fields(node.id = %node_id, request.reconcile = true))]
```
tiagolobocastro:
```suggestion
#[tracing::instrument(skip_all, level = "trace", fields(node.id = %node_id, request.reconcile = true))]
```

control-plane/agents/src/bin/core/controller/reconciler/node/purge.rs
```diff
@@ -0,0 +1,76 @@
...
+    let request = DestroyNode::purge(node_id.clone())
+        .with_accept(true)
+        .with_accept_volume_loss(true)
+        .with_accept_snapshot_loss(true);
+
+    node_guard.purge_node(context.registry(), &request).await?;
+
+    // Remove runtime wrapper.
+    {
+        let mut nodes = context.registry().nodes().write().await;
```
tiagolobocastro: Isn't purge_node doing this already? If so then it should be, so the caller doesn't need to worry about this

control-plane/agents/src/bin/core/node/specs.rs
```diff
@@ -137,6 +137,28 @@ impl ResourceSpecsLocked {
...
+    /// Worker that reconciles dirty NodeSpec's with the persistent store.
+    /// This is useful when node operations are performed but we fail to
+    /// update the spec with the persistent store.
+    pub(crate) async fn reconcile_dirty_nodes(&self, registry: &Registry) -> bool {
+        let mut pending_ops = false;
+
+        let nodes = self.nodes_rsc();
+        for node in nodes {
+            // Skip nodes with no pending operation to avoid unnecessary guard
+            // acquisition. This keeps the reconciler zero-cost for healthy nodes.
```
tiagolobocastro: I wouldn't say it's quite zero cost, you have atomics for the rsc and then the lock here

control-plane/agents/src/bin/core/tests/node/purge.rs
```diff
@@ -0,0 +1,399 @@
...
+        .expect("Failed to write Purging node spec to etcd");
+
+    // 5. Restart core agent — reconciler should detect Purging and resume the purge.
+    cluster
+        .restart_core_with_liveness(None)
+        .await
+        .expect("Core agent should restart and become live");
+
+    // 6. Wait for the reconciler to complete the purge (node spec gone).
+    let timeout = Duration::from_secs(15);
```
tiagolobocastro: the timeout above was 5, is there a reason here for taking so long?

control-plane/agents/src/bin/core/tests/node/purge.rs
```diff
@@ -0,0 +1,399 @@
...
+        }
+        assert!(
+            start.elapsed() < timeout,
+            "Timed out waiting for reconciler to complete node purge"
+        );
+        tokio::time::sleep(Duration::from_millis(100)).await;
+    }
+
+    // 7. Verify pools are gone.
+    let timeout = Duration::from_secs(5);
```
tiagolobocastro: Why are all these below using a loop, shouldn't the outcome already have been reached?

control-plane/agents/src/bin/core/tests/node/purge.rs
```diff
@@ -0,0 +1,399 @@
...
+            Err(e) if e.kind == ReplyErrorKind::NotFound => break,
+            Ok(r) if r.clone().into_inner().is_empty() => break,
+            _ => {}
+        }
+        assert!(
+            start.elapsed() < timeout,
+            "Timed out waiting for replicas to be cleaned up after node purge"
+        );
+        tokio::time::sleep(Duration::from_millis(100)).await;
+    }
```
tiagolobocastro, line 492: can we now delete the volume?

control-plane/agents/src/bin/core/tests/node/purge.rs
```diff
@@ -0,0 +1,399 @@
...
+                id: pool_id.clone(),
+                disks: vec!["malloc:///resume_disk?size_mb=100".into()],
+                ..Default::default()
+            },
+            None,
+        )
+        .await
+        .expect("Pool creation should succeed");
+
+    let vol_id = VolumeId::new();
```
tiagolobocastro, line 315: can we have another volume, with say 2 replicas (another replica on another node)

control-plane/agents/src/bin/core/controller/reconciler/node/purge.rs
niladrih: destroy is doing this. I'll have to call destroy here, instead of purge_node. Right.

control-plane/agents/src/bin/core/tests/node/purge.rs
niladrih: No, I will reduce the timeout.

control-plane/agents/src/bin/core/tests/node/purge.rs
niladrih: That's right, it was meant to be defensive. I've removed the loops. Just asserts.

control-plane/agents/src/bin/core/tests/node/purge.rs
niladrih: Added a volume delete step.

control-plane/agents/src/bin/core/tests/node/purge.rs
niladrih: Added another 2 replica volume.

control-plane/agents/src/bin/core/controller/reconciler/node/purge.rs
```diff
@@ -0,0 +1,73 @@
...
+impl TaskPoller for NodePurgeReconciler {
+    async fn poll(&mut self, context: &PollContext) -> PollResult {
+        let nodes = context.specs().nodes();
+        let mut results = Vec::with_capacity(nodes.len());
+
+        for node in nodes {
+            if !node.status.purging() {
+                continue;
+            }
+            results.push(resume_node_purge(context, node.id().clone()).await);
```
tiagolobocastro: Can't we just pass the node btw?

tiagolobocastro (approved):

control-plane/agents/src/bin/core/controller/reconciler/node/purge.rs
niladrih: If by that you mean pass the node instead of the node_id, then no. Because inside resume_node_purge we'd have to get a guarded_node, so the whole node is of little use.

niladrih: Oh right, I see how the snapshot reconciler is doing it. It's an OperationGuardArc<NodeSpec>, so that's possibly a better fit IMO. Making the change.

niladrih:
bors merge

PR #1082, opened 2026-04-10
feat(plugin): add node delete command

niladrih:
Changes:
- Move the show_impact code outside pool.rs so that we may be able to use
  it with node delete as well.
- Add new error types for pool and node delete and add hint mapping for them.
- Add node delete plugin change

### DIFF
--- a/control-plane/plugin/src/resources/error.rs
+++ b/control-plane/plugin/src/resources/error.rs
@@ -71,6 +71,11 @@ pub enum Error {
         id: String,
         source: openapi::tower::client::Error<openapi::models::RestJsonError>,
     },
+    #[snafu(display("Failed to delete node {id}. Error {source}"))]
+    DeleteNodeError {
+        id: String,
+        source: openapi::tower::client::Error<openapi::models::RestJsonError>,
+    },
     /// Error when get pool request fails.
     #[snafu(display("No state for pool {id}. Please verify if node is online"))]
     PoolStateError { id: String },
@@ -152,6 +157,9 @@ pub enum Error {
         id: String,
         source: openapi::tower::client::Error<openapi::models::RestJsonError>,
     },
+    /// A purge/delete precondition was not met.
+    #[snafu(display("{reason}"))]
+    Purge { reason: PurgeReason },
 }
 
 /// Errors related to label topology formats.
@@ -208,6 +216,77 @@ pub enum OpError {
         id: String,
         source: openapi::tower::client::Error<openapi::models::RestJsonError>,
     },
+}
+
+/// Known purge/delete precondition failures, with user-facing messages
+/// that reference the CLI flags needed to resolve the issue.
+#[derive(Debug, strum_macros::Display)]
+pub enum PurgeReason {
+    #[strum(to_string = "Node is online. Only offline nodes can be deleted.")]
+    NodeIsOnline,
+    #[strum(to_string = "Node must be cordoned first. Use: cordon node <id> <label>")]
+    NodeNotCordoned,
+    #[strum(
+        to_string = "Node has resources. Use --purge to force-remove the node and all its resources."
+    )]
+    NodeHasResources,
+    #[strum(to_string = "Node has pools with data. Confirm with --yes to proceed.")]
+    NodePurgeAcceptRequired,
+    #[strum(
+        to_string = "Volumes would lose their last healthy replica. Use --accept-volume-loss or --accept-data-loss to proceed."
+    )]
+    NodePurgeVolumeLoss,
+    #[strum(
+        to_string = "Snapshots would lose their last replica snapshot. Use --accept-snapshot-loss or --accept-data-loss to proceed."
+    )]
+    NodePurgeSnapshotLoss,
+    #[strum(
+        to_string = "Pool state is not Offline or Unknown. Only pools with Offline or Unknown state can be purged."
+    )]
+    PoolNotPurgeable,
+    #[strum(
+        to_string = "Pool must be cordoned first. Use: cordon pool <id> --replicas --snapshots"
+    )]
+    PoolNotCordoned,
+    #[strum(
+        to_string = "Pool cordon must block both replicas and snapshots. Use: cordon pool <id> --replicas --snapshots"
+    )]
+    PoolCordonInsufficient,
+    #[strum(to_string = "Pool has replicas. Confirm with --yes to proceed.")]
+    PoolPurgeAcceptRequired,
+    #[strum(
+        to_string = "Volumes would lose their last healthy replica. Use --accept-volume-loss or --accept-data-loss to proceed."
+    )]
+    PoolPurgeVolumeLoss,
+    #[strum(
+        to_string = "Snapshots would lose their last replica snapshot. Use --accept-snapshot-loss or --accept-data-loss to proceed."
+    )]
+    PoolPurgeSnapshotLoss,
+}
+
+impl PurgeReason {
+    /// Try to extract a known purge reason from an API error.
+    pub fn from_api_error(
+        source: &openapi::tower::client::Error<openapi::models::RestJsonError>,
+    ) -> Option<Self> {
+        use openapi::models::rest_json_error::Kind;
+        let kind = source.error_body().map(|b| b.kind)?;
+        match kind {
+            Kind::NodeIsOnline => Some(Self::NodeIsOnline),
+            Kind::NodeNotCordoned => Some(Self::NodeNotCordoned),
+            Kind::NodeHasResources => Some(Self::NodeHasResources),
+            Kind::NodePurgeAcceptRequired => Some(Self::NodePurgeAcceptRequired),
+            Kind::NodePurgeVolumeLossAcceptRequired => Some(Self::NodePurgeVolumeLoss),
+            Kind::NodePurgeSnapshotLossAcceptRequired => Some(Self::NodePurgeSnapshotLoss),
+            Kind::PoolNotPurgeable => Some(Self::PoolNotPurgeable),
+            Kind::PoolNotCordoned => Some(Self::PoolNotCordoned),
+            Kind::PoolCordonInsufficient => Some(Self::PoolCordonInsufficient),
+            Kind::PoolPurgeAcceptRequired => Some(Self::PoolPurgeAcceptRequired),
+            Kind::PoolPurgeVolumeLossAcceptRequired => Some(Self::PoolPurgeVolumeLoss),
+            Kind::PoolPurgeSnapshotLossAcceptRequired => Some(Self::PoolPurgeSnapshotLoss),
+            _ => None,
+        }
+    }
 }
 
 impl From<TopologyError> for Error {
--- a/control-plane/plugin/src/resources/impact.rs
+++ b/control-plane/plugin/src/resources/impact.rs
@@ -1,0 +1,343 @@
+use crate::{
+    resources::{
+        error::Error,
+        utils::{optional_cell, print_table, CreateRow, GetHeaderRow, OutputFormat},
+        NodeId, PoolId, VolumeId,
+    },
+    rest_wrapper::RestClient,
+};
+use openapi::models::{self, PoolCordonDrain};
+use prettytable::Row;
+use serde::Serialize;
+
+/// Cordon state for purge eligibility.
+#[derive(Serialize, Debug, Clone)]
+pub enum CordonState {
+    /// Pool has no cordon in place.
+    NotCordoned,
+    /// Pool is cordoned but does not block both replicas and snapshots.
+    Insufficient,
+    /// Pool is cordoned and blocks both replica and snapshot scheduling.
+    Ready,
+}
+
+impl Default for CordonState {
+    fn default() -> Self {
+        Self::NotCordoned
+    }
+}
+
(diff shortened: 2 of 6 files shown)


control-plane/plugin/src/resources/node.rs
```diff
@@ -764,3 +764,313 @@ impl Label for Node {
...
+    /// Required when --purge would cause snapshot loss.
+    #[arg(long)]
+    accept_snapshot_loss: bool,
+}
+
+#[async_trait(?Send)]
+impl Delete for Node {
+    type ID = DeleteNodeArgs;
+
+    async fn del(id: &Self::ID, _ignore_not_found: bool, output: &OutputFormat) -> PluginResult {
```
tiagolobocastro: not handling `ignore_not_found` ?

control-plane/plugin/src/resources/node.rs
```diff
@@ -764,3 +764,313 @@ impl Label for Node {
...
+            Err(source) => {
+                // Translate core agent errors into user-friendly messages with CLI flags.
+                use openapi::models::rest_json_error::Kind;
+                if let Some(kind) = source.error_body().map(|b| b.kind) {
+                    let hint = match kind {
+                        Kind::NodeIsOnline => {
+                            Some("Node is online. Only offline nodes can be deleted.")
+                        }
+                        Kind::NodeNotCordoned => {
+                            Some("Node must be cordoned first. Use: kubectl mayastor cordon node <id> <label>")
```
tiagolobocastro: `kubectl mayastor` is not the correct command for other plugins like the umbrella

```suggestion
                            Some("Node must be cordoned first. Use: cordon node <id> <label>")
```

control-plane/plugin/src/resources/node.rs
```diff
@@ -764,3 +764,313 @@ impl Label for Node {
...
+                .then(|| format!("{} snapshot(s)", self.snapshot_loss.snapshots.len())),
+        );
+
+        row![self.node_id, volume_loss, snapshot_loss]
+    }
+}
+
+/// Per-pool impact detail within a node purge.
+#[derive(Serialize, Debug, Clone)]
+pub struct PoolImpact {
```
tiagolobocastro: Can't we reuse the impact from the pool module?

control-plane/plugin/src/resources/node.rs
```diff
@@ -764,3 +764,313 @@ impl Label for Node {
...
+    /// Volumes affected by purging this pool.
+    pub affected_volumes: Vec<super::VolumeId>,
+}
+
+/// Impact analysis for a node purge operation.
+///
+/// Shown when using `--show-impact` to preview what would happen before
+/// actually purging the node and all its pools.
+#[derive(Serialize, Debug, Clone)]
+pub struct NodePurgeImpact {
```
tiagolobocastro: Do these need to be pub?

control-plane/plugin/src/resources/node.rs
```diff
@@ -764,3 +764,313 @@ impl Label for Node {
...
+            ready_for_purge,
+        };
+
+        // Print node-level summary.
+        print_table(output, impact);
+
+        // Print per-pool breakdown if there are pools.
+        if !pool_impacts.is_empty() {
+            println!();
+            println!("Per-pool breakdown:");
```
tiagolobocastro: This doesn't make sense if the output is json/yaml

control-plane/plugin/src/resources/node.rs
```diff
@@ -764,3 +764,313 @@ impl Label for Node {
...
+            Ok(response) => {
+                utils::print_table(output, response.into_body());
+                Ok(())
+            }
+            Err(source) => {
+                // Translate core agent errors into user-friendly messages with CLI flags.
+                use openapi::models::rest_json_error::Kind;
+                if let Some(kind) = source.error_body().map(|b| b.kind) {
+                    let hint = match kind {
+                        Kind::NodeIsOnline => {
```
tiagolobocastro: These should be in an error type, which is then returned

control-plane/plugin/src/resources/node.rs
niladrih: Added this.

control-plane/plugin/src/resources/node.rs
niladrih: Removed this way of handling impact.

control-plane/plugin/src/resources/node.rs
niladrih: Changed this into a Snafu type

control-plane/plugin/src/resources/node.rs
niladrih: Moved this into an impact module.

control-plane/plugin/src/resources/node.rs
niladrih: Printing if only -o option is none or table mode.

control-plane/plugin/src/resources/impact.rs
```diff
@@ -0,0 +1,344 @@
...
+
+    // Print node-level summary.
+    print_table(output, impact);
+
+    // Print per-pool breakdown only in table mode; for json/yaml the pools
+    // are already included in the serialised NodePurgeImpact.
+    if let OutputFormat::None = output {
+        if !pool_impacts.is_empty() {
+            println!();
+            println!("Per-pool breakdown:");
```
tiagolobocastro: Are these necessary? (I don't know how it looks like...)

control-plane/plugin/src/resources/impact.rs
```diff
@@ -0,0 +1,344 @@
...
+}
+
+impl GetHeaderRow for PoolPurgeImpact {
+    fn get_header_row(&self) -> Row {
+        row![
+            "POOL",
+            "STATUS",
+            "CORDON",
+            "REPLICAS",
+            "AFFECTED_VOLUMES",
```
tiagolobocastro:
```suggestion
            "AFFECTED-VOLUMES",
```
or
```suggestion
            "VOLUMES",
```

control-plane/plugin/src/resources/impact.rs
```diff
@@ -0,0 +1,344 @@
...
+    pub status: models::PoolStatus,
+    /// Number of replicas residing on this pool.
+    pub replica_count: usize,
+    /// Volumes affected by purging this pool.
+    pub affected_volumes: Vec<VolumeId>,
+}
+
+impl GetHeaderRow for PoolImpact {
+    fn get_header_row(&self) -> Row {
+        row!["POOL", "STATUS", "REPLICAS", "AFFECTED_VOLUMES"]
```
tiagolobocastro:
```suggestion
        row!["POOL", "STATUS", "REPLICAS", "AFFECTED-VOLUMES"]
```
or simply
```suggestion
        row!["POOL", "STATUS", "REPLICAS", "VOLUMES"]
```

control-plane/plugin/src/resources/impact.rs
```diff
@@ -0,0 +1,344 @@
...
+
+impl GetHeaderRow for NodePurgeImpact {
+    fn get_header_row(&self) -> Row {
+        row![
+            "NODE",
+            "STATUS",
+            "CORDONED",
+            "POOLS",
+            "TOTAL_REPLICAS",
+            "AFFECTED_VOLUMES",
```
tiagolobocastro: I'll stop these, you catch the drift
Also don't use `_`, prefer `-` or making name shorter

```suggestion
            "REPLICAS",
            "VOLUMES",
```

control-plane/plugin/src/resources/node.rs
```diff
@@ -764,3 +764,102 @@ impl Label for Node {
...
+                    }),
+                }
+            }
+        }
+    }
+}
+
+impl GetHeaderRow for models::NodeDeleteResult {
+    fn get_header_row(&self) -> Row {
+        row!["NODE", "VOLUME_LOSS", "SNAPSHOT_LOSS"]
```
```suggestion
        row!["NODE", "VOLUME-LOSS", "SNAPSHOT-LOSS"]
```

control-plane/plugin/src/resources/error.rs
```diff
@@ -210,6 +222,54 @@ pub enum OpError {
...
 
+/// Check whether an API error is a known purge/delete precondition failure
+/// and return the matching user-facing hint.
+pub fn purge_hint(
+    source: &openapi::tower::client::Error<openapi::models::RestJsonError>,
+) -> Option<&'static str> {
+    use openapi::models::rest_json_error::Kind;
+    let kind = source.error_body().map(|b| b.kind)?;
+    match kind {
+        Kind::NodeIsOnline => Some("Node is online. Only offline nodes can be deleted."),
```
tiagolobocastro: Can't this go in an error type itself?

control-plane/plugin/src/resources/error.rs
```diff
@@ -152,6 +157,13 @@ pub enum Error {
         id: String,
         source: openapi::tower::client::Error<openapi::models::RestJsonError>,
     },
+    /// A purge/delete precondition was not met. The hint tells the user
+    /// which CLI flags to use.
+    #[snafu(display("{hint}\n{source}"))]
+    PurgeError {
+        hint: &'static str,
+        source: openapi::tower::client::Error<openapi::models::RestJsonError>,
```
tiagolobocastro: I'm confused, if we have hint, why do we need the other error?

niladrih: We should have it, in case it's something we're missing in our assumptions of the error types, IMO.

control-plane/plugin/src/resources/impact.rs
niladrih: This is not necessary. It's something I meant to remove and missed.

control-plane/plugin/src/resources/impact.rs
niladrih: Going with VOLUMES.

control-plane/plugin/src/resources/error.rs
niladrih: Yes. I've added it to the enum match sequence.

control-plane/plugin/src/resources/error.rs
niladrih: On second thought, it's look prettier. Removed the source error.

control-plane/plugin/src/resources/impact.rs
```diff
@@ -0,0 +1,328 @@
...
+
+/// Fetch all volumes from the REST API (used by impact analysis).
+async fn fetch_all_volumes() -> Vec<models::Volume> {
+    RestClient::client()
+        .volumes_api()
+        .get_volumes(0, None, None)
+        .await
+        .map(|r| r.into_body().entries)
+        .unwrap_or_default()
+}
```
Abhinandan-Purkait: Please use pagination, this will overwhelm on large clusters.

control-plane/plugin/src/resources/impact.rs
```diff
@@ -0,0 +1,328 @@
...
+
+    // Print per-pool breakdown only in table mode; for json/yaml the pools
+    // are already included in the serialised NodePurgeImpact.
+    if let OutputFormat::None = output {
+        if !pool_impacts.is_empty() {
+            println!();
+            print_table(output, pool_impacts);
+        }
+    }
+
```
Abhinandan-Purkait: Can't we have an equivalent json or yaml output for the same?

tiagolobocastro (approved):

control-plane/plugin/src/resources/node.rs
```diff
@@ -764,3 +764,102 @@ impl Label for Node {
...
+        if id.show_impact {
+            return super::impact::show_node_impact(&id.node_id, output).await;
+        }
+
+        // --accept-data-loss is shorthand for both volume and snapshot loss.
+        let accept_volume_loss = id.accept_volume_loss || id.accept_data_loss;
+        let accept_snapshot_loss = id.accept_snapshot_loss || id.accept_data_loss;
+        let body = openapi::models::DeleteNodeBody {
+            purge: Some(id.purge),
+            accept: Some(true),
```
Abhinandan-Purkait, line 819: Why is this hardcoded to true? Does this field have not intent then? I don't see any ovveride for this

Abhinandan-Purkait (approved):

control-plane/plugin/src/resources/impact.rs
niladrih: Yes, indeed.

control-plane/plugin/src/resources/impact.rs
niladrih: Yes, we could. Adding that.

control-plane/plugin/src/resources/node.rs
niladrih: If we've come this far into the code, this means that we already have a confirmation that the user wants to go ahead with pool delete. We're using the same confirmation and using it for the 'accept losing repicas from pools' accept=true. I'll add a comment for this, it certainly looks like we're doing something incorrect here.

niladrih:
bors merge

niladrih:
bors merge

niladrih:
bors merge

niladrih:
bors cancel

niladrih:
bors merge

PR #1084, opened 2026-04-15
deployer: csi-node-2 as a qemu vm for rwx testing 

tiagolobocastro:
test(rwx): add basic live migration simulation
    
    Create and stage volume in csi-node-1.
    
    Then we modify volume to RWX by publishing and then staging
    to csi-node-2 (which is the qemu vm).
    
    todo: add some testing while both are connected
    
    Then we unstage from csi-node-1.
    
    todo: add some testing when only csi-node-2 is connected.
    
    Then we unstage from csi-node-2.
    
---

    feat(deployer): csi-node-2 as a qemu vm
    
    Adds --rwx-vm option to the deployer which deploys csi-node-2 and
    agent-ha-node-2 as systemd services inside a vm.
    
    If clean is set then we ensure vm teardown by using libc::PR_SET_PDEATHSIG.
    The deployer cleanup script also looks up and kills should these leak.
    
    The agent-ha-node and csi-node deployer components will attempt to
    wait for the services to be ready.
    This allows us to use this within a deployer-cluster test.
    
---

    test: add nix-vm for running as csi-node-2
    
    Define a vm with nvme-tcp, csi-node and agent-ha-node as systemd services.
    This can be used to simulate csi-node/ha-node combo running on the same node.
    
    This will be used by the rwx tests as we need the csi-node-2 to run in a different
    kernel than csi-node-1.
    
    This approach is fairly strict, with a bunch of hard-coded ports which is far from
    ideal.
    It should however be sufficient since we only intend to use this for the rwx test
    case, ensuring multiple connected hosts don't get broken.
    This also allows us to test temporary connection issues using iptables ie.
    
    To build and run the vm:
    1. cd deployer/misc/rwx-vm
    WORKSPACE_ROOT=$src
    on nixos:
    nixos-rebuild build-vm -I nixos-config=./csi-node-2.nix
    non non-nixos:
    nix-build -f '<nixpkgs/nixos>' -I nixos-config=./csi-node-2.nix config.system.build.vm
    --extra-experimental-features nix-command
    2. ./result/bin/run-nixos-vm -nographic
    
    The workspace root is required because that's how we can access the csi-node and
    agent-ha-node binaries.
    
---

    test: add csi-node tcp connection and stage with gRPC types
    
    Allows staging a volume via TCP.
    This will be required for upcoming changes where we can't use the
    unix socket directly, and have to go via trampoline.
    
    Also allow staging using the internal service types.
    
---

    test: add fake agent-ha-node endpoint reporting
    
    Useful for testing where the reply back from the cluster agent has to
    go via another endpoint.
    
    This will be useful for the coming test changes for rwx.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,5 @@ __pycache__
 /ci-report/
 pytest.log
 /bundle
+/deployer/misc/rwx-vm/nixos.qcow2
+/deployer/misc/rwx-vm/result
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1175,6 +1175,7 @@ dependencies = [
  "grpc",
  "humantime",
  "hyper-util",
+ "nix",
  "paste",
  "reqwest",
  "rpc",
--- a/control-plane/agents/src/bin/core/tests/volume/mod.rs
+++ b/control-plane/agents/src/bin/core/tests/volume/mod.rs
@@ -8,6 +8,7 @@ mod garbage_collection;
 mod helpers;
 mod hotspare;
 mod resize;
+mod rwx;
 mod snapshot;
 mod snapshot_clone;
 mod switchover;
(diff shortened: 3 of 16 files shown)


tiagolobocastro:
bors try

niladrih (approved):

tiagolobocastro:
bors merge

PR #1085, opened 2026-04-16
 feat: allow deleting snapshots from deleted pools

tiagolobocastro:
feat: allow deleting snapshots from deleted pools
    
    When users recreate pools manually and leave resources around its possible
    that the pool uuid will change, and so we'll reject snapshot deletion.
    
    In this case, we should accept the deletion and allow graceful cleanup.
    
---

    feat: add pool uuid to the transport types
    
    Note that we're not yet exposing this via the public interface nor the
    internal service though I think we should do that.


tiagolobocastro:
Resolves https://github.com/openebs/mayastor/issues/1944

Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

PR #1087, opened 2026-04-21
DiskPool CR error handling on creation

tiagolobocastro:
feat: set dsp error for failed creations
    
    When pool creation fails, we inspect the pool diagnostic error (if any).
    This errors is then set to the state error as well as the failure event.
    
    todo: add missing backoff on creation errors
    
---

    refactor: probe pool before creation
    
    Prior to this change, we only probed for pool imports.
    We now also probe for pool creations, bringing the benefits here.
    
---

    feat: return custom pool error on failed creations
    
    Massage the errors depending on the return of the create gRPC call.
    This allows the diskpool operator to add more context information
    to the CR state information.
    
---

    refactor: add custom pool create error
    
    A new pool diagnostics was added for report faulted pools but it did
    not take into account when we try to create a pool with a faulty disk.
    
    In order to do this we modify the creation to return a custom pool error
    containing the same diagnostic information we use during import.
    
---

    refactor: manual impl of RestJsonError
    
    This will be useful for the next change where we don't want custom
    information to be auto debug derived.
    
---

    chore: upgrade paperclip to v0.9.7


control-plane/rest/openapi-specs/v0_api_spec.yaml
```diff
@@ -3881,10 +3881,19 @@ components:
...
+      x-no-derive-dbg: true
       required:
         - details
         - kind
         - message
+    CustomErrorInfo:
+      type: object
+      properties:
+        poolDiag:
+          $ref: '#/components/schemas/PoolDiag'
```
Abhinandan-Purkait, line 3906: nit: This won't be very reusable as we have pool specific field here. Maybe call it PoolErrorExt? or something in that line?

Abhinandan-Purkait (approved):

control-plane/rest/openapi-specs/v0_api_spec.yaml
tiagolobocastro: We can also just add new optional fields here as need, otherwise this becomes a very deep Option<Option<Option<
wdyt?

Abhinandan-Purkait: We can use optional fields as well.

tiagolobocastro: I made pool non-optional, seems to have made it not so long: https://github.com/openebs/mayastor-control-plane/pull/1087/changes/75df328336f0b75a21db6fa061ec62787048369e

tiagolobocastro:
bors try

tiagolobocastro:
bors merge

tiagolobocastro:
Doh, data-plane needs to merge first...

tiagolobocastro:
bors merge

PR #1088, opened 2026-04-22
feat: expose pool replica and snapshot count

tiagolobocastro:
Exposes these counts from a pool, allowing for easily inspecting pool usage in terms of placement rather than just commitment/usage.

We expose both the state information, which is retrieved by counting
the replicas directly from the data-plane information and the spec
like information, which we retrieve from the pstor entries.
This latter is useful because it still allows us to gauge the pool
usage even when the node is offline for example.

### DIFF
--- a/control-plane/agents/src/bin/core/pool/specs.rs
+++ b/control-plane/agents/src/bin/core/pool/specs.rs
@@ -272,6 +272,31 @@ impl ResourceSpecs {
         }
         false
     }
+    /// Get a count of all volume replicas on the given pool.
+    fn pool_replica_count(&self, id: &PoolId) -> u64 {
+        self.replicas
+            .values()
+            .filter(|r| r.immutable_ref().pool_name() == id)
+            .count() as u64
+    }
+    /// Get a count of all volume snapshots on the given pool.
+    fn pool_snap_count(&self, id: &PoolId) -> u64 {
+        let mut snapshots = 0u64;
+        for snapshot in self.volume_snapshots.values() {
+            let snapshot = snapshot.lock();
+
+            let transactions = snapshot.metadata().transactions();
+            let this_pool = transactions
+                .values()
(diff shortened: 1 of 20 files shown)


control-plane/agents/src/bin/core/pool/specs.rs
```diff
@@ -272,6 +272,33 @@ impl ResourceSpecs {
         }
         false
     }
+    /// Get a count of all volume replicas on the given pool.
+    fn pool_replica_count(&self, id: &PoolId) -> u64 {
+        self.replicas
+            .values()
+            .filter(|r| r.immutable_ref().pool_name() == id)
```
tiagolobocastro, line 279: Shall we consider only replicas in created state?
IE if a replica is in deleting phase, do we consider it?

CC @Abhinandan-Purkait  @niladrih

Abhinandan-Purkait: Yeah, I think so only created.

control-plane/rest/openapi-specs/v0_api_spec.yaml
```diff
@@ -3686,6 +3700,16 @@ components:
           format: uint64
         errorInfo:
           $ref: '#/components/schemas/PoolErrorInfo'
+        replicaCount:
+          description: -|
+            How many replicas this pool contains.
```
Abhinandan-Purkait: nit: Change this from a question to statement?

Abhinandan-Purkait (approved):

control-plane/rest/openapi-specs/v0_api_spec.yaml
tiagolobocastro: Modified

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

PR #1089, opened 2026-04-24
test(rwx): add ha event and run fio

tiagolobocastro:
ci: log container errors before tearing them down
    
    On the deployer tests we have a wrapper over the composer library.
    
    Sometimes we pass the composer to other threads, and it may end up
    tearing down ahead of time, in a racy manner.
    
    This should also fix the missing container logs in these kinds of
    tests, as we see sometimes in CI, we now have the root cause!
    
---

    test(rwx): add ha event and run fio
    
    Adds a failover event by shutting down the target node.
    This should trigger a failover, and the volume target should then
    be moved to another available node.
    
---

    feat(csi/node): add support for reconnect-delay configuration
    
    This is mostly useful for testing, though it could be used to speed up
    retries in a cluster.
    
### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -17,18 +17,18 @@ repos:
         args: [ "--spec-changes" ]
         pass_filenames: false
         language: system
-      - id: rust-lint
-        name: Rust lint
-        description: Run cargo clippy on files included in the commit. clippy should be installed before-hand.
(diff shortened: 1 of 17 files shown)


tiagolobocastro:
bors try

Abhinandan-Purkait (approved):

susobhandey (commented):
LGTM

tiagolobocastro:
bors merge

PR #1090, opened 2026-04-28
Pool delete fixes

niladrih:
fix(reconciler): don't create empty replicas for purged published volumes

Add a guard in volume_replica_count_reconciler_traced before the
Ordering::Less branch: skip replica creation when the volume has zero
spec'd replicas and was previously published (health_info_id is set).
Replica anti-affinity guarantees at most one replica per volume per
node, so a single purge can only ever drop the spec count by one — the
0 case is the only one this guard needs to cover. Higher Less arities
imply a surviving spec with rebuild source available, and the volume
would already be Faulted and dispatched away from this reconciler.

---

feat(plugin): hint on pool delete error when pool is already Deleting/Purging

If `del_pool` returns an error (other than NOT_FOUND), GET the pool and
check its spec status. If the pool is already in `Deleting` or `Purging`
state, attach a hint to the `DeletePoolError` telling the user that the
operation will be retried automatically.

The hint is appended to the existing error display as a `\nHint: ...`
line, so it is visible in all output modes without changing the error
type or any upstream handler.

---

refactor(plugin): clarify --accept-data-loss consequences in purge error hints

The volume-loss and snapshot-loss PurgeReason messages previously listed
--accept-volume-loss/--accept-snapshot-loss and --accept-data-loss as if
they were interchangeable. Clarify that --accept-data-loss is a combined
flag that accepts both losses, so users understand the broader consequence
before reaching for it.

---

test: check if no new replicas are created for faulted vols. of purged pools

### DIFF
--- a/control-plane/agents/src/bin/core/controller/reconciler/volume/hot_spare.rs
+++ b/control-plane/agents/src/bin/core/controller/reconciler/volume/hot_spare.rs
@@ -267,6 +267,29 @@ async fn volume_replica_count_reconciler_traced(
 
     match current_replica_count.cmp(&required_replica_count) {
         Ordering::Less => {
+            // Guard: do not materialise a fresh, empty replica when the volume has
+            // already been published but currently has zero spec'd replicas.
+            //
+            // A volume that has never been published has no `health_info_id` and
+            // no data to lose, so a fresh replica there is correct.
+            if current_replica_count == 0 && volume.as_ref().health_info_id().is_some() {
+                // The reconciler cannot recover this volume on its own.
+                // Emit the warning exactly once per agent lifetime per volume to avoid flooding
+                // logs at the reconcile cadence.
+                if !volume.as_ref().data_loss_warned() {
+                    // Mark first, then log. The lock is dropped before the
+                    // tracing macro fires so the lock isn't held across logging.
+                    volume.lock().set_data_loss_warned();
+                    volume.warn_span(|| {
+                        tracing::warn!(
+                            "Volume has no remaining replicas; hot-spare will not create \
+                            additional replicas since no rebuild is possible"
+                        )
+                    });
+                }
+                return PollResult::Ok(PollerState::Idle);
+            }
+
             volume.warn_span(|| {
                 tracing::warn!(
                     "The volume has '{}' replica(s) but it should have '{}'",
--- a/control-plane/agents/src/bin/core/tests/pool/purge.rs
+++ b/control-plane/agents/src/bin/core/tests/pool/purge.rs
@@ -3,7 +3,7 @@ use grpc::operations::{
     pool::traits::PoolOperations, replica::traits::ReplicaOperations,
     volume::traits::VolumeOperations,
 };
-use std::time::Duration;
+use std::{collections::HashMap, time::Duration};
 use stor_port::{
     pstor::{etcd::Etcd, StoreObj},
     transport_api::{ReplyErrorKind, ResourceKind},
@@ -13,7 +13,8 @@ use stor_port::{
             SpecStatus,
         },
         transport::{
-            CreatePool, CreateVolume, DestroyPool, Filter, NodeStatus, Pool, PoolId, VolumeId,
+            CreatePool, CreateVolume, DestroyPool, DestroyVolume, Filter, GetVolumes, NodeStatus,
+            Pool, PoolId, PublishVolume, VolumeAccessMode, VolumeId, VolumeStatus,
         },
     },
 };
@@ -29,8 +30,14 @@ async fn pool_purge() {
     let cluster = ClusterBuilder::builder()
         .with_rest(false)
         .with_agents(vec!["core"])
-        .with_io_engines(1)
-        .with_pools(2) // pool-1 for purge tests, pool-2 for normal delete
+        // Engine 1 is intentionally idle for phases 1-3 (no pools on it). The existing
+        // phases assume the volume replica goes on the engine being stopped, which is
+        // guaranteed when engine 1 has no pool to land on. Phase 4 then provisions a
+        // pool on engine 1 to act as the surviving online destination that the bug
+        // would write a stray replica to.
+        .with_io_engines(2)
+        .with_pool(0, "malloc:///disk1?size_mb=100")
+        .with_pool(0, "malloc:///disk2?size_mb=100")
         .with_cache_period("100ms")
         .with_node_deadline("100ms")
         .with_reconcile_period(Duration::from_millis(100), Duration::from_millis(100))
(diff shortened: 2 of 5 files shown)


niladrih:
bors try

niladrih:
bors try

control-plane/agents/src/bin/core/controller/reconciler/volume/hot_spare.rs
```diff
@@ -267,6 +267,47 @@ async fn volume_replica_count_reconciler_traced(
...
+            // creates a brand new empty replica on a different node — silently masking
+            // the data loss the operator just consented to.
+            //
+            // A volume that has never been published has no `health_info_id` and no
+            // data to lose, so creating a replica is fine; we only gate published
+            // volumes.
+            if volume.as_ref().health_info_id().is_some() {
+                let volume_uuid = volume.uuid().clone();
+                let has_healthy_source =
+                    match context.registry().volume_state_health(&volume_uuid).await {
```
tiagolobocastro: I'm not sure we should do this here, this is an expensive operation
Can't we just gate 0->1 and ignore the health info here?

niladrih: I think the 0 -> 1 gate works, as we can't have more than 1 replica on a single node. Made the change. PTAL.

niladrih:
bors try

control-plane/agents/src/bin/core/controller/reconciler/volume/hot_spare.rs
```diff
@@ -267,6 +267,41 @@ async fn volume_replica_count_reconciler_traced(
...
+            // `Less` state (e.g. 1 -> 3 after a partial purge across nodes) implies
+            // at least one surviving spec, and `create_volume_replicas` will refuse
+            // to use a stale source — the volume would already be Faulted and
+            // dispatched away from this reconciler entirely.
+            //
+            // A volume that has never been published has no `health_info_id` and
+            // no data to lose, so a fresh replica there is correct.
+            if current_replica_count == 0 && volume.as_ref().health_info_id().is_some() {
+                volume.warn_span(|| {
+                    tracing::warn!(
```
tiagolobocastro: Also not sure about this, this may flood logs forever since it can't recover on its own...
You could potentially put a sticky bool in the volume runtime metadata (not persistent)

control-plane/agents/src/bin/core/controller/reconciler/volume/hot_spare.rs
```diff
@@ -267,6 +267,41 @@ async fn volume_replica_count_reconciler_traced(
...
+                        "Published volume has no replicas, needs '{}'. Skipping replica \
+                         creation to avoid materialising an empty replica that would \
+                         mask data loss.",
+                        required_replica_count
```
tiagolobocastro:
```suggestion
                        "Volume has no remaining replicas; hot-spare will not create additional replicas since no rebuild is possible"
```

control-plane/plugin/src/resources/error.rs
```diff
@@ -232,13 +234,13 @@ pub enum PurgeReason {
     NodeHasResources,
     #[strum(to_string = "Node has pools with data. Confirm with --yes to proceed.")]
     NodePurgeAcceptRequired,
-    #[strum(
-        to_string = "Volumes would lose their last healthy replica. Use --accept-volume-loss or --accept-data-loss to proceed."
-    )]
+    #[strum(to_string = "Volumes would lose their last healthy replica. \
```
tiagolobocastro, line 237: Why are these new errors using `to_string =` instead of `display(` ? 
If there's no reason then let's use display to be consistent

niladrih: These are strum attributes, not snafu.

tiagolobocastro: oh nvm, the way github was showing the diff this looked like it was part of `Error` rather than `PurgeReason`

control-plane/agents/src/bin/core/controller/reconciler/volume/hot_spare.rs
niladrih: Oh didn't know about this. I've added a VolumeRuntimeMetadata member to track this log warning.

control-plane/agents/src/bin/core/controller/reconciler/volume/hot_spare.rs
```diff
@@ -267,6 +267,52 @@ async fn volume_replica_count_reconciler_traced(
...
+            // no data to lose, so a fresh replica there is correct.
+            if current_replica_count == 0 && volume.as_ref().health_info_id().is_some() {
+                // The reconciler cannot recover this volume on its own — only the
+                // operator can (delete, or restore from backup). Emit the warning
+                // exactly once per agent lifetime per volume to avoid flooding
+                // logs at the reconcile cadence (~30s in production = 2,800
+                // duplicate lines per day, every day, until manual intervention).
+                // The flag lives in VolumeRuntimeMetadata which is `#[serde(skip)]`,
+                // so it resets on core agent restart — fine, one fresh warning per
+                // restart is informative confirmation that the protection is active.
```
tiagolobocastro: Please keep the comments short and sweet :)

Example:

```suggestion
                // The reconciler cannot recover this volume on its own.
                // Emit the warning exactly once per agent lifetime per volume to avoid flooding
                // logs at the reconcile cadence.
```

control-plane/agents/src/bin/core/controller/reconciler/volume/hot_spare.rs
```diff
@@ -267,6 +267,37 @@ async fn volume_replica_count_reconciler_traced(
...
+            // Guard: do not materialise a fresh, empty replica when the volume has
+            // already been published but currently has zero spec'd replicas.
+            //
+            // We gate `0 -> N` rather than every `Less` transition because replica
+            // anti-affinity guarantees at most one replica per volume per node, so
+            // a single purge can only ever drop the spec count by one. Any other
+            // `Less` state (e.g. 1 -> 3 after a partial purge across nodes) implies
+            // at least one surviving spec, and `create_volume_replicas` will refuse
+            // to use a stale source — the volume would already be Faulted and
+            // dispatched away from this reconciler entirely.
```
tiagolobocastro:
```suggestion
```

control-plane/agents/src/bin/core/tests/pool/purge.rs
```diff
@@ -392,3 +403,241 @@ async fn purge_reconciler_resumes_interrupted(cluster: &Cluster) {
...
+        assert!(
+            replicas.is_empty(),
+            "Reconciler created a new replica after the last healthy one was purged \
+             (replicas: {:?}). This is the data-loss-masking bug the guard prevents.",
+            replicas
+                .iter()
+                .map(|r| (r.node.clone(), r.pool_id.clone()))
+                .collect::<Vec<_>>()
+        );
+        tokio::time::sleep(Duration::from_millis(200)).await;
```
tiagolobocastro: if reconciler is 100ms then we can just wait on those?

```suggestion
        tokio::time::sleep(Duration::from_millis(100)).await;
```

tiagolobocastro (approved):
Could you please add description to the PR itself

niladrih:
bors merge

niladrih:
bors cancel

niladrih:
bors merge.

PR #1091, opened 2026-05-05
fix(csi): return FailedPrecondition instead of NotFound on freeze failure

### LINKED ISSUE #1978: CreateSnapshot fails on unpublished volume due to unnecessary freeze attempt
## Describe the bug

When `CreateSnapshot` is called for a volume that is not published/attached, the CSI controller attempts to freeze the filesystem and fails with gRPC `NotFound`. Instead of skipping the freeze and proceeding with snapshot creation (since an unpublished volume has no active I/O), it returns an error that causes the snapshot creation to fail entirely.

**Affected code:** `control-plane/csi-driver/src/bin/controller/controller.rs`, lines ~987-998

```rust
if let Some(ref app_node_endpoint) = app_node_endpoint_info {
    match issue_fs_freeze(app_node_endpoint.clone(), volume_uuid.to_string()).await
    {
        Err(error) if error.code() == Code::NotFound => {
            Err(Status::not_found(format!(
                "Failed to freeze volume {volume_uuid}, filesystem volume is not attached"
            )))
        }
        _else => _else,
    }?;
}
```

The code already checks `volume.spec.target` via `volume_app_node()` (line 123) and skips freeze when the volume is not published. However, a race condition exists: the volume can get unpublished between the target check and the actual `issue_fs_freeze` call. In this case, the freeze fails with `NotFound` and the entire `CreateSnapshot` fails.

**Observed in logs (csi-snapshotter container):**

```
createSnapshotWrapper: CreateSnapshot for content snapcontent-f587fed7-... returned error:
rpc error: code = NotFound desc = Failed to freeze volume ba075adf-..., filesystem volume is not attached
```

## To Reproduce

1. Create a volume (PVC) via any mayastor StorageClass and attach/mount it to a pod
2. Create a VolumeSnapshot of the volume (using default `mayastor-snapclass` VolumeSnapshotClass)
3. Detach/delete the pod so the volume is no longer published/attached
4. Create another VolumeSnapshot of the same volume while it is detached
5. `CreateSnapshot` fails with freeze error

## Expected behavior

- If the volume is not published/attached, skip the filesystem freeze and proceed with snapshot creation without quiesce
- An unpublished volume has no active I/O, so freezing is unnecessary — the snapshot will be consistent regardless
- The freeze should only be required when the volume is actively mounted and has potential in-flight writes

## OS info

- **Distro:** Ubuntu 24.04.3 LTS (Noble Numbat)
- **Kernel:** 6.8.0-71-generic
- **Mayastor version:** v2.10.0
  - `openebs/mayastor-csi-controller:v2.10.0`
  - `openebs/mayastor-io-engine:v2.10.0`
- **csi-snapshotter sidecar:** `registry.k8s.io/sig-storage/csi-snapshotter:v8.2.0`
- **Kubernetes:** v1.34.3

## Additional context

The fix is to catch the `NotFound` error from `issue_fs_freeze` and proceed with snapshot creation instead of failing. This handles the race condition where the volume gets unpublished between the target check and the freeze call:

```rust
Err(error) if error.code() == Code::NotFound => {
    warn!(
        volume.uuid = %volume_uuid,
        "Skipping filesystem freeze: volume is not attached, proceeding without quiesce"
    );
    Ok(())
}
```

PR: (will link after creation)

yugchaudhari:
## Summary

When `CreateSnapshot` fails because the source volume is not attached/staged, the CSI controller returns gRPC `NotFound`. While `NotFound` is treated as a final error per CSI spec, `FailedPrecondition` is more semantically correct, the operation cannot proceed in the current state of the resource (volume is not staged), but the resource itself exists.

## Changes

- Return `Status::failed_precondition` instead of `Status::not_found` when `issue_fs_freeze` fails because the volume is not attached
- The error message remains unchanged

## Why FailedPrecondition

- The volume exists (so `NotFound` is misleading)
- The freeze cannot proceed because the volume is not in a publishable/staged state — this is a precondition failure
- Makes it clearer to users and CSI consumers that the snapshot can be retried after publishing/staging the volume
- Both error codes are treated as final (non-retriable) by the external-snapshotter sidecar, so retry behavior is preserved

Fixes: https://github.com/openebs/mayastor/issues/1978

### DIFF
--- a/control-plane/csi-driver/src/bin/controller/controller.rs
+++ b/control-plane/csi-driver/src/bin/controller/controller.rs
@@ -988,7 +988,7 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
                     match issue_fs_freeze(app_node_endpoint.clone(), volume_uuid.to_string()).await
                     {
                         Err(error) if error.code() == Code::NotFound => {
-                            Err(Status::not_found(format!(
+                            Err(Status::failed_precondition(format!(
                                 "Failed to freeze volume {volume_uuid}, filesystem volume is not attached"
                             )))
                         }


Abhinandan-Purkait (approved):

tiagolobocastro (approved):

tiagolobocastro:
thank you @yugchaudhari !
bors merge

yugchaudhari:
@tiagolobocastro @Abhinandan-Purkait CI failures look unrelated to this change, failures are in pool/node fixtures (`'NoneType' object has no attribute 'status'`) and resource exhaustion (`Not enough suitable pools available`), not in the freeze code path. Could you retrigger?

Abhinandan-Purkait:
bors try

Abhinandan-Purkait:
@yugchaudhari Can you please squash your commits? The CI succeeds now.

yugchaudhari:
@Abhinandan-Purkait Done. Squashed to single commit.

Abhinandan-Purkait:
bors merge

tiagolobocastro:
@yugchaudhari could you lint your commit message please?

image

yugchaudhari:
@tiagolobocastro Done! It should be fine now.

tiagolobocastro:
bors merge

PR #1092, opened 2026-05-06
fix(core/pool_op): avoid returning error w/o completing purge op

niladrih:
After start_destroy_for_purge marks the pool as Purging and logs a pending Destroy op to etcd, the destructive work (replica operation_guard_wait + destroy_or_purge) used  to bubble errors straight out of purge_pool. This bypassed complete_destroy entirely, leaving the spec wedged: status Purging with a pending Destroy op that was never cleared. Subsequent purge attempts — including the reconciler's resume path — then tripped busy()? inside start_destroy_for_purge, so the pool became un-purgeable without manual etcd intervention.

Move the purge work into a purge_pool_resources helper returning Result<PoolDeleteResult, SvcError>, and always pipe it through complete_destroy. On Ok the spec is deleted from etcd and memory; on Err complete_destroy clears the pending op and keeps Purging

### DIFF
--- a/control-plane/agents/src/bin/core/pool/pool_operations.rs
+++ b/control-plane/agents/src/bin/core/pool/pool_operations.rs
@@ -21,8 +21,9 @@ use stor_port::{
             snapshots::replica::ReplicaSnapshot,
         },
         transport::{
-            CreatePool, CtrlPoolState, DestroyPool, ExpandPool, Pool, PoolDeleteResult, PoolDiag,
-            PoolDiskError, PoolId, PoolStatus, ReplicaOwners,
+            CreatePool, CtrlPoolState, DestroyPool, ExpandPool, NodeId, Pool, PoolDeleteResult,
+            PoolDiag, PoolDiskError, PoolId, PoolStatus, ReplicaOwners, SnapshotLossInfo,
+            VolumeLossInfo,
         },
     },
 };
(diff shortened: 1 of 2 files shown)


control-plane/agents/src/bin/core/pool/pool_operations.rs
```diff
@@ -469,10 +491,10 @@ impl OperationGuardArc<PoolSpec> {
             snapshots_affected = replica_snapshots.len(),
             volume_loss = result.has_volume_loss(),
             snapshot_loss = result.has_snapshot_loss(),
-            "Pool purged successfully. Affected volumes will be marked faulted by reconciler."
+            "Pool purge resources cleaned up. Affected volumes will be marked faulted by reconciler."
```
tiagolobocastro: nit: I would prefer the purged successfully message

But the remaining of the sentence is not correct, there's no such thing as volumes being marked, the status is always derived (at least today)

tiagolobocastro (approved):

control-plane/agents/src/bin/core/pool/pool_operations.rs
niladrih: I seem to have made that assumption. Removed that second half of the message. Restored the 'purged successfully' log message.

niladrih:
bors merge

PR #1093, opened 2026-05-06
feat: add node resource tally

tiagolobocastro:
feat: add resource tally to the rest-plugin for node objects
    
---

    feat: expose node resource tally on the public api
    
---

    feat: add node resource tallies to the internal service
    
---

    feat(agent/core): add replica and snapshot count to node
    
    Similar to the pool changes, keeps a tally of replica and snapshot count,
    and the pool count, but for the node.
    This will then be exposed by the internal and public api.


Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

PR #1095, opened 2026-05-08
fix(csi/controller): add missing type value to cli argument

tiagolobocastro:
Missing type value parse on the tls client certificate path. todo: convert to clap derive to avoid these issues...


Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

PR #1096, opened 2026-05-10
build: remove openssl and ring dependency

tiagolobocastro:
Drops hyper-client-tls feature and all related tls feats.
This was used to build the windows kubectl-client but the rls build
now works fine for windows targets, so let's drop it to avoid depending
on the openssl, and it's also good to keep things consistent across targets.

Also drops ring and favours aws-lc-rs. This will eventually make it easier
to use fips algorithms.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -131,8 +131,6 @@ dependencies = [
  "actix-service",
  "actix-utils",
  "futures-core",
- "http 0.2.12",
- "http 1.1.0",
  "impl-more",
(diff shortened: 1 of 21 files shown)


tiagolobocastro:
bors try

niladrih (approved):

Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

tiagolobocastro:
bors merge

PR #1099, opened 2026-05-12
fix: don't schedule on critical pools

tiagolobocastro:
Don't try to schedule on critical pools, since we don't expect a good outcome anyway.

Much more than this will be needed, as we should not even try to issue any I/O.
Data-plane itself should also avoid any gRPC timing out when it knows the pool is stalled anyway..


niladrih (approved):

susobhandey (approved):
LGTM

Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

PR #1100, opened 2026-05-15
fix: use kube::Config::infer() to support in-cluster config detection

datacore-gjanoti:
## Description
`config_from_kubeconfig` in `k8s/proxy/src/lib.rs` was falling back to `kube::Config::from_kubeconfig` when no kubeconfig path is provided. This only looks for `~/.kube/config` and does not try in-cluster config, causing the binary to fail when running inside a Kubernetes pod.

Changed to `kube::Config::infer()` which tries `~/.kube/config` first and automatically falls back to in-cluster config when running inside a pod. This is a single line change.

## Motivation and Context
When running the binary inside a Kubernetes pod, it fails with `failed to read kubeconfig from '/.kube/config': No such file or directory` because no kubeconfig file exists inside the pod. The standard Kubernetes approach for pods is to use in-cluster config via the service account token automatically mounted at `/var/run/secrets/kubernetes.io/serviceaccount/`. `kube::Config::infer()` handles this correctly by trying kubeconfig first and falling back to in-cluster config automatically.

## Regression
No

## How Has This Been Tested?
Tested by running the binary inside a Kubernetes pod without mounting any kubeconfig file. With the fix applied, the binary correctly uses in-cluster config and successfully collects all diagnostic data. Also verified that the existing CLI behavior is unchanged — running locally with `~/.kube/config` still works correctly since `kube::Config::infer()` tries kubeconfig first.

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added unit tests to cover my changes.

### DIFF
--- a/k8s/proxy/src/lib.rs
+++ b/k8s/proxy/src/lib.rs
@@ -28,7 +28,13 @@ pub async fn config_from_kubeconfig(
             kube::Config::from_custom_kubeconfig(kube_config, &options).await?
         }
 
-        None => kube::Config::from_kubeconfig(&options).await?,
+        None => {
+            if std::env::var("KUBERNETES_SERVICE_HOST").is_ok() {
+                kube::Config::incluster()?
+            } else {
+                kube::Config::from_kubeconfig(&options).await?
+            }
+        }
     };
     config.apply_debug_overrides();
     Ok(config)


Abhinandan-Purkait (approved):
LGTM.

Abhinandan-Purkait:
@datacore-gjanoti Please squash your commits.

k8s/proxy/src/lib.rs
```diff
@@ -28,7 +28,7 @@ pub async fn config_from_kubeconfig(
             kube::Config::from_custom_kubeconfig(kube_config, &options).await?
         }
 
-        None => kube::Config::from_kubeconfig(&options).await?,
+        None => kube::Config::infer().await?,
```
tiagolobocastro: This is not correct, we fall here when no config is explicitly provided

tiagolobocastro (changes requested):
I don't think these changes are quite there, as it's catching usual kubectl-plugin usual where we don't explicitly add the kubeconfig

Also please squash the comments, the revert should not be part of the history since it's not a revert of previous tree content

k8s/proxy/src/lib.rs
datacore-gjanoti: Makes sense. Would this be the right fix - check for KUBERNETES_SERVICE_HOST env var (set automatically by K8s in every pod) and use in-cluster config only when running inside a pod, otherwise fall back to the original from_kubeconfig behavior?


**None => {
          if std::env::var("KUBERNETES_SERVICE_HOST").is_ok() {
              kube::Config::incluster()?
          } else {
              kube::Config::from_kubeconfig(&options).await?
          }
}**

tiagolobocastro: yeah something like that could work

tiagolobocastro: Though rather than infer, probably then you need to use `kube::Config::incluster`, ie:
```rust
None if std::env::var("KUBERNETES_SERVICE_HOST").is_err() => kube::Config::from_kubeconfig(&options).await?,
None => kube::Config::incluster()
```

tiagolobocastro: Oh I see, you had already changed from infer to cluster, yes that's it

datacore-gjanoti: > Oh I see, you had already changed from infer to cluster, yes that's it

done, made the changes and squashed

tiagolobocastro (approved):

tiagolobocastro:
bors merge

tiagolobocastro (approved):

tiagolobocastro:
bors merge

tiagolobocastro:
bors try

tiagolobocastro:
bors try

tiagolobocastro:
bors merge

PR #1104, opened 2026-05-25
fix(reconciler): resize undersized replicas before retrying nexus resize

cmrmahesh:
When a volume resize partially succeeds (some replicas expand, one fails), the nexus reconciler (fixup_nexus_size) retries only the nexus resize which always fails because an undersized child replica blocks it. This creates an infinite retry loop with no self-healing.

This fix adds a replica size check before the nexus resize attempt. If any replica is below the volume spec size, it is resized first. This mirrors the inverse logic already present in the garbage collector (which shrinks over-expanded replicas).

## Problem

Issue described at [#1989](https://github.com/openebs/mayastor/issues/1989) 
When a volume resize partially succeeds (2/3 replicas expand, 1 fails due to transient gRPC error), the volume spec is committed at the new size. The reconciler then retries `resize_nexus` every 5 minutes, which always fails with:

  Child nvmf://... of nexus ... is too small: size = 314572800 x 512, required = 524288000 x 512

No `resize_replica` is ever re-attempted for the stuck replica.

  ## Fix
Before calling `resize_nexus`, iterate over volume replicas and expand any that are undersized. This uses the same pattern as the garbage collector (`volume/garbage_collector.rs`) which already handles the inverse case (shrinking over-expanded replicas).

  ## Testing
  - [x] - nix-shell --run "cargo build --bins" 
  - [x] - nix-shell --run "cargo check -p agents"
  - [ ] -  nix-shell --run "cargo test -p agents -- volume::resize --test-threads=1" - will rely on CI
  - [ ] - Existing tests in agents already cover resize, but I can add dedicated tests if required

### DIFF
--- a/control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
+++ b/control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
@@ -36,7 +36,8 @@ use stor_port::{
         },
         transport::{
             Child, ChildUri, CreateNexus, Nexus, NexusChildActionContext, NexusShareProtocol,
-            NexusStatus, NodeStatus, ReplicaId, ResizeNexus, ShareNexus, UnshareNexus,
+            NexusStatus, NodeStatus, ReplicaId, ResizeNexus, ResizeReplica, ShareNexus,
+            UnshareNexus,
         },
     },
 };
@@ -670,6 +671,45 @@ pub(super) async fn fixup_nexus_size(
     // Nexus's size doesn't need any fixup.
     if nexus.as_ref().size == required_size {
         return PollResult::Ok(PollerState::Idle);
+    }
+
+    // Ensure all nexus children (replicas) are at the required size before attempting
+    // nexus resize. Only resize replicas that are actual nexus children — disconnected
+    // volume replicas are handled separately by other reconcilers.
+    for child in nexus
+        .as_ref()
+        .children
+        .iter()
+        .filter_map(|c| c.as_replica_ref())
+    {
+        // Spec size only updates on successful resize, so if it's already at
+        // required size we can skip this child entirely.
+        let Some(rsc) = registry.specs().replica_rsc(child.uuid()) else {
+            continue;
+        };
+        if rsc.lock().size >= required_size {
+            continue;
+        }
+        let mut replica = registry.specs().replica(child.uuid()).await?;
+        let state = registry.replica(child.uuid()).await?;
+        nexus.info(&format!(
+            "Child replica {} is undersized (spec {}B < {}B), resizing before nexus resize",
+            child.uuid(),
+            replica.as_ref().size,
+            required_size
+        ));
+        replica
+            .resize(
+                registry,
+                &ResizeReplica::new(
+                    &state.node,
+                    replica.as_ref().pool_name(),
+                    None,
+                    child.uuid(),
+                    required_size,
+                ),
+            )
+            .await?;
     }
 
     let nexus_state = registry.nexus(nexus.as_ref().uid()).await?;


control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
```diff
@@ -672,6 +673,38 @@ pub(super) async fn fixup_nexus_size(
...
+                            &state.node,
+                            replica.as_ref().pool_name(),
+                            None,
+                            replica.uuid(),
+                            required_size,
+                        ),
+                    )
+                    .await?;
+            }
+        }
```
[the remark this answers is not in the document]
niladrih: @cmrmahesh any specific reason why we're ignoring this error?

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
```diff
@@ -672,6 +673,38 @@ pub(super) async fn fixup_nexus_size(
...
+                        registry,
+                        &ResizeReplica::new(
+                            &state.node,
+                            replica.as_ref().pool_name(),
+                            None,
+                            replica.uuid(),
+                            required_size,
+                        ),
+                    )
+                    .await?;
```
Abhinandan-Purkait: What happens if a replica cannot be resized for whatever reason let's say that specific pool has no more capacity. Then this will itself end up in a loop or is that handled somewhere else, I don't have it at the top of my mind atm?

Abhinandan-Purkait: I would say let's kind of test this scenario, create 4 pools, on 4 nodes, 1GB, 3GB, 3GB, 3GB, now cordon one of the 3GB pools, create a ~ 1GB Volume with 3 replicas, the replicas should land on 1GB, 3GB, 3GB pools. Uncordon the 3GB pool that you had cordoned earlier. Now expand the volume to 2GB. Can you check if this is handled atm, if not then this change might not be enough?

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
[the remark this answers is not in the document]
cmrmahesh: This was a miss. Fixed. Now we propagate the error from registry.replica() instead of silently skipping it. If any replica state lookup fails, we return early with the error which gets  propagated → poller retries next cycle a reconcile_period (10s). Nexus resize is never attempted without verifying all children sizes.

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
cmrmahesh: Thanks @Abhinandan-Purkait 
I see 2 cases:
1. Initial API call - `resizeable_replicas()` - this will precheck that all replica pools have capacity before proceeding. If 1 GB pool cant fit 2 GB -> API fails -> volume spec never updated so reconciler is never involved.
2. In the reconciler path - if replica resize fails - (with my pr) error propogates - and poller tries at fixed reconcile_period. No reconciler in the codebase has bounded retries today (can be broader design change). Pool-full is unrecoverable without operator action. 

As for the test scenario you asked - I believe the existing `resize_with_no_capacity` test already covers this path. I can add a targeted integration test for this specific scenario.

Abhinandan-Purkait: The ENOSPC was more of a generic example, the point was if any unrecoverable error happens for a replica resize then we potentially get locked in this same situation of retrying the resize. So, I feel there should be some mechanism that evicts that replica from the volume after a certain number of retries and then once a new replica is added it just finishes the retry. Anyway this is something we can tackle later, for now the changes looks good to me.

cmrmahesh: > I feel there should be some mechanism that evicts that replica from the volume after a certain number of retries and then once a new replica is added it just finishes the retry.

Agreed. Let me think of a way and add a OEP.

Abhinandan-Purkait: Sure. Please feel free to add other thoughts and bugs or enhancement you can think of. All of these fixes are making the project more robust. Thanks for the contributions!

Also please feel free to join our community meeting happening tomorrow, 7:30PM IST, for more discussions. ref:- https://github.com/openebs/community#community

Abhinandan-Purkait (approved):

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
cmrmahesh: Glad I could help. Thanks for the invite, looking forward to joining.

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
```diff
@@ -672,6 +673,47 @@ pub(super) async fn fixup_nexus_size(
         return PollResult::Ok(PollerState::Idle);
     }
 
+    // Ensure all replicas are at the required size before attempting nexus resize.
+    // A previous resize attempt may have expanded some replicas but not all, leaving
+    // the nexus unable to resize due to an undersized child.
+    for replica_spec in context.specs().volume_replicas(volume.as_ref().uid()) {
+        let mut replica = match replica_spec.operation_guard() {
+            Ok(guard) => guard,
+            Err(_) => continue,
```
tiagolobocastro: We should bail out here rather than continue since this replica would not be resized

cmrmahesh: Good one, we can skip an attempt for nexus resize, I will throw in a log as well if replica is locked.

cmrmahesh: @tiagolobocastro 

Another way to optimise this further is to
 
1. For all replicas
2. Get state (registry look up) -> check actual size 
3. if undersized -> acquire guard -> resize

Will wait for your opinion

cmrmahesh: Hi @tiagolobocastro  can you take a look at this

tiagolobocastro: I think checking the spec size would be enough, since that won't update unless the resize was a success.
The log is not likely not needed, since if the replicas was locked, the next reconciler retry will retry again, which IIRC is the current pattern we use.

Though I'm wondering now how can this even happen? I'm still travelling to can't look at the code properly but it looks like the volume size is not committed until all replica resize all succeeded?

tiagolobocastro: I see 2 potential issues, if the volume is not published, we may succeed the resize even though some healthy replicas are offline?
Also if for some blip we saw a replica offline we may also skip resize in that case.

I'm fine with this change as WA but I think the resize logic needs refactoring as it doesn't seem to take into account nexus children which should be resized as a pre-condition, and neither the healthy replicas.
Also during volume resize if a replica resize fails we seem to just bail out? (will have to double check this)
Also this nexus resize should not be running for shutdown nexuses.

> I think checking the spec size would be enough, 

Ah, but what if a replica is offline but not in use by the nexus?
Perhaps rather than trying to resize all volume replicas, we should rather resize only the nexus children?
And then consider we may need a separate reconciler for all volume replicas.

tiagolobocastro: So my proposal for this PR is to resize only the nexus children here (or at least only have those as the required ones to resize), and then we may need some other work to tackle the other issues.

cmrmahesh: > it looks like the volume size is not committed until all replica resize all succeeded?

Yes, we are optimisitcally updating the spec. see: https://github.com/openebs/mayastor-control-plane/blob/develop/control-plane/agents/src/bin/core/volume/operations.rs#L237

cmrmahesh: > my proposal for this PR is to resize only the nexus children here 

Updated the PR

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
```diff
@@ -672,6 +673,45 @@ pub(super) async fn fixup_nexus_size(
...
+        .as_ref()
+        .children
+        .iter()
+        .filter_map(|c| c.as_replica_ref())
+    {
+        // Quick spec-size check without acquiring operation guard.
+        // Spec size only updates on successful resize, so if it's already at
+        // required size we can skip this child entirely.
+        if let Some(rsc) = registry.specs().replica_rsc(child.uuid()) {
+            if rsc.immutable_ref().size >= required_size {
```
tiagolobocastro: `immutable_ref` should only be used for fields which never change
`as_ref` for fields which haven't changed since the guard has been obtained
otherwise, lock must be taken

So we could take the lock here, or just fetch the guard, see my changes above

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
```diff
@@ -672,6 +673,45 @@ pub(super) async fn fixup_nexus_size(
...
+    for child in nexus
+        .as_ref()
+        .children
+        .iter()
+        .filter_map(|c| c.as_replica_ref())
+    {
+        // Quick spec-size check without acquiring operation guard.
+        // Spec size only updates on successful resize, so if it's already at
+        // required size we can skip this child entirely.
+        if let Some(rsc) = registry.specs().replica_rsc(child.uuid()) {
```
tiagolobocastro: spec should always exist, but maybe this would be cleaner.
also including the `as_ref` as per the other comment

```suggestion
        let Some(rsc) = registry.specs().replica_rsc(child.uuid()) else {
            continue;
        };
        let mut replica = registry.specs().replica(child.uuid()).await?;
        if rsc.as_ref().size >= required_size {
            continue;
        }
        
```

tiagolobocastro (commented):
This looks good thank, just one more change on the immurable_ref

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
tiagolobocastro: @cmrmahesh would you be able to add yourselves to the adopters: https://github.com/openebs/community/blob/develop/ADOPTERS.md ?
Also we are currently applying for incubation stage, would be great if we could submit you for interview

cmrmahesh: Sure will do it. 
FYI @yugchaudhari

control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs
cmrmahesh: Done. Please check.

tiagolobocastro (approved):

tiagolobocastro:
Thanks @cmrmahesh
bors merge

cmrmahesh:
Hi @Abhinandan-Purkait / @niladrih  looks like the CI failure is unrelated. can we rerun?

Abhinandan-Purkait:
bors merge

Abhinandan-Purkait:
@cmrmahesh The tests need fixing, will merge once that is resolved.

cmrmahesh:
> @cmrmahesh The tests need fixing, will merge once that is resolved.

Sure, you can let me know if I can be of any help.

tiagolobocastro:
bors merge

PR #1106, opened 2026-05-28
build: use static CDN to pull crates

tiagolobocastro:
Use static.crates.io (CDN) instead of crates.io/api to avoid the 1 req/sec rate limit on the API servers, which currently returns intermittent 403s.

This will likely require tweaks further down.


tiagolobocastro:
@Abhinandan-Purkait this should help, give it a go.
Other repos would require similar changes...

niladrih (approved):

Abhinandan-Purkait (approved):

Abhinandan-Purkait:
Thanks! @tiagolobocastro 
bors merge

PR #1107, opened 2026-05-28
cherry pick fixes from release/2.11

Abhinandan-Purkait:
- Use static.crates.io (CDN) instead of crates.io/api to avoid the 1 req/sec rate limit on the API servers, which currently returns intermittent 403s.
- Fail fs ops if the existing fs is different from requested


niladrih (approved):

Abhinandan-Purkait:
bors merge

PR #1108, opened 2026-05-28
feat(core): offline rebuild for unpublished degraded volumes

yugchaudhari:
Iteration 2 of offline volume rebuild ([openebs/openebs#4208](https://github.com/openebs/openebs/issues/4208)), builds on the detection-only reconciler from [#1103](https://github.com/openebs/mayastor-control-plane/pull/1103).

When an unpublished volume goes degraded, the reconciler now creates a temporary unshared nexus (target_config with `protocol: None`) after a grace period. That makes the volume look published to the existing HotSpareReconciler, which rebuilds the faulted replicas. Once the volume is Online again, the reconciler tears the nexus down and the volume returns to unpublished.

Key point: no rebuild logic is reimplemented, we lean on HotSpare. We just stand up and tear down the nexus.

New config: `--offline-rebuild-grace-period` (default 10m), behind the existing `--offline-rebuild-enabled` flag.

BDD tests cover happy path, feature-disabled, and never-published precondition.

### DIFF
--- a/control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
+++ b/control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
@@ -1,16 +1,25 @@
 use crate::controller::{
     reconciler::{PollContext, TaskPoller},
-    resources::{operations_helper::OperationSequenceGuard, ResourceMutex},
+    resources::{
+        operations::ResourcePublishing, operations_helper::OperationSequenceGuard,
+        OperationGuardArc, ResourceMutex,
+    },
     task_poller::{PollResult, PollerState},
 };
 
-use stor_port::types::v0::{store::volume::VolumeSpec, transport::VolumeStatus};
+use stor_port::types::v0::{
+    store::volume::VolumeSpec,
+    transport::{PublishVolume, UnpublishVolume, VolumeState, VolumeStatus},
+};
 
-/// Offline Volume Rebuild reconciler.
+/// Offline Volume Rebuild.
 ///
-/// Detects unpublished volumes whose replicas have become unhealthy and would
-/// benefit from a rebuild. Iteration 1 only logs eligible candidates; the
-/// actual rebuild action is added in a follow-up iteration.
+/// Detects unpublished volumes whose replicas have become unhealthy and
+/// creates a temporary non-shared nexus so the existing rebuild engine
+/// can restore them. Once rebuild completes, tears the nexus down.
+///
+/// The per-volume grace-period state lives on `VolumeMetadata.runtime` so it
+/// gets cleaned up automatically when the volume is deleted.
 ///
 /// See `designs/replicated-pv/mayastor/offline-volume-rebuild.md` for the
 /// full design.
@@ -41,32 +50,161 @@ impl TaskPoller for OfflineRebuildReconciler {
 
 #[tracing::instrument(
     level = "debug",
-    skip(context, volume_spec),
+    skip_all,
     fields(volume.uuid = %volume_spec.uuid(), request.reconcile = true)
 )]
 async fn offline_rebuild_reconcile(
     volume_spec: &mut ResourceMutex<VolumeSpec>,
     context: &PollContext,
 ) -> PollResult {
-    let volume = match volume_spec.operation_guard() {
+    let mut volume = match volume_spec.operation_guard() {
         Ok(guard) => guard,
         Err(_) => return PollResult::Ok(PollerState::Busy),
     };
 
-    if !volume.as_ref().policy.self_heal
-        || !volume.as_ref().status.created()
-        || volume.as_ref().target().is_some()
-    {
+    if !volume.as_ref().policy.self_heal || !volume.as_ref().status.created() {
         return PollResult::Ok(PollerState::Idle);
     }
 
-    let volume_state = context.registry().volume_state(volume.uuid()).await?;
-    if volume_state.status == VolumeStatus::Degraded {
-        tracing::debug!(
-            volume.uuid = %volume.uuid(),
-            "Unpublished volume is Degraded; eligible for offline rebuild"
-        );
+    // Early bails that don't need the volume state: a real (shared) publish is
+    // not our concern, and a never-published volume has no data to rebuild.
+    let has_offline_rebuild_target = match volume.as_ref().target() {
+        Some(target) => {
+            if target.protocol().is_some() {
+                return PollResult::Ok(PollerState::Idle);
+            }
+            true
+        }
+        None => {
+            if volume.as_ref().health_info_id().is_none() {
+                return PollResult::Ok(PollerState::Idle);
+            }
+            false
+        }
+    };
+
+    // Must use the health-aware state: for an unpublished volume the plain
+    // volume_state() derives status from the replica spec count (unchanged when a
+    // node dies), so it reports Online and we'd never detect the degradation.
+    // volume_state_health() factors in actual online replica health.
+    let volume_state = context
+        .registry()
+        .volume_state_health(volume.uuid())
+        .await?;
+
+    if has_offline_rebuild_target {
+        return teardown_if_rebuilt(&mut volume, &volume_state, context).await;
     }
 
+    // Volume is unpublished with prior health info. Wait for it to actually be
+    // Degraded, then enforce the grace period before initiating rebuild.
+    if volume_state.status != VolumeStatus::Degraded {
+        volume.lock().metadata.clear_offline_rebuild_degraded();
+        return PollResult::Ok(PollerState::Idle);
+    }
+
+    let degraded_for = volume.lock().metadata.offline_rebuild_degraded();
+    let grace_period = context.registry().offline_rebuild_grace_period();
+    if degraded_for < grace_period {
+        tracing::debug!(
+            volume.uuid = %volume.uuid(),
+            remaining = ?grace_period.saturating_sub(degraded_for),
+            "Offline rebuild waiting for grace period"
+        );
+        return PollResult::Ok(PollerState::Busy);
+    }
+
+    initiate_offline_rebuild(&mut volume, context).await
+}
+
+/// Creates a non-shared nexus for the volume so the existing rebuild engine
+/// (HotSpareReconciler) can restore faulted replicas.
+async fn initiate_offline_rebuild(
+    volume: &mut OperationGuardArc<VolumeSpec>,
+    context: &PollContext,
+) -> PollResult {
+    let registry = context.registry();
+
+    // Respect the system-wide rebuild concurrency limit.
+    if registry.rebuild_allowed().await.is_err() {
+        tracing::debug!(
+            volume.uuid = %volume.uuid(),
+            "Offline rebuild deferred: max concurrent rebuilds reached"
+        );
+        return PollResult::Ok(PollerState::Busy);
+    }
+
+    tracing::info!(
+        volume.uuid = %volume.uuid(),
+        "Initiating offline rebuild: creating non-shared nexus"
+    );
+
+    let request = PublishVolume::new(
+        volume.uuid().clone(),
+        None, // auto-select target node (prefers node with healthy replica)
+        None, // share = None → nexus won't be shared
+        Default::default(),
+        vec![],
+        Default::default(),
+    );
+
+    match volume.publish(registry, &request).await {
+        Ok(_) => {
+            tracing::info!(
+                volume.uuid = %volume.uuid(),
+                "Offline rebuild nexus created; HotSpareReconciler will handle the rebuild"
+            );
+            volume.lock().metadata.clear_offline_rebuild_degraded();
+            PollResult::Ok(PollerState::Idle)
+        }
+        Err(error) => {
+            tracing::warn!(
+                volume.uuid = %volume.uuid(),
+                %error,
+                "Failed to create offline rebuild nexus"
+            );
+            PollResult::Ok(PollerState::Idle)
+        }
+    }
+}
+
+/// Tears down the temporary nexus once the volume is fully rebuilt. For any
+/// other state we wait — a stuck or transiently-unavailable nexus is left in
+/// place so we don't churn (creating and destroying repeatedly) or accidentally
+/// tear down a healthy in-progress rebuild during a brief node blip.
+async fn teardown_if_rebuilt(
+    volume: &mut OperationGuardArc<VolumeSpec>,
+    volume_state: &VolumeState,
+    context: &PollContext,
+) -> PollResult {
+    if volume_state.status == VolumeStatus::Online {
+        tracing::info!(
+            volume.uuid = %volume.uuid(),
+            "Offline rebuild complete; tearing down temporary nexus"
+        );
+
+        let request = UnpublishVolume::new(volume.uuid(), false, vec![]);
+        match volume.unpublish(context.registry(), &request).await {
+            Ok(_) => {
+                tracing::info!(
+                    volume.uuid = %volume.uuid(),
+                    "Temporary nexus destroyed; volume returned to unpublished state"
+                );
+            }
+            Err(error) => {
+                tracing::warn!(
+                    volume.uuid = %volume.uuid(),
+                    %error,
+                    "Failed to tear down offline rebuild nexus"
+                );
+            }
+        }
+        return PollResult::Ok(PollerState::Idle);
+    }
+
+    // Only `Online` is a clean enough signal to tear the nexus down. Anything
+    // else (Degraded mid-rebuild, transient state during a node blip, etc.)
+    // we wait on — tearing down would either churn against the grace timer
+    // or kill a healthy in-progress rebuild during a brief hiccup.
     PollResult::Ok(PollerState::Idle)
 }
--- a/control-plane/agents/src/bin/core/tests/volume/offline_rebuild.rs
+++ b/control-plane/agents/src/bin/core/tests/volume/offline_rebuild.rs
@@ -1,0 +1,164 @@
+#![cfg(test)]
+
+use super::helpers::wait_till_volume_status;
+use deployer_cluster::ClusterBuilder;
+use std::{collections::HashMap, time::Duration};
+use stor_port::types::v0::{
+    openapi::models::{self, VolumePolicy, VolumeStatus},
+    transport::VolumeId,
+};
+use uuid::Uuid;
+
+const RECONCILE_PERIOD_MS: u64 = 250;
+const GRACE_PERIOD_SECS: u64 = 2;
+const REBUILD_TIMEOUT_SECS: u64 = 30;
+
+/// Happy path: create a 2-replica volume, publish to establish health_info,
+/// unpublish, stop one io-engine node → volume becomes Degraded.
+/// With offline rebuild enabled, the reconciler creates a temp nexus,
+/// HotSpare rebuilds the replica, then the reconciler tears down the nexus.
+/// Final state: volume Online with no target (unpublished).
+#[tokio::test]
+async fn offline_rebuild_happy_path() {
+    let reconcile = Duration::from_millis(RECONCILE_PERIOD_MS);
+    let cluster = ClusterBuilder::builder()
+        .with_rest(true)
+        .with_io_engines(3)
+        .with_tmpfs_pool(52428800)
+        .with_cache_period("250ms")
+        .with_reconcile_period(reconcile, reconcile)
+        .with_options(|o| {
+            o.with_isolated_io_engine(true)
+                .with_agents_env("OFFLINE_REBUILD_ENABLED", "true")
+                .with_agents_env(
+                    "OFFLINE_REBUILD_GRACE_PERIOD",
+                    &format!("{GRACE_PERIOD_SECS}s"),
+                )
+        })
+        .build()
+        .await
+        .unwrap();
+
+    let api_client = cluster.rest_v00();
+    let volume_api = api_client.volumes_api();
+
+    let volid = VolumeId::new();
+    let body = models::CreateVolumeBody::new(VolumePolicy::new(true), 2, 10485760u64, false, false);
+    let volume = volume_api.put_volume(&volid, body).await.unwrap();
+    let uid = volume.spec.uuid;
+
+    // Also provision a second volume that is never published. With no
+    // health_info_id the reconciler must never stand up a target for it, even
+    // while the first volume is actively being rebuilt.
+    let never_pub_id = VolumeId::new();
+    let never_pub_body =
+        models::CreateVolumeBody::new(VolumePolicy::new(true), 2, 10485760u64, false, false);
+    let never_pub = volume_api
+        .put_volume(&never_pub_id, never_pub_body)
+        .await
+        .unwrap();
+    let never_pub_uid = never_pub.spec.uuid;
+
+    // Publish to establish health_info_id (nexus info persisted to etcd).
+    let volume = volume_api
+        .put_volume_target(
+            &uid,
+            models::PublishVolumeBody::new_all(
+                HashMap::new(),
+                None,
+                cluster.node(0).to_string(),
+                models::VolumeShareProtocol::Nvmf,
+                None,
+                cluster.csi_node(0),
+                None,
+            ),
+        )
+        .await
+        .expect("Should publish volume");
+
+    assert_eq!(volume.state.status, VolumeStatus::Online);
+
+    // Unpublish — volume returns to no-target state.
+    volume_api
+        .del_volume_target(&uid, None, None)
+        .await
+        .expect("Should unpublish volume");
+
+    let volume = volume_api.get_volume(&uid).await.unwrap();
+    assert!(
+        volume.state.target.is_none(),
+        "Volume should have no target after unpublish"
+    );
+
+    // Stop a node hosting a replica — makes volume Degraded.
+    let replicas = api_client.replicas_api().get_replicas().await.unwrap();
+    let victim_replica = replicas
+        .iter()
+        .find(|r| r.node != cluster.node(0).to_string())
+        .expect("Should have a replica on a non-target node");
+    let victim_node = victim_replica.node.clone();
+
+    cluster
+        .composer()
+        .stop(&victim_node)
+        .await
+        .expect("Should stop io-engine node");
+
+    // Wait for Degraded status.
+    wait_till_volume_status(
+        &cluster,
+        &uid,
+        VolumeStatus::Degraded,
+        Duration::from_secs(10),
+    )
+    .await
+    .expect("Volume should become Degraded");
+
+    // Now wait for grace period + rebuild time. The reconciler should:
+    // 1. Wait for grace period (GRACE_PERIOD_SECS)
+    // 2. Create temporary unshared nexus
+    // 3. HotSpare rebuilds replica onto remaining healthy node's pool
+    // 4. Volume goes Online
+    // 5. Reconciler tears down temporary nexus
+    //
+    // We wait for the volume to return to Online with no target.
+    let timeout = Duration::from_secs(GRACE_PERIOD_SECS + REBUILD_TIMEOUT_SECS);
+    wait_till_volume_online_no_target(&cluster, &uid, timeout)
+        .await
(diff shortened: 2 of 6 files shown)


yugchaudhari:
Tested this on a 3-node cluster, both the BDD suite and a manual run.

BDD tests pass (`cargo test -p agents --test core offline_rebuild`): happy path, feature-disabled, and the never-published precondition.

For the manual run I created a 2-replica volume, published then unpublished it (to establish the health info), and killed one of the replica's io-engine nodes. The volume state over time:

```
Degraded  target=None                ← node killed, volume degraded
Degraded  target=io-engine-3/none    ← offline rebuild created the temp unshared nexus (protocol=none)
Online    target=None                ← rebuild done, nexus torn down, back to unpublished
Online    target=None                ← stays healed
```

And the reconciler's own logs line up with that:

```
16:39:39  DEBUG  Offline rebuild waiting for grace period, remaining: 7.97s
16:39:46  DEBUG  Offline rebuild waiting for grace period, remaining: 956ms
16:39:47  INFO   Initiating offline rebuild: creating non-shared nexus
16:39:47  INFO   Offline rebuild nexus created; HotSpareReconciler will handle the rebuild
16:39:50  INFO   Offline rebuild complete; tearing down temporary nexus
16:39:50  INFO   Temporary nexus destroyed; volume returned to unpublished state
```

So the full lifecycle works end to end: degraded unpublished volume → temp nexus → HotSpare rebuild → teardown → back to Online/unpublished.

control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
```diff
@@ -1,25 +1,39 @@
...
+/// can restore them. Once rebuild completes, tears the nexus down.
 ///
 /// See `designs/replicated-pv/mayastor/offline-volume-rebuild.md` for the
 /// full design.
 #[derive(Debug)]
-pub(super) struct OfflineRebuildReconciler {}
+pub(super) struct OfflineRebuildReconciler {
+    /// Tracks when each volume was first observed as degraded, used to
+    /// enforce the grace period before initiating rebuild.
+    degraded_since: HashMap<VolumeId, Instant>,
```
abhilashshetty04: I think this is not required. each child has `faulted_at` timestamp. `faulted_at` is updated by the io-engine layer. when nexus marks a child as `Faulted`. 

we use that in handle_faulted_child.
https://github.com/openebs/mayastor-control-plane/blob/develop/control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs#L192C34-L192C44

yugchaudhari: Right, `faulted_at` is what `handle_faulted_child` uses when there's a live nexus. The issue is offline rebuild fires on *unpublished* volumes, no nexus, no children, so no `child.faulted_at` (it's on [`Child`](https://github.com/openebs/mayastor-control-plane/blob/develop/control-plane/stor-port/src/types/v0/transport/child.rs#L21), not `Replica`). When the replica's node drops there's nothing recording when it went away, so `degraded_since` is from when the reconciler first sees the volume Degraded. It's the only signal I've got without a nexus.

They're also at different points in the lifecycle. `degraded_since` decides whether to create the temp nexus in the first place, as per the OEP we don't want to spin up one for a brief node blip. `faulted_at` only exists once a nexus has children, so it's about the wait *during* a rebuild, not whether to start one. Once the reconciler creates the nexus, `handle_faulted_child` takes over and uses `faulted_at` like it does right now, I have not reimplemented any of that, just covering the gap before a nexus exists.

The in-memory map does have a downside: it resets on agent restart, so the grace window starts over. If there's a replica-level timestamp that survives without a nexus, I'd rather use that, let me know if one exists and I'll switch. Otherwise tracking it ourselves seems like the only option for the no-nexus case, but I'm open if you see a cleaner way.

control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
```diff
@@ -33,40 +47,179 @@ impl TaskPoller for OfflineRebuildReconciler {
...
+        Default::default(),
+        vec![],
+        Default::default(),
+    );
+
+    match volume.publish(registry, &request).await {
+        Ok(_) => {
+            tracing::info!(
+                volume.uuid = %volume.uuid(),
+                "Offline rebuild nexus created; HotSpareReconciler will handle the rebuild"
```
abhilashshetty04, line 155: HotSpareReconciler mainly handles Full rebuild operation. Will it not be applicable for handle_faulted_children reconciler?

From the OEP,

`The existing nexus_reconciler will detect faulted children and trigger rebuild via handle_faulted_children (with a possibly shorter faulted_child_wait for offline rebuilds, since there is no I/O cost in starting immediately).`

In partial rebuild, when `faulted_child_wait` is elapsed w.r.t `child.faulted_at` duration we initiate Full rebuild by removing the Faulted child. It will be more costly if we start immediately. If child comes back while Full rebuild is going on we don't use that (we have plans to add that as future enhancement). It gets garbage collected.

Partial rebuild was introduced to ensure we reuse existing replica if it comes back. In this case, there wont be any write log as there is no IO. So if `Faulted` child comes back while its part of nexus we just mark it online. So, I suggest having longer `faulted_child_wait` period.

Apologies, I was on leave during OEP discussions.

abhilashshetty04:
@yugchaudhari , What happens if a publish request comes in when Offline rebuild is going on?

control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
abhilashshetty04: You are right, There wont be `faulted_at` timestamp here as it gets updated when nexus see IO errors on a Child. I think in handle_faulted_child reconciler we rely on faulted_at timestamp. As the reconciler was written for rebuild when nexus was present throughout the outage. We start Full rebuild if its not present. 

Can you please verify using your image if we start Full rebuild directly using following code path.

https://github.com/openebs/mayastor-control-plane/blob/develop/control-plane/agents/src/bin/core/controller/reconciler/nexus/mod.rs#L194

We may have to change that behaviour for an unshared nexus if we want to support Partial rebuild in this feature.

yugchaudhari: Verified on a cluster. The L194 path doesn't get hit, because the temp nexus never has a faulted child.

When the reconciler creates the nexus it only pulls in the healthy/accessible replicas, so it comes up with a single Online child, the dead replica isn't added. The missing replica is then restored by HotSpare's `volume_replica_count_reconciler`, which creates a fresh replica and adds it as an `OutOfSync` child that gets rebuilt. `handle_faulted_children` does nothing here, its `child.state.faulted()` filter matches nothing.

Log from the run:

```
06:50:21.235  Initiating offline rebuild: creating non-shared nexus
06:50:21.244  nexus created → children: [Child { state: Online, faulted_at: None }]   # 1 child, the healthy replica
06:50:22.259  hot_spare: Successfully created '1' new replica(s)
06:50:22.443  child added → state: Degraded, rebuild_progress: Some(0), state_reason: OutOfSync, has_io_log: Some(false)
06:50:24.451  Offline rebuild complete; tearing down temporary nexus
06:50:24.456  Temporary nexus destroyed; volume returned to unpublished state
```

So it is effectively a full rebuild (fresh replica, no log to partial-rebuild from), but via the replica-count path, not the faulted-child-removal at L194.

control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
yugchaudhari: Following from the above, since the dead replica never becomes a child of the temp nexus, `faulted_child_wait` doesn't really come into play for this flow right now. There's no faulted child sitting in the nexus to wait on; we just build a fresh replica on a surviving pool.

Your point about reusing the original replica when its node comes back is fair, and it'd avoid a full rebuild. But that's a new behavior rather than a wait tweak, we'd have to add the faulted replica as a child up front and then lean on the wait window. I'd keep that out of this iteration: the current flow is correct, just not optimal when the node returns quickly. I'll track it as a follow-up.

yugchaudhari:
> @yugchaudhari , What happens if a publish request comes in when Offline rebuild is going on?

The publish just shares the existing temp nexus instead of making a new one, the publish flow already hits the "target_cfg exists" branch and shares it, so the publish goes through and the rebuild carries on.

There's a gap here. The temp nexus is unshared (`protocol: None`), and right now nothing flips it to the real protocol when it gets promoted. So once the volume goes Online the offline-rebuild reconciler still thinks the nexus is its own and tears it down. Fixing that, set the protocol on promote so the reconciler leaves it alone, plus the matching check on the publish path, is the concurrent-publish piece I am planning for the next iteration. This PR is just the create/rebuild/teardown lifecycle.

I did expect those changes in this PR. Went through the OEP, thought that it was not covered there so asked here.

control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
tiagolobocastro: What happens when a volume is deleted?
Can we move this to the volume's runtime information instead?

control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
```diff
@@ -33,40 +47,179 @@ impl TaskPoller for OfflineRebuildReconciler {
...
+            );
         }
         Self::squash_results(results)
     }
 }
 
 #[tracing::instrument(
     level = "debug",
-    skip(context, volume_spec),
+    skip(context, volume_spec, degraded_since),
```
tiagolobocastro:
```suggestion
    skip_all,
```

control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
```diff
@@ -33,40 +47,179 @@ impl TaskPoller for OfflineRebuildReconciler {
...
         return PollResult::Ok(PollerState::Idle);
     }
 
-    let volume_state = context.registry().volume_state(volume.uuid()).await?;
-    if volume_state.status == VolumeStatus::Degraded {
+    // Must use the health-aware state: for an unpublished volume the plain
+    // volume_state() derives status from the replica spec count (unchanged when a
+    // node dies), so it reports Online and we'd never detect the degradation.
+    // volume_state_health() factors in actual online replica health.
+    let volume_state = context
```
tiagolobocastro, line 90: can we move this down, since it's not needed if we return via L92

control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
```diff
@@ -33,40 +47,179 @@ impl TaskPoller for OfflineRebuildReconciler {
...
+                        volume.uuid = %volume.uuid(),
+                        %error,
+                        "Failed to tear down offline rebuild nexus"
+                    );
+                }
+            }
+            PollResult::Ok(PollerState::Idle)
+        }
+        VolumeStatus::Degraded => {
+            // Rebuild still in progress, HotSpare is handling it.
```
tiagolobocastro: This assumption may not be entirely accurate, the volume may be degraded due to lack of replicas
I think here we'd perhaps need to check for the target state?

control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
```diff
@@ -33,40 +47,179 @@ impl TaskPoller for OfflineRebuildReconciler {
...
+            PollResult::Ok(PollerState::Busy)
```
tiagolobocastro: I think we can return idle here, making the reconciler faster won't help the rebuild itself complete

control-plane/agents/src/bin/core/tests/volume/offline_rebuild.rs
```diff
@@ -0,0 +1,276 @@
...
+
+/// A volume that was never published has no health_info_id, so the offline
+/// rebuild reconciler must skip it: it never creates a temporary nexus.
+///
+/// Note: a never-published volume's status is always reported Online (status is
+/// derived from existing replica specs, with no nexus health baseline to compare
+/// against), so killing a replica node does not move it to Degraded. The point of
+/// this test is simply that the reconciler never stands up a target for it.
+#[tokio::test]
+async fn offline_rebuild_skips_never_published_volume() {
```
tiagolobocastro: can we put this in the first test to avoid increasing ci time too much

control-plane/agents/src/bin/core/tests/volume/offline_rebuild.rs
```diff
@@ -0,0 +1,276 @@
...
+    let timeout = Duration::from_secs(GRACE_PERIOD_SECS + REBUILD_TIMEOUT_SECS);
+    wait_till_volume_online_no_target(&cluster, &uid, timeout)
+        .await
+        .expect("Volume should be Online with no target after offline rebuild");
+}
+
+/// When offline rebuild is disabled (default), a degraded unpublished volume
+/// should remain Degraded indefinitely — no temporary nexus is created.
+#[tokio::test]
+async fn offline_rebuild_disabled_no_action() {
```
tiagolobocastro: I think not needed here, we can add an e2e test if required, to avoid increasing CI time here

control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
yugchaudhari: Moved it to VolumeMetadata.runtime. It now sits next to data_loss_warned, same #[serde(skip)] runtime-only pattern. The map is gone; degraded_since lives on the volume now, so it's cleaned up automatically when the volume is deleted.

control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
yugchaudhari: Done. Moved it below the early bails. We don't hit volume_state_health now when we'd just return at the shared-target or no-health-info check anyway.

control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
yugchaudhari: Yes, status alone wasn't enough. I've reworked teardown_if_rebuilt to look at the target itself. It now takes &VolumeState (not just status) and checks the nexus children: if children.len() < num_replicas or any child has rebuild_progress, we're still making progress so we wait. Otherwise (Degraded but no progress, all children present) it's stuck and we tear the nexus down so the next reconcile cycle can try fresh.

control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
```diff
@@ -41,32 +52,186 @@ impl TaskPoller for OfflineRebuildReconciler {
...
+        return PollResult::Ok(PollerState::Idle);
     }
 
+    tracing::warn!(
+        volume.uuid = %volume.uuid(),
+        status = ?volume_state.status,
+        "Offline rebuild nexus is not making progress; cleaning up"
+    );
+    let request = UnpublishVolume::new(volume.uuid(), true, vec![]);
+    let _ = volume.unpublish(context.registry(), &request).await;
```
Abhinandan-Purkait: I wonder if this could end up in a situation of creating and destroying offline nexus again and again?

Abhinandan-Purkait: For example, if there is node blip this will teardown the in progress nexus right?

Abhinandan-Purkait:
Also, how does the reconciler distinguish between an user manual unshared nexus vs an offline rebuilder nexus? Can it then teardown the nexus in the first case if no progress or if the volume becomes online(i.e fully rebuilt) potentially breaking the volume. Although I am not really sure if that can be done by the user unless they use direct rest calls. Just wanted to point it out. wdyt @tiagolobocastro ?

control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
yugchaudhari: Yeah, on a genuinely stuck volume (e.g., no surviving pool to host the replacement replica), tearing down would just trigger the grace timer to start over and we'd recreate on the next round. Dropped the destructive non-Online branch, only Online tears down now; anything else waits, with a warn log when the children-count and rebuild_progress show no forward progress so the operator has a signal. Stuck volume stays put with its target visible. This also covers the transient node blip case, only Online state triggers nexus tear-down now.

yugchaudhari:
> Also, how does the reconciler distinguish between an user manual unshared nexus vs an offline rebuilder nexus? Can it then teardown the nexus in the first case if no progress or if the volume becomes online(i.e fully rebuilt) potentially breaking the volume. Although I am not really sure if that can be done by the user unless they use direct rest calls. Just wanted to point it out. wdyt @tiagolobocastro ?

You're right, we don't actually distinguish them. The only signal is `target.protocol().is_none()`, and yes, a user can produce that via REST publish with `protocol: None`. After the change above (only Online tears down) the immediate destructive risk narrows, but a user-created unshared nexus that reaches Online would still get torn down by us.

The proper fix is an explicit ownership marker, set on the volume when we create the temp nexus, cleared on teardown, reconciler only acts on its own. The next iteration already needs the same flag for promote-on-publish (it has to know "this is the offline-rebuild nexus" to flip the protocol when a real publish arrives), so I'd be adding the same field either way.

Do you want it added here, or fine to land with the iteration3 promote-on-publish? Either works on my end.

control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
yugchaudhari: Yeah, no point polling harder when we can't speed up the rebuild itself. Returns Idle now.

control-plane/agents/src/bin/core/tests/volume/offline_rebuild.rs
yugchaudhari: Done, dropped the test.

control-plane/agents/src/bin/core/tests/volume/offline_rebuild.rs
yugchaudhari: Done, merged this with the first test as a parallel never-published volume + an assert that its target stays None after the rebuild.

control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
yugchaudhari: Done, switched to skip_all.

Abhinandan-Purkait:
> > Also, how does the reconciler distinguish between an user manual unshared nexus vs an offline rebuilder nexus? Can it then teardown the nexus in the first case if no progress or if the volume becomes online(i.e fully rebuilt) potentially breaking the volume. Although I am not really sure if that can be done by the user unless they use direct rest calls. Just wanted to point it out. wdyt @tiagolobocastro ?
> 
> You're right, we don't actually distinguish them. The only signal is `target.protocol().is_none()`, and yes, a user can produce that via REST publish with `protocol: None`. After the change above (only Online tears down) the immediate destructive risk narrows, but a user-created unshared nexus that reaches Online would still get torn down by us.
> 
> The proper fix is an explicit ownership marker, set on the volume when we create the temp nexus, cleared on teardown, reconciler only acts on its own. The next iteration already needs the same flag for promote-on-publish (it has to know "this is the offline-rebuild nexus" to flip the protocol when a real publish arrives), so I'd be adding the same field either way.
> 
> Do you want it added here, or fine to land with the iteration3 promote-on-publish? Either works on my end.

Sounds, good. Upto you, how you want to organise, just wanted to point it out.

Abhinandan-Purkait:
Also, on a side note, just like to highlight one thought.

With this feature wouldn't we starve the published nexus rebuilds, as the max rebuilds is constricted. Would it make sense to decide this based on published degraded vs max rebuilds and prefer published degraded. But might not be that straightforward. Its fine at the moment. We can think about optimization once the feature is done. Thanks!

yugchaudhari:
> > > Also, how does the reconciler distinguish between an user manual unshared nexus vs an offline rebuilder nexus? Can it then teardown the nexus in the first case if no progress or if the volume becomes online(i.e fully rebuilt) potentially breaking the volume. Although I am not really sure if that can be done by the user unless they use direct rest calls. Just wanted to point it out. wdyt @tiagolobocastro ?
> > 
> > 
> > You're right, we don't actually distinguish them. The only signal is `target.protocol().is_none()`, and yes, a user can produce that via REST publish with `protocol: None`. After the change above (only Online tears down) the immediate destructive risk narrows, but a user-created unshared nexus that reaches Online would still get torn down by us.
> > The proper fix is an explicit ownership marker, set on the volume when we create the temp nexus, cleared on teardown, reconciler only acts on its own. The next iteration already needs the same flag for promote-on-publish (it has to know "this is the offline-rebuild nexus" to flip the protocol when a real publish arrives), so I'd be adding the same field either way.
> > Do you want it added here, or fine to land with the iteration3 promote-on-publish? Either works on my end.
> 
> Sounds, good. Upto you, how you want to organise, just wanted to point it out.

Cool, will add it with iteration 3 then. Thanks for flagging it.

yugchaudhari:
> Also, on a side note, just like to highlight one thought.
> 
> With this feature wouldn't we starve the published nexus rebuilds, as the max rebuilds is constricted. Would it make sense to decide this based on published degraded vs max rebuilds and prefer published degraded. But might not be that straightforward. Its fine at the moment. We can think about optimization once the feature is done. Thanks!

Fair point, max_rebuilds is shared today, so an offline rebuild can grab a slot a degraded-published volume actually needs. Some prioritization (published-degraded first) would help. I'll capture it as a follow-up optimization so it doesn't get lost.

control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
abhilashshetty04: @yugchaudhari, I wonder if its worth attempting to create a nexus (unshared) with Faulted child for some period of time. I think creation will work but attach will fail for the Faulted Child. Please try this scenario. If we elapse the period at which we check if Faulted child comes back we can create Nexus without Faulted child.

yugchaudhari: Tried it on a cluster. io-engine rejects the whole nexus creation when any child URI is unreachable, atomic rollback, no nexus exists afterwards.

With both children (one live local replica + one nvmf URI pointing at the killed node):

```
PUT /v0/nodes/io-engine-1/nexuses/{id}
body: { children: [<live-bdev-uri>, <dead-nvmf-uri>], size: ... }

→ HTTP 400
   create_nexus: Failed to create child of nexus ...:
     Failed to create a BDEV '10.1.0.7:8420/nqn.2019-05.io.openebs:<uuid>'
   errno: 6 (ENXIO)
```

Same request with only the live URI succeeds (HTTP 200, child `Online`), so it's the unreachable child that's the blocker, not the call shape.

So at today's io-engine layer it's not "create succeeds, attach fails for the faulted child", it's "create fails, no nexus." For the approach you suggested to work, io-engine would need to allow nexus creation with a child whose attach failed (entering the nexus as `Faulted` with `faulted_at` set, so `handle_faulted_child` can take over). That's a data-plane change and overlaps with the Iter 4 work in the OEP plan ("Nexus code change for unshared clean-shutdown semantics" on the io-engine side). Until io-engine supports that, building the nexus from healthy replicas only and letting the replica-count reconciler add a fresh one is the only working path.

control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
```diff
@@ -41,32 +52,186 @@ impl TaskPoller for OfflineRebuildReconciler {
...
+        None => false,
+    };
+
+    if !progressing {
+        tracing::warn!(
+            volume.uuid = %volume.uuid(),
+            status = ?volume_state.status,
+            "Offline rebuild nexus shows no forward progress; leaving in place for operator visibility"
+        );
+    }
```
tiagolobocastro: I don't think this is needed, any replica fault etc should be handled via the hot spare reconcilers, so we don't need duplicate logs here?

```suggestion
```

control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
```diff
@@ -41,32 +52,186 @@ impl TaskPoller for OfflineRebuildReconciler {
...
+        return PollResult::Ok(PollerState::Idle);
+    }
+
+    let now = Instant::now();
+    volume.lock().metadata.mark_offline_rebuild_degraded(now);
+    let first_seen = volume
+        .as_ref()
+        .metadata
+        .offline_rebuild_degraded_since()
+        .unwrap_or(now);
```
tiagolobocastro: nit: would be nice to have this done by the function itself, so the caller becomes simply:

```suggestion
    let degraded_for = volume.lock().metadata.offline_rebuild_degraded();
```

tiagolobocastro (approved):

control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
tiagolobocastro: > Yeah, on a genuinely stuck volume (e.g., no surviving pool to host the replacement replica), tearing down would just trigger the grace timer to start over and we'd recreate on the next round. 

I think we could tackle this by adding another check, that we only create nexus if there's a replica to rebuild from, then we should be able to tear down, which I think we should do as there's no need to keep resources busy without a client wanting the volume
Maybe something else to add to the work queue :)

control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
yugchaudhari: Done, dropped the progressing check and the warn. Kept a trimmed comment explaining why we only act on Online.

control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
yugchaudhari: Done. Added offline_rebuild_degraded() that does the mark + return-elapsed in one shot, caller is the one-liner you suggested. Also removed the now-unused mark_offline_rebuild_degraded and offline_rebuild_degraded_since since nothing else calls them.

control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
yugchaudhari: Agreed, that's the cleaner shape. Once we know the rebuild is viable up front, the tear-down on stuck becomes safe (no churn, no dangling resources). Placing this into Iteration 5 (production-readiness) alongside the concurrency-limit / grace-period, since both benefit from the same "is this rebuild going to succeed?" precondition.

Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

yugchaudhari:
> bors merge

@tiagolobocastro fixed the lint error.

tiagolobocastro:
bors merge

yugchaudhari:
The failing bdd-ci checks look unrelated to this PR, all failures are in pool tests ('Deleted' is not a valid PoolStatus, deployer start flakes, 504/409 on pool delete). My changes are in the offline-rebuild reconciler and volume runtime metadata, no pool code touched. int-ci (which runs the offline_rebuild integration tests) is passing cleanly.

tiagolobocastro:
Raised https://github.com/openebs/mayastor-control-plane/pull/1111 to fix the flaky test
Meanwhile we can try again:
bors merge

tiagolobocastro:
bors merge

PR #972, opened 2025-05-01
DiskPool Operator: add quantity for a nicer print

tiagolobocastro:
feat(operator/dsp): add quantity for a nicer print

    Adds quantity type for pool capacities.
    These are then used for CRD printcolumns, showing a nicer humansize
    value, rather than a large byte number.
    This value is the same as the kubectl-plugin one.
    Also fix a bug where no space was added between value and unit.

---

    fix(k8s/proxy): add phase to conditions check

PR #1109, opened 2026-05-29
fix(operator): compare disk pool quantities by bytes

immanuwell:
follow-up to #972

what
`Quantity` equality was using the rounded display string too. That makes different byte values compare equal, eg `1 GiB` and `1 GiB + 25 MiB`.

why this is real
DiskPool status comes from raw byte counters, and pool blobstore cluster size is `4 MiB` by default. So this is kinda easy to hit in practice, not some impossible edge case.

repro
1. Before this patch, `Quantity::from_bytes(1073741824) == Quantity::from_bytes(1099956224)` was `true`.
2. `ResourceContext::set_status_or_unknown` only patches when `status != new_status`.
3. A real change in `used_q`, `available_q`, or `capacity_q` could get skipped.

fix
Compare `Quantity` by raw bytes only, and keep the rounding behavior only for display and serde.
Added regression tests for exact equality and for the lossy string roundtrip.

test
`cargo test -p k8s-operators`

### DIFF
--- a/k8s/operators/src/pool/diskpool/crd/quantity.rs
+++ b/k8s/operators/src/pool/diskpool/crd/quantity.rs
@@ -2,7 +2,7 @@ use schemars::JsonSchema;
 use std::convert::TryFrom;
 
 /// A quantity represented as a humansize with up to 1 floating point.
-#[derive(Default, Debug, Clone, Copy)]
+#[derive(Default, Debug, Clone, Copy, Eq, PartialEq)]
 pub struct Quantity(pub u64);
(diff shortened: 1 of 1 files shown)


tiagolobocastro (commented):
IIRC that had been intentional, since the value is stored as a String anyway.
Also this value is only used for "pretty printing", the raw bytes from the non `_q` variants would detect the changes?

abhilashshetty04:
@immanuwell , just checking in—are you still planning to work on this PR? It's been inactive for a while.

PR #1110, opened 2026-05-31
refactor(rest): streamline TLS and JWT authentication across clients

tiagolobocastro:
- Change JWT parameters from raw string tokens to file paths throughout:
  * CSI node and controller drivers
  * Pool operator
  * REST plugin
  * Test/deployer utilities
  * This defers token reading to the openapi client layer

- Simplify TLS path API:
  * TlsMode::new now accepts Option<&PathBuf> instead of &Option<PathBuf> to avoid unnecessary temporary clones
  * CsiControllerConfig getters return Option<&PathBuf> for consistency

- Reduce duplication in CSI node client:
  * Share single Configuration across AppNodes and Volumes clients
  * Remove duplicated TLS/security setup logic
  * Both wrappers now take Configuration by value (cheap clone internally)

- Add kube-forward secret discovery infrastructure:
  * Pod tls and auth discovery via annotations and volumes
  * Fixed SecretRef keys: ca.crt, tls.crt, tls.key, jwt

- Add auto-tls generation for HTTPS TLS without verification

This will allow our plugin to discover the cert and auth information from a running cluster.
Also we switch from the kube-apiserver proxy to port forwarding since otherwise we can't use HTTPS with verification.
This is because the kube-apiserver proxy terminates the request and issues a new request to the target pod, meaning our certificates or header bearer tokens cannot be send across.
So, the kube-apiserver proxy will only for in case of AutoTLS with no auth. If no TLS is configured, then we revert back to HTTP fully.

A flag has been added to the rest-client to force http usage. Eventually we can disable http binding from the rest service by default.

Also worth noting that this is only tackling the rest interface, and not the internal service gRPC nor the data-plane gRPC.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,4 @@
 {
-    "nixEnvSelector.nixFile": "${workspaceRoot}/shell.nix"
+    "nixEnvSelector.nixFile": "${workspaceFolder}/shell.nix",
+    "nixEnvSelector.useFlakes": false
 }
--- a/openapi/templates/default/tower-hyper/client/configuration.mustache
+++ b/openapi/templates/default/tower-hyper/client/configuration.mustache
@@ -5,7 +5,7 @@ pub type ReqBody = hyper_body::Body;
 pub use hyper::{body, Request, Response, Uri};
 
 use hyper_rustls::ConfigBuilderExt;
-use std::{sync::Arc, time::Duration};
+use std::{path::PathBuf, sync::Arc, time::Duration};
 use tokio::sync::Mutex;
 use tower::{util::BoxCloneService, Layer, Service, ServiceExt};
 
@@ -41,6 +41,9 @@ pub struct ApiKey {
 pub enum Error {
     NativeRootCerts(std::io::Error),
     Certificate,
+    ClientKey,
+    ClientIdentity,
+    ClientAuth(rustls::Error),
     TlsConnector,
     NoTracingFeature,
     UrlToUri(hyper::http::uri::InvalidUri),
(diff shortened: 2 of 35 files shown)


openapi/templates/default/tower-hyper/client/configuration.mustache
```diff
@@ -41,35 +41,162 @@ pub struct ApiKey {
...
+    /// TLS mode for HTTPS connections.
+    pub tls: TlsMode,
+    /// Discovery of properties. 
+    /// If true, the client will attempt to discover the required TLS and Auth configuration
+    pub discover: bool,
+}
+impl ClientSecurity {
+    /// Creates a new client security configuration.
+    pub fn try_new(jwt: &Option<PathBuf>, tls: TlsMode) -> Result<Self, std::io::Error> {
+        let jwt = jwt.as_ref().map(std::fs::read_to_string).transpose()?;
```
Abhinandan-Purkait, line 107: You probably need to trim the read content to avoid errors due to space or endline.

tiagolobocastro: Shouldn't we expect the jwt to be correct already?

Abhinandan-Purkait: Yes, ideally it should be. But it's always a good idea to kind of trim after reading contents from file to remove unnecessary characters. Just a nit.

Abhinandan-Purkait (approved):
Looks good overall.

openapi/templates/default/tower-hyper/client/configuration.mustache
tiagolobocastro: Alright, I'll put trim as will

tiagolobocastro: Done

.vscode/settings.json
niladrih, line 1: EOF should have a newline before it

tiagolobocastro:
bors merge

niladrih (approved):
LGTM

.vscode/settings.json
tiagolobocastro: Done

tiagolobocastro:
bors merge

PR #1111, opened 2026-06-02
test(pool): don't starve other io-engines

tiagolobocastro:
On CI this test failed because import pool was taking 7s seconds 
and the import was still not complete!!
It looks like the issue is that the heartbeat, when set to 0, tries to
do it in a busy loop and on CI this seems to lead into one of the
io-engines not allowing the other to run! I'm not sure what kind of
scheduler is setup there, but perhaps it's not yielding or something.

So we now ensure the heartbeat period is every second.
todo: change io-engine to allow sub-second heartbeats for testing.


tiagolobocastro:
bors try

Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

niladrih (approved):

tiagolobocastro:
bors merge

PR #1112, opened 2026-06-03
feat(core): handle CSI publish promotion on offline-rebuild volumes

yugchaudhari:
Iteration 3 of offline volume rebuild ([openebs/openebs#4208](https://github.com/openebs/openebs/issues/4208)), builds on iteration 2 ([#1108](https://github.com/openebs/mayastor-control-plane/pull/1108)) which handles the basic create/rebuild/teardown lifecycle.

When a CSI publish arrives while an offline rebuild is in progress, the publish flow now promotes the existing unshared nexus, shares it via the existing `share_nexus`, flips the protocol on the spec instead of trying to create a new nexus. The rebuild keeps going on the now-shared nexus, the reconciler stops treating it as its own, and the volume becomes a normal published volume.

To support this safely, the volume metadata records the *purpose* of the current target via a `VolumeTargetMode` enum (`ServeFrontendApp`, `OfflineRebuild`, with `MaintenanceMode` reserved for follow-up work). The reconciler's identify-our-nexus check and the publish flow's promotion path both gate on `OfflineRebuild`, so a user-created unshared nexus produced via direct REST publish (which stays at the default `ServeFrontendApp`) never gets touched. The mode is kept off the public `PublishVolume` API, internal callers like the reconciler go through `OperationGuardArc::<VolumeSpec>::publish_with_mode` instead, which records the mode on the resulting `target_config`.

BDD tests cover the promote-on-publish path and GC safety during rebuild; the existing iteration 2 happy path still passes.

### DIFF
--- a/control-plane/agents/src/bin/core/tests/volume/offline_rebuild.rs
+++ b/control-plane/agents/src/bin/core/tests/volume/offline_rebuild.rs
@@ -1,11 +1,11 @@
 #![cfg(test)]
 
-use super::helpers::wait_till_volume_status;
+use super::helpers::{wait_node_online, wait_till_volume_status};
 use deployer_cluster::ClusterBuilder;
 use std::{collections::HashMap, time::Duration};
 use stor_port::types::v0::{
     openapi::models::{self, VolumePolicy, VolumeStatus},
-    transport::VolumeId,
+    transport::{NodeId, VolumeId},
 };
 use uuid::Uuid;
 
@@ -13,13 +13,25 @@ const RECONCILE_PERIOD_MS: u64 = 250;
 const GRACE_PERIOD_SECS: u64 = 2;
 const REBUILD_TIMEOUT_SECS: u64 = 30;
 
-/// Happy path: create a 2-replica volume, publish to establish health_info,
-/// unpublish, stop one io-engine node → volume becomes Degraded.
-/// With offline rebuild enabled, the reconciler creates a temp nexus,
-/// HotSpare rebuilds the replica, then the reconciler tears down the nexus.
-/// Final state: volume Online with no target (unpublished).
+/// End-to-end coverage for the offline-rebuild reconciler. Drives three
+/// scenarios sequentially on a single shared cluster (each scenario takes a
+/// node down and brings it back up before handing off), which avoids the
+/// per-test cluster spin-up cost while still exercising each path on a clean
+/// pool layout:
+///
+/// 1. **Happy path**: degraded unpublished volume gets a temp nexus, the
+///    rebuild runs, the temp nexus is torn down, the volume returns to
+///    Online with no target. A second volume that was never published must
+///    not be touched.
+/// 2. **Promote-on-publish**: while the temp nexus exists, a CSI publish
+///    promotes the existing unshared nexus instead of creating a new one.
+///    The rebuild keeps going on the now-shared nexus and the volume reaches
+///    Online still published.
+/// 3. **GC safety**: during an active offline rebuild, the GarbageCollector
+///    must not strip ownership from the surviving healthy replicas of the
+///    volume being rebuilt.
 #[tokio::test]
-async fn offline_rebuild_happy_path() {
+async fn offline_rebuild_e2e() {
     let reconcile = Duration::from_millis(RECONCILE_PERIOD_MS);
     let cluster = ClusterBuilder::builder()
         .with_rest(true)
@@ -39,6 +51,13 @@ async fn offline_rebuild_happy_path() {
         .await
         .unwrap();
 
+    happy_path(&cluster).await;
+    promote_on_publish(&cluster).await;
+    gc_safety(&cluster).await;
+}
+
+/// Happy path scenario, see [`offline_rebuild_e2e`] doc.
+async fn happy_path(cluster: &deployer_cluster::Cluster) {
     let api_client = cluster.rest_v00();
     let volume_api = api_client.volumes_api();
 
@@ -78,7 +97,6 @@ async fn offline_rebuild_happy_path() {
 
     assert_eq!(volume.state.status, VolumeStatus::Online);
 
-    // Unpublish — volume returns to no-target state.
     volume_api
         .del_volume_target(&uid, None, None)
         .await
@@ -90,23 +108,16 @@ async fn offline_rebuild_happy_path() {
         "Volume should have no target after unpublish"
     );
 
-    // Stop a node hosting a replica — makes volume Degraded.
-    let replicas = api_client.replicas_api().get_replicas().await.unwrap();
-    let victim_replica = replicas
-        .iter()
-        .find(|r| r.node != cluster.node(0).to_string())
-        .expect("Should have a replica on a non-target node");
-    let victim_node = victim_replica.node.clone();
-
+    // Stop a node hosting a replica, this makes the volume Degraded.
+    let victim_node = pick_victim_node(&api_client, &cluster.node(0).to_string(), uid).await;
     cluster
         .composer()
         .stop(&victim_node)
         .await
         .expect("Should stop io-engine node");
 
-    // Wait for Degraded status.
     wait_till_volume_status(
-        &cluster,
+        cluster,
         &uid,
         VolumeStatus::Degraded,
         Duration::from_secs(10),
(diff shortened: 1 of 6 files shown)


abhilashshetty04:
Hi @yugchaudhari , Are we introducing `offline_rebuild_owned` to make sure we don't cleanup Offline rebuild reconciler created nexus which is now shared after csi PublishVolume ?

Why do we want to distinguish user created non-shared nexus and offline rebuild nexus?

yugchaudhari:
> Hi @yugchaudhari , Are we introducing `offline_rebuild_owned` to make sure we don't cleanup Offline rebuild reconciler created nexus which is now shared after csi PublishVolume ?
> 
> Why do we want to distinguish user created non-shared nexus and offline rebuild nexus?

Correct, but the bigger reason is that `protocol == None` on its own isn't a reliable signal for the reconciler to know this nexus is "created by me". A user can produce the same shape via a direct REST publish with `protocol: None`, and without an ownership flag the reconciler would tear that nexus down as soon as the volume goes Online, on a nexus it did not create. @Abhinandan-Purkait flagged this on iteration 2 (openebs/mayastor-control-plane#1108).

Promote-on-publish uses the same flag. When a CSI publish lands mid-rebuild and `target_config` already exists with `protocol: None`, the publish path needs to flip the protocol on the existing nexus instead of creating a new one. The marker is how that path knows "this target_config came from the offline-rebuild reconciler, it's safe to promote", and it gets cleared as part of the promotion so once the publish takes over the reconciler stops tracking it.

So you're right on the promotion case, but the main thing it solves is the user-created unshared nexus case.

control-plane/stor-port/src/types/v0/transport/volume.rs
```diff
@@ -639,6 +639,13 @@ pub struct PublishVolume {
     pub frontend_nodes: Vec<String>,
     /// The access mode (a la csi) for the volume.
     pub access_mode: VolumeAccessMode,
+    /// Internal flag set by the offline-rebuild reconciler when creating its
+    /// temporary unshared nexus. The publish flow records it on the volume so
+    /// the reconciler can later identify *its* nexus, separate from any
+    /// unshared nexus a user could produce via direct REST publish.
+    /// Not exposed over the wire.
+    #[serde(skip)]
+    pub offline_rebuild_owned: bool,
```
tiagolobocastro: If the reason for this change is to ensure a REST publish by a user can't lead into the core agent thinking the volume is in offline-rebuild then we're not truly fixing that since we're again opening up that possibility here?

How about keeping `PublishVolume` as is and adding the flag only to the core agent?
We could either add a context to the publish interface or add a new-type like PublishVolume containing both the public PublishVolume and the bool

tiagolobocastro: Also somewhat related to this, we wanted to have a maintenance mode which would make the volume available for some tasks like fs recovery etc and would prevent any app from publishing the volume until maintenance or was.
So about having an enum rather than bool, like `Option<VolumeTargetMode>`:

```rust
enum VolumeTargetMode {
    ServeFrontendApp,
    MaintnanceMode,
    OfflineRebuild
}
```

?

tiagolobocastro: In which case if we have enum, then does it make sense to keep it publicly exposed in the `PublishVolume` or would we expose separate api for entering maintenance mode?

CC @Abhinandan-Purkait

yugchaudhari: Yes, all three are fair.

On (1) and (2): I'll drop the flag from public `PublishVolume` and wrap it internally at the core agent, probably a small new-type around the request plus the mode, will see what reads cleanest at the call sites. And yes, bool will age badly with maintenance mode coming, so will switch to a `VolumeTargetMode` enum (`ServeFrontendApp` / `MaintenanceMode` / `OfflineRebuild`). I'll persist it on `VolumePersistedMetadata` as `target_mode` with default `ServeFrontendApp` so existing data still round-trips.

On (3): I agree on separating maintenance mode out from `PublishVolume`. OfflineRebuild is reconciler-internal anyway, it never comes in through the public API. Maintenance mode is operator-initiated and probably deserves its own entry point.

Will rework the PR based on this.

yugchaudhari: Reworked and pushed. `PublishVolume` is flag-free now, the offline-rebuild reconciler goes through `publish_with_mode(VolumeTargetMode::OfflineRebuild)` on the `OperationGuardArc` instead. Bool replaced by `VolumeTargetMode` enum with `MaintenanceMode` reserved for future.

Abhinandan-Purkait: I think keeping a seperate API makes more sense for maintainence. But the TargetMode does sounds good. Thanks!

Abhinandan-Purkait (approved):
LGTM

control-plane/agents/src/bin/core/tests/volume/offline_rebuild.rs
```diff
@@ -137,6 +137,242 @@ async fn offline_rebuild_happy_path() {
...
+    cluster.composer().stop(&victim_node).await.unwrap();
+
+    // Wait for the offline-rebuild reconciler to create the unshared nexus.
+    let unshared_target = wait_for_unshared_target(&cluster, &uid, Duration::from_secs(15))
+        .await
+        .expect("Offline rebuild should create unshared target");
+    let rebuild_node = unshared_target.node.clone();
+
+    // CSI publish arrives mid-rebuild — should promote (share the existing nexus).
+    let publish_ctx = HashMap::from([("io.nvme.timeout".to_string(), "30".to_string())]);
```
tiagolobocastro: Why do we need a timeout here

control-plane/agents/src/bin/core/tests/volume/offline_rebuild.rs
```diff
@@ -137,6 +137,242 @@ async fn offline_rebuild_happy_path() {
...
+    assert!(
+        volume.state.target.is_some(),
+        "Volume stays published after promotion"
+    );
+}
+
+/// GC safety: during an offline rebuild, the GarbageCollector must not disown
+/// the live (healthy) replicas. Only the faulted one (post-rebuild) may be cleaned up.
+#[tokio::test]
+async fn offline_rebuild_gc_safety() {
```
tiagolobocastro: Please try not to create more clusters than needed and try to reuse them instead, since there's a timing cost for creating new cluster and all these tests run sequentially at the moment

control-plane/agents/src/bin/core/tests/volume/offline_rebuild.rs
```diff
@@ -137,6 +137,242 @@ async fn offline_rebuild_happy_path() {
...
+    // Wait until offline rebuild creates the unshared nexus.
+    wait_for_unshared_target(&cluster, &uid, Duration::from_secs(15))
+        .await
+        .expect("Offline rebuild should start");
+
+    // Sleep briefly to let GC poll at least once with the rebuild active.
+    tokio::time::sleep(Duration::from_secs(3)).await;
+
+    // Confirm live replicas are still owned by the volume.
+    let post_replicas = api_client.replicas_api().get_replicas().await.unwrap();
```
tiagolobocastro: I'm confused, what is this actually checking for?

control-plane/stor-port/src/types/v0/store/volume.rs
```diff
@@ -339,6 +362,14 @@ pub struct VolumePersistedMetadata {
...
     #[serde(skip_serializing_if = "Option::is_none")]
     snapshot_as_thin: Option<bool>,
+    /// Purpose the current `target_config` is serving. Used by the
+    /// offline-rebuild reconciler to safely identify its own nexus, by the
+    /// promote-on-publish path to decide whether to flip the protocol on an
+    /// incoming CSI publish, and as the extension point for future
+    /// `MaintenanceMode` flows. Defaults to `ServeFrontendApp` so existing
+    /// persisted state round-trips unchanged.
+    #[serde(default, skip_serializing_if = "VolumeTargetMode::is_default")]
+    target_mode: VolumeTargetMode,
```
tiagolobocastro: hmm shouldn't this exist in the VolumeTarget itself?

yugchaudhari: Moved to `VolumeTarget`. The mode now lives on the target via `serde(flatten)` in `TargetConfig`; accessors on `VolumeSpec` proxy through. `deactivate_target` no longer needs an explicit reset since the mode dies with the target on the next publish.

control-plane/agents/src/bin/core/tests/volume/offline_rebuild.rs
yugchaudhari: The old assertion only caught destruction, not disown. Tightened to verify ownership. The new check looks up the replica in specs_api().get_specs().replicas and asserts owners.volume == Some(uid). A disown-without-destroy would have slipped past the previous check.

control-plane/agents/src/bin/core/tests/volume/offline_rebuild.rs
yugchaudhari: Merged all three tests into a single `offline_rebuild_e2e`. Three phases run sequentially on the same cluster: happy_path, promote_on_publish and gc_safety. Each phase restarts the stopped node before handing off, so the next phase starts from the full 3-node pool.

control-plane/agents/src/bin/core/tests/volume/offline_rebuild.rs
yugchaudhari: That was an `io.nvme.timeout` in the CSI publish_context, not a test-level timeout. Not needed for this test, dropped it.

control-plane/agents/src/bin/core/tests/volume/offline_rebuild.rs
```diff
@@ -90,51 +108,290 @@ async fn offline_rebuild_happy_path() {
...
+/// Restart a node previously stopped by `composer().stop()` so the next
+/// scenario starts from a healthy 3-node cluster.
+async fn restart_node(cluster: &deployer_cluster::Cluster, node: &str) {
+    cluster
+        .composer()
+        .start(node)
+        .await
+        .expect("Should restart io-engine node");
+    // Give the node time to re-register before the next scenario lands.
+    tokio::time::sleep(Duration::from_secs(5)).await;
```
tiagolobocastro: This should only be needed if we had stopped the node?
And also there's existing helpers IIRC to wait for a node to become online rather than a hard sleep

control-plane/agents/src/bin/core/volume/specs.rs
```diff
@@ -360,7 +360,19 @@ fn validate_publish(
...
-        if volume.publish_context.as_ref() != Some(args.publish_context()) {
+        // Promotion of an offline-rebuild nexus on incoming CSI publish. The internal
+        // reconciler had no source of CSI publish_context (cleared on prior unpublish), so
+        // accept the incoming context as the truth. Single-node-writer + empty-to-single
+        // frontend is also the normal CSI publish shape; allow it here. Only applies when the
+        // current target_config is one we own — a user-created unshared target is not
+        // promotable through this path.
+        let promoting_offline_rebuild = target.protocol().is_none()
+            && args.protocol().is_some()
+            && volume.is_offline_rebuild_target();
```
tiagolobocastro: Shouldn't it just be this now?
And we might need another check, if volume is already published for offline rebuild and another publish is sent for offline rebuild how do we handle it?

```suggestion
        let promoting_offline_rebuild = volume.is_offline_rebuild_target();
```

control-plane/stor-port/src/types/v0/transport/volume.rs
tiagolobocastro: Another task that could be added, is an api for starting an offline rebuild via the kubectl-mayastor plugin; in case we wanted to kick it before the wait time?

control-plane/agents/src/bin/core/tests/volume/offline_rebuild.rs
yugchaudhari: Replaced the sleep with `wait_node_online` from helpers. Returns as soon as the node reports Online (with a timeout), rather than blindly waiting 5s every time.

control-plane/agents/src/bin/core/volume/specs.rs
yugchaudhari: Applied the suggestion. Dropped the two redundant conditions since `is_offline_rebuild_target()` already implies them.

On the double-publish: added an explicit guard at the top of `publish_with_mode` that returns `VolumeAlreadyPublished` when the incoming mode is `OfflineRebuild` and the current target is already in that mode. The reconciler self-prevents this in its own poll loop today, so it's a defensive net for any internal caller we add later (like offline rebuild api via the kubectl-mayastor plugin).

control-plane/stor-port/src/types/v0/transport/volume.rs
yugchaudhari: Good idea. Tracking it as a follow-up. Should be small once this is merged: a thin REST/gRPC endpoint that calls `publish_with_mode(OfflineRebuild)` directly.

tiagolobocastro (approved):

yugchaudhari:
@tiagolobocastro should we merge this if everything looks good? so I can get started on iteration 4.

tiagolobocastro:
bors merge

yugchaudhari:
@tiagolobocastro

Root cause for the test failure was in the test helper. `pick_victim_node` was returning the first non-target node from the replica list across all volumes. In `promote_on_publish` the test volume's 2 replicas land on a subset of nodes (e.g. io-engine-1 + io-engine-3). With leftover replicas from `happy_path` still being cleaned up, the helper could pick io-engine-2 as victim, but io-engine-2 doesn't host any of the test volume's data. Stopping it leaves the volume Online, the reconciler never sees a Degraded state, no offline-rebuild target gets created, and `wait_for_unshared_target` panics at the 15s timeout.

Fix: scope `pick_victim_node` to the specific volume's replicas via the spec's `owners.volume`, same thing I'm already using in `gc_safety`.

Ran the test 10 times in isolation with the fix, all passed. Runtime also dropped from ~31s to ~25s since we no longer eat the 15s timeout. `cargo fmt --check` and `cargo clippy --workspace --tests -- -D warnings` clean.

Pushed. Ready for `bors merge` whenever.

tiagolobocastro:
bors try

tiagolobocastro:
bors merge

PR #1114, opened 2026-06-05
feat(core): add bestEffort snapshot restore policy

yugchaudhari:
## Summary

Adds a new StorageClass parameter `snapshotRestorePolicy` to control how a snapshot restore behaves when not every replica pool can host a clone of the source snapshot.

- `strict` (default): preserves current behavior. Every requested replica must be cloned from the snapshot.
- `bestEffort`: the restore proceeds as long as at least one clone succeeds. The volume comes up under-replicated and the regular replica reconciler fills in the missing replicas via a normal rebuild.

Also maps `ResourceExhausted` to `FailedPrecondition` on the snapshot create path so the external-snapshotter side-car stops retrying forever on permanent capacity failures. Reverts naturally once kubernetes-csi/external-snapshotter#1334 lands.

Refs openebs/mayastor#1987

## Test plan

- [x] New `snapshot_restore_best_effort` test under `agents::tests::volume::snapshot_clone`. Pauses one of two io-engines and asserts `Strict` returns `ReplicaCreateNumber` while `BestEffort` succeeds with one clone (and `num_replicas == 2` preserved on the spec).
- [x] Full `snapshot_clone` test set still passing (`cargo test --package agents --test core volume::snapshot_clone -- --test-threads=1`).
- [x] `cargo check --workspace` clean.
- [x] `cargo clippy --workspace --tests -- -D warnings` clean.

### DIFF
--- a/control-plane/csi-driver/src/bin/controller/controller.rs
+++ b/control-plane/csi-driver/src/bin/controller/controller.rs
@@ -456,6 +456,9 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
                         // thin:false explicitly.
                         let thin_arg_passed = args.parameters.contains_key("thin");
                         let thin = !thin_arg_passed || thin;
+                        let snapshot_restore_policy = context
+                            .snapshot_restore_policy()
+                            .map(models::SnapshotRestorePolicy::from);
                         RestApiClient::get_client()
                             .create_snapshot_volume(
                                 &parsed_vol_uuid,
@@ -467,6 +470,7 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
                                 affinity_group_name.map(AffinityGroup::new),
                                 max_snapshots,
                                 encrypted,
+                                snapshot_restore_policy,
                             )
                             .await?
                     }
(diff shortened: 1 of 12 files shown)


control-plane/csi-driver/src/bin/controller/controller.rs
```diff
@@ -1000,8 +1004,19 @@ impl rpc::csi::controller_server::Controller for CsiControllerSvc {
                     .create_volume_snapshot(&volume_uuid, &snap_uuid)
                     .await
                     .map_err(|error| match error {
+                        // The external-snapshotter side-car treats
```
tiagolobocastro: IIRC the limit is 100
These are 70, so you can use longer lines, up to 100 to make it easier to read

control-plane/grpc/src/operations/volume/traits.rs
```diff
@@ -1062,6 +1069,15 @@ impl CreateVolumeInfo for ValidatedCreateVolumeRequest {
     fn cluster_size(&self) -> Option<u32> {
         self.inner.cluster_size
     }
+
+    fn snapshot_restore_policy(&self) -> SnapshotRestorePolicy {
+        match self.inner.snapshot_restore_policy {
+            Some(value) => volume::SnapshotRestorePolicy::try_from(value)
+                .map(SnapshotRestorePolicy::from)
+                .unwrap_or_default(),
```
tiagolobocastro: This needs to be validated before building the `ValidatedCreateVolumeRequest`, otherwise unknown policies would become default instead of failing, ie when running mismatch of versions

yugchaudhari: Moved the conversion into `ValidateRequestTypes::validated()`. Unknown enum values now fail fast with `ReplyError::invalid_argument` instead of silently defaulting. The accessor on `ValidatedCreateVolumeRequest` just reads the pre-parsed typed field.

tiagolobocastro (approved):

Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

yugchaudhari:
> bors merge

@tiagolobocastro fixed lint issues

tiagolobocastro:
bors merge

PR #1117, opened 2026-06-14
feat: add event-consumer crate for NATS event consumption

rohan2794:
Adds utils/event-consumer — library crate that connects to a NATS server and forwards raw events over a tokio mpsc channel.

Supports JetStream pull consumer and Core NATS subscription modes, selectable via ConsumerConfig. Includes capped exponential backoff and retry logic for connection and consumer recovery.

Intended for use by events-aggregator in mayastor-extensions, and available to any control-plane component consuming events for debug.

### DIFF
--- a/utils/event-consumer/src/consumer.rs
+++ b/utils/event-consumer/src/consumer.rs
@@ -1,0 +1,256 @@
+use crate::ConsumerConfig;
+use anyhow::bail;
+use async_nats::jetstream::consumer::{pull, Consumer};
+use events_api::common::retry::{backoff_with_options, BackoffOptions};
+use futures::StreamExt;
+use std::time::Duration;
+use tokio::sync::mpsc;
+use tracing::{error, info, warn};
+
+/// A message received from NATS, abstracting over JetStream and Core NATS delivery.
+pub enum UnifiedMessage {
+    /// A JetStream message that must be ACKed or NAKed by the caller.
+    JetStream(async_nats::jetstream::Message),
+    /// A Core NATS message — fire-and-forget, no acknowledgement needed.
+    Core(async_nats::Message),
+}
+
+/// Manages the NATS connection and forwards incoming messages to a channel.
+pub struct NatsConsumer {
+    client: async_nats::Client,
+    config: ConsumerConfig,
+}
+
+fn consumer_backoff_options() -> BackoffOptions {
+    BackoffOptions::new()
+        .with_init_delay(Duration::from_millis(500))
+        .with_cutoff(2)
+        .with_delay_step(Duration::from_secs(2))
+        .with_max_delay(Duration::from_secs(30))
+}
+
+impl NatsConsumer {
+    /// Connect to the NATS server using the provided config.
+    /// Returns an error if the connection cannot be established.
+    pub async fn connect(config: ConsumerConfig) -> Result<Self, async_nats::Error> {
+        info!("Connecting to NATS at {}...", config.nats_url);
+
+        let connection_timeout = config.connection_timeout;
+        let request_timeout = config.request_timeout;
+        let backoff = consumer_backoff_options();
+
+        let options = async_nats::ConnectOptions::new()
+            .connection_timeout(connection_timeout)
+            .request_timeout(Some(request_timeout))
+            .retry_on_initial_connect()
+            .event_callback(|event| async move {
+                match event {
+                    async_nats::Event::Connected => info!("NATS connection established"),
+                    async_nats::Event::Disconnected => warn!("NATS connection lost"),
+                    async_nats::Event::LameDuckMode => warn!("NATS server entering lame duck mode"),
+                    async_nats::Event::SlowConsumer(_) => warn!("NATS slow consumer detected"),
+                    _ => (),
+                }
+            })
+            .reconnect_delay_callback(move |attempts| backoff.delay(attempts as u32));
+        let client = options.connect(config.nats_url.as_str()).await?;
+        Ok(Self { client, config })
+    }
+
+    /// Start consuming messages from NATS and forward them to `tx`.
+    /// Dispatches to JetStream pull consumer or Core NATS depending on `config.jetstream_enabled`.
+    pub async fn subscribe(self, tx: mpsc::Sender<UnifiedMessage>) -> anyhow::Result<()> {
+        if self.config.jetstream_enabled {
+            info!("JetStream enabled — starting pull consumer");
+            self.setup_jetstream(tx).await
+        } else {
+            info!("JetStream disabled — using Core NATS subscription");
+            self.setup_core_nats(tx);
+            Ok(())
+        }
+    }
+
+    // Wires together consumer acquisition and the message pump.
+    async fn setup_jetstream(self, tx: mpsc::Sender<UnifiedMessage>) -> anyhow::Result<()> {
+        let consumer = self.create_pull_consumer().await?;
+        tokio::spawn(run_jetstream_loop(consumer, tx));
+        Ok(())
+    }
+
+    // Connects to the JetStream stream and creates (or attaches to) the durable pull consumer.
+    // Retries with backoff; fails after max_retries.
+    async fn create_pull_consumer(&self) -> anyhow::Result<Consumer<pull::Config>> {
+        let js = async_nats::jetstream::new(self.client.clone());
+        let stream_name = &self.config.jetstream_stream_name;
+        let consumer_name = &self.config.jetstream_consumer_name;
+        let subject = &self.config.subject_filter;
+        let request_timeout = self.config.request_timeout;
+        let backoff = consumer_backoff_options();
+        let mut tries = 0u32;
+
+        loop {
+            if self.client.connection_state() != async_nats::connection::State::Connected {
+                warn!(
+                    tries,
+                    max = backoff.max_retries,
+                    "NATS client not connected; retrying"
+                );
+                if tries >= backoff.max_retries {
+                    bail!(
+                        "JetStream setup failed after {} attempts: client disconnected",
+                        tries + 1
+                    );
+                }
+                backoff_with_options(&mut tries, &backoff).await;
+                continue;
+            }
+
+            match tokio::time::timeout(request_timeout, js.get_stream(stream_name)).await {
+                Ok(Ok(stream)) => {
+                    match stream
+                        .get_or_create_consumer(
+                            consumer_name,
+                            pull::Config {
+                                durable_name: Some(consumer_name.to_string()),
+                                filter_subject: subject.to_string(),
+                                max_deliver: self.config.jetstream_max_deliver,
+                                ..Default::default()
+                            },
+                        )
+                        .await
+                    {
+                        Ok(consumer) => return Ok(consumer),
+                        Err(e) => {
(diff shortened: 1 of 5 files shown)


utils/event-consumer/src/consumer.rs
```diff
@@ -0,0 +1,256 @@
...
+                ),
+            }
+
+            if attempt >= MAX_ATTEMPTS {
+                bail!("Unable to set up JetStream after {MAX_ATTEMPTS} attempts");
+            }
+        };
+
+        tokio::spawn(async move {
+            let mut recovery_attempt = 0u32;
```
tiagolobocastro: this `setup_jetstream` is quite large, please this to separate function, and just pass it here

utils/event-consumer/src/consumer.rs
```diff
@@ -0,0 +1,256 @@
...
+        info!("Connecting to NATS at {}...", config.nats_url);
+
+        let connection_timeout = config.connection_timeout;
+        let request_timeout = config.request_timeout;
+
+        let options = async_nats::ConnectOptions::new()
+            .connection_timeout(connection_timeout)
+            .request_timeout(Some(request_timeout))
+            .retry_on_initial_connect()
+            .reconnect_delay_callback(|attempts| {
```
tiagolobocastro: async-nats doesn't provide a default one themselves?

utils/event-consumer/src/lib.rs
```diff
@@ -0,0 +1,36 @@
+mod consumer;
+
+pub use consumer::{NatsConsumer, UnifiedMessage};
+
+/// Configuration for the NATS event consumer.
+#[derive(Debug, Clone)]
+pub struct ConsumerConfig {
+    /// NATS server URL.
+    pub nats_url: String,
```
tiagolobocastro:
```suggestion
    pub nats_url: url::Url,
```

utils/event-consumer/src/consumer.rs
```diff
@@ -0,0 +1,256 @@
...
+                        recovery_attempt += 1;
+                        let delay = backoff_delay_ms(recovery_attempt);
+                        error!(recovery_attempt, error = %e, delay_ms = delay, "Failed to get JetStream message iterator; retrying");
+
+                        if recovery_attempt >= MAX_ATTEMPTS {
+                            error!(
+                                max = MAX_ATTEMPTS,
+                                "JetStream consumer failed to bind; exiting"
+                            );
+                            std::process::exit(1);
```
tiagolobocastro: A library should not be exiting by itself, the binary using the library should make the decision

utils/event-consumer/src/consumer.rs
tiagolobocastro: Also we seem to already have something here: https://github.com/openebs/mayastor-dependencies/blob/develop/apis/events/src/mbus_nats.rs
`BackoffOptions` 
Unless there's a reason, we should reuse it?

tiagolobocastro (commented):
There's already a subscribe in the api library: <https://github.com/openebs/mayastor-dependencies/blob/d3d21e69ba401aefcf2308f4eabe9e61b6869415/apis/events/src/mbus_nats.rs#L329>
Can we not reuse that?

utils/event-consumer/src/consumer.rs
rohan2794: Yes, BackoffOptions can be reused.

rohan2794:
> There's already a subscribe in the api library: https://github.com/openebs/mayastor-dependencies/blob/d3d21e69ba401aefcf2308f4eabe9e61b6869415/apis/events/src/mbus_nats.rs#L329 Can we not reuse that?

Existing subscribe can’t be used as is.  Event-consumer uses a pull consumer where it explicitly fetch and The Messages type is different between push and pull. Every message is ACKed unconditionally but event-consumer does explicit ACK/NAK so that  transient failures would be retried.

utils/event-consumer/src/consumer.rs
rohan2794: fixed

utils/event-consumer/src/lib.rs
rohan2794: changed to url type

utils/event-consumer/src/consumer.rs
rohan2794: reused BackoffOptions from events api module

utils/event-consumer/src/consumer.rs
rohan2794: refactored it into three function

Abhinandan-Purkait (approved):

utils/event-consumer/src/consumer.rs
```diff
@@ -0,0 +1,272 @@
...
+                match event {
+                    async_nats::Event::Connected => info!("NATS connection established"),
+                    async_nats::Event::Disconnected => warn!("NATS connection lost"),
+                    async_nats::Event::LameDuckMode => warn!("NATS server entering lame duck mode"),
+                    async_nats::Event::SlowConsumer(_) => warn!("NATS slow consumer detected"),
+                    _ => (),
+                }
+            })
+            .reconnect_delay_callback(move |attempts| {
+                // Replicate BackoffOptions delay formula synchronously; reconnect callback is not async.
```
tiagolobocastro: why... again it's duplicating... refactor `BackoffOptions` to allow getting the backoff without sleeping

utils/event-consumer/src/consumer.rs
```diff
@@ -0,0 +1,272 @@
...
+                "JetStream setup failed; retrying"
+            );
+            backoff_with_options(&mut tries, &backoff).await;
+        }
+    }
+
+    // Core NATS Subscription Setup with Resilient Reconnection Logic.
+    async fn setup_core_nats(self, tx: mpsc::Sender<UnifiedMessage>) -> anyhow::Result<()> {
+        let client = self.client.clone();
+        let subject = self.config.subject_filter.clone();
```
tiagolobocastro: not necessary to clone, can just move because we're consuming self here

utils/event-consumer/src/consumer.rs
```diff
@@ -0,0 +1,272 @@
...
+            let backoff = consumer_backoff_options();
+            let mut tries = 0u32;
+
+            loop {
+                match client.subscribe(subject.clone()).await {
+                    Ok(mut subscription) => {
+                        tries = 0;
+                        while let Some(message) = subscription.next().await {
+                            if tx.send(UnifiedMessage::Core(message)).await.is_err() {
+                                return;
```
tiagolobocastro, line 168: isn't this bailing out if the buffer is full?

utils/event-consumer/src/consumer.rs
```diff
@@ -0,0 +1,272 @@
...
+                tries,
+                max = backoff.max_retries,
+                "JetStream setup failed; retrying"
+            );
+            backoff_with_options(&mut tries, &backoff).await;
+        }
+    }
+
+    // Core NATS Subscription Setup with Resilient Reconnection Logic.
+    async fn setup_core_nats(self, tx: mpsc::Sender<UnifiedMessage>) -> anyhow::Result<()> {
```
tiagolobocastro: Why is this returning a result when it's only returning Ok(()) ?

utils/event-consumer/src/consumer.rs
```diff
@@ -0,0 +1,272 @@
...
+                match client.subscribe(subject.clone()).await {
+                    Ok(mut subscription) => {
+                        tries = 0;
+                        while let Some(message) = subscription.next().await {
+                            if tx.send(UnifiedMessage::Core(message)).await.is_err() {
+                                return;
+                            }
+                        }
+
+                        if tries >= backoff.max_retries {
```
tiagolobocastro, line 172: How is the receiver coping with this? Does it then shutdown?

utils/event-consumer/src/consumer.rs
```diff
@@ -0,0 +1,272 @@
...
+
+        tokio::spawn(async move {
+            let backoff = consumer_backoff_options();
+            let mut tries = 0u32;
+
+            loop {
+                match client.subscribe(subject.clone()).await {
+                    Ok(mut subscription) => {
+                        tries = 0;
+                        while let Some(message) = subscription.next().await {
```
tiagolobocastro, line 165: What does None signify here? 
Does the client subscribe handle shutdown requests?

utils/event-consumer/src/consumer.rs
```diff
@@ -0,0 +1,272 @@
...
+
+            match tokio::time::timeout(request_timeout, js.get_stream(stream_name)).await {
+                Ok(Ok(stream)) => {
+                    match stream
+                        .get_or_create_consumer(
+                            consumer_name,
+                            pull::Config {
+                                durable_name: Some(consumer_name.to_string()),
+                                filter_subject: subject.to_string(),
+                                max_deliver: 3,
```
tiagolobocastro: what is this option doing?

tiagolobocastro (commented):
As another PR would be good to add the fetching of messages (one time, and wait) to the rest-plugin so we can then interact with deployer clusters

utils/event-consumer/src/consumer.rs
rohan2794: `tx.send().await`  waits until there is space in the buffer. It only returns Error when the receiver has been dropped (channel closed). Buffer full is not an error here, it's just backpressure.

So the `return` only fires when the downstream consumer has stopped reading — i.e., the whole pipeline is shutting down.

utils/event-consumer/src/consumer.rs
rohan2794: Yes, the receiver shuts down cleanly through channel closure.

utils/event-consumer/src/consumer.rs
rohan2794: `subscription.next()` returns `None` to an unexpected loss of connection. In case of `None`, it will try to re-subscribe.
 No — there is no shutdown path over client. The only ways this loop exits are:
1. tx.send().is_err() — the downstream channel closed 
2. tries >= backoff.max_retries — exhausted retries

utils/event-consumer/src/consumer.rs
rohan2794: `max_deliver: 3` sets the maximum number of times NATS JetStream will attempt to deliver a message to the consumer before giving up on it.

rohan2794: Moved it to consumer config and removed hard coded value

utils/event-consumer/src/consumer.rs
rohan2794: refactored `BackoffOptions` to add delay https://github.com/openebs/mayastor-dependencies/pull/159

utils/event-consumer/src/consumer.rs
rohan2794: fixed

utils/event-consumer/src/consumer.rs
rohan2794: fixed

tiagolobocastro (approved):
Please squash the commits

rohan2794:
@mergifyio queue

rohan2794:
bors merge

tiagolobocastro:
bors merge

PR #1121, opened 2026-06-18
fix(health): don't cache readiness failure before first successful probe

niladrih:
CachedCoreState previously probed core-grpc in its constructor and
cached the result. If the gRPC service wasn't up yet, a false value
was cached with a fresh timestamp, and no retry happened until the
cache expired. This caused the readiness probe to report not-ready
for the entire cache_duration even if core-grpc came up moments later.

Make last_updated an Option that starts as None. While None, every
readiness check makes the actual gRPC call instead of serving a cached
false. On the first successful probe last_updated becomes Some, and
is refreshed on every probe thereafter so that cache expiry continues
to work correctly when core-grpc later goes down. Also make
CachedCoreState::new() synchronous since the eager probe is no longer
needed.

### DIFF
--- a/control-plane/rest/service/src/health/core_state.rs
+++ b/control-plane/rest/service/src/health/core_state.rs
@@ -13,7 +13,10 @@ pub struct CachedCoreState {
 /// This type remembers a liveness state, and when this data was refreshed.
 struct ServerState {
     is_live: bool,
-    last_updated: Instant,
+    /// Initially `None`, causing every readiness check to make the actual gRPC
+    /// call. Set to `Some` on the first successful probe, and updated on every
+    /// probe thereafter so that cache expiry works correctly for failures too.
+    last_updated: Option<Instant>,
 }
 
 impl ServerState {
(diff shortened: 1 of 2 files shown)


control-plane/rest/service/src/health/core_state.rs
```diff
@@ -14,6 +14,9 @@ pub struct CachedCoreState {
 struct ServerState {
     is_live: bool,
     last_updated: Instant,
+    /// Whether a probe has ever succeeded. Until this is true, we always make
+    /// the actual gRPC call rather than serving a cached `false`.
+    has_succeeded: bool,
```
tiagolobocastro: this feels awkward, why not just make `last_updated` an option?

niladrih: Good idea

tiagolobocastro (approved):

Abhinandan-Purkait (approved):

niladrih:
bors merge

niladrih:
bors merge

niladrih:
bors merge

PR #1122, opened 2026-06-20
refactor(tls): read secret name directly from annotation

tiagolobocastro:
Reads the client secret name directly from annotation. This means we can use a separate certificate with client auth EKU only.

todo: would we want different secrets per client? Right now it's one for all.


Abhinandan-Purkait (approved):

tiagolobocastro:
bors merge

PR #1127, opened 2026-06-25
feat(volume/republish): wait for importing replicas before child selection

susobhandey:
feat(volume/republish): wait for importing replicas before child selection

If a pool import is slow and a republish arrives before it completes, the
local replica is not yet Online and gets silently dropped from the nexus
child candidate set. The nexus is then recreated without it, forcing a full
rebuild of what is usually the most up-to-date copy.

Add a bounded wait with early exit before healthy_volume_replicas() in the
republish path: read the previously-healthy children from the persisted
NexusInfo and poll until they report Online, or until a deadline elapses.

- Healthy imports pay ~0: returns immediately when nothing is missing.
- Never adds an offline child; only delays the point-in-time candidate
  snapshot so it is more complete.
- On timeout, proceeds exactly as before, so a never-returning replica
  cannot stall recovery.

The wait deadline must stay below the HA failover trigger and client I/O
timeout to avoid trading a rebuild for a failover; it is currently a const
and should be moved to config.

### DIFF
--- a/control-plane/agents/src/bin/core/volume/operations.rs
+++ b/control-plane/agents/src/bin/core/volume/operations.rs
@@ -21,7 +21,8 @@ use crate::{
         snapshot_operations::DestroyVolumeSnapshotRequest,
         specs::{
             create_volume_replicas, healthy_volume_replicas, resizeable_replicas,
-            volume_move_replica_candidates, CreateReplicaCandidate,
+            volume_move_replica_candidates, wait_for_expected_replicas, CreateReplicaCandidate,
+            REPUBLISH_REPLICA_POLL, REPUBLISH_REPLICA_WAIT,
         },
     },
 };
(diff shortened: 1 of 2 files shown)


control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -471,6 +472,22 @@ impl ResourcePublishing for OperationGuardArc<VolumeSpec> {
...
+        // rebuild. On timeout this proceeds exactly as before.
+        if let Ok(nexus_info) = registry
+            .nexus_info(Some(&request.uuid), Some(older_nexus.uuid()), false, None)
+            .await
+        {
+            wait_for_expected_replicas(
+                registry,
+                &nexus_info,
+                REPUBLISH_REPLICA_WAIT,
+                REPUBLISH_REPLICA_POLL,
```
tiagolobocastro, line 487: I'm not sure if waiting in-line here is the right approach (vs fail & retry)

Also this wait-time would not be sufficient for the large pool slow import, which is the main culprit for these; as the node is up and ready, but pool importing still taking time, especially when pool needs to be "fsck'ed"

Abhinandan-Purkait: Yeah, we usually don't have inline wait in the operations workflow. So, we should maybe do a fail and retry or wait on delegate the work to something else, otherwise the guard might be taken for too long.

PR #1129, opened 2026-06-26
feat(event-consumer): create JetStream stream on startup if absent

rohan2794:
- Switch from get_stream() to get_or_create_stream() in create_pull_consumer
so whichever pod starts first (events-aggregator or obs-callhome-stats)
creates the events-stream. Eliminates all 404 "stream not found" WARNs
during startup caused by racing against callhome-stats initialisation.

- Adds jetstream_stream_replicas to ConsumerConfig (default 1) and wires
--events-replicas CLI arg in events-aggregator, using the same
events_replicas Helm helper as other eventing components to ensure the
replica count always matches the NATS cluster size.

- Downgrade transient JetStream setup errors from error to warn since
  stream-not-found during startup is an expected transient condition
- Use a shorter jetstream_setup_timeout (3s) for stream discovery probes
  instead of the 10s request_timeout, eliminating long stalls per retry

### DIFF
--- a/utils/event-consumer/src/consumer.rs
+++ b/utils/event-consumer/src/consumer.rs
@@ -1,6 +1,9 @@
 use crate::ConsumerConfig;
 use async_nats::jetstream::consumer::{pull, Consumer};
-use events_api::common::retry::{backoff_with_options, BackoffOptions};
+use events_api::common::{
+    constants::{MAX_MSGS_PER_SUBJECT, STREAM_SIZE, SUBJECTS},
+    retry::{backoff_with_options, BackoffOptions},
+};
 use futures::StreamExt;
 use std::time::Duration;
 use tokio::{sync::mpsc, task::JoinHandle};
@@ -106,7 +109,7 @@ impl NatsConsumer {
         let stream_name = &self.config.jetstream_stream_name;
         let consumer_name = &self.config.jetstream_consumer_name;
         let subject = &self.config.subject_filter;
-        let request_timeout = self.config.request_timeout;
+        let setup_timeout = self.config.jetstream_setup_timeout;
         let backoff = consumer_backoff_options();
         let mut tries = 0u32;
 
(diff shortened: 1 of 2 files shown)


utils/event-consumer/src/consumer.rs
```diff
@@ -70,6 +73,16 @@ impl NatsConsumer {
             })
             .reconnect_delay_callback(move |attempts| backoff.delay(attempts as u32));
         let client = options.connect(config.nats_url.as_str()).await?;
+        // async_nats returns the client before the TCP handshake completes when
+        // retry_on_initial_connect() is set. Poll until Connected so the JetStream
+        // setup loop doesn't log spurious "not connected" warnings.
+        let deadline = tokio::time::Instant::now() + connection_timeout;
+        while client.connection_state() != async_nats::connection::State::Connected {
```
tiagolobocastro: Nats docs state: `The client should not wait for the first connection to succeed or fail, as in some network scenarios, this can take much time. If the first attempt fails, a standard [Reconnect process] should be performed.`

If we really want to ensure connected, maybe the binary should make this decision and not the library

utils/event-consumer/src/consumer.rs
```diff
@@ -126,7 +139,17 @@ impl NatsConsumer {
...
             }
 
-            match tokio::time::timeout(request_timeout, js.get_stream(stream_name)).await {
+            let stream_config = async_nats::jetstream::stream::Config {
+                name: stream_name.to_string(),
+                subjects: vec![SUBJECTS.into()],
+                max_messages_per_subject: MAX_MSGS_PER_SUBJECT,
+                max_bytes: STREAM_SIZE,
+                storage: async_nats::jetstream::stream::StorageType::Memory,
+                num_replicas: self.config.jetstream_stream_replicas,
```
tiagolobocastro, line 138: Is this the same config which the other event components create?

rohan2794: Yes, it's same one which **obs-callhome-stats** creates. ref: https://github.com/openebs/mayastor-extensions/blob/develop/chart/templates/mayastor/obs/obs-callhome-deployment.yaml#L74C18-L74C24

tiagolobocastro: But where are we setting the replicas here?

rohan2794: It will be set via Helm chart https://github.com/openebs/mayastor-extensions/pull/902/changes#diff-0d900dd3c462837fe56fdf112c1ee2a5099909560ee1acba04c72b5041ced022R50

utils/event-consumer/src/consumer.rs
rohan2794: I will remove polling loop from here as  JetStream setup loop already retries on "not connected".

Abhinandan-Purkait (approved):

Cargo.lock
```diff
@@ -3209,7 +3220,7 @@ dependencies = [
  "opentelemetry",
  "opentelemetry-proto",
  "opentelemetry_sdk",
- "prost",
+ "prost 0.13.3",
```
tiagolobocastro: @rohan2794, can you pick a version of pbjson which uses the same prost version we're currently using?

rohan2794: Downgraded to `0.7.0` version of pbjson . Respective PR in dependency repo [PR](https://github.com/openebs/mayastor-dependencies/pull/170), PTAL

tiagolobocastro (approved):

tiagolobocastro:
bors merge

PR #1131, opened 2026-06-29
feat(core): pre-flight viability check and safe teardown for offline rebuild

yugchaudhari:
Iteration 5 of offline volume rebuild ([openebs/openebs#4208](https://github.com/openebs/openebs/issues/4208)), builds on iteration 3 ([#1112](https://github.com/openebs/mayastor-control-plane/pull/1112)) which wired the full create/rebuild/teardown lifecycle plus promote-on-publish. Iteration 4 was the data-plane half (clean_shutdown semantics on unshared nexuses) and landed in [openebs/mayastor#2007](https://github.com/openebs/mayastor/pull/2007); no control-plane change was needed for that goal once io-engine flipped the flag (`scheduling.rs::healthy_children` already does the right thing). This PR picks up the iter5 production-readiness items, starting with the pre-flight viability check @tiagolobocastro raised on iteration 2 (#1108 [comment](https://github.com/openebs/mayastor-control-plane/pull/1108#discussion_r3343131975)).

Today's `initiate_offline_rebuild` jumps straight to `publish_with_mode` once the grace period has elapsed. If there's no healthy replica to copy from, or no candidate pool to host the replacement replica, the underlying publish fails at scheduling time and the reconciler churns through retries each poll cycle. There's also no real teardown path for the in-flight case: if the source node goes offline mid-rebuild, the temp nexus stays up indefinitely because the only teardown trigger is `Online`.

This PR adds two related changes to the reconciler. Before publishing, it now checks viability: at least one replica must be `Online` (source) and `volume_pool_candidates` must return at least one pool (target). If either check fails, the reconciler returns `Busy` so the next poll re-evaluates, and the grace-period stamp is left in place so a recovering cluster doesn't have to wait a fresh grace window. After publishing, the teardown path now treats `VolumeStatus::Faulted` as a real failure signal, with the precondition in place, hitting Faulted while the temp nexus exists means no healthy source children remain, so we unpublish and clear the grace stamp instead of holding resources. `Degraded` and `Unknown` still fall through to the existing wait branch, those are normal in-progress / node-blip states and tearing them down would churn against the rebuild engine.

Both checks reuse existing scheduling helpers (`volume_pool_candidates` / `GetSuitablePools`) and the `volume_state_health` result we already compute earlier in the same poll cycle. No spec, persistent-store, or gRPC changes.

BDD coverage adds one scenario to the `offline_rebuild_e2e`: a 3-replica volume on a 3-node cluster (no spare pool), victim node down → asserts no temp nexus is created during the deferred window, victim node returns → volume goes Online without any rebuild having fired. Two other branches are intentionally left without BDD in this PR:

- **Source-not-viable**: the existing `if status != Degraded` early-bail catches `Faulted` before the precondition runs, so the source check is defensive code that isn't reachable on the normal Degraded path. Worth keeping for safety, but not naturally testable through the volume API.
- **Faulted-teardown**: the auto-selected nexus placement co-locates the temp nexus with the surviving replica's node on a 3-node cluster, so killing the source also kills the nexus host and the volume reports Degraded/Unknown instead of Faulted via the path-2 fallback in `volume_state_health`. Triggering the Faulted branch needs topology-controlled placement, leaving that as a follow-up.

The remaining deferred follow-ups (rebuild slot prioritization, unpublish-during-rebuild, operator-initiated trigger) stay scoped for separate iter5 PRs.

Refs:
- OEP: https://github.com/openebs/openebs/blob/develop/designs/replicated-pv/mayastor/offline-volume-rebuild.md
- Tracking: openebs/openebs#4208

### DIFF
--- a/control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
+++ b/control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
@@ -1,10 +1,14 @@
-use crate::controller::{
-    reconciler::{PollContext, TaskPoller},
-    resources::{
-        operations::ResourcePublishing, operations_helper::OperationSequenceGuard,
-        OperationGuardArc, ResourceMutex,
+use crate::{
+    controller::{
+        reconciler::{PollContext, TaskPoller},
+        resources::{
+            operations::ResourcePublishing, operations_helper::OperationSequenceGuard,
+            OperationGuardArc, ResourceMutex,
+        },
+        scheduling::volume::GetSuitablePools,
+        task_poller::{PollResult, PollerState},
     },
-    task_poller::{PollResult, PollerState},
+    volume::volume_pool_candidates,
 };
 
 use stor_port::types::v0::{
@@ -116,13 +120,14 @@ async fn offline_rebuild_reconcile(
         return PollResult::Ok(PollerState::Busy);
     }
 
-    initiate_offline_rebuild(&mut volume, context).await
+    initiate_offline_rebuild(&mut volume, &volume_state, context).await
 }
 
 /// Creates a non-shared nexus for the volume so the existing rebuild engine
 /// (HotSpareReconciler) can restore faulted replicas.
 async fn initiate_offline_rebuild(
     volume: &mut OperationGuardArc<VolumeSpec>,
+    volume_state: &VolumeState,
     context: &PollContext,
 ) -> PollResult {
     let registry = context.registry();
@@ -134,6 +139,43 @@ async fn initiate_offline_rebuild(
             "Offline rebuild deferred: max concurrent rebuilds reached"
         );
         return PollResult::Ok(PollerState::Busy);
+    }
+
+    // Pre-flight viability: only stand up the temp nexus if the rebuild can
+    // actually make progress. Needs both a source (a healthy replica to copy
+    // from) and a target (a candidate pool with room for the replacement
+    // replica).
+    //
+    // Prefer the strict health signal — `online_clean_replicas > 0` — when
+    // available; that's the CP's own trust decision and already accounts for
+    // NexusInfo state. Volume health is optional (can be disabled via
+    // `--no-volume-health`), so fall back to a plain online-status check on
+    // the replica topology when it's missing rather than deferring the
+    // rebuild indefinitely.
+    let source_viable = volume_state
+        .health
+        .as_ref()
+        .map(|h| h.online_clean_replicas > 0)
+        .unwrap_or_else(|| {
+            volume_state
+                .replica_topology
+                .values()
+                .any(|r| r.status().online())
+        });
+    let target_viable = || async {
+        !volume_pool_candidates(GetSuitablePools::new(volume.as_ref(), None), registry)
+            .await
+            .is_empty()
+    };
+
+    if !source_viable || !target_viable().await {
+        tracing::debug!(
+            volume.uuid = %volume.uuid(),
+            source_viable,
+            "Offline rebuild deferred: rebuild not viable \
+            (no healthy source replica and/or no candidate pool)"
+        );
+        return PollResult::Ok(PollerState::Idle);
     }
 
     tracing::info!(
@@ -207,7 +249,30 @@ async fn teardown_if_rebuilt(
         return PollResult::Ok(PollerState::Idle);
     }
 
-    // Only `Online` is a clean enough signal to tear the nexus down. Anything
+    // Safe teardown when the rebuild has no chance of progressing: nexus is up
+    // but `Faulted` means no healthy source children remain (e.g. the source
+    // node went offline mid-rebuild).
+    if volume_state.status == VolumeStatus::Faulted {
+        let request = UnpublishVolume::new(volume.uuid(), false, vec![]);
+        if let Err(error) = volume.unpublish(context.registry(), &request).await {
+            tracing::warn!(
+                volume.uuid = %volume.uuid(),
+                %error,
+                "Failed to tear down stuck offline rebuild nexus"
+            );
+        } else {
+            tracing::warn!(
+                volume.uuid = %volume.uuid(),
+                "Offline rebuild source no longer available; tore down temporary nexus"
+            );
+        }
+        // Restart the grace timer from scratch when viability returns, so a
+        // recovering node has to clear the same wait window a fresh degradation would.
+        volume.lock().metadata.clear_offline_rebuild_degraded();
+        return PollResult::Ok(PollerState::Idle);
+    }
+
+    // Only `Online` and `Faulted` are clean enough signals to act on. Anything
     // else (Degraded mid-rebuild, transient state during a node blip, etc.)
     // we wait on — tearing down would either churn against the grace timer
     // or kill a healthy in-progress rebuild during a brief hiccup.
--- a/control-plane/agents/src/bin/core/tests/volume/offline_rebuild.rs
+++ b/control-plane/agents/src/bin/core/tests/volume/offline_rebuild.rs
@@ -13,7 +13,7 @@ const RECONCILE_PERIOD_MS: u64 = 250;
 const GRACE_PERIOD_SECS: u64 = 2;
 const REBUILD_TIMEOUT_SECS: u64 = 30;
 
-/// End-to-end coverage for the offline-rebuild reconciler. Drives three
+/// End-to-end coverage for the offline-rebuild reconciler. Drives four
 /// scenarios sequentially on a single shared cluster (each scenario takes a
 /// node down and brings it back up before handing off), which avoids the
 /// per-test cluster spin-up cost while still exercising each path on a clean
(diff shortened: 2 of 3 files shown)


control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
```diff
@@ -136,6 +141,32 @@ async fn initiate_offline_rebuild(
...
+    // Pre-flight viability: only stand up the temp nexus if the rebuild can
+    // actually make progress. Needs both a source (a healthy replica to copy
+    // from) and a target (a candidate pool with room for the replacement
+    // replica). Without this check, a create-then-fail loop is possible and
+    // safe teardown on stuck (below) would have to choose between churn and
+    // dangling resources.
+    let source_viable = volume_state
+        .replica_topology
+        .values()
+        .any(|r| r.status().online());
```
Abhinandan-Purkait: I wonder if this check is enough. @tiagolobocastro I wonder will this be always true, as the last child is never faulted?

Abhinandan-Purkait (approved):
Looks good otherwise

control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
yugchaudhari: The `.status()` here is the pool-side replica state (Online / Degraded / Faulted / Unknown), not the nexus's view of children. `.online()` is only true when the replica actually reports `Online`, so a replica on an offline node comes back as `Unknown/Offline`, and a corrupted one as `Faulted`. So the check does catch those cases.

The "last child never faulted" rule you're thinking of is over in `nexus_bdev_children`: the nexus refuses to fault its last healthy child so it doesn't lose data. That's a nexus invariant on child state and doesn't touch `ReplicaTopology.status()`. And we only reach this reconciler path when the volume is unpublished (no nexus, no children), so that rule doesn't apply here.

Practically, `source_viable = false` when every healthy replica is on an offline node, or the survivors are `Faulted`/`Degraded`/`Unknown`. That's the exact case we want to defer on rather than stand up a temp nexus that will fail.

If you'd like it stricter, I can add `r.healthy == Some(true)` on top so we're checking both runtime status and the persistent healthy flag.

control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
```diff
@@ -136,6 +141,32 @@ async fn initiate_offline_rebuild(
...
+    let target_viable =
```
tiagolobocastro: This should be lazy, to avoid even attempting when there's no source viable

```suggestion
    let target_viable = ||
```

control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
tiagolobocastro: I think here online is not enough, we have to check that it's clean/healthy
But we already have this information, something like this:

```rust
let source_viable = volume_state.health.as_ref().map(|h| h.online_clean_replicas).unwrap_or_default() > 0;
```

control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
```diff
@@ -207,7 +238,36 @@ async fn teardown_if_rebuilt(
...
+    // Safe teardown when the rebuild has no chance of progressing: nexus is up
+    // but `Faulted` means no healthy source children remain (e.g. the source
+    // node went offline mid-rebuild). Since the pre-flight viability check
+    // already validated source/target at create time, hitting `Faulted` here is
+    // a real failure signal, not transient churn — tear down so resources
+    // aren't held indefinitely and the grace-period can restart cleanly when
+    // viability returns. `Unknown` and `Degraded` still fall through to the
+    // wait branch: those are node-blip / in-progress states.
+    if volume_state.status == VolumeStatus::Faulted {
+        tracing::warn!(
```
tiagolobocastro: It is worth moving this below the If let Err as a past log message to avoid thrashing the logs in case of unpublish failures?

control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
```diff
@@ -207,7 +238,36 @@ async fn teardown_if_rebuilt(
...
+        if let Err(error) = volume.unpublish(context.registry(), &request).await {
+            tracing::warn!(
+                volume.uuid = %volume.uuid(),
+                %error,
+                "Failed to tear down stuck offline rebuild nexus"
+            );
+        }
+        // Restart the grace timer from scratch when viability returns, so a
+        // recovering node has to clear the same wait window a fresh
+        // degradation would.
```
tiagolobocastro:
```suggestion
        // recovering node has to clear the same wait window a fresh degradation would.
```

control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
```diff
@@ -207,7 +238,36 @@ async fn teardown_if_rebuilt(
...
 
-    // Only `Online` is a clean enough signal to tear the nexus down. Anything
+    // Safe teardown when the rebuild has no chance of progressing: nexus is up
+    // but `Faulted` means no healthy source children remain (e.g. the source
+    // node went offline mid-rebuild). Since the pre-flight viability check
+    // already validated source/target at create time, hitting `Faulted` here is
+    // a real failure signal, not transient churn — tear down so resources
+    // aren't held indefinitely and the grace-period can restart cleanly when
+    // viability returns. `Unknown` and `Degraded` still fall through to the
+    // wait branch: those are node-blip / in-progress states.
```
tiagolobocastro: nit: let's not overcomplicate the comments

```suggestion
    // Safe teardown when the rebuild has no chance of progressing: nexus is up
    // but `Faulted` means no healthy source children remain (e.g. the source
    // node went offline mid-rebuild).
```

control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
yugchaudhari: Switched to `volume_state.health.online_clean_replicas > 0`

control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
yugchaudhari: Applied. Moved the log below the `unpublish` call and into the `else` branch, so on failure only the "Failed to tear down" warn fires.

control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
Abhinandan-Purkait: But @tiagolobocastro wouldn't that need the health info to be enabled or am I mis remembering the switch? But yeah this looks plausible.

yugchaudhari: Can we do something like : Strict `online_clean_replicas` when health is on, degrades to the online-status check when it isn't?

```rust
let source_viable = volume_state
    .health
    .as_ref()
    .map(|h| h.online_clean_replicas > 0)
    .unwrap_or_else(|| {
        volume_state.replica_topology.values().any(|r| r.status().online())
    });
```

tiagolobocastro: > But @tiagolobocastro wouldn't that need the health info to be enabled or am I mis remembering the switch? But yeah this looks plausible.

Ah you're right, health can be disabled!

tiagolobocastro: > Can we do something like : Strict `online_clean_replicas` when health is on, degrades to the online-status check when it isn't?
> 
> ```rust
> let source_viable = volume_state
>     .health
>     .as_ref()
>     .map(|h| h.online_clean_replicas > 0)
>     .unwrap_or_else(|| {
>         volume_state.replica_topology.values().any(|r| r.status().online())
>     });
> ```

Yeah this seems reasonable then, adding the fallback

yugchaudhari: Added the fallback.

control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
```diff
@@ -136,6 +141,43 @@ async fn initiate_offline_rebuild(
...
+    };
+
+    if !source_viable || !target_viable().await {
+        tracing::debug!(
+            volume.uuid = %volume.uuid(),
+            source_viable,
+            "Offline rebuild deferred: rebuild not viable \
+            (no healthy source replica and/or no candidate pool)"
+        );
+        return PollResult::Ok(PollerState::Busy);
```
tiagolobocastro: Let's no return busy here, replica availability is beyond our control, so checking this more often is not likely to help

```suggestion
        return PollResult::Ok(PollerState::Idle);
```

yugchaudhari: Done.

tiagolobocastro (approved):

PR #1137, opened 2026-07-07
feat(core): treat unshared crashed nexus as clean when scheduling replicas

yugchaudhari:
Companion to openebs/mayastor#2030, which refactored the io-engine's `NexusInfo` to track "was the target actually shared" as a dedicated `shared` field rather than overloading `clean_shutdown`. This PR is the control-plane consumer for that new flag, and closes out the CP-side work for iteration 4 of the offline volume rebuild ([openebs/openebs#4208](https://github.com/openebs/openebs/issues/4208)).

A nexus that crashed while unshared has no possibility of in-flight front-end I/O, so its children are safe to trust as in-sync even when `clean_shutdown` is false. Before this PR, the CP couldn't distinguish "crashed while quiet" from "crashed while serving I/O" and always fell back to the conservative "only trust one child" path, forcing a full rebuild of the others every time an offline-rebuild temp nexus crashed.

## What this PR does

Adds a `shared: Option<bool>` field to the CP-side `NexusInfo` mirroring the field io-engine now writes to etcd, and introduces a `NexusInfo::clean_state()` helper that centralises the trust-all rule:

```rust
pub fn clean_state(&self) -> bool {
    self.clean_shutdown || !self.shared.unwrap_or(true)
}
```

Four places that previously gated on `!clean_shutdown` are refactored to use this helper instead:

- `nexus::scheduling::healthy_children` (nexus recreation)
- `volume::scheduling::healthy_volume_replicas` (volume nexus creation, functional mirror of the first)
- `volume::scheduling::snapshoteable_replica` (snapshot creation, preserves its existing `!published` gate)
- `volume::registry` `VolumeHealth` `clean_replicas` / `online_clean_replicas` computation

All four need to determine "how many children can we trust" and needed to stay consistent, so I extracted the helper to avoid duplicating the rule and its doc across four places.

## Backward compatibility

The `shared` field is `Option<bool>` with `#[serde(default)]`. Entries persisted by an io-engine version that predates openebs/mayastor#2030, or by a mixed-version rollout mid-upgrade, deserialize with `shared = None`. `unwrap_or(true)` in `clean_state()` treats `None` as if the nexus was shared, preserving the pre-existing conservative behaviour for those entries. No upgrade-window regression and no risk of assuming dirty data as clean.

The truth table:

| `clean_shutdown` | `shared`      | `clean_state()` | Behaviour                                     |
|------------------|---------------|-----------------|-----------------------------------------------|
| `true`           | any           | `true`          | Trust all (clean shutdown, was already the case) |
| `false`          | `Some(true)`  | `false`         | Trust one (was actively shared, was already the case) |
| `false`          | `Some(false)` | `true`          | Trust all, new safe case                     |
| `false`          | `None`        | `false`         | Trust one, conservative fallback for old / mid-upgrade entries |

## Tests

Inline unit tests in `nexus_persistence.rs`:

- Serde compatibility: absent field deserializes to `None`, `"shared": true` to `Some(true)`, `"shared": false` to `Some(false)`.
- `clean_state()` truth table: cleanly-shut-down, unclean-but-unshared, unclean-and-shared, unclean-and-shared-missing.

The existing `tests/deserializer.rs::NexusInfo` case continues to pass unchanged (it uses `..Default::default()` in the expected struct literal, so the new field slots in automatically), which is an implicit end-to-end confirmation that older etcd entries still deserialize cleanly.

## Deferred to follow-up

A BDD scenario in `tests/volume/offline_rebuild.rs` that exercises the full "unshared nexus crash, no unnecessary rebuild on healthy replicas" path. That test needs an io-engine build carrying openebs/mayastor#2030 so the `shared` field is actually written to etcd during the test cluster's lifecycle. Planning to add it as a small follow-up PR once the submodule bump lands.

## Refs

- OEP: https://github.com/openebs/openebs/blob/develop/designs/replicated-pv/mayastor/offline-volume-rebuild.md
- Tracking: openebs/openebs#4208
- Iter4 io-engine (initial): openebs/mayastor#2007
- Iter4 io-engine (refactor): openebs/mayastor#2030

### DIFF
--- a/control-plane/agents/src/bin/core/tests/volume/mod.rs
+++ b/control-plane/agents/src/bin/core/tests/volume/mod.rs
@@ -133,9 +133,17 @@ async fn nexus_persistence_test_iteration(
                 uuid: volume.spec().uuid.clone(),
                 // publish it on the remote first, to complicate things
                 target_node: Some(remote.clone()),
-                share: None,
+                // Publish shared over Nvmf so io-engine records `shared: Some(true)`
+                // on the nexus. Destroy on unpublish preserves the last-persisted
+                // shared value, so the fault mutations below run against a shared
+                // nexus — matching the "trust one child" scenario the assertions
+                // encode. Under `clean_state()`, an unshared nexus would count as
+                // clean and the "trust one" path would be unreachable.
+                share: Some(VolumeShareProtocol::Nvmf),
                 publish_context: HashMap::new(),
-                frontend_nodes: vec![],
+                // SingleNodeWriter + Nvmf share requires at least one frontend
+                // node when the cluster runs with `no_deprecated_access_mode(true)`.
+                frontend_nodes: vec!["a".into()],
                 access_mode: VolumeAccessMode::SingleNodeWriter,
             },
             None,
@@ -1453,9 +1461,13 @@ async fn health() {
             &PublishVolume::new(
                 volume.spec().uuid.clone(),
                 Some(cluster.node(0)),
-                None,
+                // Publish shared over Nvmf. Under the `clean_state()` rule an
+                // unshared nexus is considered clean (no possible in-flight I/O)
+                // and its children are all trusted, which would make the initial
+                // "one clean replica" assertion below unreachable.
+                Some(VolumeShareProtocol::Nvmf),
                 HashMap::new(),
-                vec![],
+                vec!["a".into()],
                 VolumeAccessMode::SingleNodeWriter,
             ),
             None,
@@ -1553,9 +1565,13 @@ async fn health() {
             &PublishVolume::new(
                 volume.spec().uuid.clone(),
                 Some(cluster.node(1)),
-                None,
+                // Match the earlier publish: shared over Nvmf keeps the health
+                // assertions below on the "trust one child" branch. The
+                // unshared/`clean_state() == true` path is exercised further
+                // down in this same test.
+                Some(VolumeShareProtocol::Nvmf),
                 HashMap::new(),
-                vec![],
+                vec!["a".into()],
                 VolumeAccessMode::SingleNodeWriter,
             ),
             None,
@@ -1614,6 +1630,66 @@ async fn health() {
                 && h.clean_replicas == 2
                 && h.live_healthy_replicas == 0
                 && h.online_healthy_replicas == 1
+        },
+    )
+    .await
+    .unwrap();
+
+    // "Trust all when unshared" branch of `NexusInfo::clean_state()`: with no
+    // share, no front-end I/O can be in flight, so all healthy replicas are
+    // safe as rebuild sources even if `clean_shutdown` is false.
+    // Bring the two stopped nodes back so we have a live target to publish on,
+    // then republish the same volume without a share protocol.
+    cluster.composer().start(&cluster.node(0)).await.unwrap();
+    cluster.composer().start(&cluster.node(1)).await.unwrap();
+    cluster
+        .wait_node_status(cluster.node(0), transport::NodeStatus::Online)
+        .await
+        .unwrap();
+    cluster
+        .wait_node_status(cluster.node(1), transport::NodeStatus::Online)
+        .await
+        .unwrap();
+    cluster.wait_pool_online(cluster.pool(0, 0)).await.unwrap();
+    cluster.wait_pool_online(cluster.pool(1, 0)).await.unwrap();
+
+    let volume = volume_client
+        .unpublish(
+            &UnpublishVolume::new(&volume.spec().uuid, false, vec![]),
+            None,
+        )
+        .await
+        .unwrap();
+
+    let volume = volume_client
+        .publish(
+            &PublishVolume::new(
+                volume.spec().uuid.clone(),
+                Some(cluster.node(0)),
+                None,
+                HashMap::new(),
+                vec![],
+                VolumeAccessMode::SingleNodeWriter,
+            ),
+            None,
+        )
+        .await
+        .unwrap();
+
+    cluster.composer().kill(&cluster.node(2)).await.unwrap();
+
+    wait_for_health(
+        &volume.state(),
+        &volume_client,
+        Duration::from_secs(2),
+        |h| {
+            h.healthy_replicas == 2
+                && !h.clean_shutdown
+                && h.all_healthy_replicas_clean
+                && h.clean_replicas == h.healthy_replicas
+                && h.online_clean_replicas == h.online_healthy_replicas
+                && h.live_healthy_replicas == 2
+                && h.online_healthy_replicas == 2
         },
     )
     .await
--- a/control-plane/agents/src/bin/core/volume/registry.rs
+++ b/control-plane/agents/src/bin/core/volume/registry.rs
@@ -134,16 +134,17 @@ impl Registry {
                     .any(|(r, t)| r == &c.uuid && t.status().online())
             });
             let online_healthy_replicas = online_healthy.count() as u8;
+            let all_healthy_replicas_clean = info.clean_state();
             VolumeHealth {
                 clean_shutdown: info.clean_shutdown,
                 healthy_replicas: info.nr_healthy_replicas(),
-                clean_replicas: if info.clean_shutdown {
+                clean_replicas: if all_healthy_replicas_clean {
                     info.nr_healthy_replicas()
                 } else {
                     info.nr_healthy_replicas().min(1)
                 },
                 online_healthy_replicas,
-                online_clean_replicas: if info.clean_shutdown {
+                online_clean_replicas: if all_healthy_replicas_clean {
                     online_healthy_replicas
                 } else {
                     online_healthy_replicas.min(1)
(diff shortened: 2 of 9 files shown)


yugchaudhari:
@tiagolobocastro please take a look whenever you have a minute.

control-plane/stor-port/src/types/v0/store/nexus_persistence.rs
```diff
@@ -305,3 +323,83 @@ pub async fn delete_all_v1_nexus_info<S: Store>(
...
+    fn deserialize_without_shared_field() {
+        let json = r#"{"clean_shutdown":false,"children":[]}"#;
+        let info: NexusInfo = serde_json::from_str(json).unwrap();
+        assert_eq!(info.shared, None);
+        assert!(!info.clean_shutdown);
+    }
+
+    /// New entries carry an explicit `shared` value which round-trips through serde.
+    #[test]
+    fn deserialize_with_shared_true() {
```
tiagolobocastro: We don't need to test serde

control-plane/stor-port/src/types/v0/store/nexus_persistence.rs
```diff
@@ -305,3 +323,83 @@ pub async fn delete_all_v1_nexus_info<S: Store>(
...
+    /// An unclean shutdown with the `shared` field absent (old entry / mixed-version rollout)
+    /// must be treated conservatively as if the nexus was shared.
+    #[test]
+    fn dirty_state_when_unclean_and_shared_missing() {
+        let info = NexusInfo {
+            clean_shutdown: false,
+            shared: None,
+            ..Default::default()
+        };
+        assert!(!info.clean_state());
```
tiagolobocastro: you can test this on `deserialize_without_shared_field` itself

tiagolobocastro (approved):

control-plane/stor-port/src/types/v0/store/nexus_persistence.rs
yugchaudhari: Done, dropped both `deserialize_with_shared_true` and `deserialize_with_shared_false`.

control-plane/stor-port/src/types/v0/store/nexus_persistence.rs
yugchaudhari: Done, added the `clean_state()` assertion into `deserialize_without_shared_field` and dropped the separate test.

Abhinandan-Purkait (approved):

yugchaudhari:
@tiagolobocastro @Abhinandan-Purkait let's merge this if this looks good.

tiagolobocastro:
bors merge

yugchaudhari:
> bors merge

@tiagolobocastro 

`volume::volume` and `volume::health` were still gated on the old `!clean_shutdown` semantics. Under the new `clean_state()` rule an unshared nexus (which is what `share: None` gives you) counts as clean, so the "trust one child" paths those tests were exercising became unreachable.

Fixed in the tests:
- nexus_persistence_test_iteration: forced `nexus_info.shared = Some(true)` before the write-back so all three fault variants stay in the shared-nexus scenario the assertions encode.
- health: publish with `Some(VolumeShareProtocol::Nvmf)` so io-engine actually records `shared: Some(true)` on the initial poll, leaving the first wait_for_health (kill one replica while target up) able to see the "trust one" case. Later polls (after unpublish) were already consistent with the new rule.

The bdd failures look unrelated, not from this PR. Please queue again whenever you have a moment.

control-plane/agents/src/bin/core/tests/volume/mod.rs
```diff
@@ -201,6 +201,14 @@ async fn nexus_persistence_test_iteration(
...
         }
     }
+    // The persistence test publishes with `share: None`, so io-engine records
+    // `shared: Some(false)` on the nexus. Under the new `clean_state()` rule that
+    // makes the fault scenarios "unshared and clean" and healthy_volume_replicas
+    // returns all children instead of the "trust one" path the test is exercising.
+    // Force `shared: Some(true)` to keep the scenario as "shared nexus with a
+    // fault" — which is what the assertions below (single child, per-fault URI)
+    // encode.
+    nexus_info.shared = Some(true);
```
tiagolobocastro: Then let's share the nexus instead so it can be more representative?

tiagolobocastro:
bors merge

control-plane/agents/src/bin/core/volume/registry.rs
```diff
@@ -134,16 +134,18 @@ impl Registry {
                     .any(|(r, t)| r == &c.uuid && t.status().online())
             });
             let online_healthy_replicas = online_healthy.count() as u8;
+            // See `NexusInfo::clean_state` for the trust-all rule.
+            let all_children_trusted = info.clean_state();
             VolumeHealth {
                 clean_shutdown: info.clean_shutdown,
```
tiagolobocastro, line 139: Thinking about this more, you can't walk backwards to the health here without the share
So should we include the share here, or perhaps expose the fact children are in-sync some other way?

CC @Abhinandan-Purkait

yugchaudhari: Two options, let me know which one you prefer:

A. Add `shared: Option<bool>` on `VolumeHealth`, mirroring the `NexusInfo` field. Callers keep both `clean_shutdown` and `shared` and compute the trust-all rule externally.

B. Add a first-class `all_children_trusted: bool` (or similar name) on `VolumeHealth`. Exposes the `clean_state()` output directly, so callers don't need to know the combination rule.

I prefer B, the whole point of `NexusInfo::clean_state()` was to centralize the rule and re-implementing it in every caller seems to defeat the purpose of the helper. But A is more literal to what io-engine actually persists. Happy to do both if that's cleaner: `shared` for observability + `all_children_trusted` for the derived signal.

tiagolobocastro:
FYI BDD tests fixed: <https://github.com/openebs/mayastor-control-plane/commit/99884ef8d65d26f9be400182ca36e6b3379e6044>

control-plane/agents/src/bin/core/volume/registry.rs
tiagolobocastro: Ideally I'd like to rename the existing nomenclature, but it's public api so we'd have to break with a major version change :/

I suggest B but with slight naming change: `all_healthy_replicas_clean` ?

yugchaudhari: Sounds good, I've named the field `all_healthy_replicas_clean`. Pushed the field on `VolumeHealth` + wiring in `registry.rs` so it comes straight from `NexusInfo::clean_state()`.

tiagolobocastro (approved):

control-plane/agents/src/bin/core/volume/registry.rs
tiagolobocastro: @Abhinandan-Purkait ptal

Abhinandan-Purkait: `all_healthy_replicas_clean` or `all_healthy_replicas_synced` ? Either way sounds good.

Abhinandan-Purkait (approved):

yugchaudhari:
Both int-test failures (`volume::volume` and `volume::health`) came from the same root cause: this PR's `NexusInfo::clean_state()` change (`clean_shutdown || !shared`) means an unshared crashed nexus now trusts all healthy replicas. The `volume::health` flow republishes the volume with `share: None` mid-test and then asserts on `clean_replicas == 1`, assertions written for the old "always trust one" behavior that no longer hold under the new semantics.

Two changes to close this:

1. Keep the existing test on the "trust one" branch.
Changed the republish at `volume/mod.rs:1587` from `None` back to `Some(VolumeShareProtocol::Nvmf)` (with `frontend_nodes: ["a"]`). The rest of the assertions below it are all about the shared/`clean_state()==false` path, sharing the nexus keeps them relevant.

2. Add explicit coverage for the new "trust all when unshared" branch.
New test `health_unshared_trusts_all_replicas` publishes an unshared nexus on a 3-replica volume, kills one io-engine, and waits until health reports `healthy_replicas == 2`, `!clean_shutdown`, and `all_healthy_replicas_clean == true`. That last assertion is the one that only holds under the new semantics, so it locks in the intended behaviour rather than leaving it implicit.

Verified locally on VM: full agents-core int-test binary runs green.

cc @tiagolobocastro

control-plane/agents/src/bin/core/tests/volume/mod.rs
```diff
@@ -1620,6 +1636,75 @@ async fn health() {
...
+/// Coverage for the "trust all when unshared" branch of
+/// `NexusInfo::clean_state()`: when the nexus is published without a share, no
+/// front-end I/O can be in flight, so all healthy replicas are safe as rebuild
+/// sources even if `clean_shutdown` is false. `clean_replicas` and
+/// `online_clean_replicas` should equal their healthy/online counterparts, and
+/// `all_healthy_replicas_clean` should report true.
+#[tokio::test]
+async fn health_unshared_trusts_all_replicas() {
+    let cluster = ClusterBuilder::builder()
+        .with_rest(true)
```
tiagolobocastro: can we run this on the existing health test above?

PR #1141, opened 2026-07-09
Auto reload tls secrets

tiagolobocastro:
feat(rest/server): hot-reload TLS certificates without restart
    
    Serve the REST server certificate through a swappable ResolvesServerCert
    and (for mTLS) a swappable ClientCertVerifier.
    
    A background inotify watcher reloads both when the files change on disk
    (e.g. cert-manager rotation or secret recreation), falling back to polling
    when inotify is unavailable.
    
    A catch-up reload after (re-)arming the watches closes the race where
    a rotation lands before the watch is armed.
    
    Also add the same catch-up reload to the openapi tower-hyper client
    template's TLS reload watcher.
    
---

    feat(rest/client): add auto-reload by watching for cert changes
    
    Adds opt-in automatic reloading of file-based TLS certificates for the
    tower-hyper OpenAPI client. TlsMode now retains the source certificate
    paths alongside the loaded bytes, so renewed certificates (e.g.
    cert-manager rotations) can be re-read at runtime without rebuilding the
    Configuration.
    
    A background TlsReloadService watches the certificate paths via inotify
    (Linux) and swaps the inner HTTP client once a change is observed.
    A last-modified fingerprint guards against duplicate/spurious events so the
    connector is rebuilt at most once per actual change, and a polling
    fallback covers the cases where inotify cannot be armed.
    When all certs share a parent directory that directory is watched, which
    catches atomic symlink-swap rotations.
    
    Reload is enabled automatically whenever the TLS mode carries certificate
    paths (ClientSecurity::try_new/new), and can be toggled explicitly via
    ClientSecurity::with_tls_reload.
    Byte-only modes and non-Linux platforms keep the existing static connector.

### DIFF
--- a/control-plane/rest/service/src/main.rs
+++ b/control-plane/rest/service/src/main.rs
@@ -1,5 +1,6 @@
 mod authentication;
 mod health;
+mod tls;
 mod v0;
 
 use crate::{
@@ -27,7 +28,7 @@ use rustls::{
     RootCertStore, ServerConfig,
 };
 use rustls_pemfile::{certs, private_key};
-use std::{fs::File, io::BufReader, path::PathBuf, sync::Arc, time::Duration};
+use std::{io::BufReader, path::PathBuf, sync::Arc, time::Duration};
 use stor_port::transport_api::{RequestMinTimeout, TimeoutOptions};
 use utils::{
     tracing_telemetry::{FmtLayer, FmtStyle, KeyValue},
@@ -72,7 +73,7 @@ pub(crate) struct CliArgs {
     dummy_certificates: bool,
 
     /// Auto-generate an ephemeral self-signed server certificate for HTTPS.
-    #[clap(long, conflicts_with_all = ["dummy_certificates", "cert_file", "key_file"])]
+    #[clap(long, conflicts_with_all = ["dummy_certificates", "cert_file", "key_file", "client_ca_file"])]
     auto_tls: bool,
 
     /// Path to the certificate authority (CA) bundle used to authenticate TLS clients.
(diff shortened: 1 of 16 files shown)


control-plane/rest/service/src/main.rs
```diff
@@ -122,17 +123,53 @@ impl CliArgs {
...
 
-/// default timeout options for every bus request
-fn timeout_opts() -> TimeoutOptions {
-    let timeout_opts =
-        TimeoutOptions::new_no_retries().with_req_timeout(CliArgs::args().request_timeout.into());
+    fn workers(&self) -> usize {
+        let max_workers = match self.max_workers {
+            0 => num_cpus::get_physical(),
+            max => max,
+        };
```
niladrih, line 169: Does this not require a default case handling?

control-plane/rest/service/src/main.rs
```diff
@@ -122,17 +123,53 @@ impl CliArgs {
...
 
-    if CliArgs::args().no_min_timeouts {
-        timeout_opts.with_min_req_timeout(None)
-    } else {
-        timeout_opts.with_min_req_timeout(RequestMinTimeout::default())
+        let workers = match self.workers {
+            0 => num_cpus::get_physical(),
+            workers => workers,
+        };
```
niladrih, line 174: Also, does this not require a default case handling?

niladrih (approved):
Could this build for anything which is not linux?

control-plane/rest/service/src/main.rs
tiagolobocastro: function not modified, only moved it to the CliArgs impl to use self

control-plane/rest/service/src/main.rs
tiagolobocastro: function not modified, only moved it to the CliArgs impl to use self

Abhinandan-Purkait (approved):

tiagolobocastro:
> Could this build for anything which is not linux?

The certificate auto-reload no, only for linux since it's using [inotify](https://www.google.com/url?sa=t&source=web&rct=j&opi=89978449&url=https://man7.org/linux/man-pages/man7/inotify.7.html)
It's probably possible to use another crate or system which is compatible with other OS, but it didn't seem worth the effort since we only build linux containers only anyway

PR #1142, opened 2026-07-10
allow selective reset on pool clear-error api

abhilashshetty04:
Integrates extension of clear-errors api changes.

Refer https://github.com/openebs/mayastor/pull/2034 for related change in mayastor


.gitmodules
```diff
@@ -1,4 +1,4 @@
 [submodule "utils/dependencies"]
 	path = utils/dependencies
 	url = ../mayastor-dependencies.git
-	branch = develop
+	branch = extend_clear_errors
```
abhilashshetty04: Breaking change on `develop` introduced by https://github.com/openebs/mayastor-dependencies/pull/162

cc @yugchaudhari

control-plane/grpc/proto/v1/pool/pool.proto
```diff
@@ -454,9 +454,14 @@ message ClearErrorsRequest {
 }
 
 enum ClearErrors {
-  // Clears all counted errors and related alerts
-  // Example, it clears the io stall transitions, but doesn't clear an io stall
-  ClearAll = 0;
+  // Clears all counted errors and stall transitions.
+  // Note: It doesn't clear an io stall state if the pool is currently in a stalled state.
+  // The stall state will be cleared when the pool is no longer stalled.
+  All = 0;
```
tiagolobocastro: Why remove the `Clear` ? This is also not mentioned in the empty commit message...

In protobuf it's good practice to keep a prefix because otherwise the enum value may clash

Abhinandan-Purkait (approved):

control-plane/grpc/proto/v1/pool/pool.proto
abhilashshetty04: I had tried to keep it consistent with io-engine protobuf but rust-lint was failing as all variants starts with same prefix.

error: all variants have the same prefix: `Clear`
   --> /home/admin/mayastor-control-plane/target/debug/build/grpc-a0ecb768991d3844/out/v1.pool.rs:783:1
    |
783 | / pub enum ClearErrors {
784 | |     /// Clears all counted errors and stall transitions.
785 | |     /// Note: It doesn't clear an io stall state if the pool is currently in a stalled state.
786 | |     /// The stall state will be cleared when the pool is no longer stalled.
...   |
791 | |     ClearIoStallTransitions = 2,
792 | | }
    | |_^
    |
    = help: remove the prefixes and use full paths to the variants instead of glob imports
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#enum_variant_names
    = note: `-D clippy::enum-variant-names` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(clippy::enum_variant_names)]`
   
   
   This works on openebs/mayastor maybe we skip linter on generated files there?

tiagolobocastro: Add `#![allow(clippy::enum_variant_names)]` to <https://github.com/openebs/mayastor-control-plane/blob/372010f5a1bcd8c7a33c1940f1f72559f07f03a7/control-plane/grpc/src/lib.rs#L41>

tiagolobocastro (approved):

abhilashshetty04:
@mergify queue

PR #1152, opened 2026-07-14
feat(app-node): plumb transport capabilities through registration

yugchaudhari:
## Summary

Adds an optional `TransportCaps { rdma_capable, ana_capable }` payload the csi-node reports at registration time, so the control-plane has each node's NVMe-oF RDMA and ANA multipath availability recorded on the `AppNodeSpec`.

## What's added

- **proto**: new `v1.common.TransportCaps` + optional field on `AppNodeSpec` / `RegisterRequest`
- **openapi**: matching `TransportCaps` schema + optional property
- **store / transport**: `TransportCaps` struct; `Option<TransportCaps>` on `AppNodeSpec` and `RegisterAppNode` with `#[serde(default)]` so existing etcd entries deserialize cleanly
- **gRPC conversion layer**: `AppNodeRegisterInfo::transport_caps()` trait method + `From` impls between proto ↔ store, openapi ↔ store, and through the `AppNodeSpec` conversions
- **core registry**: persists caps on register, and treats a change in caps as a spec change on re-registration
- **REST handler**: passes caps through from the openapi body to the transport type
- **csi-node**: reuses the RDMA/ANA detection already done at startup and passes it into the registration loop

## Backward compatibility

Optional end-to-end:
- an older csi-node that never populates the field just sends `None` and registers as before
- a control-plane reading an older csi-node's payload sees `None` and behaves as today
- existing etcd entries without the field deserialize cleanly (`#[serde(default)]` on the store type)

## Testing

- [x] Unit: `TransportCaps` round-trips proto ↔ store and openapi ↔ store
- [x] Unit: `AppNodeSpec` serde back-compat: a payload without `transport_caps` deserializes with the field set to `None`
- [x] Int: extended the existing `app_node_registration` test to register one node without caps and one with caps, and asserts the caps round-trip through gRPC/etcd, including across a `restart_core`

### DIFF
--- a/control-plane/grpc/proto/v1/misc/common.proto
+++ b/control-plane/grpc/proto/v1/misc/common.proto
@@ -240,3 +240,15 @@ message MapWrapper {
 message EncryptionSecret {
   string name = 1;
 }
+
+// Transport-layer capabilities reported by the csi-node at registration time.
+message TransportCaps {
+  // RDMA HCA hardware is present on this node (reported by `ibv_devinfo`).
+  bool rdma_hca_present = 1;
+  // The `nvme_rdma` kernel module is loaded on this node. Both this and
+  // `rdma_hca_present` are required for NVMe-oF RDMA to actually be usable.
+  bool nvme_rdma_module_loaded = 2;
+  // NVMe ANA multipath is enabled on this node
+  // (/sys/module/nvme_core/parameters/multipath = Y).
+  bool ana_capable = 3;
+}
--- a/control-plane/rest/openapi-specs/v0_api_spec.yaml
+++ b/control-plane/rest/openapi-specs/v0_api_spec.yaml
@@ -5116,6 +5116,10 @@ components:
         labels:
           org.io/csi-node.nvme-ana: true,
           org.io/zone: us-east-1a
+        transportCaps:
+          rdmaHcaPresent: true
+          nvmeRdmaModuleLoaded: true
+          anaCapable: true
       type: object
       properties:
         endpoint:
@@ -5129,6 +5133,8 @@ components:
           type: object
           additionalProperties:
             type: string
+        transportCaps:
+          $ref: '#/components/schemas/TransportCaps'
       required:
         - id
         - endpoint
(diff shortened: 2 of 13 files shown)


Abhinandan-Purkait (approved):

control-plane/grpc/proto/v1/misc/common.proto
```diff
@@ -240,3 +240,13 @@ message MapWrapper {
 message EncryptionSecret {
   string name = 1;
 }
+
+// Transport-layer capabilities reported by the csi-node at registration time.
+message TransportCaps {
+  // NVMe-oF RDMA transport is usable on this node
+  // (HCAs present and the nvme_rdma kernel module is loaded).
+  bool rdma_capable = 1;
```
tiagolobocastro: maybe split into 2 bools
would be good to differentiate between kernel module loaded and HCA present

control-plane/rest/openapi-specs/v0_api_spec.yaml
```diff
@@ -5126,9 +5131,30 @@ components:
...
         - endpoint
+    TransportCaps:
+      description: |-
+        Transport-layer capabilities reported by the csi-node at registration time.
+      type: object
+      example:
+        rdma_capable: true
+        ana_capable: true
+      properties:
+        rdma_capable:
```
tiagolobocastro: we've been using camel case for new fields IIRC

```suggestion
        rdmaCapable:
```

control-plane/stor-port/src/types/v0/store/app_node.rs
```diff
@@ -39,6 +39,21 @@ impl StorableObject for AppNodeSpec {
...
 pub type AppNodeLabels = std::collections::HashMap<String, String>;
 
+/// Transport-layer capabilities reported by the csi-node at registration time.
+/// Room to grow: further transport-related flags slot in here without needing
+/// another `AppNodeSpec` field per capability.
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq, Default)]
+pub struct TransportCaps {
+    /// NVMe-oF RDMA transport is usable on this node (HCAs present and the
+    /// `nvme_rdma` kernel module is loaded).
+    #[serde(default)]
```
tiagolobocastro:
```suggestion
```

This is not required because this entire object itself is new

tiagolobocastro (commented):
just a couple comments, otherwise lgtm

yugchaudhari:
> just a couple comments, otherwise lgtm

Addressed all three: split `rdma_capable` into `rdma_hca_present` + `nvme_rdma_module_loaded`; camelCase for the new openapi fields (`rdmaHcaPresent`, `nvmeRdmaModuleLoaded`, `anaCapable`, `transportCaps`); dropped the redundant `#[serde(default)]` on `TransportCaps` fields. csi-node now computes both signals separately so the CP can differentiate hardware vs kernel-module readiness.

cc @tiagolobocastro

tiagolobocastro (approved):

yugchaudhari:
@tiagolobocastro the int-test failure seems unrelated, it passed on last run, you can retry whenever you want.

PR #1156, opened 2026-07-18
test: fix racy health check

tiagolobocastro:
test: fix racy health check

    Fix racyness by avoiding rebuilds and faulting during rebuilds.
    We do this by ensuring only the healthy replicas are online when the nexus
    is created.

    We can then safely kill the nexus node and verify all replicas are clean.

---

    ci: ensure Cargo.lock is up-to-date in PR

    When we modify the deps Cargo.lock may need updating and we should ensure
    the CI catches any missing changes.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -32,3 +32,9 @@ jobs:
         run: nix-shell --run "black --diff --check tests/bdd/"
       - name: Lint nix code
         run: nix-shell --run "nixpkgs-fmt --check ."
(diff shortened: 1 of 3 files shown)


tiagolobocastro:
@yugchaudhari the health test was still flaky - I think this fixes it, ptal

abhilashshetty04 (approved):

yugchaudhari (approved):
Looks good, thanks for the clean up :)

PR #1158, opened 2026-07-28
Flaky ci

tiagolobocastro:
test(deployer): enable interrupt when not-isolated
    
    When we don't use one core per io-engine tests may fail when
    one of them doesn't get to run.
    To try and tackle this, let's use the interrupt mode in this
    case which should make cpu usage fairer.
    
---

    test: wait for non online status
    
    In case node is offline|unknown
    We probably need to clarify when a node can be offline vs unknown.

---

    test(deployer): stop should cleanup rwx vm
    
    Add rust impl for cleaning up the rwx vm.
    
---

    test(deployer): fix missing rwx vm cleanup after flag refactor
    
    In a previous PR the composer flags were silenced in favour of a composer wrapper
    which cleaned up by itself (without using composer).
    
    This introduced a regression where the rwx vm was not getting killed after the test
    was complete.
    This may affect the subsequent tests.
    
    To tackle this we move the wrapper to the deployer itself, and let the deployer
    query the wrapper for cleanup.
    
---

    fix: run fio on non/io-engine cores using taskset
    
    On recent kernel with the new scheduler, running 2 processes on the same
    core seems to be less fair and second process may not run sufficiently.
    
    We need a better approach for this, but for now at least modify fio runs
    to use a non-engine core.
    
---

    fix(agent/ha-node): slightly more robust new connection handling
    
    Seen during the tests, we may receive device add by udev and the state of the connection
    is connecting rather than new or live.
    
    In this case, we add a new state to avoid marking the path as suspected right away, since
    we know we've just added the path.
    
    Also add the subsystem to the logging as it could be useful for debugging.

### DIFF
--- a/control-plane/agents/src/bin/ha/node/detector.rs
+++ b/control-plane/agents/src/bin/ha/node/detector.rs
@@ -16,7 +16,9 @@ use tokio::{
 /// Possible states of every path record.
 #[derive(Debug, Clone, Eq, PartialEq)]
 enum PathState {
+    Added,
     Good,
+    AddedSuspected,
     Suspected,
     Failed,
     Removing,
@@ -55,7 +57,13 @@ impl PathRecords {
         self.paths
             .entry(path.clone())
             .or_insert_with(|| {
-                PathRecord::new(ctrlr.nqn().to_owned(), epoch, path, reporter.clone())
+                PathRecord::new(
+                    ctrlr.nqn().to_owned(),
+                    epoch,
+                    path,
+                    reporter.clone(),
+                    ctrlr.is_new(),
+                )
             })
             .with_epoch(epoch)
     }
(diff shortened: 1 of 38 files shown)


tiagolobocastro:
bors try

control-plane/agents/src/bin/ha/node/detector.rs
```diff
@@ -16,7 +16,9 @@ use tokio::{
 /// Possible states of every path record.
 #[derive(Debug, Clone, Eq, PartialEq)]
 enum PathState {
+    Added,
     Good,
+    AddedSuspected,
```
Abhinandan-Purkait, line 21: 👍🏻

Abhinandan-Purkait (approved):

niladrih (approved):

PR #1161, opened 2026-08-03
fix(core): offline rebuild reconciler picks up scale-up replicas

### LINKED ISSUE #2043: Scaled up unpublished but previously published volume remains in Degraded state
**Describe the bug**
When a published volume is unpublished. NexusInfo will have the list of children when it was being unpublished. When user scales up the volume a new replica/child is not part of the persisted NexusInfo. 

Even though Replica is Online from availability perspective its data is inconsistent most likely. under replica topology it has `state: Online` and `healthy: false `which marks volume `state: Degraded`

**To Reproduce**

scale up the previously published volume

**Expected behavior**
Offline volume rebuild reconciler should handle this too. Preferably without waiting for the --offline-rebuild-grace-period. This is needed for Mayastor pool drain feature also.

**Screenshots**

```
{
  "spec": {
    "num_replicas": 3,
    "size": 1073741824,
    "sizeBdev": 1073741824,
    "status": "Created",
    "uuid": "ec4e66fd-3b33-4439-b504-d49aba53da25",
    "policy": {
      "self_heal": true
    },
    "thin": false,
    "num_snapshots": 0,
    "max_snapshots": 10,
    "encrypted": false,
    "labelVersion": 1
  },
  "state": {
    "size": 1073741824,
    "status": "Degraded",
    "uuid": "ec4e66fd-3b33-4439-b504-d49aba53da25",
    "health": {
      "cleanShutdown": true,
      "healthyReplicas": 2,
      "cleanReplicas": 2,
      "onlineHealthyReplicas": 2,
      "onlineCleanReplicas": 2,
      "liveHealthyReplicas": 0,
      "allHealthyReplicasClean": true
    },
    "replica_topology": {
      "87bca435-5767-4d0b-84c5-bc50b56aacaa": {
        "node": "io-engine-3",
        "pool": "pool-3",
        "state": "Online",
        "usage": {
          "capacity": 1073741824,
          "allocated": 1073741824,
          "allocated_snapshots": 0,
          "allocated_all_snapshots": 0
        },
        "healthy": false,
        "encrypted": false
      },
      "769e821f-cc4e-4a1d-8b32-1dd12925d949": {
        "node": "io-engine-1",
        "pool": "pool-1",
        "state": "Online",
        "usage": {
          "capacity": 1073741824,
          "allocated": 1073741824,
          "allocated_snapshots": 0,
          "allocated_all_snapshots": 0
        },
        "healthy": true,
        "encrypted": false
      },
      "b61d1e05-60ea-480e-871f-a5bc52462e1a": {
        "node": "io-engine-2",
        "pool": "pool-2",
        "state": "Online",
        "usage": {
          "capacity": 1073741824,
          "allocated": 1073741824,
          "allocated_snapshots": 0,
          "allocated_all_snapshots": 0
        },
        "healthy": true,
        "encrypted": false
      }
    },
    "usage": {
      "capacity": 1073741824,
      "allocated": 1073741824,
      "allocated_replica": 1073741824,
      "allocated_snapshots": 0,
      "allocated_all_snapshots": 0,
      "total_allocated": 3221225472,
      "total_allocated_replicas": 3221225472,
      "total_allocated_snapshots": 0
    }
  }
}
```

NexusInfo (previous)

Image

Additional Context:

Mayastor version: v2.11.1

yugchaudhari:
Fixes openebs/mayastor#2043.

When someone scales up an unpublished volume, the new replica is created but isn't in the persisted `NexusInfo` yet, so it comes back Online with `healthy: false` and the volume flips to `Degraded`. The offline-rebuild reconciler *does* pick this up eventually, but it treats the situation like a transient replica blip and waits out the full `--offline-rebuild-grace-period` first. Then, even after the grace window elapses, `initiate_offline_rebuild` bails out because `volume_pool_candidates` sees no free pool for a "replacement", every pool already has a replica.

Both of those checks assume we're recovering from a lost replica. In the scale-up case (and, per the issue, in pool-drain relocations too) the "unhealthy" replica is right there in the topology; there's no transient period to absorb and no need for a new pool, the extra replica *is* the rebuild destination.

The reconciler now compares the current replica topology against the persisted `NexusInfo.children`. If any replica UUID is missing from `NexusInfo`, that's the signal:

  * skip the grace-period gate: the unhealthy flag is definitional, not transient;
  * skip the candidate-pool check in `initiate_offline_rebuild`: the extra replica is the in-place target, so `HotSpareReconciler` can rebuild it once the temp nexus is stood up.

If `NexusInfo` can't be loaded (missing key, transient store error) the code falls back to the existing conservative behaviour.

### Testing

Added `offline_rebuild_scale_up_bypasses_grace` alongside the existing e2e. It configures a 30s grace period, publishes → unpublishes → scales 2 → 3, and asserts the temp target appears well inside the window. On develop, the test times out at 8s waiting for the temp nexus; with the fix the rebuild starts within a couple of reconcile ticks and the volume returns to `Online`.

Both the new test and the four existing scenarios (`happy_path`, `promote_on_publish`, `gc_safety`, `no_pool_candidate_defers_rebuild`) pass.

cc: @abhilashshetty04

### DIFF
--- a/control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
+++ b/control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
@@ -13,8 +13,14 @@ use crate::{
 
 use stor_port::types::v0::{
     store::volume::VolumeSpec,
-    transport::{PublishVolume, UnpublishVolume, VolumeState, VolumeStatus, VolumeTargetMode},
+    transport::{
+        PublishVolume, ReplicaId, ReplicaTopology, UnpublishVolume, VolumeState, VolumeStatus,
+        VolumeTargetMode,
+    },
 };
+
+use agents::errors::SvcError;
+use std::collections::{HashMap, HashSet};
 
 /// Offline Volume Rebuild.
 ///
@@ -109,26 +115,97 @@ async fn offline_rebuild_reconcile(
         return PollResult::Ok(PollerState::Idle);
     }
 
-    let degraded_for = volume.lock().metadata.offline_rebuild_degraded();
-    let grace_period = context.registry().offline_rebuild_grace_period();
-    if degraded_for < grace_period {
-        tracing::debug!(
+    // Scale-up added a replica that was never part of the last-published
+    // nexus, so it is definitively out-of-sync — not a transient blip the
+    // grace period is meant to absorb. If that replica is also currently
+    // online it can serve as an in-place rebuild target, so no fresh pool is
+    // needed either.
+    let has_in_place_target =
+        has_scale_up_rebuild_target(&volume, &volume_state.replica_topology, context).await?;
+
+    if !has_in_place_target {
+        let degraded_for = volume.lock().metadata.offline_rebuild_degraded();
+        let grace_period = context.registry().offline_rebuild_grace_period();
+        if degraded_for < grace_period {
+            tracing::debug!(
+                volume.uuid = %volume.uuid(),
+                remaining = ?grace_period.saturating_sub(degraded_for),
+                "Offline rebuild waiting for grace period"
+            );
+            return PollResult::Ok(PollerState::Busy);
+        }
+    } else {
+        tracing::info!(
             volume.uuid = %volume.uuid(),
-            remaining = ?grace_period.saturating_sub(degraded_for),
-            "Offline rebuild waiting for grace period"
+            "Bypassing grace period: replica added since last publish"
         );
-        return PollResult::Ok(PollerState::Busy);
     }
 
-    initiate_offline_rebuild(&mut volume, &volume_state, context).await
+    initiate_offline_rebuild(&mut volume, &volume_state, context, has_in_place_target).await
+}
+
+/// Returns `true` when a scale-up left an Online replica that the
+/// last-published nexus never had, making it an in-place rebuild destination.
+///
+/// Such a replica is known-out-of-sync by definition rather than transiently
+/// degraded, and being Online it is also reachable, so the rebuild can start
+/// without waiting on the grace period or standing up a fresh pool.
+///
+/// The Online guard matters: a spec-only replica whose node is currently down
+/// would still be in `replica_topology` and absent from `NexusInfo`, but the
+/// temp nexus would come up short one child and the rebuild would never make
+/// progress.
+///
+/// Known caveat: if the last publish itself came up short because a replica
+/// was unavailable then, `children.len()` already trails `num_replicas` and a
+/// replacement reads as a scale-up. That only costs an earlier rebuild.
+///
+/// If the `NexusInfo` cannot be loaded (missing key, transient store error),
+/// falls back to the safe default of *not* bypassing the grace period.
+async fn has_scale_up_rebuild_target(
+    volume: &OperationGuardArc<VolumeSpec>,
+    replica_topology: &HashMap<ReplicaId, ReplicaTopology>,
+    context: &PollContext,
+) -> Result<bool, SvcError> {
+    let Some(nexus_info) = context
+        .registry()
+        .nexus_info(
+            Some(volume.uuid()),
+            volume.as_ref().health_info_id(),
+            false,
+            None,
+        )
+        .await?
+    else {
+        return Ok(false);
+    };
+
+    // The replica count reconciler also adds replicas that are absent from
+    // NexusInfo, to replace a failed one. Those must still serve out the grace
+    // period in case the original comes back, so require the desired count to
+    // have outgrown the last-published child count.
+    if usize::from(volume.as_ref().num_replicas) <= nexus_info.children.len() {
+        return Ok(false);
+    }
+
+    let known: HashSet<_> = nexus_info.children.iter().map(|c| &c.uuid).collect();
+    Ok(replica_topology
+        .iter()
+        .any(|(uuid, topo)| topo.status().online() && !known.contains(uuid)))
 }
 
 /// Creates a non-shared nexus for the volume so the existing rebuild engine
 /// (HotSpareReconciler) can restore faulted replicas.
+///
+/// `has_in_place_target` skips the candidate-pool check: when an extra
+/// replica already sits in the topology (e.g. from scale-up while
+/// unpublished), that replica is itself the rebuild destination and no fresh
+/// pool is required.
 async fn initiate_offline_rebuild(
     volume: &mut OperationGuardArc<VolumeSpec>,
     volume_state: &VolumeState,
     context: &PollContext,
+    has_in_place_target: bool,
 ) -> PollResult {
     let registry = context.registry();
 
@@ -143,8 +220,8 @@ async fn initiate_offline_rebuild(
 
     // Pre-flight viability: only stand up the temp nexus if the rebuild can
     // actually make progress. Needs both a source (a healthy replica to copy
-    // from) and a target (a candidate pool with room for the replacement
-    // replica).
+    // from) and a target (either an in-place replica flagged unhealthy, or
+    // a candidate pool with room for a replacement replica).
     //
     // Prefer the strict health signal — `online_clean_replicas > 0` — when
     // available; that's the CP's own trust decision and already accounts for
@@ -162,18 +239,18 @@ async fn initiate_offline_rebuild(
                 .values()
                 .any(|r| r.status().online())
         });
-    let target_viable = || async {
-        !volume_pool_candidates(GetSuitablePools::new(volume.as_ref(), None), registry)
+    let target_viable = has_in_place_target
+        || !volume_pool_candidates(GetSuitablePools::new(volume.as_ref(), None), registry)
             .await
-            .is_empty()
-    };
+            .is_empty();
 
-    if !source_viable || !target_viable().await {
+    if !source_viable || !target_viable {
         tracing::debug!(
             volume.uuid = %volume.uuid(),
             source_viable,
+            has_in_place_target,
             "Offline rebuild deferred: rebuild not viable \
-            (no healthy source replica and/or no candidate pool)"
+            (no healthy source replica and/or no rebuild destination)"
         );
         return PollResult::Ok(PollerState::Idle);
     }
--- a/control-plane/agents/src/bin/core/tests/volume/offline_rebuild.rs
+++ b/control-plane/agents/src/bin/core/tests/volume/offline_rebuild.rs
@@ -61,6 +61,106 @@ async fn offline_rebuild_e2e() {
     no_pool_candidate_defers_rebuild(&cluster).await;
 }
 
+/// Scale-up on an unpublished volume adds a replica that is absent from the
+/// persisted `NexusInfo`. It appears in the topology as Online but
+/// healthy=false, flipping the volume to Degraded. The offline-rebuild
+/// reconciler must treat this as "known needs rebuild" and skip the grace
+/// period.
+///
+/// A long grace period is configured so a passing test proves the bypass: if
+/// the reconciler still waited on the grace timer, the temp nexus would not
+/// appear within the assertion window.
+#[tokio::test]
+async fn offline_rebuild_scale_up_bypasses_grace() {
+    const LONG_GRACE_SECS: u64 = 30;
+    const START_WINDOW_SECS: u64 = 8;
+
+    let reconcile = Duration::from_millis(RECONCILE_PERIOD_MS);
+    let cluster = ClusterBuilder::builder()
+        .with_rest(true)
+        .with_io_engines(3)
+        .with_tmpfs_pool(52428800)
+        .with_cache_period("250ms")
+        .with_reconcile_period(reconcile, reconcile)
+        .with_options(|o| {
+            o.with_isolated_io_engine(true)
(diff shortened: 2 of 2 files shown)


abhilashshetty04:
Thanks for the fix :) @yugchaudhari

abhilashshetty04 (approved):

control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
```diff
@@ -109,26 +112,83 @@ async fn offline_rebuild_reconcile(
...
 
-    let degraded_for = volume.lock().metadata.offline_rebuild_degraded();
-    let grace_period = context.registry().offline_rebuild_grace_period();
-    if degraded_for < grace_period {
-        tracing::debug!(
+    // Scale-up (or pool-drain relocation) added a replica that was never part
+    // of the last-published nexus, so it is definitively out-of-sync — not a
+    // transient blip the grace period is meant to absorb. The extra replica
+    // also acts as an in-place rebuild target so no fresh pool is needed
+    // (openebs/mayastor#2043).
```
abhilashshetty04: no need to tag the issue doc.

control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
```diff
@@ -109,26 +112,83 @@ async fn offline_rebuild_reconcile(
...
-    initiate_offline_rebuild(&mut volume, &volume_state, context).await
+    initiate_offline_rebuild(&mut volume, &volume_state, context, has_in_place_target).await
+}
+
+/// Returns `true` when the volume's current replica topology contains at
+/// least one replica whose UUID is absent from the persisted `NexusInfo`.
+///
+/// Such a replica cannot have been part of the last-published nexus (typical
+/// causes: user scaled `num_replicas` up while the volume was unpublished, or
+/// a pool drain moved a replica onto a fresh pstor slot). Its unhealthy flag
```
abhilashshetty04: what is fresh pstor slot?

control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
```diff
@@ -109,26 +112,83 @@ async fn offline_rebuild_reconcile(
...
 
-    initiate_offline_rebuild(&mut volume, &volume_state, context).await
+    initiate_offline_rebuild(&mut volume, &volume_state, context, has_in_place_target).await
+}
+
+/// Returns `true` when the volume's current replica topology contains at
+/// least one replica whose UUID is absent from the persisted `NexusInfo`.
+///
+/// Such a replica cannot have been part of the last-published nexus (typical
+/// causes: user scaled `num_replicas` up while the volume was unpublished, or
```
abhilashshetty04:
```suggestion
/// causes: user scaled `num_replicas` up while the volume was unpublished or
```

abhilashshetty04 (changes requested):

control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
```diff
@@ -109,26 +112,82 @@ async fn offline_rebuild_reconcile(
...
+        .await?
+    else {
+        return Ok(false);
+    };
+
+    let known: HashSet<_> = nexus_info.children.iter().map(|c| &c.uuid).collect();
+    Ok(volume_state
+        .replica_topology
+        .keys()
+        .any(|uuid| !known.contains(uuid)))
```
abhilashshetty04: Check if replica is online also.

Absent from NexusInfo" ≠ "is a viable rebuild destination". Consider: unpublished 3-replica volume, R3 was added while unpublished, then R3's node goes offline. R3 is still in replica_topology (spec-derived) and still absent from NexusInfo → has_in_place_target == true → the candidate-pool check is skipped entirely. source_viable is satisfied by R1/R2, so we publish a temp nexus that comes up with 2 children as we dont consider adding R3 Degraded, children.len() != num_replicas → Degraded, and teardown_if_rebuilt only acts on Online or Faulted. The temp nexus sits there indefinitely.

control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
```diff
@@ -109,26 +112,82 @@ async fn offline_rebuild_reconcile(
...
+/// causes: user scaled `num_replicas` up while the volume was unpublished or
+/// a pool drain moved a replica to a different pool). Its unhealthy flag is
+/// definitional rather than transient, so the offline-rebuild reconciler can
+/// skip the grace window and start the rebuild immediately.
+///
+/// If the `NexusInfo` cannot be loaded (missing key, transient store error),
+/// falls back to the safe default of *not* bypassing the grace period.
+async fn has_replica_missing_from_nexus_info(
+    volume: &OperationGuardArc<VolumeSpec>,
+    volume_state: &VolumeState,
```
abhilashshetty04: Can we pass just `ReplicaTopology`?

control-plane/agents/src/bin/core/tests/volume/offline_rebuild.rs
```diff
@@ -61,6 +61,106 @@ async fn offline_rebuild_e2e() {
     no_pool_candidate_defers_rebuild(&cluster).await;
 }
 
+/// Scale-up on an unpublished volume adds a replica that is absent from the
+/// persisted `NexusInfo`. It appears in the topology as Online but
+/// healthy=false, flipping the volume to Degraded. The offline-rebuild
+/// reconciler must treat this as "known needs rebuild" and skip the grace
+/// period (see openebs/mayastor#2043).
```
abhilashshetty04: Ticket referred here as well.

control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
yugchaudhari: I meant _different pool_, fixed.

control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
yugchaudhari: Done, it now takes `&HashMap<ReplicaId, ReplicaTopology>` directly.

control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
yugchaudhari: Done, added an `Online` guard. `topo.status().online() && !known.contains(uuid)`, so a spec-only replica whose node is down no longer flips `has_in_place_target` and the candidate-pool check runs as before.

abhilashshetty04 (approved):

Abhinandan-Purkait (approved):

tiagolobocastro (commented):
Doesn't this "fight" against the existing timer logic if a replica is added by the replica count reconciler? Because it also won't be having an entry in etcd

yugchaudhari:
> Doesn't this "fight" against the existing timer logic if a replica is added by the replica count reconciler? Because it also won't be having an entry in etcd

Are you referring to the case where the replica count reconciler adds a replacement during the grace window (e.g. node went down, reconciler adds R4 before we know if it's transient)? If so, should we gate on `spec.num_replicas > nexus_info.children.len()` to distinguish intentional scale-up from replacement additions?

tiagolobocastro:
> > Doesn't this "fight" against the existing timer logic if a replica is added by the replica count reconciler? Because it also won't be having an entry in etcd
> 
> Are you referring to the case where the replica count reconciler adds a replacement during the grace window (e.g. node went down, reconciler adds R4 before we know if it's transient)? If so, should we gate on `spec.num_replicas > nexus_info.children.len()` to distinguish intentional scale-up from replacement additions?

Yes, though not sure if `nexus_info.children.len()` would help in case nexus was last published with less children due to unavailability for example
We could add it and live with this caveat I suppose

yugchaudhari:
> Yes, though not sure if `nexus_info.children.len()` would help in case nexus was last published with less children due to unavailability for example We could add it and live with this caveat I suppose

Added it, the gate is now `num_replicas > nexus_info.children.len()` on top of the existing "Online replica absent from NexusInfo" check. A replacement from the replica count reconciler leaves `num_replicas` untouched, so it no longer bypasses and keeps waiting out the grace period.

On the caveat: agreed, and I've written it into the doc comment. If the last publish itself came up short because a replica was unavailable at the time, `children.len()` already trails `num_replicas` and a later replacement reads as a scale-up. The cost is just starting the rebuild earlier than the grace period would have, and that replica does need rebuilding either way, so it seemed worth living with rather than tracking extra state.

One more thing I noticed, the overlap may be narrower than it first appears. `volume_replica_count_reconciler` compares against `context.specs().volume_replicas()`, i.e. spec'd replicas rather than online ones, so a node simply going offline doesn't drop the count or trigger a replacement. It takes an actual replica or pool removal to get there. So the window where the two reconcilers can disagree is real, but smaller than a transient node-down.

Also renamed the helper to `has_scale_up_rebuild_target`, since the old name no longer described what it returns.

Happy to add an integration test for the negative case (replacement during the grace window must not bypass) if you think it's worth it, `no_pool_candidate_defers_rebuild` is close to the right shape, it'd mainly need a 4th io-engine and a pool removal to make the spec'd count actually drop. Let me know.

yugchaudhari:
> # Merge Queue Status
> * ✅ **Entered queue** — `2026-08-07 12:21 UTC` · Rule: `default` · triggered by @abhilashshetty04 with the [merge queue checkbox](https://github.com/openebs/mayastor-control-plane/pull/1161#issuecomment-5164885153)
> * ❌ **Checks failed** · on draft [merge queue: checking develop (0dcbd0f) and #1161 together #1164](https://github.com/openebs/mayastor-control-plane/pull/1164)
> * 🚫 **Left the queue** — `2026-08-07 12:43 UTC` · at `5a7fc1d33624b2b8b85d3edfce4e68b2dac2270b`
> 
> This pull request spent **22 minutes 31 seconds** in the queue, including **21 minutes 59 seconds** running CI.
> 
> **Waiting for**
> All conditions
> ## Reason
> The merge conditions cannot be satisfied due to failing checks
> 
> * `bors-ci`
> * `@github-actions/commitlint`
> * `@github-actions/submodule-branch`
> * `DCO`
> * `@github-actions/bors-ci`
> 
> Failing checks:
> 
> * ❌ [bors-ci](https://github.com/openebs/mayastor-control-plane/actions/runs/31177849660/job/92868673746) ([job log](https://github.com/openebs/mayastor-control-plane/actions/runs/31177849660/job/92868673746))
> 
> ## Hint
> You may have to fix your CI before adding the pull request to the queue again. If you update this pull request, to fix the CI, it will automatically be requeued once the queue conditions match again. If you think this was a flaky issue instead, you can requeue the pull request, without updating it, by posting a `@mergifyio queue` comment.
> 
> Tick the box to put this pull request back in the merge queue (same as `@mergifyio queue`).
> 
> * [ ]  Requeue this pull request

@abhilashshetty04 
Failed `int-test` (`volume::rwx::rwx`) seems unrelated to this PR.

abhilashshetty04:
> > # Merge Queue Status
> > 
> > * ✅ **Entered queue** — `2026-08-07 12:21 UTC` · Rule: `default` · triggered by @abhilashshetty04 with the [merge queue checkbox](https://github.com/openebs/mayastor-control-plane/pull/1161#issuecomment-5164885153)
> > * ❌ **Checks failed** · on draft [merge queue: checking develop (0dcbd0f) and #1161 together #1164](https://github.com/openebs/mayastor-control-plane/pull/1164)
> > * 🚫 **Left the queue** — `2026-08-07 12:43 UTC` · at `5a7fc1d33624b2b8b85d3edfce4e68b2dac2270b`
> > 
> > This pull request spent **22 minutes 31 seconds** in the queue, including **21 minutes 59 seconds** running CI.
> > **Waiting for**
> > All conditions
> > ## Reason
> > The merge conditions cannot be satisfied due to failing checks
> > 
> > * `bors-ci`
> > * `@github-actions/commitlint`
> > * `@github-actions/submodule-branch`
> > * `DCO`
> > * `@github-actions/bors-ci`
> > 
> > Failing checks:
> > 
> > * ❌ [bors-ci](https://github.com/openebs/mayastor-control-plane/actions/runs/31177849660/job/92868673746) ([job log](https://github.com/openebs/mayastor-control-plane/actions/runs/31177849660/job/92868673746))
> > 
> > ## Hint
> > You may have to fix your CI before adding the pull request to the queue again. If you update this pull request, to fix the CI, it will automatically be requeued once the queue conditions match again. If you think this was a flaky issue instead, you can requeue the pull request, without updating it, by posting a `@mergifyio queue` comment.
> > Tick the box to put this pull request back in the merge queue (same as `@mergifyio queue`).
> > 
> > * [x]   Requeue this pull request
> 
> @abhilashshetty04 Failed `int-test` (`volume::rwx::rwx`) seems unrelated to this PR.

@yugchaudhari , seems like an infra issue. failed while building cluster. Requeued.

abhilashshetty04:
@mergifyio queue

PR #1143, opened 2026-07-10
fix: add read_only field to v1 Nexus/PublishNexus struct literals

yugchaudhari:
Bumps utils/dependencies to include openebs/mayastor-dependencies#162 which added an `optional bool read_only` field to both the v1 `Nexus` proto message and `PublishNexusRequest`. Two sites in the control-plane construct these as struct literals without `..Default::default()`, so bumping the submodule alone breaks the build with `E0063: missing field read_only`.

Add the field explicitly as `None` at both sites:

  - `utils/bundle-sim/src/bin/io-engine.rs:190`: the `Nexus { ... }` literal in the bundle-sim io-engine simulator. bundle-sim replays a snapshot of nexus state and doesn't model publish transitions, so `None` matches the "not currently tracked" shape.
  - `control-plane/agents/src/bin/core/controller/io_engine/v1/translation.rs:530`: the `PublishNexusRequest { ... }` literal in the `ShareNexus -> PublishNexusRequest` translation. ROX is not modelled on the CP side yet, so publish always requests RWO here until the consumer PR lands and threads a per-volume `read_only` decision through this call.

Companion to the io-engine fix: openebs/mayastor#2033.

Refs: openebs/mayastor#2033
Refs: openebs/mayastor-dependencies#162

PR #1162, opened 2026-08-05
feat(rox): plumb read_only through publish path

yugchaudhari:
Iteration 2 of the ROX-fs OEP ([openebs/openebs#4242](https://github.com/openebs/openebs/issues/4242)). The io-engine landed the nexus-level `read_only` flag in [openebs/mayastor#2035](https://github.com/openebs/mayastor/pull/2035): writes reach `Nexus::submit_request` and are rejected with `SPDK_NVME_SC_COMMAND_NAMESPACE_IS_PROTECTED`, and `reservation_acquire` is skipped so multi-host attach works. This PR is the control-plane consumer for that flag: it plumbs a `read_only` decision from the publish access mode all the way down to `PublishNexusRequest`, and keeps the flag around across republish, reconciler re-share, and offline-rebuild promotion.

## What this PR does

Adds a `read_only` field on three type layers so the ROX intent survives every publish-shaped transition:

- **`TargetConfig.read_only`** (persisted per-publish): set from the publish access mode, carried across `republish`, and re-applied by the offline-rebuild promotion branch when a CSI publish arrives mid-rebuild.
- **`NexusSpec.read_only`** (persisted per-nexus): updated on `Share`, cleared on `Unshare`. The nexus reconciler's re-share paths (`fixup_nexus_protocol`) forward this into `ShareNexus` so a re-share after a target-node bounce doesn't silently revert the mode to RWO.
- **`ShareNexus.read_only`** (transport): propagated to the io-engine via `PublishNexusRequest.read_only`, replacing the `None` placeholder from #1143.

Also introduces a `VolumeAccessMode::MultiNodeReaderOnly` variant on the CP-side enum plus the matching proto/openapi `AccessMode::MultiNodeReaderOnly` so `PublishVolume` can distinguish ROX from the existing RWO/RWX modes.

### Mode-conflict guard

A volume already published as ROX cannot be concurrently republished as RWO, and vice versa, same target cannot serve as writer and reader-only at the same time. `next_target_config` returns a new `SvcError::VolumeAccessModeConflict` (`FailedPrecondition`) when the requested mode disagrees with the currently-published mode. Callers must unpublish first, then re-publish with the new mode. The guard is skipped for offline-rebuild targets, which are unshared scratch state and don't serve front-end I/O, a CSI publish mid-rebuild is a promotion, not a switch, and the promotion branch reconciles the mode itself.

## Backward compatibility

Etcd-persisted:

- `TargetConfig` and `NexusSpec` both gain `#[serde(default)]` on `read_only`, so persisted entries from pre-ROX builds deserialize to `read_only: false` rather than erroring.
- Explicit round-trip + missing-key deserialization unit tests cover this in both modules.

Wire:

- `TargetConfig`, `NexusSpec`, and `ShareNexusRequest` all carry `read_only` as `optional`, so older senders/receivers default to RWO, same policy the io-engine iteration 1 PR used.

## Testing

New `volume::rox::rox_publish_and_mode_conflict_guard` int-test covers:

- A `MultiNodeReaderOnly` publish sets `TargetConfig.read_only == true`.
- The publish's `frontend_nodes` all appear in the target's `frontend` nqns, reuses the existing RWX-block multi-initiator fan-out with no separate code path.
- RWO on a ROX-published volume → `VolumeAccessModeConflict`.
- ROX on an RWO-published volume → same error.
- Both transitions work cleanly after an unpublish (ROX → unpublish → RWO succeeds and clears `read_only`; RWO → unpublish → ROX succeeds and sets it back).

## What is deferred for future iterations

- CSI controller: accept `MultiNodeReaderOnly` in `check_volume_capability`'s `AccessType::Mount` branch, drop the unconditional `args.readonly` reject in `ControllerPublishVolume`, add the `rox_fs` StorageClass parameter (mirroring `rwx_block`).
- Add the `ro,noload` / `ro,norecovery` mount options in CSI node.

Until iteration 3 lands, the new access-mode variant is only reachable via the internal gRPC path, the CSI side still rejects `MultiNodeReaderOnly` for `Mount`, so end-user behavior is unchanged.

## References

- OEP: [openebs/openebs#4231](https://github.com/openebs/openebs/pull/4231), tracking [openebs/openebs#4242](https://github.com/openebs/openebs/issues/4242)
- io-engine iteration 1: [openebs/mayastor#2035](https://github.com/openebs/mayastor/pull/2035)

### DIFF
--- a/control-plane/agents/src/bin/core/tests/volume/rox.rs
+++ b/control-plane/agents/src/bin/core/tests/volume/rox.rs
@@ -1,0 +1,189 @@
+#![cfg(test)]
+
+use deployer_cluster::ClusterBuilder;
+use grpc::operations::volume::traits::VolumeOperations;
+use std::{collections::HashMap, time::Duration};
+use stor_port::{
+    transport_api::{ReplyErrorKind, ResourceKind},
+    types::v0::transport::{
+        CreateVolume, PublishVolume, UnpublishVolume, VolumeAccessMode, VolumeId,
+        VolumeShareProtocol,
+    },
+};
+
+/// End-to-end plumbing test for read-only (ROX) publishes. Covers:
+///
+/// - A `MultiNodeReaderOnly` publish sets `TargetConfig.read_only == true` on
+///   the volume spec. The resulting nexus records one `allowed_hosts` entry per
+///   `frontend_node` passed on the publish, which is the RWX-block multi-
+///   initiator fan-out — ROX inherits it without a separate code path.
+/// - RWO ↔ ROX mode switches on an already-published volume are rejected with
+///   `SvcError::VolumeAccessModeConflict` (→ `FailedPrecondition`).
+/// - Unpublish followed by a publish in the other mode succeeds: the guard is
+///   about concurrent conflicting modes on one target, not about forbidding
+///   transitions.
+#[tokio::test]
+async fn rox_publish_and_mode_conflict_guard() {
+    // `no_deprecated_access_mode(true)` flips core into strict ACL mode; without
+    // it the `--deprecated-access-mode` arg makes `frontend_nodes` a no-op and
+    // `allowed_hosts` stays empty. NQNs get synthesised from the nodename
+    // string via `HostNqn::from_nodename`, so we don't need real csi-node
+    // containers.
+    let cluster = ClusterBuilder::builder()
+        .with_rest(false)
+        .with_io_engines(2)
+        .with_tmpfs_pool(100 * 1024 * 1024)
+        .with_cache_period("150ms")
+        .with_options(|o| o.no_deprecated_access_mode(true))
+        .with_reconcile_period(Duration::from_secs(100), Duration::from_secs(100))
+        .build()
+        .await
+        .unwrap();
+
+    let volume_client = cluster.grpc_client().volume();
+    let volume = volume_client
+        .create(
+            &CreateVolume {
+                uuid: VolumeId::try_from("2e3cf927-80c2-47a8-adf0-95c481000000").unwrap(),
+                size: 5242880,
+                replicas: 2,
+                ..Default::default()
+            },
+            None,
+        )
+        .await
+        .unwrap();
+
+    // === ROX publish sets read_only + fans allowed_hosts to every frontend_node ===
+    let frontends: Vec<String> = vec!["reader-a".into(), "reader-b".into()];
+    let published = volume_client
+        .publish(
+            &PublishVolume {
+                uuid: volume.spec().uuid.clone(),
+                target_node: Some(cluster.node(0)),
+                share: Some(VolumeShareProtocol::Nvmf),
+                publish_context: HashMap::new(),
+                frontend_nodes: frontends.clone(),
+                access_mode: VolumeAccessMode::MultiNodeReaderOnly,
+            },
+            None,
+        )
+        .await
+        .unwrap();
+
+    assert!(
+        published
+            .spec_ref()
+            .target_cfg()
+            .expect("target config present")
+            .read_only(),
+        "MultiNodeReaderOnly publish must set TargetConfig.read_only"
+    );
+
+    // `Nexus::allowed_hosts` returned via gRPC is hardcoded to `vec![]` in the
+    // wire→transport conversion, so the CP-observable check is the target's
+    // frontend config, which is what actually gets forwarded to `share_nvmf`
+    // via `ShareNexus::allowed_hosts` on the io-engine side.
+    let cfg_hosts = published
+        .spec_ref()
+        .target_cfg()
+        .expect("target config present")
+        .frontend()
+        .node_nqns();
+    assert_eq!(
+        cfg_hosts.len(),
+        frontends.len(),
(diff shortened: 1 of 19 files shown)


control-plane/stor-port/src/types/v0/store/nexus.rs
```diff
@@ -332,7 +340,7 @@ pub enum NexusOperation {
     Create,
     Destroy,
     Shutdown,
-    Share(NexusShareProtocol, Vec<HostNqn>),
+    Share(NexusShareProtocol, Vec<HostNqn>, bool),
```
tiagolobocastro: Is this safe, will the old Share deserialise correctly with the new bool field?

control-plane/stor-port/src/types/v0/store/nexus.rs
```diff
@@ -485,3 +498,46 @@ impl NexusStatusInfo {
         self.reshutdown
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::types::v0::transport::{CreateNexus, NexusId};
+
+    fn sample_nexus_spec(read_only: bool) -> NexusSpec {
```
tiagolobocastro: We don't need to test serde
OTOH we should test the Share where there's no serde default there, I suspect it won't work

control-plane/stor-port/src/types/v0/store/nexus.rs
yugchaudhari: Fixed, added `#[serde(default)]` on the bool.

control-plane/stor-port/src/types/v0/store/nexus.rs
yugchaudhari: Done.

tiagolobocastro (approved):

tiagolobocastro:
@Abhinandan-Purkait @abhilashshetty04 PTAL

control-plane/agents/src/bin/core/tests/volume/rox.rs
```diff
@@ -0,0 +1,189 @@
...
+use std::{collections::HashMap, time::Duration};
+use stor_port::{
+    transport_api::{ReplyErrorKind, ResourceKind},
+    types::v0::transport::{
+        CreateVolume, PublishVolume, UnpublishVolume, VolumeAccessMode, VolumeId,
+        VolumeShareProtocol,
+    },
+};
+
+/// Iter 2 of the ROX-fs OEP end-to-end plumbing test. Covers:
```
abhilashshetty04: Maybe don't need to mention which iteration the patch belongs to in the code.

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -667,6 +706,14 @@ impl OperationGuardArc<VolumeSpec> {
                 if let Some(share) = request.share {
                     target_cfg.target_mut().set_protocol(Some(share));
                 }
+                // The reconciler-created target defaults to RWO; the CSI publish
```
abhilashshetty04: By "reconciler-created target" do you mean target which we create for offline rebuild?

control-plane/agents/src/common/errors.rs
```diff
@@ -176,6 +176,15 @@ pub enum SvcError {
         node: String,
         protocol: String,
     },
+    #[snafu(display(
```
abhilashshetty04, line 179: How does user switch access mode from k8s interfaces? As i read, only RWO -> RWOP switch is supported by PVC update.

https://kubernetes.io/docs/tasks/administer-cluster/change-pv-access-mode-readwriteoncepod/

control-plane/agents/src/bin/core/volume/operations.rs
yugchaudhari: Yes, it's gated on `is_offline_rebuild_target()`, i.e. `VolumeTargetMode::OfflineRebuild`. I'll reword it to say "offline-rebuild target" rather than "reconciler-created", since `MaintenanceMode` is also reconciler-shaped and this branch deliberately doesn't cover it.

control-plane/agents/src/bin/core/tests/volume/rox.rs
yugchaudhari: Fixed.

control-plane/agents/src/common/errors.rs
yugchaudhari: Fair question, and on this PR they can't, `check_volume_capability` still rejects `MultiNodeReaderOnly` and `controller_publish_volume` still rejects `readonly`, so ROX is only reachable over the internal gRPC/REST publish path. CSI acceptance and the `rox_fs` SC parameter are iter3.

And you're right that there's no PVC-edit path even then. Rebinding a PV to a new PVC needs the old one released first, which means an `unpublish` happened, and `unpublish` → `publish` in the other mode is allowed. So this isn't guarding a user workflow, it's guarding against two conflicting publishes landing on the same live target, which is really only reachable over the internal API today.

Happy to drop it to iter3 with the CSI work if you'd rather not carry an error variant nothing user-facing can hit yet.

abhilashshetty04 (approved):

PR #1163, opened 2026-08-07
refresh node rebuild count on single nexus update

abhilashshetty04:
NodeWrapper caches a per-node num_rebuilds counter, which the registry sums across all nodes in rebuild_allowed() to enforce the system-wide --max-rebuilds limit.

That counter was refreshed in update_resources() for ResourceType::All and ResourceType::Nexuses, but not for the single-nexus ResourceType::Nexus path. So whenever a nexus state was updated individually (e.g. on a nexus-specific state refresh rather than a full node poll), the cached rebuild count went stale.

Because of this newly started rebuilds weren't reflected, so more rebuilds than max_rebuilds could be admitted.

This PR fixes this by updating rebuild count after every online_child operation.

### DIFF
--- a/control-plane/agents/src/common/errors.rs
+++ b/control-plane/agents/src/common/errors.rs
@@ -556,6 +556,7 @@ impl SvcError {
             Self::ReplicaSetPropertyFailed { .. } => tonic::Code::DataLoss,
             Self::ReplaceNqnNotFound { .. } => tonic::Code::FailedPrecondition,
             Self::PoolCreateError { code, .. } => *code,
+            Self::MaxRebuilds { .. } => tonic::Code::OutOfRange,
             _ => tonic::Code::Internal,
         }
     }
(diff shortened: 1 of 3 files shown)


control-plane/agents/src/common/errors.rs
```diff
@@ -556,6 +556,7 @@ impl SvcError {
             Self::ReplicaSetPropertyFailed { .. } => tonic::Code::DataLoss,
             Self::ReplaceNqnNotFound { .. } => tonic::Code::FailedPrecondition,
             Self::PoolCreateError { code, .. } => *code,
+            Self::MaxRebuilds { .. } => tonic::Code::ResourceExhausted,
```
tiagolobocastro, line 559: Could also consider `OUT_OF_RANGE` or `FAILED_PRECONDITION` ?

tiagolobocastro (approved):

rohan2794 (approved):

PR #1169, opened 2026-08-14
add drain_spec to PoolUSpec

abhilashshetty04:
Adds the persisted drain specification to the pool spec, laying the groundwork for pool drain. Nothing sets it yet.

### DIFF
--- a/control-plane/grpc/proto/v1/pool/pool.proto
+++ b/control-plane/grpc/proto/v1/pool/pool.proto
@@ -2,6 +2,8 @@ syntax = "proto3";
 
 import "v1/misc/common.proto";
 import "google/protobuf/wrappers.proto";
+import "google/protobuf/timestamp.proto";
+import "google/protobuf/duration.proto";
 
 package v1.pool;
 
@@ -122,9 +124,37 @@ message PoolSpec {
   // Cordon Drain state
   oneof CordonDrain {
     CordonedState cordoned = 7;
+    DrainSpec drain = 9;
   }
   // Blobstore cluster size required for this pool
   optional uint32 cluster_size = 8;
+}
+
+message DrainSpec {
+  // The time at which the drain request was made.
+  google.protobuf.Timestamp request_timestamp = 1;
+  // The user's drain policy, each of which alters what the drain operation will do.
+  DrainPolicy policy = 2;
+  // Holds user applied cordon configs if present before starting drain.
+  CordonedState user_cordon = 3;
+}
+
+message DrainPolicy {
+  // What the drain does with the snapshots left on the pool once all replicas are evacuated.
+  SnapshotPolicy snapshot_policy = 1;
+  // Grace period, after which an unplaceable replica is force-evicted.
+  optional google.protobuf.Duration unsafe_rebuild_otherwise_evict = 2;
+  // Skip the safe over-replicate flow and evict replicas directly.
+  bool unsafe_evict = 3;
+}
+
+// What a drain does with the snapshots left on the pool once all replicas are evacuated.
+// Snapshots cannot be moved, so the caller has to say up front what should happen to them.
+enum SnapshotPolicy {
+  // Move replicas only, leaving the snapshots in place.
+  Ignore = 0;
+  // Destroy the snapshots left on the pool once all replicas are evacuated.
+  AcceptLoss = 1;
 }
 
 // Pool information
--- a/control-plane/stor-port/src/types/v0/store/pool.rs
+++ b/control-plane/stor-port/src/types/v0/store/pool.rs
@@ -28,6 +28,7 @@ use std::{
     convert::From,
     fmt::Debug,
     ops::{Deref, DerefMut},
+    time::{Duration, SystemTime},
 };
 
 /// Pool data structure used by the persistent store.
@@ -280,6 +281,13 @@ impl PoolSpec {
                         self.cordon_drain = None;
                     }
                 }
+                CordonDrainState::Drain(spec) => {
+                    if let Some(uc) = &spec.user_cordon {
+                        if !uc.cordoned() {
+                            spec.user_cordon = None;
+                        }
+                    }
+                }
             }
         }
     }
@@ -306,13 +314,11 @@ impl PoolSpec {
     }
 
     /// Returns whether the pool is cordoned and its state.
-    pub fn cordoned(&self) -> Option<&CordonedState> {
-        self.cordon_drain.as_ref().map(|s| match s {
-            CordonDrainState::Cordoned(cordoned) => cordoned,
-        })
+    pub fn cordoned(&self) -> Option<CordonedState> {
+        self.effective_cordon()
     }
 
-    /// Check if the pool is cordoned for imports.
+    /// Check if the pool is cordoned for imports, as per the effective cordon policy.
     pub fn cordoned_imports(&self) -> bool {
         self.cordoned().map(|s| s.import).unwrap_or_default()
     }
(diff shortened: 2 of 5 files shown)


Abhinandan-Purkait (approved):

control-plane/grpc/proto/v1/pool/pool.proto
```diff
@@ -125,6 +126,35 @@ message PoolSpec {
   }
   // Blobstore cluster size required for this pool
   optional uint32 cluster_size = 8;
+  // Drain spec applied on the pool.
+  optional DrainSpec drain_spec = 9;
```
tiagolobocastro: I see above there's CordonDrain oneof, shouldn't it go there?

control-plane/stor-port/src/types/v0/store/pool.rs
```diff
@@ -754,3 +799,65 @@ impl From<CordonDrainState> for models::PoolCordonDrain {
...
+    /// Grace period after which an unplaceable replica is force-evicted.
+    /// `None` disables forced eviction entirely.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub unsafe_rebuild_otherwise_evict: Option<Duration>,
+    /// Skip the safe over-replicate flow and evict replicas directly.
+    pub unsafe_evict: bool,
+}
+
+impl DrainPolicy {
+    pub fn new(
```
tiagolobocastro, line 875: doc comment all pubs

control-plane/stor-port/src/types/v0/store/pool.rs
```diff
@@ -754,3 +799,65 @@ impl From<CordonDrainState> for models::PoolCordonDrain {
         }
     }
 }
+
+/// Drain spec applied on the pool.
+#[derive(Clone, Serialize, Deserialize, Debug, PartialEq)]
```
tiagolobocastro, line 838: Is the `PartialEq` because the `PoolSpec` has it?

control-plane/stor-port/src/types/v0/store/pool.rs
```diff
@@ -754,3 +799,65 @@ impl From<CordonDrainState> for models::PoolCordonDrain {
...
+}
+
+/// The user's drain policy, each of which alters what the drain is permitted to do.
+#[derive(Clone, Serialize, Deserialize, Debug, PartialEq)]
+pub struct DrainPolicy {
+    /// What to do with the snapshots left on the pool once all replicas are evacuated.
+    /// `None` when the user chose neither, which is only accepted if the pool holds no
+    /// snapshots, so the drain never has to fall back on a default here.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub snapshot_policy: Option<SnapshotPolicy>,
```
tiagolobocastro: We should just have a default policy

```suggestion
    pub snapshot_policy: SnapshotPolicy,
```

control-plane/grpc/proto/v1/pool/pool.proto
abhilashshetty04: cordon_drain holds user applied cordon configs in oneof. We want to keep it intact when we layer self-cordon on top of it during drain. When user cancels drain we remove self-cordon and existing user applied cordon stays intact.

Putting drain_spec in oneof would make it mutually exclusive with user applied cordon.

control-plane/stor-port/src/types/v0/store/pool.rs
abhilashshetty04: Yes, doesn't compile without it as `PoolSpec` derives it and holds `DrainSpec`.

control-plane/stor-port/src/types/v0/store/pool.rs
abhilashshetty04: User need not pass `--ignore-snapshot` or `--allow-snapshot-loss` if pool doesn't host snapshots or if they don't use snapshots at all. If pool has them they should explicitly define what they want to do.

control-plane/grpc/proto/v1/pool/pool.proto
tiagolobocastro: But we can keep it intact, just need to ensure the cordon is kept inside the Drain, like we have done for the node

control-plane/stor-port/src/types/v0/store/pool.rs
tiagolobocastro: But still this should become a policy with a certain value right? What is the default policy?

abhilashshetty04: I was thinking of not having any default policy for this, if user does not pass either of these flags and snapshots are present we will fail the drain request. Then user has to execute drain cmd again by passing snapshot specific flag.

https://github.com/openebs/openebs/blob/develop/designs/replicated-pv/mayastor/pool-drain.md?plain=1#L78

logical default could be --ignore-snapshot. If we make it default now and when we eventually add snapshot rebuild to this won't it be a breaking change to remove this default.

tiagolobocastro: Ok, so if policy has to be set, then it should not be optional in the pstor, should always be set to something

control-plane/grpc/proto/v1/pool/pool.proto
abhilashshetty04: Sure. will change this

control-plane/stor-port/src/types/v0/store/pool.rs
abhilashshetty04: Done, Ignore is default now.

control-plane/stor-port/src/types/v0/store/pool.rs
```diff
@@ -721,36 +775,119 @@ impl CordonDrainState {
...
+        }
+    }
+}
+
+/// The user's drain policy, each of which alters what the drain is permitted to do.
+#[derive(Clone, Serialize, Deserialize, Debug, PartialEq, Default)]
+pub struct DrainPolicy {
+    /// What to do with the snapshots left on the pool once all replicas are evacuated.
+    /// Defaults to `Ignore`, which leaves them in place.
+    #[serde(default)]
```
tiagolobocastro:
```suggestion
```

control-plane/stor-port/src/types/v0/store/pool.rs
```diff
@@ -721,36 +775,119 @@ impl CordonDrainState {
...
+    pub snapshot_policy: SnapshotPolicy,
+    /// Grace period after which an unplaceable replica is force-evicted.
+    /// `None` disables forced eviction entirely.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
```
tiagolobocastro:
```suggestion
    #[serde(skip_serializing_if = "Option::is_none")]
```

control-plane/stor-port/src/types/v0/store/pool.rs
tiagolobocastro: No I mean there should be no default here, default is only for when we're upgrading; this will be set what whatever the user specified right?

abhilashshetty04: Im confused, What if user does not specify either of --ignore-snapshots or --allow-snapshot-loss. Thats why i was thinking of having this as Option. We will accept drain request without snapshot specific flags if there are no snapshots. If there are, then reject the request stating its an required arg.

If we don't want to do that we need to have `Ignore` as default right?

abhilashshetty04: default is in `SnapshotPolicy::Ignore` only now.

tiagolobocastro: Maybe I misread but I thought you said we'd reject that, so it'll never never reach the pstor?

abhilashshetty04: `drain` cmd itself is rejected if user does not pass snapshot specific flags when pool has snapshots on them. No changes on spec at all.

control-plane/grpc/proto/v1/pool/pool.proto
```diff
@@ -122,11 +123,39 @@ message PoolSpec {
...
+  DrainPolicy policy = 2;
+  // Holds user applied cordon configs if present before starting drain.
+  CordonedState user_cordon = 3;
+}
+
+message DrainPolicy {
+  // What the drain does with the snapshots left on the pool once all replicas are evacuated.
+  SnapshotPolicy snapshot_policy = 1;
+  // Grace period, after which an unplaceable replica is force-evicted.
+  optional uint64 unsafe_rebuild_otherwise_evict = 2;
```
```suggestion
  optional google.protobuf.Duration unsafe_rebuild_otherwise_evict = 2;
```

control-plane/stor-port/src/types/v0/store/pool.rs
```diff
@@ -341,6 +347,44 @@ impl PoolSpec {
...
+    }
+
+    /// Returns the cordon policy in effect: the self-cordon while a drain is applied, otherwise
+    /// the cordon the user applied.
+    pub fn effective_cordon_policy(&self) -> Option<&CordonedState> {
+        self.cordon_drain.as_ref().map(|s| match s {
+            CordonDrainState::Cordoned(cordoned) => cordoned,
+            CordonDrainState::Drain(_) => &CordonedState::SELF_CORDON,
+        })
+    }
```
tiagolobocastro: No policy here, it's the cordon configuration or state

```suggestion
    /// Returns the cordon config in effect: the self-cordon while a drain is applied, otherwise
    /// the cordon the user applied.
    pub fn effective_cordon(&self) -> Option<&CordonedState> {
        self.cordon_drain.as_ref().map(|s| match s {
            CordonDrainState::Cordoned(cordoned) => cordoned,
            CordonDrainState::Drain(_) => &CordonedState::SELF_CORDON,
        })
    }
```

control-plane/stor-port/src/types/v0/store/pool.rs
```diff
@@ -341,6 +347,44 @@ impl PoolSpec {
...
+            CordonDrainState::Cordoned(_) => None,
+        }
+    }
+
+    /// Returns the cordon policy in effect: the self-cordon while a drain is applied, otherwise
+    /// the cordon the user applied.
+    pub fn effective_cordon_policy(&self) -> Option<&CordonedState> {
+        self.cordon_drain.as_ref().map(|s| match s {
+            CordonDrainState::Cordoned(cordoned) => cordoned,
+            CordonDrainState::Drain(_) => &CordonedState::SELF_CORDON,
```
tiagolobocastro: hmm this should be an additive value of user cordon + the inferred cordon as the user may have pool cordoned for imports as well

abhilashshetty04: Thought about it seems reasonable. However, If user has blocked import then the pool will not make progress and occupy a slot in Draining queue until user changes their cordon config.

Behind the scenes replicas may have got fully rebuild somewhere else also.

tiagolobocastro: Yeah but if the user blocked imports we should respect that, in which case there's no drain possible

tiagolobocastro: I mean if the node restarts and pool can't be imported

tiagolobocastro (approved):

PR #1172, opened 2026-08-18
ci: run image tests on arm runners

tiagolobocastro:
test: skip rwx if kvm is not available
    
    On CI ARM runners kvm is not available, so skip the test.
    
---

    build: fix bindgen and stdenv
    
    Use the platform bindgenHook to setup the BINDGEN_EXTRA_CLANG_ARGS and the
    LIBCLANG_PATH correctly
    Also clear all propagated deps/build information from the oxalic rust platform

---

    ci: run image tests on arm runners
    
    We can't run int/bdd until we build io-engine images
    We can't also run unit because there's no easy way atm to run
    unit and not run the int tests...
    
---

    ci: remove hardcoded user
    
    This used to be required (IIRC) for some mac runs (no longer it seems).

### DIFF
--- a/.github/actions/setup-nix/action.yml
+++ b/.github/actions/setup-nix/action.yml
@@ -27,6 +27,15 @@ inputs:
 runs:
   using: composite
   steps:
+    - name: Linux Runner Info
+      shell: bash
+      if: ${{ runner.os == 'Linux' }}
+      run: |
+        lscpu
+        free -h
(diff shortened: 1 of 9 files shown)


tiagolobocastro:
Delaying the unit/int tests until we build aarch64 io-engine images...

niladrih (approved):

PR #1174, opened 2026-08-19
add pool drain record

abhilashshetty04:
This PR Introduces the persisted state that will drive the pool drain procedure.

There is no rest changes yet. only core type and gRPC changes.

### DIFF
--- a/control-plane/grpc/proto/v1/pool/pool.proto
+++ b/control-plane/grpc/proto/v1/pool/pool.proto
@@ -107,6 +107,117 @@ message Metadata {
   // Note that this number may not match the volume-snapshots api, as it will include all snapshots,
   // configured to reside in the pool, sometimes even the failed ones until such time they are deleted.
   optional uint64 snap_count = 4;
+  // Populated when drain request is submitted.
+  // Stores states driving the drain procedure.
+  optional PoolDrainRecord drain_record = 5;
+}
+
+// Record of an in-progress pool drain, driving the drain procedure and tracking its progress.
+message PoolDrainRecord {
+  // Current phase of the drain state machine.
+  DrainPhase phase = 1;
+  // Why pool is in the aforementioned phase.
+  optional PhaseReason phase_reason = 2;
+  // Statistics about the pool's usage at the time the drain was requested, and a live snapshot
+  // of its current usage.
+  optional PoolDrainStatistics drain_statistics = 3;
+  // In-flight replica moves for this drain.
+  repeated DrainConfig replica_moves = 4;
+}
+
+// Why the pool is in its current drain phase.
+enum PhaseReason {
+  // The pool is waiting for a slot to start draining, as the number of concurrent
+  // drains is capped cluster-wide.
+  WaitingForSlot = 0;
+  // The pool is degraded and cannot safely drain without risking data loss.
+  OfflinePool = 1;
+  // The pool has a single replica that cannot be safely evicted without risking data loss.
+  SingleReplicaUnsafeEviction = 2;
+  // The node containing pool is back but pool is cordoned for import, so it cannot be drained.
+  ImportCordoned = 3;
+  // Snapshots are left behind due to user chosen drain policy.
+  SnapshotsRetained = 4;
+}
+
+// Contains the reason for unwinding a spare replica, which is used to determine how
+// to proceed with the drain procedure.
+enum UnwindSpare {
+  // The drain was aborted (`DrainPhase → Aborted`): remove the spare and end the
+  // move, keeping `draining_replica` where it is.
+  Abort = 0;
+  // The pool hosting the still-rebuilding spare has itself entered a drain: remove
+  // the spare and let this move place a fresh one on another eligible pool.
+  Respare = 1;
+}
+
+// Configuration of a single replica move carried out as part of a pool drain.
+message DrainConfig {
+  // When this move first attempted to place its spare.
+  optional google.protobuf.Timestamp placement_started_at = 1;
+  // Id of the volume draining_replica belongs to.
+  string volume = 2;
+  // Id of the pool draining_replica belongs to.
+  string pool = 3;
+  // Id of the draining replica. It's None when it's evacuated successfully.
+  optional string draining_replica = 4;
+  // Id of the spare replica created as part of drain procedure.
+  optional SpareReplica spare_replica = 5;
+  // Why a move's over-replicated spare is being removed.
+  optional UnwindSpare unwind_spare = 6;
+}
+
+// Spare replica reference for this drain.
+message SpareReplica {
+  // Replica Id of the spare, if placed.
+  optional string replica_id = 1;
+}
+
+// The phase of the pool drain state machine.
+enum DrainPhase {
+  // Could not determine the actual drain state.
+  DrainUnknown = 0;
+  // The drain has been admitted and the pool self-cordoned, but no replica has been moved yet:
+  // the pool is waiting for a slot.
+  Queued = 1;
+  // Replicas are actively being evacuated, each one rebuilt onto a spare replica elsewhere
+  // before the copy on this pool is dropped.
+  Draining = 2;
+  // All replicas are evacuated from this pool from the volume's perspective, but replicas are
+  // pending cleanup on the pool itself.
+  AwaitingCleanup = 3;
+  // The user has cancelled the drain and the control plane is undoing the changes it made.
+  // Transient rather than terminal: once cleanup completes the drain record is cleared and the
+  // self-cordon dropped. A new drain is rejected while the pool is in this phase.
+  DrainAborted = 4;
+  // Transitioned from Draining when we can't make any progress on drain operation. Awaiting on
+  // state change or user intervention to continue.
+  PartiallyDrained = 5;
+  // The pool has reached zero allocation: all replicas evacuated and no snapshots left behind,
+  // either because there were none or because they were destroyed under --accept-snapshot-loss.
+  Drained = 6;
+}
+
+/// Gets applied when the pool transitions into Draining state.
+message PoolDrainStatistics {
+  // The initial usage stats of the pool when the pool transitions into Draining state.
+  PoolUsage initial = 1;
+  // A live pool usage snapshot, as the drain's progress counterpart to `initial`.
+  // `None` when the pool has no reported state (node offline / not imported): progress is
+  // then unknown, not zero.
+  optional PoolUsage current = 2;
+}
+
+message PoolUsage {
+  // How many replicas in the pool.
+  uint64 repl_count = 1;
+  // How many replica-snapshots in the pool.
+  uint64 snap_count = 2;
+  // Used bytes from the pool.
+  uint64 used = 3;
+  // Total size of pool replicas.
+  // Unset when the pool state does not report a commitment.
+  optional uint64 committed = 4;
 }
 
 // User specification of a pool.
--- a/control-plane/grpc/src/operations/pool/traits.rs
+++ b/control-plane/grpc/src/operations/pool/traits.rs
@@ -13,19 +13,20 @@ use stor_port::{
     transport_api::{v0::Pools, ReplyError, ResourceKind},
     types::v0::{
         store::pool::{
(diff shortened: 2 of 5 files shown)


Abhinandan-Purkait (dismissed):

control-plane/grpc/proto/v1/pool/pool.proto
```diff
@@ -107,6 +107,114 @@ message Metadata {
...
+
+// The phase of the pool drain state machine.
+// The pool stays self-cordoned for as long as it carries a drain record, terminal phases
+// included; the only exit from any phase is the user cancelling the drain, which moves it
+// to DrainAborted for cleanup.
+enum DrainPhase {
+  // The drain has been admitted and the pool self-cordoned, but no replica has been moved yet:
+  // the pool is waiting for a slot, as the number of concurrent drains is capped cluster-wide.
+  // Queued pools are promoted to Draining in FIFO order, longest-waiting first.
+  Queued = 0;
```
tiagolobocastro: 0 should be unknown, to handle version mismatch

control-plane/grpc/proto/v1/pool/pool.proto
```diff
@@ -107,6 +107,114 @@ message Metadata {
...
+  // before the copy on this pool is dropped.
+  Draining = 1;
+  // The user has cancelled the drain and the control plane is undoing the changes it made.
+  // Transient rather than terminal: once cleanup completes the drain record is cleared and the
+  // self-cordon dropped. A new drain is rejected while the pool is in this phase.
+  DrainAborted = 2;
+  // Every replica has been evacuated, but snapshots deliberately remain on the pool
+  // (drain requested with --ignore-snapshots), so allocation cannot reach zero.
+  // A successful outcome, as opposed to Failed.
+  PartiallyDrained = 3;
```
tiagolobocastro: Wondering now if `PartiallyDrained` is a bit confusing, ie what does it mean, some replicas moved but others haven't?
I know it means snapshots haven't been, but for users may not be completely clear
Any other suggestions?
CC @Abhinandan-Purkait

control-plane/grpc/src/operations/pool/traits.rs
```diff
@@ -578,11 +582,237 @@ impl From<PoolDef> for pool::PoolDefinition {
...
+    }
+}
+
+impl From<PoolDrainRecord> for pool::PoolDrainRecord {
+    fn from(value: PoolDrainRecord) -> Self {
+        Self {
+            phase: pool::DrainPhase::from(value.phase) as i32,
+            drain_statistics: Some(pool::PoolDrainStatistics {
+                initial: value.initial_stats.into_opt(),
+                current: None,
```
tiagolobocastro, line 599: Can't current also be the initial here?

control-plane/stor-port/src/types/v0/store/pool.rs
```diff
@@ -198,8 +201,103 @@ pub struct PoolMetadata {
...
+    /// The move was requested by a pool drain.
+    PoolDrain(DrainConfig),
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+pub struct DrainConfig {
+    /// When this move first attempted to place its spare.
+    pub placement_started_at: Option<SystemTime>,
+    /// Id of the volume draining_replica belongs to.
+    pub volume: VolumeId,
```
tiagolobocastro, line 283: hmm have you considered what happens if you find a replica which has no owner volume?

control-plane/stor-port/src/types/v0/transport/pool.rs
```diff
@@ -621,6 +624,17 @@ impl Pool {
...
     }
+    /// Get a live pool usage statistics, if available.
+    pub fn current_usage(&self) -> Option<PoolUsage> {
+        let state = self.state.as_ref()?;
+        let def = self.config.as_ref().map(|config| &config.definition);
+        Some(PoolUsage {
+            repl_count: def.and_then(|def| def.replica_count)?,
+            snap_count: def.and_then(|def| def.snapshot_count)?,
+            used: state.used,
+            committed: state.committed?,
```
tiagolobocastro: seems like `committed` in `PoolUsage` perhaps should be optional to match `state.committed` ?

control-plane/stor-port/src/types/v0/store/pool.rs
abhilashshetty04: It can happen in scenario when Pool comes back after partial rebuild wait period. We will not apply ReplicaMove for such replicas.They will be garbage collected.

control-plane/grpc/src/operations/pool/traits.rs
abhilashshetty04: Current gets populated later in the conversion as it comes from transport::PoolState. Hence, its None here.

https://github.com/openebs/mayastor-control-plane/pull/1174/changes#diff-7c5a865d688c32b91be1a59ed04a6d4f115bede809bbe03584b1dda4e3d0782eR973

control-plane/grpc/proto/v1/pool/pool.proto
Abhinandan-Purkait: Yeah, I can't think of anything as such.

control-plane/grpc/proto/v1/pool/pool.proto
abhilashshetty04: Its always going to be set to some variant. I have removed default on Queued as that does not make sense as we persist it.

control-plane/grpc/proto/v1/pool/pool.proto
```diff
@@ -107,6 +107,125 @@ message Metadata {
...
+  DegradedPool = 1;
+  // The pool has a single replica that cannot be safely evicted without risking data loss.
+  SingleReplicaUnsafeEviction = 2;
+}
+
+message ReplicaMoveConfig {
+  // Holds a variant with config specific to a type of entity requesting the move.
+  ReplicaMoveRequester requester = 1;
+}
+
```
Abhinandan-Purkait: I honestly feel it should be flattened and the requester is not needed atm, also I can't see a situation where it differs in future?

Abhinandan-Purkait (approved):
Looks fine to me

control-plane/grpc/proto/v1/pool/pool.proto
tiagolobocastro: Think of updates, where an old client version may not be aware of a new phase added on server

control-plane/grpc/proto/v1/pool/pool.proto
abhilashshetty04: For eg: Cluster capacity rebalance where user may want to run the rebalance once they add new nodes or pools so some replicas move off of more occupied pools.

We could have a way of moving replicas matching certain params. These ops will have a different requester and configs.

Abhinandan-Purkait: Hmm, rebalance is indeed a valid use case, okay, maybe fine then

control-plane/grpc/proto/v1/pool/pool.proto
tiagolobocastro: How about `HaltedWithSnapshots` ? 
Or perhaps just `Halted` or even `PartiallyDrained` but with a new field reason enum containing `SnapshotsRetained` ?

abhilashshetty04: DrainedSnapshotsRetained, ReplicasDrained, SnapshotsRetained, DrainedWithSnapshots. any of these?

abhilashshetty04: @niladrih any suggestions?

abhilashshetty04: PartiallyDrained is a terminal state. Should not require a reason enum like Queued does. Let us finalise on better state for this. @tiagolobocastro

niladrih: Discussed with @abhilashshetty04. How about `DrainedSnaphotsRetained`. Why choose this.. it says Drained, which is a terminal state, because `Drained` also a terminal state. Retained is a known word for, data left alone, in k8s or otherwise. 'WithSnapshots' might be ambiguous, `Retained` is not. Snapshots or Snapshot is trivial. Or even `DrainedRetainedSnaphots`.  

This is assuming we're not creating a reason type.

abhilashshetty04: `DrainedSnaphotsRetained` makes sense. But `ReplicasDrained` seems more concise. It has a terminal word `Drained` in it. Also conveys only replicas are drained.

tiagolobocastro: `DrainedSnaphotsRetained` sounds good to me, because it makes it clear snapshots are retained
though I'd still slightly prefer `PartiallyDrained` with `SnapshotsRetained` as a `reason`
Either sounds good

control-plane/grpc/proto/v1/pool/pool.proto
```diff
@@ -107,6 +107,119 @@ message Metadata {
...
+  // before the copy on this pool is dropped.
+  Draining = 2;
+  // All replicas are evacuted from this pool from the volume's perspsective, but replicas are
+  // pending cleanup on the pool itself.
+  AwaitingCleanup = 3;
+  // The user has cancelled the drain and the control plane is undoing the changes it made.
+  // Transient rather than terminal: once cleanup completes the drain record is cleared and the
+  // self-cordon dropped. A new drain is rejected while the pool is in this phase.
+  DrainAborted = 4;
+  // Every replica has been evacuated, but snapshots deliberately remain on the pool
```
tiagolobocastro: This is not true, we'd come here with different reasons

control-plane/stor-port/src/types/v0/store/pool.rs
```diff
@@ -198,8 +201,112 @@ pub struct PoolMetadata {
 
 /// Pool meta information.
 #[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Default)]
-pub struct PoolPersistedMetadata {}
+pub struct PoolPersistedMetadata {
+    /// Populated when drain request is submitted.
+    /// Stores states driving the drain procedure.
+    pub drain_record: Option<PoolDrainRecord>,
```
tiagolobocastro, line 208:
```suggestion
    #[serde(default, skip_serializing_if = "super::is_default")]
    pub drain_record: Option<PoolDrainRecord>,
```

control-plane/stor-port/src/types/v0/store/pool.rs
```diff
@@ -198,8 +201,112 @@ pub struct PoolMetadata {
...
+    /// snapshots are left behind.
+    Drained,
+    /// The user has cancelled the drain and the control plane is undoing the changes the drain
+    /// made, unwinding in-flight moves and their spare replicas.
+    Aborted,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Default)]
+pub struct PoolUsage {
+    /// Num replicas present on pool.
```
tiagolobocastro:
```suggestion
    /// Number of replicas present on pool.
```

control-plane/stor-port/src/types/v0/store/pool.rs
```diff
@@ -198,8 +201,112 @@ pub struct PoolMetadata {
...
+    pub repl_count: u64,
+    /// Num snapshots present on pool.
```
tiagolobocastro:
```suggestion
    /// Number of snapshots present on pool.
```

control-plane/grpc/proto/v1/pool/pool.proto
```diff
@@ -107,6 +107,119 @@ message Metadata {
...
+  // Populated when drain request is submitted.
+  // Stores states driving the drain procedure.
+  optional PoolDrainRecord drain_record = 5;
+}
+
+// Record of an in-progress pool drain, driving the drain procedure and tracking its progress.
+message PoolDrainRecord {
+  // Current phase of the drain state machine.
+  DrainPhase phase = 1;
+  // Why pool is in the Queued or PartiallyDrained phase.
```
```suggestion
  // Why pool is in the aforementioned phase.
```

tiagolobocastro (approved):

PR #1178, opened 2026-08-20
fix(csi-node): self-heal NodePublishVolume when staging mount is missing

susobhandey:
GLCP-379583: after a node power-cycle, a race between the CSI publish flow and force_unstage (or any other actor that tears down the staging mount after NodeStageVolume succeeds) can leave a volume's staging path unmounted while kubelet's ActualStateOfWorld still believes staging succeeded. Per the CSI spec kubelet treats staging as one-shot and never re-invokes NodeStageVolume on its own, so NodePublishVolume then fails forever with "no mount for staging path", and the only recovery today is deleting the pod to force kubelet through a full unstage/restage cycle.

Extract the attach+stage body of node_stage_volume into Node::do_stage_volume(), callable without acquiring its own VolumeOpGuard (the guard is not reentrant - a second acquisition for the same volume returns Aborted instead of blocking, see limiter.rs). In node_publish_volume, detect a missing staging mount before publishing and re-run staging inline (the caller already holds the guard for this volume_id), instead of failing hard.

Block volumes are unaffected: publish_block_volume re-resolves the device from the URI on every call and never depends on the staging mount surviving.

Full validation log in GLCP-379583-fix-proposal.md.

### DIFF
--- a/control-plane/csi-driver/src/bin/node/main.rs
+++ b/control-plane/csi-driver/src/bin/node/main.rs
@@ -38,6 +38,10 @@ mod nodeplugin_grpc;
 mod nodeplugin_nvme;
 #[cfg(target_os = "linux")]
 mod nodeplugin_svc;
+/// Tracks recently-staged volumes so `force_unstage_volume` can avoid
+/// tearing down a freshly (re)established mount. See GLCP-379583.
+#[cfg(target_os = "linux")]
+mod recent_stage;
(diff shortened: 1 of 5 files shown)


tiagolobocastro (commented):
I don't understand how node publish can compete with controller publish's clean of the unstage
This seems like a kubelet bug

PR #1179, opened 2026-08-20
feat(grpc): support TLS across the agent gRPC stack

tiagolobocastro:
The agent gRPC servers and clients now support TLS end to end. Servers accept both TLS and plaintext on a single port via connection sniffing, clients can dial over TLS transparently, and core selects http/https per io-engine based on the capability advertised at registration.

This dual support is only meant for upgrades to ensure smooth transition. After all io-engines are upgraded, only TLS should be enabled.

Single-port TLS + plaintext
- Add `MaybeTlsConnection` and `sniff_and_accept()`: peek the first byte and route a TLS ClientHello (0x16) to the TLS acceptor, otherwise serve plaintext h2c; `incoming`/`incoming_with_server_config` now yield sniffed connections.
- Remove the obsolete `TlsConnection` type.
- Collapse the core listener to a single port (50051); drop the separate registration listener/port (50053) and `DEFAULT_GRPC_REGISTRATION_SERVER_ADDR`, and update deployer portmap/wait addresses accordingly.

Client transport
- `Client::new` now returns `anyhow::Result<Self>` and delegates to `new_with_tls`; auto-TLS is derived from an `https` scheme. The infallible call sites are preserved via `.expect`.
- Add an auto-TLS connector so clients can dial TLS endpoints transparently.

Per-node io-engine scheme selection
- Add `optional bool grpcTls` to `MayastorFeatures` and map it through registration into transport `NodeFeatures.grpc_tls`, persisted on the node spec/state.
- `NodeWrapper::node_grpc_tls()` reads the flag (false for sim nodes, defaults to plaintext when absent) and threads it into `GrpcContext`.
- `GrpcContext::connect_channel()` dials https when set, else http; v0/v1 io-engine clients use it. On core restart, carry the persisted features so the first liveness probe uses the correct scheme.

### DIFF
--- a/control-plane/agents/src/bin/ha/cluster/main.rs
+++ b/control-plane/agents/src/bin/ha/cluster/main.rs
@@ -2,7 +2,7 @@ use clap::Parser;
 use grpc::client::CoreClient;
 use http::Uri;
 use once_cell::sync::OnceCell;
-use std::net::SocketAddr;
+use std::{net::SocketAddr, path::PathBuf};
 use tracing::info;
 use utils::{
     package_description,
@@ -34,6 +34,19 @@ struct Cli {
     /// Core gRPC server URL or address.
     #[clap(long, short, default_value = DEFAULT_GRPC_CLIENT_ADDR)]
     core_grpc: Uri,
+    /// Path to the CA bundle used to verify Core and HA Node Agent gRPC servers and authenticate
+    /// HA Cluster Agent gRPC clients.
+    #[clap(long = "grpc-tls-ca-file")]
+    grpc_tls_ca_file: Option<PathBuf>,
+    /// Path to the TLS certificate used by the HA Cluster Agent gRPC server and client.
+    #[clap(long = "grpc-tls-cert-file", requires = "grpc_tls_key_file")]
+    grpc_tls_cert_file: Option<PathBuf>,
+    /// Path to the TLS private key used by the HA Cluster Agent gRPC server and client.
+    #[clap(long = "grpc-tls-key-file", requires = "grpc_tls_cert_file")]
+    grpc_tls_key_file: Option<PathBuf>,
+    /// Auto-generate an ephemeral self-signed certificate for the HA Cluster Agent gRPC server.
+    #[clap(long = "grpc-auto-tls", conflicts_with_all = ["grpc_tls_ca_file", "grpc_tls_cert_file", "grpc_tls_key_file"])]
+    grpc_auto_tls: bool,
 
     /// Sends opentelemetry spans to the Jaeger endpoint agent.
     #[clap(long, short)]
@@ -68,16 +81,41 @@ impl Cli {
     fn args() -> Self {
         Cli::parse()
     }
+
+    fn grpc_tls(&self) -> anyhow::Result<Option<grpc::tls::TlsConfig>> {
+        if self.grpc_tls_ca_file.is_some() && self.grpc_tls_cert_file.is_none() {
+            anyhow::bail!(
+                "a TLS-enabled HA Cluster Agent requires both gRPC TLS certificate and private key files"
+            );
+        }
+        let tls = grpc::tls::TlsConfig::new(
+            self.grpc_tls_ca_file.clone(),
+            self.grpc_tls_cert_file.clone(),
+            self.grpc_tls_key_file.clone(),
+        )?;
+        Ok(tls.enabled().then_some(tls))
+    }
 }
 
 /// Once cell static variable to store the grpc client and initialize once at startup.
 pub static CORE_CLIENT: OnceCell<CoreClient> = OnceCell::new();
+pub static GRPC_TLS: OnceCell<Option<grpc::tls::TlsConfig>> = OnceCell::new();
+pub static GRPC_AUTO_TLS: OnceCell<bool> = OnceCell::new();
 
 /// Get Core gRPC Client
 pub(crate) fn core_grpc<'a>() -> &'a CoreClient {
     CORE_CLIENT
         .get()
         .expect("gRPC Core Client should have been initialised")
+}
+
+pub(crate) fn grpc_tls() -> Option<grpc::tls::TlsConfig> {
+    GRPC_TLS.get().cloned().flatten()
+}
+
+/// Whether the HA Node Agent gRPC servers are dialled over TLS (file-backed or auto-TLS).
+pub(crate) fn grpc_tls_enabled() -> bool {
+    grpc_tls().is_some() || GRPC_AUTO_TLS.get().copied().unwrap_or(false)
 }
 
 fn initialize_tracing(args: &Cli) {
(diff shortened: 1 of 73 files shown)


control-plane/agents/src/bin/ha/cluster/main.rs
```diff
@@ -93,16 +125,28 @@ fn initialize_tracing(args: &Cli) {
 
 #[tokio::main]
 async fn main() -> anyhow::Result<()> {
+    utils::init_rustls_crypto_provider();
```
tiagolobocastro, line 134: we shouldn't need these but the features are enabling multiple tls backends
we can either try to fix that here or leave it for now, and remove it once we move out of ring and switch fully to rustls with aws-lc-rs

abhilashshetty04 (approved):

control-plane/grpc/src/tls.rs
```diff
@@ -0,0 +1,764 @@
...
+/// The transport is chosen by peeking the first byte of the stream: a TLS `ClientHello` record
+/// always starts with `0x16`, whereas cleartext HTTP/2 (h2c) starts with the connection preface
+/// `PRI * HTTP/2.0` (i.e. `0x50`). This lets a single listener accept legacy plaintext clients and
+/// TLS clients at the same time, which is what allows a non-TLS to TLS rolling upgrade.
+pub enum MaybeTlsConnection {
+    /// A plaintext connection.
+    Plain(Pin<Box<TcpStream>>),
+    /// A TLS connection.
+    Tls(Pin<Box<TlsStream<TcpStream>>>),
+}
```
StealthBadger747, line 355: I’ve been looking at the TLS/plaintext coexistence design. I understand why accepting both on one port helps rolling upgrades, but I’m wondering what the intended post-upgrade state is and whether plaintext can eventually be rejected.

I've done a bit of research on what other applications have chosen to do in this scenario and there are a couple options.

1. **Istio:** They have chosen to expose an explicit permissive-to-strict policy transition https://istio.io/latest/docs/reference/config/security/peer_authentication (`DISABLE`, `PERMISSIVE`, and `STRICT` policies)
  a. PostgreSQL seems to also do a very similiar policy with `host/hostssl/hostnossl` policy values.
2. **Redis:** separate TCP and TLS listeners, with ordinary TCP listeners disabled for a TLS-only deployment. https://redis.io/docs/latest/operate/oss_and_stack/management/security/encryption

Compliance and security frameworks will want to see a mode where TLS is fully enforced.

StealthBadger747: Re-reading your PR body I see you said

> This dual support is only meant for upgrades to ensure smooth transition. After all io-engines are upgraded, only TLS should be enabled.

But wouldn't it make sense to deal with this now?

control-plane/grpc/src/tls.rs
```diff
@@ -0,0 +1,764 @@
...
+    socket: SocketAddr,
+    tls: TlsConfig,
+) -> anyhow::Result<impl futures::Stream<Item = Result<MaybeTlsConnection, io::Error>>> {
+    let listener = tokio::net::TcpListener::bind(socket).await?;
+    let tls = ReloadableServerTls::new(tls)?;
+
+    Ok(TcpListenerStream::new(listener).then(move |connection| {
+        let acceptor = tls.acceptor();
+        async move { sniff_and_accept(connection?, acceptor).await }
+    }))
```
StealthBadger747: It seems like `.then()` waits for each returned future before polling the next connection. This can lead to blocking every subsequent client on this listener. I made a toy connection that opened a socket without sending anything, and confirmed that this is an issue.

The same pattern appears in incoming_with_server_config.

control-plane/grpc/src/tls.rs
```diff
@@ -0,0 +1,764 @@
...
+        builder
+            .with_client_cert_verifier(WebPkiClientVerifier::builder(Arc::new(roots)).build()?)
+            .with_single_cert(certificates, private_key)?
+    } else {
+        builder
+            .with_no_client_auth()
+            .with_single_cert(certificates, private_key)?
+    };
+
+    Ok(config)
```
StealthBadger747, line 551: It seems like this manually constructed rustls configuration never sets `alpn_protocols` to `h2`.

The same omission appears in `auto_client_config` and `auto_server_config`.

StealthBadger747 (commented):
Found a couple of issues while familiarizing myself with the core code/libraries used.

control-plane/grpc/src/tls.rs
tiagolobocastro: > but I’m wondering what the intended post-upgrade state is and whether plaintext can eventually be rejected.

Once all data-plane nodes are upgraded then only tls would be enabled
I haven't covered this in this PR, that I'll cover on another as this one is already rather large

> Compliance and security frameworks will want to see a mode where TLS is fully enforced.

Yeah a policy here sounds good, especially for new installs where this mode would not be required to begin with

tiagolobocastro: > But wouldn't it make sense to deal with this now?

Sorry didn't get what you mean here?

control-plane/grpc/src/tls.rs
tiagolobocastro: hmm good catch
Looks like can use `futures::StreamExt::buffer_unordered` here

StealthBadger747: Thanks. And yeah a `futures::StreamExt::buffer_unordered` makes sense. Perhaps a timeout as well?

control-plane/grpc/src/tls.rs
StealthBadger747: > Once all data-plane nodes are upgraded then only tls would be enabled
I haven't covered this in this PR, that I'll cover on another as this one is already rather large

Sounds good.

> Sorry didn't get what you mean here?

Sorry I could've been more verbose. I meant that it might make sense to add the server policy (e.g. the `DISABLE/PERMISSIVE/STRICT` policies) now, but a follow up is perfectly fine as well.

control-plane/grpc/src/tls.rs
tiagolobocastro: I think tonic has an `assume_http2`, so it allows it to be missing
Though we could probably add it here anyhow

control-plane/grpc/src/tls.rs
tiagolobocastro: Yes indeed, I'll follow up with PR on this
Thanks for the reviews @StealthBadger747 !

control-plane/grpc/src/tls.rs
tiagolobocastro: hmm today I think we only have timeouts on the client side, though I guess makes sense to have server-side timeout as well?

tiagolobocastro: looks like there's no timeout server-side on connect even without my changes, so will take this work in another PR as well

control-plane/grpc/src/tls.rs
tiagolobocastro: Done

PR #1186, opened 2026-08-28
build: switch to rustls for coming fips feature

tiagolobocastro:
build: switch to rustls for coming fips feature
    
    Drops hyper-client-tls feature and all related tls feats.
    This was used to build the windows kubectl-client but the rls build
    now works fine for windows targets, so let's drop it to avoid depending
    on the openssl, and it's also good to keep things consistent across targets.
    
    Also drops ring and favours aws-lc-rs. This will eventually make it easier
    to use fips algorithms.
    Also removes openssl dependency for windows plugins.
    
---

    build: upgrade to rust 1.97.1
    
    This will be required to update kube to 0.99.0

### DIFF
--- a/openapi/templates/default/actix/mod.mustache
+++ b/openapi/templates/default/actix/mod.mustache
@@ -1,4 +1,2 @@
-#[cfg(feature = "actix-client")]
-pub mod client;
 #[cfg(feature = "actix-server")]
 pub mod server;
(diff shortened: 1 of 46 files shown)


Abhinandan-Purkait (approved):

openapi/templates/default/actix/mod.mustache
```diff
@@ -1,4 +1,2 @@
-#[cfg(feature = "actix-client")]
-pub mod client;
 #[cfg(feature = "actix-server")]
 pub mod server;
```
Abhinandan-Purkait, line 2: Blank line?

abhilashshetty04 (approved):

PR #1196, opened 2026-09-09
feat(core): cap concurrent offline rebuilds

yugchaudhari:
Offline rebuilds and published-degraded rebuilds share one budget — both go through `registry.rebuild_allowed()` — so a batch of offline rebuilds can occupy every slot `max_rebuilds` allows and leave nothing for volumes that are actually serving I/O. This is the starvation @Abhinandan-Purkait raised on #1108.

## What this adds

`--max-offline-rebuilds` (env `MAX_OFFLINE_REBUILDS`), an *additional* ceiling rather than a separate allowance: offline rebuilds still count against `max_rebuilds`, so setting this lower reserves the remainder for published volumes. Unset means only `max_rebuilds` applies, so nothing changes for existing deployments.

The reconciler counts the offline-rebuild targets currently standing, works out the headroom once per poll, and spends it as it creates new ones, so a single poll cannot overshoot.

## Two deliberate choices

**Counting standing targets, not rebuild jobs.** A target exists briefly before its rebuild starts and again after it finishes but before teardown, so the count runs slightly high and errs toward admitting fewer. An exact count is reachable (`Nexus.rebuilds` is per-nexus on v1, `NexusSpec.owner` gives the volume) but on v0 that field is a node-wide total, so it would need handling. Discussed on the OEP.

**No preemption.** A running offline rebuild keeps its slot even if a published volume degrades a moment later. Cancelling a rebuild that has already copied data trades one problem for another, and @tiagolobocastro agreed on the OEP that this belongs with a future rebuild-request queue, which would also allow priority ordering by volume size rather than just the offline/published split.

## Testing

New `offline_rebuild_respects_max_offline_rebuilds` int-test runs with the cap at 0: the volume is degraded and otherwise immediately eligible, so a temp nexus would appear if the cap were not consulted. It reuses the scale-up path so no node has to be stopped and no grace period waited out.

Attribution comes from pairing with the existing `offline_rebuild_scale_up_bypasses_grace`, which shows the identical scenario standing a target up well within 8s when uncapped — so the deferral here is the cap rather than the grace period or a viability check.

Verified locally: `cargo check --all --all-targets`, `cargo clippy --all --all-targets -- -D warnings`, `cargo fmt --all --check` all clean; the new int-test passes.


control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
```diff
@@ -51,13 +52,41 @@ impl TaskPoller for OfflineRebuildReconciler {
...
+    // Must read through the lock: `immutable_ref` is a snapshot of the spec as first
+    // loaded, so it would miss targets this reconciler stood up since then.
+    let in_flight = volumes
+        .iter()
+        .filter(|volume| {
+            let volume = volume.lock();
+            volume.is_offline_rebuild_target()
+                && volume.target().is_some_and(|t| t.protocol().is_none())
+        })
+        .count();
```
Abhinandan-Purkait: You seem to be considering max number of offline nexuses rather than rebuilds, isn't that different from what max rebuilds mean?

Abhinandan-Purkait: cc @abhilashshetty04  @tiagolobocastro  ?

control-plane/agents/src/bin/core/main.rs
```diff
@@ -59,6 +59,13 @@ pub(crate) struct CliArgs {
     #[clap(long, env = "OFFLINE_REBUILD_GRACE_PERIOD", default_value = "10m")]
     pub(crate) offline_rebuild_grace_period: humantime::Duration,
 
+    /// The maximum number of offline rebuilds permitted at any given time.
+    /// These also count towards `max_rebuilds`, so setting this lower than
+    /// `max_rebuilds` reserves the remainder for volumes which are published
+    /// and serving I/O. If `None` only `max_rebuilds` applies.
+    #[clap(long, env = "MAX_OFFLINE_REBUILDS")]
+    pub(crate) max_offline_rebuilds: Option<NumRebuilds>,
```
Abhinandan-Purkait, line 68: What happens if this value is more than he max_rebuild? I did not see any check for it?

control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
yugchaudhari: > You seem to be considering max number of offline nexuses rather than rebuilds, isn't that different from what max rebuilds mean?

Good catch, you're right and I've changed it. `max_rebuilds` counts jobs, I was counting nexuses, so a volume restoring two replicas through one temp nexus scored 1 against the cap where `max_rebuilds` would score it 2.

It now sums `nexus.rebuilds` across the standing offline-rebuild targets, with a floor of one per target. The floor matters: a target whose rebuild hasn't started yet reports `rebuilds: 0`, and taking that literally would admit another volume and let both exceed the cap once they got going. Over-counting only delays a rebuild, under-counting breaks the limit the knob exists for. Same floor covers a nexus we can't read.

One thing I haven't covered: the int-test uses a cap of 0, where nexuses and jobs are indistinguishable, so it doesn't exercise the multi-job case you found. Testing that needs a volume losing two replicas at once. Happy to add it if you'd like it pinned down.

control-plane/agents/src/bin/core/main.rs
yugchaudhari: No check, and nothing breaks. `initiate_offline_rebuild` calls `rebuild_allowed()` before consulting the offline budget, so both have to pass and the effective limit is `min(max_rebuilds, max_offline_rebuilds)`. The larger value is just inert.

It is a confusing thing to configure though, so I've added a startup warning when the offline cap isn't below `max_rebuilds`, since in that case it reserves nothing and silently isn't doing what was asked. I went with a warning rather than clamping or rejecting: clamping overrides what the operator wrote without telling them, and rejecting turns a harmless misconfiguration into a failed start. Say if you'd rather it were stricter.

control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
Abhinandan-Purkait: Yeah, would be good to have a test to cover these.

yugchaudhari: > Yeah, would be good to have a test to cover these.

Added. The counting rule is now a small function with unit tests on it: a standing target costs at least one slot when its rebuild hasn't started or its nexus can't be read, and its real job count beyond that, so `Some(2)` costs 2.

I tried to do this as an int-test first and backed off. The discriminating case is a volume whose temp nexus carries two jobs blocking a second volume under a cap of 2, but that only discriminates for as long as the rebuild runs, and these are 10MB tmpfs volumes. The assertion that looks robust, "never two temp nexuses at once", is actually wrong even for correct code: once the first volume's rebuild finishes but before teardown its nexus reports zero jobs, floors to one, the budget frees and the second volume legitimately gets a nexus. So that test would fail intermittently against a working implementation.

The unit tests pin the rule deterministically, but they don't prove the end-to-end path, so the gap is narrower rather than gone. If you want the end-to-end case covered too I'll do it, just flagging that it'd likely need a slow-rebuild knob or a much larger volume to be reliable.

control-plane/agents/src/bin/core/tests/volume/offline_rebuild.rs
```diff
@@ -161,6 +161,107 @@ async fn offline_rebuild_scale_up_bypasses_grace() {
...
+/// not consulted. This is deliberately the same scale-up scenario as
+/// [`offline_rebuild_scale_up_bypasses_grace`], which shows the target arriving
+/// well inside `START_WINDOW_SECS` when no cap is set; the pair is what
+/// attributes the deferral here to the cap rather than to the grace period or a
+/// viability check.
+#[tokio::test]
+async fn offline_rebuild_respects_max_offline_rebuilds() {
+    // Comfortably longer than the window the uncapped test needs to stand a
+    // target up, so "nothing appeared" means blocked rather than merely slow.
+    const OBSERVE_SECS: u64 = 12;
```
tiagolobocastro: We only need to sleep for GRACE_PERIOD_SECS + 2x the reconcile period duration?

control-plane/agents/src/bin/core/tests/volume/offline_rebuild.rs
```diff
@@ -161,6 +161,107 @@ async fn offline_rebuild_scale_up_bypasses_grace() {
...
+
+    let reconcile = Duration::from_millis(RECONCILE_PERIOD_MS);
+    let cluster = ClusterBuilder::builder()
+        .with_rest(true)
+        .with_io_engines(3)
```
tiagolobocastro: Can we use 2 engines only?

control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
```diff
@@ -51,13 +52,70 @@ impl TaskPoller for OfflineRebuildReconciler {
...
+
+    let mut in_flight: NumRebuilds = 0;
+    for nexus_id in targets {
+        let reported = context
+            .registry()
+            .nexus(&nexus_id)
+            .await
+            .map(|nexus| nexus.rebuilds)
+            .ok();
+        in_flight = in_flight.saturating_add(target_slots(reported));
```
tiagolobocastro, line 113: Do we currently allow multiple concurrent rebuilds for offline rebuild?

control-plane/agents/src/bin/core/tests/volume/offline_rebuild.rs
yugchaudhari: Yes, and less than that even. This test scales up, which bypasses the grace period, so the reconciler would act within a poll or two if it weren't capped. Changed to `GRACE_PERIOD_SECS + 2 * reconcile`, keeping the grace period as headroom rather than because it applies.

control-plane/agents/src/bin/core/tests/volume/offline_rebuild.rs
yugchaudhari: Done. It needed the scenario adjusting rather than just the count: it was scaling 2→3 replicas, which wants three pools. It's now a 1-replica volume scaled to 2, which exercises the same thing (replica absent from `NexusInfo` → Degraded → grace bypassed → cap blocks it).

control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
yugchaudhari: Yes. `attach_replicas_to_nexus` keeps attaching candidates until `nexus_children >= volume_children`, and each `attach_replica` does an `add_child` with `auto_rebuild`, so a volume that has lost two replicas gets both attached in the same hot-spare pass and the one temp nexus carries two rebuild jobs. They're still each gated by `rebuild_allowed()`, so `max_rebuilds` throttles them, but nothing restricts an offline-rebuild nexus to one.

That's what the job-vs-nexus counting is for: if it were one rebuild per nexus the distinction wouldn't matter.

tiagolobocastro (approved):

control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
tiagolobocastro: ok just wanted to check, because otherwise we didn't have to check the rebuilds per nexus

Abhinandan-Purkait (approved):

PR #1197, opened 2026-09-09
add abort drain api implementation

abhilashshetty04:


Abhinandan-Purkait (approved):

control-plane/agents/src/bin/core/controller/resources/operations.rs
```diff
@@ -58,6 +58,8 @@ pub(crate) trait ResourceDrain {
     ) -> Result<Self::Output, SvcError>;
     /// Mark the resource as drained.
     async fn set_drained(&mut self, registry: &Registry) -> Result<Self::Output, SvcError>;
+    /// Abort the drain in progress.
+    async fn abort_drain(&mut self, rgistry: &Registry) -> Result<Self::Output, SvcError>;
```
tiagolobocastro, line 62:
```suggestion
    async fn abort_drain(&mut self, registry: &Registry) -> Result<Self::Output, SvcError>;
```

control-plane/grpc/proto/v1/pool/pool.proto
```diff
@@ -649,4 +664,5 @@ service PoolGrpc {
   // Clear error counters and alerts
   rpc ClearErrors (ClearErrorsRequest) returns (ClearErrorsReply) {}
   rpc DrainPool (DrainPoolRequest) returns (DrainPoolReply) {}
+  rpc AbortPoolDrain(AbortPoolDrainRequest) returns (AbortPoolDrainReply) {}
```
tiagolobocastro, line 667: `AbortDrainPool`  to match with `DrainPool`
```suggestion
  rpc AbortDrainPool(DrainPoolAbortRequest) returns (DrainPoolAbortReply) {}
```

tiagolobocastro: Oh Actually, this should be done in uncordon ?

control-plane/rest/openapi-specs/v0_api_spec.yaml
```diff
@@ -1557,6 +1557,35 @@ paths:
           $ref: '#/components/responses/ServerError'
       security:
         - JWT: [ ]
+  '/nodes/{node_id}/pools/{pool_id}/abort_drain':
```
tiagolobocastro: this doesn't make sense in REST, REST is resource based, it'd have to be del against a drain
an abort_drain being deleted is aborting the abort almost

tiagolobocastro (commented):
I thought we had discussed and it'd be the uncordon aborting the drain or am I misremembering
CC @Abhinandan-Purkait

control-plane/rest/openapi-specs/v0_api_spec.yaml
abhilashshetty04: Makes sense.

tiagolobocastro:
> > I thought we had discussed and it'd be the uncordon aborting the drain or am I misremembering CC @Abhinandan-Purkait
> 
> Yes, that's true, uncordon with --drain, but I also get the point, that it might not be possible to extend the UncordonBody in a proper way. A new endpoint does makes sense. But yeah, we can discuss if we have a better way of fitting that in.

I we use uncordon --drain, then I think we should just reuse the uncordon endpoint, keep the api consistent
Also this is what the nodes api does IIRC

Now if we're saying that uncordon starts to look odd here, then we should revisit the original request to make `uncordon --drain`
IE, if we want to break out of having api similar to the nodes, then we could consider changing it at the top instead?

abhilashshetty04:
> > > I thought we had discussed and it'd be the uncordon aborting the drain or am I misremembering CC @Abhinandan-Purkait
> > 
> > 
> > Yes, that's true, uncordon with --drain, but I also get the point, that it might not be possible to extend the UncordonBody in a proper way. A new endpoint does makes sense. But yeah, we can discuss if we have a better way of fitting that in.
> 
> I we use uncordon --drain, then I think we should just reuse the uncordon endpoint, keep the api consistent Also this is what the nodes api does IIRC
> 
> Now if we're saying that uncordon starts to look odd here, then we should revisit the original request to make `uncordon --drain` IE, if we want to break out of having api similar to the nodes, then we could consider changing it at the top instead?

Makes sense, we decided to implement this via uncordon because k8s kubectl does not have abort subcommand. It's fine to have abort verb since it is our use case. @tiagolobocastro 

for mayastor node drain also we can add abort when pool drain is integrated with node drain.

abhilashshetty04:
@tiagolobocastro , drain is cancelled using `uncordon --drain` end to end.

niladrih (approved):

tiagolobocastro (approved):

PR #1198, opened 2026-09-14
feat(fips): add --enable-fips to our components

tiagolobocastro:
feat(fips): add --enable-fips to our components
    
    The flag puts the TLS our components do on a FIPS validated crypto module and
    fails startup if the module this was built against turns out not to be one. It
    is off by default, so nothing changes for a deployment which does not ask for
    it.
    
    It lives in utils as CryptoArgs, for binaries to flatten into their own args,
    so that all eight spell the flag and its help the same way; the three which
    build their Command by hand reuse it through augment_args rather than restating
    the argument. Further crypto options belong in there rather than in a struct of
    their own.
    
    The deployer runs the components in FIPS mode by default, --no-fips opts out,
    and it passes this on as ENABLE_FIPS rather than an argument, the way a chart
    would set it.
    
---

    feat: add fips helper to utils crate
    
    Pulls the latest dependencies with the fips enablement crate.
    Adds helpers for enabling fips mode and verifying the global crypto
    provider and both default client and server configs.
    
    init_rustls_crypto_provider() goes: ring is no longer compiled in, so
    aws_lc_rs is the only provider left and rustls installs it from the crate
    features by itself - which the pins below make the FIPS one. utils no longer
    needs rustls at all.
    
    The aws-lc-rs and rustls entries in workspace.dependencies are what the fips
    crate's `workspace = true` resolves against when it is built as part of this
    workspace, so they have to match the ones in utils/dependencies.
    
    aws-lc-fips-sys builds the validated module from source, so the dev shell and
    the nix build gain cmake, go and perl.


Abhinandan-Purkait:
> Not sure what you mean here, why would a user be running control-plane binaries/containers directly and not via helm?
> 
I did not say they will be running directly, what I meant is if they are running a specific version after upgrade which is fips compliant, i.e build with the fips packages, then it should work, why would it be missing from the images after upgrade?

Abhinandan-Purkait:
Its all helm chart based, every helm chart is tied to an image version, so how will it start breaking?

tiagolobocastro:
> > Not sure what you mean here, why would a user be running control-plane binaries/containers directly and not via helm?
> 
> I did not say they will be running directly, what I meant is if they are running a specific version after upgrade which is fips compliant, i.e build with the fips packages, then it should work, why would it be missing from the images after upgrade?

Again it depends on what specific options may be enable-eable; enabling fips may mean certain features cannot be used (ie diskpool encryption)

Abhinandan-Purkait:
How will things pan out if someone wants to use fips but had previously used encryption, the newer dataplanes would stop supporting encryption all of a sudden, should we have some checks in upgrade?

tiagolobocastro:
> How will things pan out if someone wants to use fips but had previously used encryption, the newer dataplanes would stop supporting encryption all of a sudden, should we have some checks in upgrade?

this is the whole point of the enable-fips option... if you use diskpool-encryption then you can't enable it!

Abhinandan-Purkait:
> > How will things pan out if someone wants to use fips but had previously used encryption, the newer dataplanes would stop supporting encryption all of a sudden, should we have some checks in upgrade?
> 
> this is the whole point of the enable-fips option... if you use diskpool-encryption then you can't enable it!

Right. LGTM

Abhinandan-Purkait (approved):

abhilashshetty04 (approved):

PR #1199, opened 2026-09-17
feat(core): keep the target alive when unpublished mid-rebuild

yugchaudhari:
Unpublishing a volume destroys its target. If a rebuild was running at the time, that work is thrown away — and for an unpublished volume the offline-rebuild reconciler then has to redo it from scratch once the grace period expires.

This drops the share instead and hands the target to the offline-rebuild reconciler, which already knows how to let a rebuild finish and tear the target down afterwards. The volume ends up unpublished either way; the difference is that the rebuild survives.

This is point 6 from openebs/mayastor#1977, carried into the OEP's Concurrent Operations section, and the shape @tiagolobocastro picked on openebs/openebs#4208.

## Scope, worth being explicit about

It demotes on **any** unpublish that lands while a rebuild is in progress, not only targets that began as offline rebuilds. Promotion overwrites the target mode, so by unpublish time there is nothing left to tell the two apart — that limitation is why option (a) was chosen over tracking provenance.

So this does change behaviour for volumes that never involved offline rebuild: unpublish one mid-rebuild with the feature enabled and the nexus now lingers, unshared, until the rebuild completes, rather than being destroyed. That seemed the better trade, since the alternative is discarding partial work that would be redone later anyway, but it is a wider blast radius than the OEP wording alone implies.

It is gated on the offline-rebuild reconciler being enabled. Without that guard nothing would own the demoted target and it would be left running indefinitely.

## Implementation note

The demoted target has to stay `active` in the spec. `target_cfg()` filters on that flag, so deactivating it — which is what a normal unpublish does — would hide the target from the reconciler and leave the nexus with nothing to tear it down. Both paths are unit tested for exactly that.

`UnpublishOperation` gains a `#[serde(default)]` bool. It is a struct rather than a tuple variant, so pre-existing persisted operations still load.

## Testing

Two unit tests on the commit path: a plain unpublish still retires the target, and a demoting one leaves it visible, unshared, and owned by the offline-rebuild reconciler.

No int-test: the discriminating moment is an unpublish landing while a rebuild is genuinely in flight, and on 10MiB tmpfs volumes that window is too short to hit reliably. The failure I was most concerned about — the target being hidden from the reconciler and the nexus orphaned — is deterministic and is what the unit tests pin down. Happy to add an end-to-end one if you would rather have it.

Verified locally: `cargo check --all --all-targets`, `cargo clippy --all --all-targets -- -D warnings`, `cargo fmt --all --check` all clean.


Abhinandan-Purkait (approved):

control-plane/agents/src/bin/core/volume/operations.rs
```diff
@@ -386,6 +399,12 @@ impl ResourcePublishing for OperationGuardArc<VolumeSpec> {
 
         let result = match specs.nexus_opt(volume_target.nexus()).await? {
             None => Ok(()),
+            // Demote: drop the share but leave the target running for the rebuild.
+            Some(mut nexus) if last_node && demote => {
+                // `demote` was derived from `state.target`, so it is present here.
+                let target = state.target.as_ref().expect("demote implies a target");
```
tiagolobocastro: hmm let's change demote to be a Option<Target>, then we don't need an expect here, with Some(x) signalling demote is true

control-plane/stor-port/src/types/v0/store/volume.rs
```diff
@@ -1177,11 +1189,29 @@ impl PublishOperation {
 #[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
 pub struct UnpublishOperation {
     host_acls: Vec<InitiatorAC>,
+    /// Keep the target alive, unshared, so an in-progress rebuild can finish.
+    /// Defaulted so entries persisted before this existed still load.
```
```suggestion
```

tiagolobocastro (approved):

control-plane/agents/src/bin/core/volume/operations.rs
yugchaudhari: Done.

control-plane/stor-port/src/types/v0/store/volume.rs
yugchaudhari: Dropped.

PR #1200, opened 2026-09-17
feat(core): let an operator trigger an offline rebuild

yugchaudhari:
A degraded unpublished volume waits out `offline_rebuild_grace_period` before the reconciler rebuilds it. That window exists to absorb a node which is only briefly away, but an operator who knows it is not coming back — planned decommission, confirmed hardware failure — currently has no way to say so and has to wait it out.

```
kubectl mayastor rebuild volume <uuid>
```

## What it does, and what it deliberately does not

The request skips the grace period **and nothing else**. Viability (is there a healthy source, is there a pool with room), the system-wide `max_rebuilds` limit, and `max_offline_rebuilds` all still apply, so a rebuild that could not succeed is still not started. The marker is set on the volume and the reconciler acts on it at its next poll.

An earlier sketch of this item, in the iteration notes, had the endpoint call `publish_with_mode(OfflineRebuild)` directly. I did not do that: it would bypass all four checks above and let an operator stand up a temp nexus with no healthy replica to copy from — exactly the create/destroy churn the pre-flight viability check in openebs/mayastor-control-plane#1131 was added to prevent. Going through the reconciler instead reuses that logic rather than duplicating or skipping it, and is less code.

## Marker lifetime

Runtime-only, so it does not survive a core-agent restart. That is deliberate: it is a one-shot intent, cleared once the rebuild is under way, and re-issuing it is cheaper than persisting transient intent. It is also cleared if the volume recovers on its own, so a stale request cannot silently skip the grace period on some later, unrelated degradation.

It is *kept* when a rebuild is torn down because viability was lost. Nothing about losing viability says the operator stopped wanting the rebuild, and it still only skips the wait. That one is commented in place so it does not read as an oversight.

## Layers

`clear-errors` on pools was the closest precedent for an imperative verb: plugin subcommand, openapi spec (the Rust client is generated from it, and the generated crate is gitignored so nothing generated lands in this diff), REST handler, gRPC message and service method, core agent.

## Testing

`offline_rebuild_can_be_requested_by_operator`: a 60s grace period with the volume degraded by stopping a node, so the scale-up bypass cannot account for the result. The volume is first observed sitting idle, then the request is issued and the temp nexus has to appear within 15s. The "nothing, then request, then something" sequence is what attributes it to the request rather than to a timer expiring.

Verified locally: `cargo check --all --all-targets`, `cargo clippy --all --all-targets -- -D warnings`, `cargo fmt --all --check` all clean, and the new int-test passes.


control-plane/agents/src/bin/core/volume/service.rs
```diff
@@ -175,6 +175,18 @@ impl VolumeOperations for Service {
         Ok(volume)
     }
 
+    async fn request_offline_rebuild(
```
Abhinandan-Purkait: nit, request_offline_rebuild --> trigger_offline_rebuild or maybe just offline_rebuild_volume

Abhinandan-Purkait (approved):

control-plane/agents/src/bin/core/volume/service.rs
yugchaudhari: Renamed to `trigger_offline_rebuild`, including the transport type, the gRPC method and the proto messages so the whole path matches.

abhilashshetty04:
@yugchaudhari , Is this really needed? We don't expose this command even for the published volume. We let system handle the rebuild trigger.

Also, after pool drain gets in, user will do drain on the pool before planned maintenance which will use the offline rebuild to move replicas part of unpublished volume without waiting for `offline_rebuild_grace_period`.

You added the support for it yourself.

https://github.com/openebs/mayastor-control-plane/pull/1161

control-plane/rest/openapi-specs/v0_api_spec.yaml
```diff
@@ -1903,6 +1903,35 @@ paths:
           $ref: '#/components/responses/ServerError'
       security:
         - JWT: [ ]
+  '/volumes/{volume_id}/rebuild':
```
abhilashshetty04, line 1906: @yugchaudhari , Is this really needed? We don't expose this command even for the published volume. We let system handle the rebuild trigger.

Also, after pool drain gets in, user will do drain on the pool before planned maintenance which will use the offline rebuild to move replicas part of unpublished volume without waiting for offline_rebuild_grace_period.

You added the support for it yourself.

https://github.com/openebs/mayastor-control-plane/pull/1161

yugchaudhari:
> @yugchaudhari , Is this really needed? We don't expose this command even for the published volume. We let system handle the rebuild trigger.
> 
> Also, after pool drain gets in, user will do drain on the pool before planned maintenance which will use the offline rebuild to move replicas part of unpublished volume without waiting for `offline_rebuild_grace_period`.
> 
> You added the support for it yourself.
> 
> #1161

Fair question. Looking into it, I found something that argues your way harder than the published-volume point does. Published volumes have a wait of their own: `--faulted-child-wait-period`, 10 minutes by default. During it a faulted child is left alone, in the hope it comes back and we can do a partial rebuild instead of a full one. There is no command to cut that short either. So the precedent is not just "no trigger for published rebuilds", it is "no trigger for any of our waits", and this would be the first exception.

One thing worth checking when pool drain lands, though. I am not sure the openebs/mayastor-control-plane#1161 bypass fires for a relocation. The gate ended up as `num_replicas > NexusInfo.children.len()`, which is the scale-up signature. A relocation that keeps the replica count the same would not match it. The PR description did say it covered drain relocations, but that gate went in later to stop the reconciler fighting the replica count reconciler. If it does not fire, drained pools would sit through the grace period, which is the opposite of what we want.

On whether this PR is needed: @tiagolobocastro asked for it before the release, so I would rather you two decide on the answer. Happy to close it if the answer is no. It does not block openebs/mayastor-control-plane#1199 or the docs, and the docs section is a few lines to drop.

abhilashshetty04:
> > @yugchaudhari , Is this really needed? We don't expose this command even for the published volume. We let system handle the rebuild trigger.
> > Also, after pool drain gets in, user will do drain on the pool before planned maintenance which will use the offline rebuild to move replicas part of unpublished volume without waiting for `offline_rebuild_grace_period`.
> > You added the support for it yourself.
> > #1161
> 
> Fair question. Looking into it, I found something that argues your way harder than the published-volume point does. Published volumes have a wait of their own: `--faulted-child-wait-period`, 10 minutes by default. During it a faulted child is left alone, in the hope it comes back and we can do a partial rebuild instead of a full one. There is no command to cut that short either. So the precedent is not just "no trigger for published rebuilds", it is "no trigger for any of our waits", and this would be the first exception.
> 
> One thing worth checking when pool drain lands, though. I am not sure the #1161 bypass fires for a relocation. The gate ended up as `num_replicas > NexusInfo.children.len()`, which is the scale-up signature. A relocation that keeps the replica count the same would not match it. The PR description did say it covered drain relocations, but that gate went in later to stop the reconciler fighting the replica count reconciler. If it does not fire, drained pools would sit through the grace period, which is the opposite of what we want.
> 
> On whether this PR is needed: @tiagolobocastro asked for it before the release, so I would rather you two decide on the answer. Happy to close it if the answer is no. It does not block #1199 or the docs, and the docs section is a few lines to drop.

Should be fine to have this. Lets take it in.

For drain, We will have a mechanism to over-replicate which will do a scale up on volume. 

https://github.com/openebs/openebs/blob/develop/designs/replicated-pv/mayastor/pool-drain.md

If we see an additional replica on volume which is degraded then it should work right? The new replica will not be in previous NexusInfo. I think that was your fix.

control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
```diff
@@ -182,7 +186,23 @@ async fn offline_rebuild_reconcile(
     let has_in_place_target =
         has_scale_up_rebuild_target(&volume, &volume_state.replica_topology, context).await?;
 
-    if !has_in_place_target {
+    // An operator who knows the node is not coming back should not have to wait out
+    // a window that exists to absorb the case where it does. The request skips the
```
abhilashshetty04: We don't need this documentation on the code? Can be added on command help perhaps.

control-plane/stor-port/src/types/v0/transport/mod.rs
```diff
@@ -158,6 +158,8 @@ pub enum MessageIdVs {
     SetVolumeReplica,
     /// Set volume property.
     SetVolumeProperty,
+    /// Request an offline rebuild, skipping the grace period.
+    TriggerOfflineRebuild,
```
abhilashshetty04:
```suggestion
    TriggerRebuild,
```

So we can reuse the same op for online rebuild if we want to add there..

yugchaudhari:
> Should be fine to have this. Let's take it in.
> For drain, We will have a mechanism to over-replicate which will do a scale up on volume. [...] If we see an additional replica on volume which is degraded then it should work right? The new replica will not be in previous NexusInfo. I think that was your fix.

@abhilashshetty04 

Thanks. On drain, I think the bypass still would not fire.

`has_scale_up_rebuild_target` needs two things: a replica absent from `NexusInfo`, and `num_replicas > NexusInfo.children.len()`. Drain gives the first but not the second. The doc keeps `num_replicas` as it is and introduces `effective_replica = num_replicas + N` for the count reconcilers, so the extra copy is not pruned as excess. Since my gate reads `num_replicas`, an over-replicated volume still looks like "3 wanted, 3 last published" and the grace period applies.

The doc already says drain has to bypass the `offline_rebuild_enabled()` flag, so that reconciler is in scope for the drain work anyway. This gate is probably one more thing on that list: either teach it about `effective_replica`, or let the `replica_move` marker bypass the grace period directly. Happy to add a comment on openebs/openebs#4245, so it is not lost once drain lands.

control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
yugchaudhari: Trimmed to one line.

control-plane/stor-port/src/types/v0/transport/mod.rs
yugchaudhari: Done. Worth noting the outward-facing names are already generic: the endpoint is `PUT /volumes/{id}/rebuild` and the command is `kubectl mayastor rebuild volume`. Only the internals still say offline, `TriggerOfflineRebuild` and `trigger_offline_rebuild`. Say the word if you want those generic too.

abhilashshetty04:
> > Should be fine to have this. Let's take it in.
> > For drain, We will have a mechanism to over-replicate which will do a scale up on volume. [...] If we see an additional replica on volume which is degraded then it should work right? The new replica will not be in previous NexusInfo. I think that was your fix.
> 
> @abhilashshetty04
> 
> Thanks. On drain, I think the bypass still would not fire.
> 
> `has_scale_up_rebuild_target` needs two things: a replica absent from `NexusInfo`, and `num_replicas > NexusInfo.children.len()`. Drain gives the first but not the second. The doc keeps `num_replicas` as it is and introduces `effective_replica = num_replicas + N` for the count reconcilers, so the extra copy is not pruned as excess. Since my gate reads `num_replicas`, an over-replicated volume still looks like "3 wanted, 3 last published" and the grace period applies.
> 
> The doc already says drain has to bypass the `offline_rebuild_enabled()` flag, so that reconciler is in scope for the drain work anyway. This gate is probably one more thing on that list: either teach it about `effective_replica`, or let the `replica_move` marker bypass the grace period directly. Happy to add a comment on [openebs/openebs#4245](https://github.com/openebs/openebs/issues/4245), so it is not lost once drain lands.

Yes effective_replica will be read moving forward for all internal decision making w.r.t replica count.  Do add the comment.

control-plane/agents/src/bin/core/volume/service.rs
```diff
@@ -175,6 +175,18 @@ impl VolumeOperations for Service {
         Ok(volume)
     }
 
+    async fn trigger_offline_rebuild(
```
abhilashshetty04: keep these generic.

control-plane/agents/src/bin/core/volume/service.rs
```diff
@@ -428,6 +440,25 @@ impl Service {
         volume.set_property(&self.registry, request).await?;
         self.registry.volume(&request.uuid).await
     }
+    /// Ask for the volume's offline rebuild to skip the grace period. The reconciler
+    /// still applies its viability and concurrency checks, so this only shortens the
+    /// wait, it does not force a rebuild that could not otherwise run.
+    #[tracing::instrument(level = "info", skip(self, request), err, fields(volume.uuid = %request.uuid))]
+    pub(super) async fn trigger_offline_rebuild(
```
abhilashshetty04: @yugchaudhari , I think till here we can keep it generic. if volume is `Degraded` and no target proceed with request_offline_rebuild(). In future, we can extend this function for online volume rebuild trigger as well.

control-plane/agents/src/common/errors.rs
```diff
@@ -415,6 +415,10 @@ pub enum SvcError {
     DrainNotAllowedWhenHAisDisabled {},
     #[snafu(display("Target switchover is not allowed without HA"))]
     SwitchoverNotAllowedWhenHAisDisabled {},
+    #[snafu(display(
```
abhilashshetty04, line 418: What happens if this gets executed for online volume or Degraded but published volume.? They all should be validated and Failed.

control-plane/agents/src/bin/core/volume/service.rs
yugchaudhari: Done. Renamed the whole path.

control-plane/agents/src/bin/core/volume/service.rs
yugchaudhari: Done.

control-plane/agents/src/common/errors.rs
yugchaudhari: Both are rejected now with `FailedPrecondition`: `RebuildTriggerVolumePublished` and `RebuildTriggerVolumeNotDegraded`.

control-plane/stor-port/src/types/v0/transport/mod.rs
Abhinandan-Purkait: What happens if someone issues it on a published volume?

abhilashshetty04: That gets rejected

FailedPrecondition: RebuildTriggerVolumePublished and RebuildTriggerVolumeNotDegraded.

abhilashshetty04 (approved):

control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
```diff
@@ -170,7 +170,11 @@ async fn offline_rebuild_reconcile(
     // Volume is unpublished with prior health info. Wait for it to actually be
     // Degraded, then enforce the grace period before initiating rebuild.
     if volume_state.status != VolumeStatus::Degraded {
-        volume.lock().metadata.clear_offline_rebuild_degraded();
+        let mut volume = volume.lock();
+        volume.metadata.clear_offline_rebuild_degraded();
+        // The volume recovered without us. Drop any outstanding request too, so it
```
tiagolobocastro, line 175:
```
        // The volume recovered without us
```

This is not entirely accurate, volume may in fact be faulted here, maybe just drop this part of the comment

control-plane/agents/src/bin/core/tests/volume/offline_rebuild.rs
```diff
@@ -260,6 +260,115 @@ async fn offline_rebuild_respects_max_offline_rebuilds() {
...
 
+/// An operator who knows a node is not coming back can ask for the rebuild to start
+/// without serving out the grace period.
+///
+/// Configured with a long grace period and degraded by stopping a node, so nothing
+/// else would bypass the wait: the scale-up bypass does not apply here. The volume is
+/// first observed sitting idle, then the request is issued and the temp nexus has to
+/// appear well inside the grace window. That gap is what shows the request did it.
+#[tokio::test]
+async fn offline_rebuild_can_be_requested_by_operator() {
```
tiagolobocastro, line 271: Can we reuse an existing cluster?

control-plane/agents/src/bin/core/tests/volume/offline_rebuild.rs
```diff
@@ -260,6 +260,115 @@ async fn offline_rebuild_respects_max_offline_rebuilds() {
...
+        &cluster,
+        &uid,
+        VolumeStatus::Degraded,
+        Duration::from_secs(10),
+    )
+    .await
+    .expect("Volume should become Degraded");
+
+    // Nothing should happen on its own: the grace period has barely started.
+    tokio::time::sleep(Duration::from_secs(SETTLE_SECS)).await;
```
tiagolobocastro, line 340: We don't need to wait, we already set wait for a high value

control-plane/rest/openapi-specs/v0_api_spec.yaml
```diff
@@ -1903,6 +1903,35 @@ paths:
...
+    put:
+      tags:
+        - Volumes
+      operationId: put_volume_rebuild
+      description: |-
+        Request an offline rebuild for a degraded, unpublished volume without waiting
```
tiagolobocastro, line 1912: Can we make this a property of the request whether we want to trigger an offline/online or any rebuild? As this is also useful for scenario when volume is published but waiting for the grace

control-plane/agents/src/bin/core/volume/service.rs
```diff
@@ -428,6 +440,49 @@ impl Service {
...
+            });
+        }
+        if state.status != VolumeStatus::Degraded {
+            return Err(SvcError::RebuildTriggerVolumeNotDegraded {
+                vol_id: request.uuid.to_string(),
+            });
+        }
+
+        let volume = self.specs().volume(&request.uuid).await?;
+        volume.lock().metadata.request_offline_rebuild();
```
tiagolobocastro, line 483: I think this should be part of the operation interface; we can make specific operations no log into the pstor so we can still keep this runtime metadata only

tiagolobocastro (commented):
What's the pros/cons here in having this as a declarative rather than imperative api?
(imperative - trigger the offline rebuild in-line with the request)

control-plane/agents/src/bin/core/controller/reconciler/volume/offline_rebuild.rs
yugchaudhari: Dropped, thanks.

control-plane/agents/src/bin/core/tests/volume/offline_rebuild.rs
yugchaudhari: Right, removed.

yugchaudhari:
> What's the pros/cons here in having this as a declarative rather than imperative api? (imperative - trigger the offline rebuild in-line with the request)

I went declarative to avoid duplicating the reconciler's checks, but having looked at what imperative would actually cost, I think you are right and it is the better shape. It removes more than it adds.

The coupling to the poll loop turned out to be thinner than I assumed. Inside `initiate_offline_rebuild` the only use of `PollContext` is `let registry = context.registry()`, and the same holds for the two helpers that produce its inputs, `has_scale_up_rebuild_target` and `offline_rebuild_budget`. The only place in the file that needs anything else from the context is `poll` itself, for `context.specs().volumes_rsc()`, and the service can reach that through `self.specs()`. So taking `&Registry` instead of `&PollContext` is enough for the service to call the very same functions, and there is no duplicated decision logic to worry about.

Going imperative then deletes the marker and everything hanging off it: the runtime field and its accessors, the extra branch in the reconciler's grace-period check, the two places it gets cleared, and the whole question of what happens to a pending request across a restart. It also answers your other comment about the operation interface, since there is no metadata mutation left to route through it, and `publish_with_mode` already goes through `VolumeOperation::Publish` and `start_update`.

The caller gets a real answer too. Right now a success means "recorded, it may happen"; imperative means "started", or a specific error saying the rebuild is not viable or the concurrency limit is reached, neither of which the caller can see today. Latency is fine, the call returns once the temp nexus exists, the rebuild itself still runs asynchronously under hot-spare.

Happy to rework it that way. Flagging one interaction first: @abhilashshetty04 asked for published volumes to be rejected, which is in the current revision, and your other comment suggests the request should eventually carry a kind so it can cover a published volume waiting on `faulted-child-wait-period`. Those two want reconciling, so tell me which way you want it and I will do the rework.

control-plane/agents/src/bin/core/tests/volume/offline_rebuild.rs
yugchaudhari: Can do. The natural one is `offline_rebuild_scale_up_bypasses_grace`, since it is the other test that needs a long grace period, so both scenarios can share a cluster the way `offline_rebuild_e2e` shares one across its four.

One thing to work out when I do it. The grace period is an agent env var, so a shared cluster means one value for both tests. The requested-rebuild scenario needs the grace to outlast the degrade detection (10s timeout) plus the start window (15s), so roughly 25s of headroom from the moment the volume degrades, and `offline_rebuild_scale_up_bypasses_grace` currently sets 30s. That is cutting it fine, so the shared value likely wants to go up, with the scale-up test keeping its own tighter number in its assertion so its bound does not get loosened.

Will fold this in with the imperative rework rather than doing it twice.

control-plane/rest/openapi-specs/v0_api_spec.yaml
yugchaudhari: Makes sense, and the published case is a real one. When a faulted child is partial-rebuild eligible, it waits out `faulted-child-wait-period`, which falls back to `TWAIT_SPERF`, 600s, when the flag is unset. It can be shortened, but only with the core agent's `--faulted-child-wait-period`, which is cluster-wide and needs a restart. The nearest thing today is removing the faulted child on the nexus, which stops the wait but forces a full rebuild and throws away the partial one. That is a nexus-level operation, not a volume-level one, and the plugin does not expose it. A kind on the request would give the operator the supported version of that.

Worth noting it pulls against the validation @abhilashshetty04 asked for, which currently rejects a published volume outright. If the request carries a kind then published stops being invalid and becomes the online case instead. I would rather settle that with the two of you than pick one and have it bounce. Raised the same point on the declarative/imperative thread.

control-plane/agents/src/bin/core/volume/service.rs
yugchaudhari: You are right that the mechanism is there. `log_op` returns a pair meaning "log to pstor before attempting" and "update pstor on completion", and `DestroySnapshot` and `RemoveUnusedReplica` already return `(false, false)`, with `CreateSnapshot` deciding it at runtime. So a `TriggerRebuild` variant could go through `start_update` and `complete_update` and still touch nothing persistent.

The part I had not thought about is that the operation interface is not only bookkeeping. `start_update_inner` gates on the spec status, so an operation is rejected with `PendingCreation` or `PendingDeletion` when the volume is still creating or already being deleted. Setting the flag directly skips that check.

That said, this only matters if the request stays declarative. If we go imperative, as your other comment suggests, the metadata mutation disappears entirely and the trigger runs the rebuild in line through `publish_with_mode`, which already goes through the operation interface as a proper `VolumeOperation::Publish`. Then there is nothing left here to route.

So I would rather settle the declarative question first and let this follow from it, than convert the marker to an operation now and delete it in the next round. If you decide to keep it declarative, I will make it an operation with `log_op` returning `(false, false)`.

tiagolobocastro:
Let's see what others have to say on imperative vs declarative; cc @Abhinandan-Purkait @abhilashshetty04

abhilashshetty04:
@yugchaudhari , Should be fine to extend this for published volume rebuild as well.

